PR #31, opened 2019-11-20
BREAKING CHANGE: Include collection pages in the `GET pages` endpoint

taufiqmmhd:
This PR now shows all collection pages, along with the existing simple pages on the `GET pages` endpoint.

### Context
Initially only simple pages were shown in the `GET pages` endpoint. The team decided that collection pages would be included as well as we were scraping the 'Collections' tab in the frontend.

### Implications
This pull request needs to be merged into `master` along with the frontend [PR](https://github.com/isomerpages/isomercms-frontend/pull/64) or else the frontend will break.

### DIFF
--- a/routes/pages.js
+++ b/routes/pages.js
@@ -41,14 +41,10 @@ router.get('/:siteName/pages', async function(req, res, next) {
       const CollectionPage = new File(access_token, siteName)
       const collectionPageType = new CollectionPageType(collectionName)
       CollectionPage.setFileType(collectionPageType)
-      const collectionPages = await Bluebird.map(await CollectionPage.list(), (item) => {
-        return {
-          ...item,
-          type: 'collection',
-          collectionName
-        }
-      })
-      return accumulator.concat(collectionPages)
+      const collectionPages = await CollectionPage.list()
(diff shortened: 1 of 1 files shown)


routes/pages.js
```diff
@@ -15,8 +17,41 @@ router.get('/:siteName/pages', async function(req, res, next) {
...
+     *  a) Lists all the collection pages with `CollectionPage.list()`
+     *  b) Map it to tag it with type `collection` [for frontend to know]
+     * 2) Concatenate it into the `accumulator`
+     * This then returns a flattened array of all collections pages from all collections (`allCollectionPages`)
+     */
+    const allCollectionPages = await Bluebird.reduce(collections, async (accumulator, collectionName) => {
+      const CollectionPage = new File(access_token, siteName)
+      const collectionPageType = new CollectionPageType(collectionName)
+      CollectionPage.setFileType(collectionPageType)
+      const collectionPages = await Bluebird.map(await CollectionPage.list(), (item) => {
```
prestonlimlianjie: You don't need to use `await Bluebird.map` here. A simple map over the result of `await CollectionPage.list()` works because there is no async call within the map.

prestonlimlianjie: The PR looks good otherwise

taufiqmmhd: ohhh yeah good point alright will fix that

PR #37, opened 2019-11-26
Refactor/fix environment variables

kwajiehao:
This PR achieves three things:
1. It introduces two new environment variables: `GITHUB_ORG_NAME` and `BRANCH_REF`. Existing code was refactored to accommodate these env variables.
2. Code was refactored so that the `baseEndpoints` do not end with a `/` (for example, for `PageType` the endpoint is now `pages` instead of `pages/`). This effectively undoes #19 but fixes the homepage `baseEndpoint` problem by using a ternary operator in `setFileType()`
2. It explicitly requests for the staging branch when making calls to the Github API

### DIFF
--- a/classes/Config.js
+++ b/classes/Config.js
@@ -1,6 +1,7 @@
 const axios = require('axios');
 
-const GITHUB_ORG_NAME = 'isomerpages'
+const GITHUB_ORG_NAME = process.env.GITHUB_ORG_NAME
+const BRANCH_REF = process.env.BRANCH_REF
 
 // validateStatus allows axios to handle a 404 HTTP status without rejecting the promise.
 // This is necessary because GitHub returns a 404 status when the file does not exist.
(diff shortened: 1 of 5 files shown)


classes/Config.js
```diff
@@ -18,8 +19,12 @@ class Config {
     try {
     	const endpoint = `https://api.github.com/repos/${GITHUB_ORG_NAME}/${this.siteName}/contents/_config.yml`
 
-	    const resp = await axios.get(endpoint, {
-	      validateStatus: validateStatus,
+			const params = {
+        validateStatus: validateStatus,
+        "branch": this.branchRef,
```
prestonlimlianjie: this.branchRef is not defined. We can just set this to be BRANCH_REF

classes/File.js
```diff
@@ -14,11 +15,7 @@ class File {
     this.accessToken = accessToken
     this.siteName = siteName
     this.baseEndpoint = null
-    this.branchRef = 'staging'
-  }
-
-  setBranchRef(branchRef) {
-    this.branchRef = branchRef
+    this.branchRef = BRANCH_REF
```
prestonlimlianjie: I don’t have strong feelings about this but maybe we could just omit the use of this.branchRef and just assign the param “branch” to BRANCH_REF in all the calls

classes/Config.js
prestonlimlianjie: There’s another param with branch on line 52

kwajiehao: Removed all mentions of `this.branchRef` and use `BRANCH_REF` instead

classes/File.js
kwajiehao: Removed all mentions of `this.branchRef` and use `BRANCH_REF` instead

PR #41, opened 2019-11-28
[Fix - Sites] Display correct sites for users

kwajiehao:
### Objective
This PR will restrict the repositories the user sees on the `Sites` page to the repositories that belong to the teams they are on.

### Context
Currently, all users receive access to all repositories when they log into IsomerCMS. This is not the intended behavior - we want to show users only the repositories they have write access to.

The way we retrieve repositories right now is to simply list all the repositories a user is associated with through Github's `user/repos` endpoint and filter for `isomerpages` organization repositories.

~Instead, we can use the `user/teams` [endpoint](https://developer.github.com/v3/teams/#list-user-teams) to retrieve all teams that the user is a part of, and then use the `teams/:team_id/repos` [endpoint](https://developer.github.com/v3/teams/#list-team-repos) to retrieve the repos. Note: using the teams API requires "`user`, `repo`, or `read:org` scope when authenticating via OAuth."~

Since technically, each user needs to belong to at least one team, and each member in the team has `write` access to all the repos associated with the team, we can filter based on which repos the user has `write` access to. 

We can use the `orgs/:orgs/repos` endpoint to get a full list of the repos in the `isomerpages` organization that the user has at least `read` access to, and then filter the response based on `permissions: { push: true }` to find the repos the user has `write` access to.

### DIFF
--- a/routes/sites.js
+++ b/routes/sites.js
@@ -59,11 +59,15 @@ router.get('/', async function(req, res, next) {
     // Variables to track pagination of user's repos in case user has more than 100
     let pageCount = 1
     let hasNextPage = true;
-    const filePath = `https://api.github.com/user/repos?per_page=100&page=`;
+    const filePath = `https://api.github.com/user/repos`;
 
     // Loop through all user repos
     while (hasNextPage) {
-      const resp = await axios.get(filePath + pageCount, {
+      const resp = await axios.get(filePath, {
+        params: {
+          per_page: 100,
+          page: pageCount,
+        },
         headers: {
           Authorization: `token ${access_token}`,
           "Content-Type": "application/json"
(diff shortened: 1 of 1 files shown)


routes/sites.js
```diff
@@ -59,22 +59,27 @@ router.get('/', async function(req, res, next) {
...
-          "Content-Type": "application/json"
+          Accept: "application/vnd.github.baptiste-preview+json",
         }
       })
 
       // Filter for isomer repos
       const isomerRepos = resp.data.reduce((acc, repo) => {
-        const { updated_at, full_name } = repo
+        const { permissions, updated_at, full_name } = repo
         const fullName = full_name.split('/')
```
prestonlimlianjie: Remove `full_name`.

We can extract `name` from the repo API response and use it. `name` is the repo name without the org name

kwajiehao: done

PR #48, opened 2019-12-06
[Feat] Create settings route and class

kwajiehao:
### Note
This PR is to be reviewed with PR#[99](https://github.com/isomerpages/isomercms-frontend/pull/99) in the frontend.

### Overview
This PR introduces the `settings` route and `Settings` class which will retrieve the relevant settings information to be rendered in the frontend. The type of settings data falls into two classes:
- general config and colors (retrieved from `_config.yml`)
- social media pages (retrieved from `_data/social-media.yml`)

### DIFF
--- a/classes/Settings.js
+++ b/classes/Settings.js
@@ -35,8 +35,6 @@ class Settings {
       const { content: socialMediaResp, sha: socialMediaSha } = await socialMedia
 
       return ({ configResp, socialMediaResp, configSha, socialMediaSha })
-      // res.status(200).json({ configContent, socialMediaContent, colorContent })
-
     } catch (err) {
       console.log(err)
     }
@@ -57,7 +55,6 @@ class Settings {
       const newConfigContent = base64.encode(yaml.safeDump(configSettings))
       const newSocialMediaContent = base64.encode(yaml.safeDump(socialMediaSettings))
       await config.update(newConfigContent, configSha)
-      console.log('abc')
       await IsomerDataFile.update('social-media.yml', newSocialMediaContent, socialMediaSha)
       return
     } catch (err) {
--- a/classes/Settings.js
+++ b/classes/Settings.js
@@ -1,4 +1,3 @@
-const yaml = require('js-yaml')
 const base64 = require('base-64')
 
 // import classes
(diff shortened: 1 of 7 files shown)


classes/Settings.js
```diff
@@ -0,0 +1,67 @@
...
+      const dataType = new DataType()
+      IsomerDataFile.setFileType(dataType)
+
+      // extract data
+      const { configSettings, configSha, socialMediaSettings, socialMediaSha } = payload
+
+      // update files
+      const newConfigContent = base64.encode(yaml.safeDump(configSettings))
+      const newSocialMediaContent = base64.encode(yaml.safeDump(socialMediaSettings))
+      await config.update(newConfigContent, configSha)
```
taufiqmmhd: likewise when updating, I feel that the frontend should only send in selective fields and not the entire file content to the backend.

routes/settings.js
```diff
@@ -0,0 +1,38 @@
...
+// Import Classes
+const { Settings } = require('../classes/Settings.js')
+
+router.get('/:siteName/settings', async function(req, res, next) {
+  try {
+    const { oauthtoken } = req.cookies
+    let { access_token } = jwtUtils.verifyToken(oauthtoken)
+    const { siteName } = req.params
+
+    const settingsFile = new Settings(access_token, siteName)
```
taufiqmmhd: Should it be called `settingsFile`? Since technically, under the hood, we're altering and retrieving content from two files. Maybe just `settings`?

classes/Settings.js
```diff
@@ -0,0 +1,67 @@
...
+      const { content: configResp, sha: configSha } = await config.read()
+      const socialMedia = IsomerDataFile.read('social-media.yml').catch((err) => {
+        // social-media.yml doesn't exist so we create a social-media.yml
+        const content = {
+          facebook: '',
+          linkedin: '',
+          twitter: '',
+          youtube: '',
+          instagram: '',
+        }
```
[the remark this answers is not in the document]
kwajiehao: In `File.js` when using the `read` method, if the response status is 404 we throw an error anyway, so this code will still trigger

routes/settings.js
kwajiehao: done

classes/Settings.js
```diff
@@ -0,0 +1,67 @@
...
+        const socialMediaYml = base64.encode(yaml.safeDump(content))
+        const { sha } = IsomerDataFile.create('social-media.yml', socialMediaYml)
+        return { content, sha }
+      })
+      const { content: socialMediaResp, sha: socialMediaSha } = await socialMedia
+
+      return ({ configResp, socialMediaResp, configSha, socialMediaSha })
```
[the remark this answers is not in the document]
kwajiehao: Modified to transmit only relevant fields

classes/Settings.js
kwajiehao: Modified to receive only relevant fields

kwajiehao:
Made changes so that only the relevant fields are transmitted - please remember to pull from the latest `feat/settings` branch in the frontend

PR #49, opened 2019-12-16
fix: use validateStatus correctly

prestonlimlianjie:
This PR fixes the bug where axios throws an error if the response status is 404 despite having a `validateStatus` param. The error is still thrown because `validateStatus` should exist as a field outside of `params` - this fix moves `validateStatus` out of the `params` field.

### DIFF
--- a/classes/Directory.js
+++ b/classes/Directory.js
@@ -29,11 +29,11 @@ class Directory {
       const endpoint = `${this.baseEndpoint}`
 
       const params = {
-        validateStatus: validateStatus,
         "ref": BRANCH_REF,
       }
 
       const resp = await axios.get(endpoint, {
+        validateStatus: validateStatus,
         params,
         headers: {
           Authorization: `token ${this.accessToken}`,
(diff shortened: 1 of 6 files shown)


classes/Directory.js
```diff
@@ -29,11 +29,11 @@ class Directory {
       const endpoint = `${this.baseEndpoint}`
 
       const params = {
-        validateStatus: validateStatus,
         "ref": BRANCH_REF,
       }
 
       const resp = await axios.get(endpoint, {
+        validateStatus: validateStatus,
```
kwajiehao: This can just be `validateStatus,`

prestonlimlianjie: > 1. The `validateStatus` function used is replicated in several files - it should exist in a utils file and be imported instead.
> 2. Only the `Directory.js` and `File.js` files have been modified, however `validateStatus` exists also in `Config.js` as part of `params`.
> 3. Finally, `sites.js` needs to implement `validateStatus` in its axios calls

Do we need `validateStatus` in `sites.js`?

kwajiehao: No we don't, ignore that

PR #50, opened 2019-12-20
feat: return site url in Settings endpoint

prestonlimlianjie:
This PR returns the site `url` field from `config.yml` in the `GET /settings` endpoint.


kwajiehao:
LGTM

PR #52, opened 2019-12-26
feat: logout button

prestonlimlianjie:
This PR adds the logout button feature to IsomerCMS.


routes/auth.js
```diff
@@ -9,9 +9,15 @@ const REDIRECT_URI = process.env.REDIRECT_URI
 const COOKIE_DOMAIN = process.env.COOKIE_DOMAIN
 const AUTH_TOKEN_EXPIRY_MS = process.env.AUTH_TOKEN_EXPIRY_DURATION_IN_MILLISECONDS.toString()
 const FRONTEND_URL = process.env.FRONTEND_URL
+const COOKIE_NAME = 'oauthtoken'
```
kwajiehao: nitpick, but is there a reason the cookie name should not also be in the env file?

prestonlimlianjie: No reason - just added it in as an env var

kwajiehao:
Closing legacy PRs as part of repo cleanup

PR #54, opened 2020-02-17
fix: allow larger image uploads

taufiq:
This is to fix the inability to upload large images. There were two
stoppers, the express server receiving it and sending it to github's
API. The solution for both issues were to increase the size limits.


lisatjide:
looks like this has been fixed but to check

prestonlimlianjie:
Issue is outdated - we want to keep the file size at 5MB max

Issue #57, opened 2020-10-07
Missing /auth/logout endpoint


kwajiehao:
We should have an endpoint which invalidates the cookie when the client tries to logout. We can refer to the Homer Native Backend implementation for how to do this.

In fact, we already have a reference to a `/auth/logout` endpoint on the frontend but we've not yet implemented it on the backend.

PR #59, opened 2020-10-14
Fix: add logout endpoint

alexanderleegs:
This PR fixes the bug identified in issue #57. Currently, the frontend attempts to call a non-existent endpoint to log the user out. This PR adds in an endpoint to handle the user log out by clearing the existing cookie. To be reviewed with PR #[154](https://github.com/isomerpages/isomercms-frontend/pull/154) on the isomercms-frontend repo.


kwajiehao (approved):
lgtm

Issue #61, opened 2020-10-27
[bug] pages should display on cms even though there is no collection in a repository


lisatjide:

lisatjide:
has been addressed by `Pages` section on CMS

PR #62, opened 2020-10-27
fix: allow collections to be empty in _config.yml

kwajiehao:
## Overview
Currently, when faced with a repo which does not have any collections, our `Collection` class' `list` method throws an error since it tries to use `Object.keys` on the non-existent `collection` attribute.

This commit fixes that by only performing the operation if the `collections` attribute exists on the config object, otherwise an empty array is returned. This allows us to accommodate repositories without any collections. This PR resolves issue #61

### DIFF
--- a/classes/Collection.js
+++ b/classes/Collection.js
@@ -16,10 +16,9 @@ class Collection {
     try {
     	const config = new Config(this.accessToken, this.siteName)
     	const { content, sha } = await config.read()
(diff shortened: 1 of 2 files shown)


gweiying (dismissed):
lgtm

alexanderleegs (dismissed):
lgtm

alexanderleegs (approved):
lgtm

PR #63, opened 2020-10-29
Feat/centralized error handler

kwajiehao:
## Overview

This PR accomplishes a few things:
- implement a central error handler for the backend
- create auth middleware which extracts the GitHub access token from the cookie's JWT and attach it to the `req` object
- change frontend redirection URL from `/sites` to `/auth` (refer to the relevant commit message for details)

### Central error handler

Currently, every single route function is wrapped in a try-catch block to catch any errors that occur in the function. This means that we would need to write code for every route function to handle their errors. Instead of doing so, we can simply remove all these try-catch blocks and use a central error handler instead by wrapping each route function with the route handler, which will catch any errors thrown by the route function and forward it to the error handler. The error handler can then take care of the HTTP response. The general HTTP response will be:

```
{
  error: {
    name,
    code,
    message,
  }
}
```

### Auth middleware

Currently, in **every** route function, we run the following code to extract the GitHub access token:

```
    const { oauthtoken } = req.cookies
    const { access_token } = jwtUtils.verifyToken(oauthtoken)
```

Instead of doing this in every route function, we can simply extract it in the middleware layer, and pass it onto the route by attaching it to the `req` object.


alexanderleegs (approved):
lgtm

PR #64, opened 2020-11-03
Setup GitHub actions

kwajiehao:
This PR sets up GitHub Actions for the CMS backend. In the process, we also remove our Travis configuration file since we don't want to double build and subsequently deploy simultaneously to elastic beanstalk. One thing to note with this change is that we will not be able to specify the S3 bucket name to which our package will be deployed. However, since this isn't a huge downside + Travis is no longer free for open-source, we will be sticking to GitHub actions.

These changes were tested with the `DEV_BRANCH` set to this branch (`refs/heads/setup-github-actions`). Refer to the `Actions` tab on this repo for example workflows run.


LoneRifle (approved):
No tests? =(

kwajiehao:
> No tests? =(

Will put some in place after beta is released

PR #66, opened 2020-11-04
Ci/refine gh actions

kwajiehao:
## Overview

This PR refines the repo's GitHub actions workflow file in the following ways:

1. Add the `test` job, which starts the application server as a test
2. Update `gatekeep` to also return `true` if the GitHub event is a `pull_request` in addition to the existing branch conditions
3. Allow the `test` job to run if the output of `gatekeep` is `true`
4. Make the `deploy` job dependent on the `test` job instead of `gatekeep`

The rationale is that we don't want the `test` job to be run when making random, mostly meaningless pushes to feature branches, but only if we're pushing to a major branch or if we're in a pull request. And also we want to run tests and make sure the server can start before deploying.

## Other notes

Since we run now start the server as a test, we also need to add environment variables into the repository's GitHub Secrets. This PR renames the `GITHUB_ORG_NAME` env var to `GH_ORG_NAME` since GitHub Secrets prevents users from setting secrets which start with the substring `GITHUB`. The elastic beanstalk environments have also been updated with the new environment variable.

## Tips

If you want to check what the full GitHub context is (to check what event is being pushed, what the event name is, what the action is, etc.), add the following snippet to your workflow file:

```
     - name: Dump GitHub context
        env:
          GITHUB_CONTEXT: ${{ toJson(github) }}
        run: echo "$GITHUB_CONTEXT"
```

This is taken from the official GitHub documentation ([link](https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions)). Do note that when using the whole `github` context, be mindful that it includes sensitive information such as `github.token`. GitHub masks secrets when they are printed to the console, but you should be cautious when exporting or printing the context.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,7 +6,7 @@ on:
 env:
   PRODUCTION_BRANCH: refs/heads/master
   STAGING_BRANCH: refs/heads/staging
-  DEV_BRANCH: refs/heads/staging-dev
+  DEV_BRANCH: refs/heads/ci/refine-gh-actions
   EB_APP: isomer-cms
   EB_ENV_PRODUCTION: isomercms-backend-prod
   EB_ENV_STAGING: isomercms-backend-staging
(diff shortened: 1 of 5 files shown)


kwajiehao:
Closing pull request because this isn't urgent right now. Leaving this here for reference.

PR #68, opened 2020-11-05
feat: add endpoint for unlinked pages

kwajiehao:
## Overview

This PR adds a new endpoint which returns only unlinked pages, that is, pages in the `pages` folder of the repo instead of all
pages in the repo. This is to accommodate our move to a file system hierarchy structure in the frontend.


alexanderleegs (approved):
lgtm

PR #70, opened 2020-11-11
Feat/retrieve csp

gweiying:
This PR creates a route to retrieve the `netlify.toml` file from the site's repo via the Github API and parse it to return its header values. This route is useful on the frontend because we are using the CSP rules specified in the header values to sanitise the page preview when users edit pages ([link](https://github.com/isomerpages/isomercms-frontend/pull/205)).

An example of the headers retrieved in the `netlify.toml` file:

```
[[headers]]
  for = "/*"
  [headers.values]
    X-XSS-Protection = "1; mode=block"
    Referrer-Policy = "no-referrer"
    X-Content-Type-Options = "nosniff"
    X-Frame-Options = "deny"
    Content-Security-Policy = "default-src 'self'; script-src 'self' blob: ..."
```

An example of the response sent to the frontend:

```
{  
    "X-XSS-Protection": "1; mode=block",
    "Referrer-Policy": "no-referrer",
    "X-Content-Type-Options": "nosniff",
    "X-Frame-Options": "deny",
    "Content-Security-Policy": "default-src 'self'; script-src 'self' blob: ..."
}
```

A new `NetlifyToml`  class was created because Isomer has a future goal to make the `netlify.toml` file a centralized setting across all sites, so we wouldn't need the other methods of the `File` class : `create`, `update`, `delete`. By creating a separate file, this allows us to maintain/ remove this functionality easily in the future.

***Note that we use the simplifying assumption that only one set of header values are specified and applied to all paths under each site.*** This is currently true for all Isomer repos, where `[[headers]]` is an array of length one, and `for = "/*"` which indicates that all paths on the site are affected by the corresponding set of `[header.values]`.

### DIFF
--- a/classes/NetlifyToml.js
+++ b/classes/NetlifyToml.js
@@ -1,0 +1,44 @@
+const axios = require('axios')
+const validateStatus = require('../utils/axios-utils')
+
+// Import error
+const { NotFoundError } = require('../errors/NotFoundError')
+
+const GITHUB_ORG_NAME = process.env.GITHUB_ORG_NAME
+const BRANCH_REF = process.env.BRANCH_REF
+
+class NetlifyToml {
+  constructor(accessToken, siteName) {
+    this.accessToken = accessToken
+    this.siteName = siteName
+  }
+
+  async read() {
+    try {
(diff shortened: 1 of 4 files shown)


kwajiehao (approved):
lgtm

PR #75, opened 2020-11-19
Fix/creation and renaming of resource categories

alexanderleegs:
This PR fixes several existing bugs related to resources. 
- Previously, when creating a resource page, we did not check for the existence of the resource category, hence it was possible to create a resource category with no index file.
- Similarly, when deleting the last resource page in a category, the index file was not cleared, so the category continued to exist.
- The endpoints related to deleting the index file of resource categories were not working correctly - this was due to us attempting to read and retrieve the sha of a file belonging to a subdirectory, which did not work. Instead, we have created a new File type (ResourceCategoryFile) which we use to query the file details.

### DIFF
--- a/routes/resourcePages.js
+++ b/routes/resourcePages.js
@@ -7,6 +7,7 @@ const { attachRouteHandlerWrapper } = require('../middleware/routeHandler')
 // Import classes 
 const { File, ResourcePageType } = require('../classes/File.js')
 const { ResourceRoom } = require('../classes/ResourceRoom.js')
+const { Resource } = require('../classes/Resource.js')
 
 // List pages in resource
 async function listResourcePages (req, res, next) {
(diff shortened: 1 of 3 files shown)


routes/resourcePages.js
```diff
@@ -97,6 +107,13 @@ async function deleteResourcePage(req, res, next) {
   IsomerFile.setFileType(resourcePageType)
   await IsomerFile.delete(pageName, sha)
 
+  // Check if collection has any pages left, and delete if none left
+  const resources = await IsomerFile.list()
+  if (_.isEmpty(resources)) {
+    const IsomerResource = new Resource(accessToken, siteName)
+    await IsomerResource.delete(resourceRoomName, resourceName)
+  }
+
```
kwajiehao, line 116: Instead of checking if the collection has any pages left, would it be feasible to check if there's only one page left and to delete the entire category if that's the case?

kwajiehao (commented):
left a comment

routes/resourcePages.js
alexanderleegs: Makes sense, resolved in https://github.com/isomerpages/isomercms-backend/pull/75/commits/c11d9fa80cbad6862658d990416b5dbf10ad3930

kwajiehao (approved):
lgtm, we will need to update the frontend to react accordingly (for example, when last page from resource category is deleted and the page reloads, it should go back to the main resources page)

PR #78, opened 2020-11-23
Feat: add new InputNameConflictError

alexanderleegs:
This PR introduces a new error for handling the case where we fail to create a new file due to conflicting file names. This PR is to be reviewed in conjunction with PR #[247](https://github.com/isomerpages/isomercms-frontend/pull/247) on the isomercms-frontend repo.

One thing to note is that we convert the error thrown by the Github API when processing - this is due to the endpoint that we are calling to create a new file also being used to modify existing files, thus the Github API returns us a `422 Unprocessable Entity` error due to the missing sha, when we actually want to reflect that there is a conflict in file names.

### DIFF
--- a/errors/InputValidationError.js
+++ b/errors/InputValidationError.js
@@ -1,0 +1,14 @@
+// Import base error
+const { BaseIsomerError } = require('./BaseError')
+
+class InputNameConflictError extends BaseIsomerError {
+  constructor (fileName) {
+    super(
+      409,
+      `A file with ${fileName} already exists.`
+    )
+  }
+}
+module.exports = {
+    InputNameConflictError,
+}
--- a/classes/File.js
+++ b/classes/File.js
@@ -4,6 +4,7 @@ const validateStatus = require('../utils/axios-utils')
 
 // Import error
 const { NotFoundError  } = require('../errors/NotFoundError')
+const { InputNameConflictError  } = require('../errors/InputValidationError')
 
 const GITHUB_ORG_NAME = process.env.GITHUB_ORG_NAME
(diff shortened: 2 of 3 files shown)


errors/InputValidationError.js
```diff
@@ -0,0 +1,14 @@
+// Import base error
+const { BaseIsomerError } = require('./BaseError')
+
+class InputNameConflictError extends BaseIsomerError {
```
kwajiehao, line 4: For the isomerCMS project, we've been trying to switch to a different custom error system where the error is solely identified by the HTTP error code, to avoid a situation where we have way too many custom errors as we did in Homer. In keeping with this new strategy, this should be a `ConflictError` with an argument in the constructor for the error message.

To standardize the error messages, we can define the types of `ConflictError` messages and export them. For example, we can have:

```
const inputNameConflictErrorMsg = (fileName) => `A file with ${fileName} already exists.`

class ConflictError extends BaseIsomerError { ... }

module.exports = {
  inputNameConflictErrorMsg,
  ConflictError,
}
```

and then throw the error like this:

```
throw new ConflictError(inputNameConflictErrorMsg(fileName))
```

kwajiehao (changes requested):
left a comment

errors/InputValidationError.js
alexanderleegs: Resolved with https://github.com/isomerpages/isomercms-backend/pull/78/commits/10ab10383ce3b458a66e66b0a49b2688647f6229

kwajiehao (approved):
lgtm

PR #80, opened 2020-11-23
Feat/update navigation file on creation of new collection and resource room

alexanderleegs:
This PR adds new functionality to automatically add newly created collections and resource rooms to the navigation bar, and meant to partially address issue [#234](https://github.com/isomerpages/isomercms-frontend/issues/234) on the isomercms-frontend repo. To be reviewed with PR [#256](https://github.com/isomerpages/isomercms-frontend/pull/256) on the isomercms-frontend repo.

I merged the frontend PR too early - there's a small extra fix on the frontend involving deleting of folders, which is fixed in PR [#257](https://github.com/isomerpages/isomercms-frontend/pull/257) instead


classes/Navigation.js
```diff
@@ -0,0 +1,66 @@
...
+    await axios.put(endpoint, params, {
+      headers: {
+        Authorization: `token ${this.accessToken}`,
+        "Content-Type": "application/json"
+      }
+    })
+  }
+}
+
+module.exports = { Navigation }
```
kwajiehao, line 66: Actually there's no need to create a `Navigation` class, to prevent duplication of code, we can create a new `File` and apply the existing `DataType` type to retrieve the `navigation.yml` file

kwajiehao (changes requested):
left a comment

classes/Navigation.js
alexanderleegs: Makes sense, resolved in https://github.com/isomerpages/isomercms-backend/pull/80/commits/e9c5760ba9d38faf01154c164d6107a3e6f22ac2, https://github.com/isomerpages/isomercms-backend/pull/80/commits/4a0b24c018f577edf0de877ed446648ac5f056ed, https://github.com/isomerpages/isomercms-backend/pull/80/commits/a1c294d7968429eae3e85cc2a01ea821fe7d426d

classes/Collection.js
```diff
@@ -57,6 +73,22 @@ class Collection {
...
+      const navContentObject = yaml.safeLoad(base64.decode(navContent))
+
+      for (let i = 0; i < navContentObject.links.length; i++) {
+        if (navContentObject.links[i].collection === collectionName) {
+          navContentObject.links.splice(i,1)
+          const newNavContent = base64.encode(yaml.safeDump(navContentObject))
+          await nav.update(NAV_FILE_NAME, newNavContent, navSha)
+          break
+        }
+      }
```
kwajiehao, line 90: Nit: generally it's safer to practice immutable programming (for a more detailed reason why: https://stackoverflow.com/questions/34385243/why-is-immutability-so-important-or-needed-in-javascript) so what we could do here, instead of unnaturally breaking a for loop, is:

```
const newNavLinks = navContentObject.links.filter(< filter out collection name>)
const newNavContentObject = {
  ...navContentObject,
  links: newNavLinks,
}
```

Since the code here works and is performant I don't have a problem with it so I'll leave it up to you

kwajiehao (approved):
lgtm, left a nit

classes/Collection.js
alexanderleegs: Oh yeah this way makes more sense. I've changed this in both resourceroom and collections in https://github.com/isomerpages/isomercms-backend/pull/80/commits/fb58d147288767db47d61d013dc4d15afd96c57d

kwajiehao (approved):
lgtm

PR #81, opened 2020-11-24
fix: update config settings and footer settings separately

gweiying:
This PR modifies the `post` update method on the `Settings` class to update each component (`footerSettings`, `configSettings`) separately, only if the relevant component is defined in the received payload.

This is because on the Edit Contact-Us page ([link](https://github.com/isomerpages/isomercms-frontend/pull/260)), we only update a key-value pair stored in `footerSettings`, so there is no need to store and resend `configSettings` on the `post` update. 

The implementation is a currently rough implementation for the frontend PR, and will be cleaned up, see (frontend TODOs list [link](https://github.com/isomerpages/isomercms-frontend/pull/260))

### DIFF
--- a/classes/Settings.js
+++ b/classes/Settings.js
@@ -135,19 +135,27 @@ class Settings {
     const footerContent = footer.content
     const navigationContent = navigation.content
 
-    const settingsObj = {
-      config: {
+    const settingsObj = {}
+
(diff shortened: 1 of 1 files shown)


kwajiehao (approved):
lgtm

PR #82, opened 2020-11-25
chore: log error message with distinct string for metric filter

kwajiehao:
This PR logs the serialized error with a distinct string, "Unrecognized internal server error" so that we can create a CloudWatch metric filter based on this string to identify and trigger alerts on unrecognized / non-Isomer errors


alexanderleegs (approved):
lgtm

PR #85, opened 2020-12-01
Feat/add settings fields

kwajiehao:
Note: this PR is to be reviewed with PR #[275](https://github.com/isomerpages/isomercms-frontend/pull/275) in the frontend.

## Overview

This PR introduces major changes in the `Settings` class to accommodate business requirements on the frontend. The new requirements are:
1. Allow users to retrieve and update the following values: `shareicon`, facebook pixel, google analytics, `is_government`
2. Reintroduce `title` field
3. Allow users to modify the agency logo
4. Allow users to set `title` of both the browser tab and the footer title at the same time

Requirements 1 & 2 are straightforward to achieve, since it simply involves sending new fields in the HTTP response.

Requirement 3 requires us to retrieve from and write to a new file, `_data/navigation.yml`, in addition to the `_config.yml` and `_data/footer.yml` file that we currently use.

Finally, requirement 4 requires us to write to a new file, `index.md`, if the `title` field is found to have changed.

This PR accomplishes all of the above, and also refactors the code for readability by creating a utility function for retrieving files concurrently since this code is repeated in both the `get` and `post` methods.

### DIFF
--- a/classes/Settings.js
+++ b/classes/Settings.js
@@ -1,4 +1,5 @@
 const { Base64 } = require('js-base64')
+const _ = require('lodash')
 const yaml = require('js-yaml')
 const Bluebird = require('bluebird')
 
@@ -10,6 +11,31 @@ const { File, DataType } = require('../classes/File.js')
 const FOOTER_PATH = 'footer.yml'
 const NAVIGATION_PATH = 'navigation.yml'
 
+const retrieveSettingsFiles = async (accessToken, siteName) => {
+  const configResp = new Config(accessToken, siteName)
+
+  const FooterFile = new File(accessToken, siteName)
+  const dataType = new DataType()
+  FooterFile.setFileType(dataType)
+
+  const NavigationFile = new File(accessToken, siteName)
+  NavigationFile.setFileType(dataType)
+
+  const fileRetrievalArr = [configResp.read(), FooterFile.read(FOOTER_PATH), NavigationFile.read(NAVIGATION_PATH)]
+
+  const fileContentsArr = await Bluebird.map(fileRetrievalArr, async (fileOp) => {
+    const { content, sha } = await fileOp
+    return { content: yaml.safeLoad(Base64.decode(content)), sha}
+  })
+
+  return {
+    configResp,
+    FooterFile,
+    NavigationFile,
+    fileContentsArr,
+  }
+}
+
 class Settings {
   constructor(accessToken, siteName) {
     this.accessToken = accessToken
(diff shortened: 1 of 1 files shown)


classes/Settings.js
```diff
@@ -16,66 +64,100 @@ class Settings {
...
+      {
+        payload: navigationSettings,
+        retrievedData: navigationContent,
+      },
+    ]
+
+    const updatedSettingsObjArr = settingsObjArr.map(({ payload, retrievedData}) => {
+      const settingsObj = _.cloneDeep(payload);
+      Object.keys(retrievedData).forEach((setting) => (settingsObj[setting] = payload[setting]));
+      return settingsObj
```
alexanderleegs, line 137: This portion seems incorrect - we should be cloning the original data (retrievedData) first, then updating it with the key value pairs from the new data (payload)?

So this function should be `const settingsObj = _.cloneDeep(retrievedData);
      Object.keys(payload).forEach((setting) => (settingsObj[setting] = payload[setting]));`

This also allows missing fields in the original data to be filled in

classes/Settings.js
```diff
@@ -1,13 +1,61 @@
...
+      return { content: yaml.safeLoad(Base64.decode(content)), sha}
+    }
+
+    // homepage requires special extraction as the content is wrapped in front matter
+    if (idx === 3) {
+      const homepageContent = Base64.decode(content)
+      const homepageFrontMatterObj = yaml.safeLoad(homepageContent.split('---')[1])
+      return { content: homepageFrontMatterObj, sha }
+    }
+  })
```
alexanderleegs, line 49: Nit: Could we consider having fileContentsArr as an object instead, with the appropriate key-value pair e.g. config: { content : ... }

Currently this part of the code is a little harder to understand/care needs to be taken when extending because of the magic number, although I think it's fine to leave it as is if it's difficult to change, since it's not that difficult to understand.

alexanderleegs (changes requested):
Left some comments.

classes/Settings.js
kwajiehao: You're right, I got confused with the naming of `retrievedData` - resolved with [f5f0d01](https://github.com/isomerpages/isomercms-backend/pull/85/commits/f5f0d01ee925d7e32001fe5a00361f6bda65e64a)

classes/Settings.js
kwajiehao: good point, resolved with [
d792730](https://github.com/isomerpages/isomercms-backend/pull/85/commits/d792730bc81048e64e6ac12423b2f597309abc3c)

alexanderleegs (approved):
lgtm

PR #87, opened 2020-12-04
Fix: throw NotFoundError for files

alexanderleegs:
This PR fixes the behaviour for throwing NotFoundErrors in File. Previously, it was not possible to reach the condition to throw the NotFoundError in read, as a file which could not be retrieved would have have a defined endpoint. This resulted in a different error being thrown. This PR fixes that and also adds in the check during deleting and updating files.

To be reviewed in conjunction with PR [#280](https://github.com/isomerpages/isomercms-frontend/pull/280) on the isomercms-frontend repo.

### DIFF
--- a/classes/File.js
+++ b/classes/File.js
@@ -89,6 +89,7 @@ class File {
     try {
       const files = await this.list()
       const fileToRead = files.filter((file) => file.fileName === fileName)[0]
+      if (fileToRead === undefined) throw new NotFoundError ('File does not exist')
       const endpoint = `${this.baseBlobEndpoint}/${fileToRead.sha}`
 
       const params = {
@@ -103,9 +104,7 @@ class File {
           "Content-Type": "application/json"
         }
       })
-  
-      if (resp.status === 404) throw new NotFoundError ('File does not exist')
-  
+
       const { content, sha } = resp.data
   
       return { content, sha }
@@ -134,6 +133,8 @@ class File {
 
       return { newSha: resp.data.commit.sha }
     } catch (err) {
+      const status = err.response.status
+      if (status === 404) throw new NotFoundError ('File does not exist')
       throw err
     }
   }
@@ -156,6 +157,8 @@ class File {
         }
       })
     } catch (err) {
+      const status = err.response.status
+      if (status === 404) throw new NotFoundError ('File does not exist')
       throw err
     }
   }


kwajiehao (changes requested):
~We are missing a case for the `NotFoundError`. Steps to reproduce:

1. Navigate to your workspace on the CMS frontend
2. Select an existing normal page in your repo in unlinked pages and delete it
3. Now attempt to delete it from the CMS. An error toast should appear - this is the correct behavior
4. Inspect the error message in the backend. You will observe the following error:

```
Unrecognized internal server error: Mon Dec 07 2020 15:46:05 GMT+0800 (Singapore Standard Time): {"name":"TypeError","message":"Cannot read property 'sha' of undefined","stack":"TypeError: Cannot read property 'sha' of undefined\n    at File.read (/Users/jiehaokwa/Documents/isomer-cms/isomercms-backend/classes/File.js:92:63)\n    at processTicksAndRejections (internal/process/task_queues.js:97:5)\n    at async readPage (/Users/jiehaokwa/Documents/isomer-cms/isomercms-backend/routes/pages.js:99:28)"}
```

This is because when we delete a file on the frontend, we first make a `GET` request to obtain the latest file SHA before submitting the delete request. However, if the file has already been deleted, then it doesn't appear in the file list, `files`, and so the variable `fileToRead` becomes undefined:

```
      ~const files = await this.list()~
      const fileToRead = files.filter((file) => file.fileName === fileName)[0]
      const endpoint = `${this.baseBlobEndpoint}/${fileToRead.sha}`
```

kwajiehao:
> ~We are missing a case for the `NotFoundError`. Steps to reproduce:~
> 
> ~1. Navigate to your workspace on the CMS frontend~
> ~2. Select an existing normal page in your repo in unlinked pages and delete it~
> ~3. Now attempt to delete it from the CMS. An error toast should appear - this is the correct behavior~
> ~4. Inspect the error message in the backend. You will observe the following error:~
> 
> ```
> Unrecognized internal server error: Mon Dec 07 2020 15:46:05 GMT+0800 (Singapore Standard Time): {"name":"TypeError","message":"Cannot read property 'sha' of undefined","stack":"TypeError: Cannot read property 'sha' of undefined\n    at File.read (/Users/jiehaokwa/Documents/isomer-cms/isomercms-backend/classes/File.js:92:63)\n    at processTicksAndRejections (internal/process/task_queues.js:97:5)\n    at async readPage (/Users/jiehaokwa/Documents/isomer-cms/isomercms-backend/routes/pages.js:99:28)"}
> ```
> 
> ~This is because when we delete a file on the frontend, we first make a `GET` request to obtain the latest file SHA before submitting the delete request. However, if the file has already been deleted, then it doesn't appear in the file list, `files`, and so the variable `fileToRead` becomes undefined:~
> 
> ```
>       const files = await this.list()
>       const fileToRead = files.filter((file) => file.fileName === fileName)[0]
>       const endpoint = `${this.baseBlobEndpoint}/${fileToRead.sha}`
> ```

Ignore the above comment, I was on the wrong branch

kwajiehao (approved):
lgtm

PR #88, opened 2020-12-07
Fix/handle renaming of folders with multiple files

alexanderleegs:
This PR addresses issue [#282](https://github.com/isomerpages/isomercms-frontend/issues/282) on the isomercms-frontend repo.

Currently, we encounter an issue when attempting to rename collections with multiple items inside. After a single file was moved, subsequent files would encounter an error due to a difference in current state? The error message shown is 
`"message":"is at 3e3261a537a9ca6b50615e4cc11fcfcc3406c1d4 but expected 2f51524a07f82e7225da00c0c652263457ed2919","documentation_url":"https://docs.github.com/rest/reference/repos#delete-a-file"}},"isAxiosError":true,"name":"Error","message":"Request failed with status code 409"` 
I'm not entirely sure what causes this issue, and why we don't have this problem when deleting folders, which also chains commits. 

The solution introduced in this PR involves the usage of git trees, which also provide us with a more efficient way of retrieving file information in a single API call. The endpoints to rename Collections and Resources have been changed to use git trees, as well as the unused endpoint to rename Resource Room.

One issue I faced for the rename resource category endpoint was the nesting of the resource folders - if I recursively retrieved the subtrees with a single API call, then directly modified the pathname of all items in that subtree, for some reason the old folder would not be deleted. I got around this by making 2 non-recursive calls instead, so I could rename the subtree directly, but would like to hear if a better method exists.

### DIFF
--- a/classes/Collection.js
+++ b/classes/Collection.js
@@ -5,7 +5,7 @@ const _ = require('lodash')
 
 const { Config } = require('./Config.js')
 const { File, CollectionPageType, DataType } = require('./File.js')
-const { deslugifyCollectionName } = require('../utils/utils.js')
+const { getRootTree, sendTree, deslugifyCollectionName } = require('../utils/utils.js')
 
 const NAV_FILE_NAME = 'navigation.yml'
 
@@ -146,28 +146,20 @@ class Collection {
       const newNavContent = base64.encode(yaml.safeDump(newNavContentObject))
       await nav.update(NAV_FILE_NAME, newNavContent, navSha)
 
-      // Get all collectionPages
-      const OldIsomerFile = new File(this.accessToken, this.siteName)
-      const oldCollectionPageType = new CollectionPageType(oldCollectionName)
-      OldIsomerFile.setFileType(oldCollectionPageType)
-      const collectionPages = await OldIsomerFile.list()
-      
-      // If the object is empty (there are no pages in the collection), do nothing
-      if (_.isEmpty(collectionPages)) return 
-
-      // Set up new collection File instance
-      const NewIsomerFile = new File(this.accessToken, this.siteName)
-      const newCollectionPageType = new CollectionPageType(newCollectionName)
-      NewIsomerFile.setFileType(newCollectionPageType)
-
-      // Rename all collectionPages
-      await Bluebird.map(collectionPages, async(collectionPage) => {
-        let pageName = collectionPage.fileName
-        const { content, sha } = await OldIsomerFile.read(pageName)
-        await OldIsomerFile.delete(pageName, sha)
-        return NewIsomerFile.create(pageName, content)
+      const { gitTree, currentCommitSha } = await getRootTree(this.siteName, this.accessToken);
+      const oldCollectionDirectoryName = `_${oldCollectionName}`
+      const newCollectionDirectoryName = `_${newCollectionName}`
+      const newGitTree = gitTree.map(item => {
+        if (item.path === oldCollectionDirectoryName) {
+          return {
+            ...item,
+            path: newCollectionDirectoryName
+          }
+        } else {
+          return item
+        }
       })
-      
+      await sendTree(newGitTree, currentCommitSha, this.siteName, this.accessToken, `Rename collection from ${oldCollectionName} to ${newCollectionName}`);
     } catch (err) {
       throw err
     }
--- a/classes/Collection.js
+++ b/classes/Collection.js
@@ -109,6 +109,7 @@ class Collection {
 
   async rename(oldCollectionName, newCollectionName) {
     try {
+      const commitMessage = `Rename collection from ${oldCollectionName} to ${newCollectionName}`
       // Rename collection in config
       const config = new Config(this.accessToken, this.siteName)
       const { content, sha } = await config.read()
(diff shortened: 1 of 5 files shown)


classes/Collection.js
```diff
@@ -146,28 +146,20 @@ class Collection {
...
+          return {
+            ...item,
+            path: newCollectionDirectoryName
+          }
+        } else {
+          return item
+        }
       })
-      
+      await sendTree(newGitTree, currentCommitSha, this.siteName, this.accessToken, `Rename collection from ${oldCollectionName} to ${newCollectionName}`);
```
kwajiehao, line 162: Can we make the text `Rename collection from ${oldCollectionName} to ${newCollectionName}` a constant and reference that instead?

classes/Resource.js
```diff
@@ -38,42 +39,35 @@ class Resource {
...
+            path: `${resourceRoomName}/${newResourceName}`
+          })
+        } else {
+          newGitTree.push({
+            ...item,
+            path: `${resourceRoomName}/${item.path}`
+          })
+        }
       })
+      await sendTree(newGitTree, currentCommitSha, this.siteName, this.accessToken, `Rename resource category from ${resourceName} to ${newResourceName}`);
```
kwajiehao, line 70: refer to above comment on commit message

classes/Collection.js
```diff
@@ -146,28 +146,20 @@ class Collection {
...
-      const newCollectionPageType = new CollectionPageType(newCollectionName)
-      NewIsomerFile.setFileType(newCollectionPageType)
-
-      // Rename all collectionPages
-      await Bluebird.map(collectionPages, async(collectionPage) => {
-        let pageName = collectionPage.fileName
-        const { content, sha } = await OldIsomerFile.read(pageName)
-        await OldIsomerFile.delete(pageName, sha)
-        return NewIsomerFile.create(pageName, content)
+      const { gitTree, currentCommitSha } = await getRootTree(this.siteName, this.accessToken);
```
kwajiehao, line 149: One thing to note is that we have an atomicity problem in this endpoint: it's possible for one of the API calls to succeed and the rest to fail (for example, config is updated, but nav is not), leading to an inconsistent state. 

I think one way we can fix this is that we retrieve the git tree first so that we have the current commit sha that our branch is pointing to. And then whenever there is an error, we can point the branch back to the original commit sha with an API call in the catch block.

utils/utils.js
```diff
@@ -24,6 +27,98 @@ function deslugifyCollectionPage(collectionPageName) {
...
+      params: {
+        ref: branchRef,
+      },
+      headers,
+    });
+    // Get the tree sha of the latest commit
+    const { commit: { tree: { sha: treeSha } } } = commits[0];
+    const currentCommitSha = commits[0].sha;
+
+    const { data: { tree: gitTree } } = await axios.get(`https://api.github.com/repos/${GITHUB_ORG_NAME}/${repo}/git/trees/${treeSha}${isRecursive ? `?recursive=1` : ''}`, {
```
kwajiehao, line 48: `isRecursive` is undefined

classes/Collection.js
kwajiehao: Ideally, it would have been great if we could update the pages all together with the git tree, but git tree does not allow you to modify the blob contents of individual files in a single api call

kwajiehao:
Can you specify the error message (regarding the sha) that you encountered?

kwajiehao (changes requested):
> One issue I faced for the rename resource category endpoint was the nesting of the resource folders - if I recursively retrieved the subtrees with a single API call, then directly modified the pathname of all items in that subtree, for some reason the old folder would not be deleted. I got around this by making 2 non-recursive calls instead, so I could rename the subtree directly, but would like to hear if a better method exists.

Let's have a call offline to discuss this issue. To clarify, by the above comment you mean that simply retrieving the root tree and renaming the appropriate files is insufficient?

utils/utils.js
alexanderleegs: Thanks for pointing that out - I was playing around with recursive retrieval earlier and forgot to remove it. Resolved in https://github.com/isomerpages/isomercms-backend/pull/88/commits/04d90118a7fd4c1f31f397f1d7079b5307845d4c

classes/Collection.js
alexanderleegs: Good idea, resolved in https://github.com/isomerpages/isomercms-backend/pull/88/commits/ac2b6d714a0cb9949b5405f9c8af6855b80ea11e

classes/Resource.js
alexanderleegs: Resolved in https://github.com/isomerpages/isomercms-backend/pull/88/commits/ac2b6d714a0cb9949b5405f9c8af6855b80ea11e

classes/Collection.js
kwajiehao: Let's do this in a separate PR for clarity

alexanderleegs: As discussed, we will be resolving this in a separate PR

alexanderleegs:
> To clarify, by the above comment you mean that simply retrieving the root tree and renaming the appropriate files is insufficient?

Yeah, if we simply rename all the files, new copies of the files with the new names are created, but the old ones are not deleted. I suspect this has something to do with the treeShas being unchanged.

kwajiehao (approved):
lgtm

PR #89, opened 2020-12-08
Feat: add endpoint to check if user has access to site

alexanderleegs:
This PR adds a new endpoint to check if users have access to a given site. To be reviewed in conjunction with PR[#280](https://github.com/isomerpages/isomercms-frontend/pull/280) on the isomercms-frontend repo.

To check if a user has access to a site, we check if the user is one of the collaborators on that repo, and throw an error message if either the site does not exist or the user does not have the appropriate permissions. We have opted to throw the same 404 error in both of these cases.

### DIFF
--- a/routes/sites.js
+++ b/routes/sites.js
@@ -99,13 +99,12 @@ async function checkHasAccess (req, res, next) {
     const { siteName } = req.params
     
     const endpoint = `https://api.github.com/repos/${ISOMER_GITHUB_ORG_NAME}/${siteName}/collaborators/${userId}`
-    const resp = await axios.get(endpoint, {
+    await axios.get(endpoint, {
       headers: {
         Authorization: `token ${accessToken}`,
         "Content-Type": "application/json",
       }
     })
-    console.log(resp)
 
     res.status(200).json()
   } catch (err) {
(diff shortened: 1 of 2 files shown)


routes/sites.js
```diff
@@ -90,6 +92,32 @@ async function getSites (req, res, next) {
...
+    const { siteName } = req.params
+    
+    const endpoint = `https://api.github.com/repos/${ISOMER_GITHUB_ORG_NAME}/${siteName}/collaborators/${userId}`
+    const resp = await axios.get(endpoint, {
+      headers: {
+        Authorization: `token ${accessToken}`,
+        "Content-Type": "application/json",
+      }
+    })
+    console.log(resp)
```
kwajiehao, line 108: Can we remove this log statement?

kwajiehao (approved):
lgtm

leaving a comment here for posterity: we can choose to rethrow the error as a 401 or 403 in the future, but a 404 will do for now

routes/sites.js
alexanderleegs: Whoops thanks for catch that, resolved in https://github.com/isomerpages/isomercms-backend/pull/89/commits/ba92d4b3ca2f41848a714c49e6c0d1ee89dbfea5

kwajiehao (approved):
lgtm

Issue #83, opened 2020-11-26
Paginate/optimize site retrieval code


kwajiehao:
Currently, it takes a whopping **7** seconds to retrieve a user's repos (for Isomer admin users). Even for users with access to less repositories, it still takes a not insignificant amount of time. This isn't a great user experience, especially for most users who only have 1 or 2 sites.

PR #90, opened 2020-12-08
feat: improve `/sites` endpoint performance

kwajiehao:
## Overview

This PR addresses issue #83 

Previously, site retrieval was slow because the GitHub API for retrieving org repos was paginated, and we retrieved the data sequentially, one page at a time. This meant that it often took up to 7 or even 8 seconds each time this endpoint is accessed (each page took around 3 seconds, perhaps due to the large amount of data being sent).

Since we know the number of repos our `isomerpages` GitHub organization has, we can use this information to speed up our endpoint by making concurrent calls instead of stepping through the API pagination. This PR improves performance by making concurrent API calls to retrieve repo info so that it now only takes around 3 seconds for the endpoint to respond, more than halving the amount of time take previously.

This commit also introduces an optional env var, `ISOMERPAGES_REPO_PAGE_COUNT`, which determines how many pages of the GitHub API to comb simultaneously.

### DIFF
--- a/routes/sites.js
+++ b/routes/sites.js
@@ -1,11 +1,14 @@
 const express = require('express');
 const router = express.Router();
 const axios = require('axios');
+const Bluebird = require('bluebird');
 const _ = require('lodash');
 const { attachRouteHandlerWrapper } = require('../middleware/routeHandler');
 // Import error
 const { NotFoundError } = require('../errors/NotFoundError')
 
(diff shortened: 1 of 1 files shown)


alexanderleegs (dismissed):
lgtm

alexanderleegs (approved):
lgtm

PR #91, opened 2020-12-08
Feat/rollback changes if not all successful

alexanderleegs:
This PR builds on PR https://github.com/isomerpages/isomercms-backend/pull/88. Currently, when calling a particular endpoint, it is possible for some of the API calls to succeed, while others fail (for example, config is updated, but nav is not), leading to an inconsistent state. 

To prevent this behaviour, this PR introduces a new route handler for endpoints which involve multiple API calls. The sha of the initial state of the repo is stored before any changes have been made, and if an error is thrown at any point, the branch is reset to its original state. Only the endpoints which involve multiple calls to the Github API utilise this new route handler, to prevent unnecessary querying. The endpoints which have been changed are:

- Collection Page - create, delete
- Collection - create, delete, rename
- Resource Page - create, delete
- Resource room - create, delete, rename
- Resource Category - create, delete, rename
- Settings - update

For reference, the time taken to retrieve the latest commit sha is around 1.5-2s, while the time taken to retrieve the git tree from its sha is around 1s for the a-test repo. The time taken to retrieve the git tree might be longer for repos with more items, although it should be noted that size of subtrees does not impact the time taken, since we currently do not retrieve items recursively.

### DIFF
--- a/middleware/routeHandler.js
+++ b/middleware/routeHandler.js
@@ -1,3 +1,5 @@
+const { backOff } = require('exponential-backoff')
+
 const { getCommitAndTreeSha, revertCommit } = require('../utils/utils.js')
 
 const attachRouteHandlerWrapper = (routeHandler) => async (req, res, next) => {
@@ -16,8 +18,12 @@ const attachRollbackRouteHandlerWrapper = (routeHandler) => async (req, res, nex
   } catch (err) {
     next(err)
   }
-  routeHandler(req, res).catch((err) => {
-    revertCommit(originalCommitSha, siteName, accessToken)
+  routeHandler(req, res).catch(async (err) => {
+    try {
+      await backOff(() => revertCommit(originalCommitSha, siteName, accessToken))
+    } catch (err) {
+      next(err)
+    }
     next(err)
   })
 }
--- a/middleware/routeHandler.js
+++ b/middleware/routeHandler.js
@@ -1,10 +1,28 @@
+const { getCommitAndTreeSha, revertCommit } = require('../utils/utils.js')
+
 const attachRouteHandlerWrapper = (routeHandler) => async (req, res, next) => {
-    routeHandler(req, res).catch((err) => {
-      next(err)
-    })
+  routeHandler(req, res).catch((err) => {
+    next(err)
+  })
+}
+
+const attachRollbackRouteHandlerWrapper = (routeHandler) => async (req, res, next) => {
+  const { accessToken } = req
+  const { siteName } = req.params
+  let originalCommitSha
+  try {
+    const { currentCommitSha } = await getCommitAndTreeSha(siteName, accessToken)
+    originalCommitSha = currentCommitSha
+  } catch (err) {
+    next(err)
   }
+  routeHandler(req, res).catch((err) => {
+    revertCommit(originalCommitSha, siteName, accessToken)
+    next(err)
+  })
+}
   
-  module.exports = {
-    attachRouteHandlerWrapper,
-  }
-  
+module.exports = {
+  attachRouteHandlerWrapper,
+  attachRollbackRouteHandlerWrapper,
+}
--- a/middleware/routeHandler.js
+++ b/middleware/routeHandler.js
@@ -21,8 +21,8 @@ const attachRollbackRouteHandlerWrapper = (routeHandler) => async (req, res, nex
   routeHandler(req, res).catch(async (err) => {
     try {
       await backOff(() => revertCommit(originalCommitSha, siteName, accessToken))
-    } catch (err) {
-      next(err)
+    } catch (retryErr) {
+      next(retryErr)
     }
     next(err)
   })
--- a/routes/collectionPages.js
+++ b/routes/collectionPages.js
@@ -6,7 +6,7 @@ const base64 = require('base-64');
 const _ = require('lodash');
 
 // Import middleware
-const { attachRouteHandlerWrapper } = require('../middleware/routeHandler')
+const { attachRouteHandlerWrapper, attachRollbackRouteHandlerWrapper } = require('../middleware/routeHandler')
 
 // Import classes 
 const { Collection } = require('../classes/Collection.js')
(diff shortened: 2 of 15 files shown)


utils/utils.js
```diff
@@ -119,6 +111,28 @@ async function sendTree(gitTree, currentCommitSha, repo, accessToken, message, b
   });
 }
 
+// send the new tree object back to Github and point the latest commit on the staging branch to it
```
kwajiehao, line 114: Can we update the comment? We're not sending the tree object back

routes/resourcePages.js
```diff
@@ -140,10 +140,10 @@ async function renameResourcePage (req, res, next) {
   res.status(200).json({ resourceName, pageName: newPageName, content, sha: newSha })
 }
 router.get('/:siteName/resources/:resourceName', attachRouteHandlerWrapper(listResourcePages))
-router.post('/:siteName/resources/:resourceName/pages', attachRouteHandlerWrapper(createNewResourcePage))
+router.post('/:siteName/resources/:resourceName/pages', attachRollbackRouteHandlerWrapper(createNewResourcePage))
 router.get('/:siteName/resources/:resourceName/pages/:pageName', attachRouteHandlerWrapper(readResourcePage))
 router.post('/:siteName/resources/:resourceName/pages/:pageName', attachRouteHandlerWrapper(updateResourcePage))
-router.delete('/:siteName/resources/:resourceName/pages/:pageName', attachRouteHandlerWrapper(deleteResourcePage))
+router.delete('/:siteName/resources/:resourceName/pages/:pageName', attachRollbackRouteHandlerWrapper(deleteResourcePage))
 router.post('/:siteName/resources/:resourceName/pages/:pageName/rename/:newPageName', attachRouteHandlerWrapper(renameResourcePage))
```
kwajiehao, line 147: `renameResourcePage` should use `attachRollbackRouteHandlerWrapper` as well in case the creation of the new file succeeds but the delete of the old file fails

routes/collectionPages.js
```diff
@@ -203,10 +203,10 @@ async function renameCollectionPage (req, res, next) {
 
 router.get('/:siteName/collections/:collectionName', attachRouteHandlerWrapper(listCollectionPages))
 router.get('/:siteName/collections/:collectionName/pages', attachRouteHandlerWrapper(listCollectionPagesDetails))
-router.post('/:siteName/collections/:collectionName/pages', attachRouteHandlerWrapper(createNewcollectionPage))
+router.post('/:siteName/collections/:collectionName/pages', attachRollbackRouteHandlerWrapper(createNewcollectionPage))
 router.get('/:siteName/collections/:collectionName/pages/:pageName', attachRouteHandlerWrapper(readCollectionPage))
 router.post('/:siteName/collections/:collectionName/pages/:pageName', attachRouteHandlerWrapper(updateCollectionPage))
-router.delete('/:siteName/collections/:collectionName/pages/:pageName', attachRouteHandlerWrapper(deleteCollectionPage))
+router.delete('/:siteName/collections/:collectionName/pages/:pageName', attachRollbackRouteHandlerWrapper(deleteCollectionPage))
 router.post('/:siteName/collections/:collectionName/pages/:pageName/rename/:newPageName', attachRouteHandlerWrapper(renameCollectionPage))
```
kwajiehao, line 210: `renameCollectionPage` should use `attachRollbackRouteHandlerWrapper` as well in case the creation of the new file succeeds but the delete of the old file fails

middleware/routeHandler.js
```diff
@@ -1,10 +1,28 @@
...
+
+const attachRollbackRouteHandlerWrapper = (routeHandler) => async (req, res, next) => {
+  const { accessToken } = req
+  const { siteName } = req.params
+  let originalCommitSha
+  try {
+    const { currentCommitSha } = await getCommitAndTreeSha(siteName, accessToken)
+    originalCommitSha = currentCommitSha
+  } catch (err) {
+    next(err)
```
kwajiehao, line 17: Currently no error will be caught because `getCommitAndTreeSha` does not throw any errors, it simply console.logs errors

middleware/routeHandler.js
```diff
@@ -1,10 +1,28 @@
...
   }
+  routeHandler(req, res).catch((err) => {
+    revertCommit(originalCommitSha, siteName, accessToken)
```
kwajiehao, line 20: We might need a retry mechanism for `revertCommit` if it fails

kwajiehao (changes requested):
left comments

utils/utils.js
alexanderleegs: Whoops, fixed in https://github.com/isomerpages/isomercms-backend/pull/91/commits/8983a82c26aaa8552d21142f108a82364d20bf4f

routes/resourcePages.js
alexanderleegs: Good catch, resolved in https://github.com/isomerpages/isomercms-backend/pull/91/commits/967eaa01b0087bbd047e511dc28ccc8993a1eadf

routes/collectionPages.js
alexanderleegs: Resolved in https://github.com/isomerpages/isomercms-backend/pull/91/commits/967eaa01b0087bbd047e511dc28ccc8993a1eadf

middleware/routeHandler.js
alexanderleegs: Right, changed to throw error instead in https://github.com/isomerpages/isomercms-backend/pull/91/commits/16362d1bc82615e3eac2e44234650a11d52ccbc5

middleware/routeHandler.js
alexanderleegs: As discussed, we can retry this with exponential backoff. Resolved in https://github.com/isomerpages/isomercms-backend/pull/91/commits/004d904fb79cd59dda82937d0e1b1a41d8483d2b

alexanderleegs:
> this is confusing, there are now two `err` objects referenced. Can we rename it so it is more clear which `err` we are forwarding?

Good point, fixed in https://github.com/isomerpages/isomercms-backend/pull/91/commits/05c1fbd768183753858045ff1fc608fa9a50056e

kwajiehao (approved):
lgtm

PR #84, opened 2020-12-01
Feat: throw ConflictError if 409 received

alexanderleegs:
This PR allows ConflictError to be thrown if a 409 error is received from the Github API. To be reviewed with PR [#247]https://github.com/isomerpages/isomercms-frontend/pull/247 on the isomercms-frontend repo.

PR #93, opened 2020-12-09
Merge to Prod

kwajiehao:
This release to production contains the following changes:

### Bugs
- Renaming of directories with multiple files (#88)
- Increase nginx client max body size so that images up to 5mb can be uploaded (#101)

### Error-handling
- Adding new custom errors and applying them to application code (404, 409) (#78, #84, #87)

### Error-logging and debugging
- Add custom string in error handler so that we can filter for unknown errors on AWS CloudWatch (#82)

### Misc. improvements
- Automatically update navigation file when new collection/resource category is created (#80)
- Add new fields to `Settings` class (#85)
- Endpoint for checking whether user has push access to repo (#89)

### Performance optimization
- Improve `/sites` route loading time (#90)
- Improve `Settings` class `post` method load time (#81, #92)

### DIFF
--- a/.ebextensions/nginx-max-body-size.config
+++ b/.ebextensions/nginx-max-body-size.config
@@ -1,0 +1,11 @@
+files:
+    "/etc/nginx/conf.d/proxy.conf" :
+        mode: "000755"
+        owner: root
+        group: root
+        content: |
+           client_max_body_size 10M;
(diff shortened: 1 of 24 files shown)


alexanderleegs (approved):
lgtm

PR #94, opened 2020-12-09
Fix/error handling for large payload

alexanderleegs:
This PR adds error handling for large payloads. In addition, it fixes a bug where errors thrown by the express json body parser would always result in a network error for the frontend. To be reviewed with PR #[289](https://github.com/isomerpages/isomercms-frontend/pull/289) on the isomercms-frontend repo.

The error handling for large payloads is different from the regular IsomerErrors, since this check is done by the express json body parser (`app.use(express.json({ limit: '5mb'}))`), it goes directly to our errorhandler before an IsomerError can be thrown. As such, a separate check in the error handler has been added.

For the network error, this was being caused by the cors middleware being initialised after the json body parser - this caused the response to not have the proper headers set, resulting in a network error always occurring. The initialisation of the cors middleware has been moved up to account for this.

### DIFF
--- a/middleware/errorHandler.js
+++ b/middleware/errorHandler.js
@@ -22,15 +22,27 @@ function errorHandler (err, req, res, next) {
             },
         })
     } else {
-        logger.info(`Unrecognized internal server error: ${errMsg}`)
-        res.status(500).json({
-            error: {
-                code: 500,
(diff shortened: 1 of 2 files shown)


kwajiehao (approved):
lgtm

PR #95, opened 2020-12-11
chore: add `/v1` to all URI paths

kwajiehao:
## Overview

This PR adds a `/v1` prefix to all URI paths to version our API. This will be useful for several reasons:
- allows us to gracefully accommodate a v2 api which uses github's graphql API
- allows us to craft tight firewall rules (currently, we search for whether a
URI path contains `/sites` or `/auth`, which would allow nonsense URI paths like
`/authors`)

**Note**: things which we need to do before we merge the PR:
- ~Change the `REDIRECT_URI` env var on backend to include `/v1`~ (if the `REDIRECT_URI` is unspecified it will use the callback URL from the GitHub app)
- Update the `REACT_APP_BACKEND_URL` env var on frontend to include `/v1`
- Update the callback URL for the staging app (and eventually the production app) to include `/v1`

### DIFF
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -213,7 +213,7 @@ components:
 
 
 paths:
-  /:
+  /v1:
     get:
       tags:
         - Authentication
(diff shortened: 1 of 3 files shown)


alexanderleegs (approved):
lgtm

PR #108, opened 2021-02-16
Feat/page reordering endpoints

kwajiehao:
## Overview

This PR adds a new endpoint, `GET /v1/sites/:siteName/folders`, which returns all files and directories contained within a collection/folder or third-nav section/sub-folder. Previously, we had only returned data of type `file` - with this change, we will return data of both types `file` and `dir` (refer to the GitHub contents API for sample responses - [link](https://docs.github.com/en/rest/reference/repos#get-repository-content)). We do so by creating a new `FolderType` class and expanding the `Directory` class to accept `FolderType` as a directory type. 

Note that this endpoint is intended for use with both top-level collection folders as well as nested sub-folders, since we have determined that we will need to accept and deal with nested sub-folders going forward (for images and files). 

### Rationale for using query params instead of path params

Since this endpoint is designed to be used with both top-level folders and nested folders, we cannot use path params because any nested folder will contain the `/` character, which messes with Express' path routing mechanism. Therefore, to avoid routing confusion, we use query params here instead.

Note: since our research has shown that users prefer the term `folder` to `collection`, and `sub-folder` to `third nav`, we will be using these new terms more frequently going forward. As an interim, we can use the terms together (like `collection/folder`) as I have done so above to facilitate the transition.


gweiying (commented):
As discussed, currently the distinction between files and folders is not conceptually clear on our backend. We currently use the File class to retrieve and list files (in Collection, Resource etc), but we should move that functionality to the separate Directory class. 

As a first step, we can move this listAll functionality to the Directory class, and do the refactoring for collections and images at a later stage.

routes/folders.js
```diff
@@ -0,0 +1,36 @@
...
+
+// List pages and directories in folder
+async function listFolderContent (req, res, next) {
+    const { accessToken } = req
+    const { siteName } = req.params
+    const { path } = req.query // paths begin with _
+
+    // Validation of path
+    const pathArr = path.split('/')
+    if (path.slice(0,1) !== '_' || pathArr.length > 2 || (pathArr.length === 2 && pathArr[1].includes('.'))) {
```
gweiying, line 21: This validation maybe can be left to a "higher level"

kwajiehao: Validation has been shifted to the `list` method for the `Directory` class so that it is performed only conditionally (when the `dirType` is `FolderType`) - see commit [3619f17](https://github.com/isomerpages/isomercms-backend/pull/108/commits/3619f174823f9754dc16335ec33a823521e6fb8b)

classes/Directory.js
```diff
@@ -36,20 +38,56 @@ class Directory {
...
+          type,
+        } = fileOrDir
+        return {
+          name,
+          path,
+          sha,
+          size,
+          content,
+          type,
         }
```
alexanderleegs, line 62: This should occur inside the `FolderType` check after the validation

alexanderleegs (changes requested):
Left minor comments

routes/folders.js
```diff
@@ -0,0 +1,25 @@
...
+const { attachRouteHandlerWrapper } = require('../middleware/routeHandler')
+
+// Import classes 
+const { Directory, FolderType } = require('../classes/Directory.js');
+
+// List pages and directories in folder
+async function listFolderContent (req, res, next) {
+    const { accessToken } = req
+    const { siteName } = req.params
+    const { path } = req.query // paths begin with _
```
gweiying, line 14: nit: from our previous discussions, we're using this route to access both collections and nested folders (can also be generalised to other image, resource folders?), so the comment here `// paths begin with _` is misleading

gweiying (commented):
I agree with Alex about moving lines 46-62 to after line 70 in `classes/Directory.js`, apart from that and some small nits, it all looks good.

routes/folders.js
kwajiehao: fixed with [c6f99d1](https://github.com/isomerpages/isomercms-backend/pull/108/commits/83f314d0dde7095a6ebb321679a5d0c719a3d0d4)

classes/File.js
```diff
@@ -3,6 +3,7 @@ const _ = require('lodash')
 const validateStatus = require('../utils/axios-utils')
 
 // Import error
+const { BadRequestError } = require('../errors/BadRequestError')
```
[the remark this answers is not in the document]
kwajiehao: fixed with [c6f99d1](https://github.com/isomerpages/isomercms-backend/pull/108/commits/83f314d0dde7095a6ebb321679a5d0c719a3d0d4)

classes/Directory.js
kwajiehao: Thanks for pointing this out - fixed with [83f314d](https://github.com/isomerpages/isomercms-backend/pull/108/commits/c6f99d1b32006316f97f364694e2e622d3222bea)

alexanderleegs (approved):
lgtm!

PR #109, opened 2021-02-17
feat: add transaction lock using DynamoDB as mutex provider

prestonlimlianjie:
This PR adds a transaction lock for multi-step operations using the `rollbackRouteHandler` using DynamoDB as described in this [design document](https://docs.google.com/document/d/1t0jUvRgCvDfes3jEGDYRoF7p42D3aoFO2deUOHnVInM/edit#).

The existing `routeHandler` is split into two types `readRouteHandler` and `writeRouteHandler`. The backend APIs that use the `readRouteHandler` only make `GET` API calls to GitHub (or do not make any API calls to GitHub), and the backend APIs using the `writeRouteHandler` make `POST/PUT/DELETE` API calls to GitHub.

This PR also adds tests for the transaction lock feature using the jest test framework.

### Environment Var Changes
This PR adds a new `MUTEX_TABLE_NAME` variable that will be used by the `rollbackRouteHandler` when not in `LOCAL_DEV` environment mode. This `MUTEX_TABLE_NAME` refers to the DynamoDB table that contains the mutexes.

There will be two `MUTEX_TABLE_NAME` tables - one for the staging+CI test environment, and one for production.

### To-dos Before Merging to Staging
Add the `MUTEX_TABLE_NAME` variable to the Elastic Beanstalk config.


utils/mutex-utils.js
```diff
@@ -0,0 +1,69 @@
...
+        repo_id: siteName
+      }
+    }
+    await docClient.delete(params).promise()
+    return logger.info(`Successfully unlocked repo ${siteName}`)
+  } catch (err) {
+    logger.error(`Failed to unlock repo ${siteName}: ${JSON.stringify(serializeError(err))}`)
+    throw new ReleaseMutexError(`Something went wrong.`)
+  }
+}
```
alexanderleegs, line 64: We could consider retrying the unlocking (using exponential backoff?), in case the unlock call fails for some reason. As discussed, we shouldn't do this for locking, since the user should just be asked to retry if the call fails.

One other possible alternative we could explore would be setting a short TTL on dynamodb, which ensures that other users can access the db after some time, even if the call to unlock fails.

.env-example
```diff
@@ -5,4 +5,5 @@ export NODE_ENV="LOCAL_DEV"
 export COOKIE_DOMAIN="localhost"
 export AUTH_TOKEN_EXPIRY_DURATION_IN_MILLISECONDS=3600000
 export JWT_SECRET=mysecretblah
-export FRONTEND_URL='http://localhost:8081'
\ No newline at end of file
+export FRONTEND_URL='http://localhost:8081'
+export MUTEX_TABLE_NAME=""
```
kwajiehao, line 9: Do we need 
```
export AWS_ACCESS_KEY_ID=""
export AWS_SECRET_ACCESS_KEY=""
``` 
to test dynamodb from a local server?

errors/MutexError.js
```diff
@@ -0,0 +1,19 @@
+// Import base error
+const { BaseIsomerError } = require('./BaseError')
+
+class GetMutexError extends BaseIsomerError {
```
kwajiehao, line 4: For the backend, we decided to implement more general errors based on HTTP status codes because we ended up with too many custom errors when working on Homer. In general, we decided to follow Google's error response format ([link](https://cloud.google.com/storage/docs/json_api/v1/status-codes)) and status response codes ([link](https://developers.google.com/maps-booking/reference/rest-api-v3/status_codes))

Also, I feel like a situation where we are unable to obtain the mutex lock should be a `409` error

utils/mutex-utils.js
```diff
@@ -0,0 +1,69 @@
...
+const { NODE_ENV, MUTEX_TABLE_NAME } = process.env
+const IS_LOCAL_DEV = NODE_ENV === 'LOCAL_DEV'
+const mockMutexObj = {}
+
+// Dynamodb constants
+const AWS_REGION_NAME = 'ap-southeast-1'
+AWS.config.update({region: AWS_REGION_NAME})
+const docClient = new AWS.DynamoDB.DocumentClient();
+
+const mockLock = (siteName) => {
```
kwajiehao, line 17: Using a local object is to store data is slightly different to how DynamoDB works - would we be open to using a local version of DynamoDB? https://medium.com/@vschroeder/install-a-local-dynamodb-development-database-on-your-machine-82dc38d59503

One issue I can see with this is that it's not straightforward to run tests for our CI/CD workflow since there aren't many GitHub actions for setting up local DynamoDB / there are issues associated with testing with the actual DynamoDB

kwajiehao (commented):
left comments

utils/mutex-utils.js
prestonlimlianjie: I meant to set it up as a hack for now (where CI would run a test against the staging DB) - but yes I'll do this instead

alexanderleegs (commented):
Actually given that we're using `attachRollbackRouteHandlerWrapper` to handle concurrency issues, we might need to change some of the existing endpoints (which still use the normal `attachRouteHandlerWrapper` because they are single step operations) - this will also help to prevent concurrency issues where user A attempts to perform a single step operation in between B's steps. 

E.g. B calls a 2-step endpoint, and the first step executes
A calls a post endpoint which modifies one of B's files
B's second step attempts to execute, but fails due to the change made, rolling back both A and B's changes

utils/mutex-utils.js
kwajiehao: Ah sorry I wasn't clear, I meant that we could do that in a subsequent PR

utils/mutex-utils.js
prestonlimlianjie: Addressed with 1dfa14c - each DynamoDB record has an `expdate` of current date + 60 seconds. However, it is important to note that DynamoDB does not delete these records immediately upon expdate.

From the [docs](https://aws.amazon.com/premiumsupport/knowledge-center/dynamodb-ttl-items-not-deleted/):
```
Wait at least for 48 hours for DynamoDB to delete the item. DynamoDB deletes expired items on a best-effort basis to be sure that there's enough throughput for other data operations. Processing takes place automatically, in the background, and doesn't affect read or write traffic to the table.
```

.env-example
prestonlimlianjie: Resolved in 50da021

errors/MutexError.js
prestonlimlianjie: Resolved in df09fce

utils/mutex-utils.js
prestonlimlianjie: ### Update
We had a discussion about this and decided that the current set up is okay:
- local dev: the mutex table is represented by a local object
- test (CI test): use the staging DynamoDB mutex table (since the records should be transient)
- staging: use the staging DynamoDB mutex table
- prod: use the prod DynamoDB mutex table

prestonlimlianjie:
> Actually given that we're using `attachRollbackRouteHandlerWrapper` to handle concurrency issues, we might need to change some of the existing endpoints (which still use the normal `attachRouteHandlerWrapper` because they are single step operations) - this will also help to prevent concurrency issues where user A attempts to perform a single step operation in between B's steps.
> 
> E.g. B calls a 2-step endpoint, and the first step executes
> A calls a post endpoint which modifies one of B's files
> B's second step attempts to execute, but fails due to the change made, rolling back both A and B's changes

Addressed with 4fd4f43

middleware/routeHandler.js
```diff
@@ -1,34 +1,53 @@
...
+// Used when there are no write API calls to the repo on GitHub
+const attachReadRouteHandlerWrapper = (routeHandler) => async (req, res, next) => {
   routeHandler(req, res).catch((err) => {
     next(err)
   })
 }
 
+// Used when there are write API calls to the repo on GitHub
+const attachWriteRouteHandlerWrapper = (routeHandler) => async (req, res, next) => {
+  await lock(siteName)
```
alexanderleegs, line 15: We need to retrieve the siteName from req.params

alexanderleegs (approved):
lgtm!

PR #110, opened 2021-02-18
feat: sends user id through redirect upon successful login

gweiying:
This PR adds functionality to send the username to the frontend for display when users successfully log in.


alexanderleegs (approved):
lgtm

PR #121, opened 2021-03-04
fix: unlock site after successful API call for rollback route handler

prestonlimlianjie:
@gweiying pointed out a bug in the transaction lock feature as well as a fix for it.

The bug was that a successful API call that used the rollback route handler did not unlock the repo because the handler was missing an `await unlock()` call after success.


gweiying (approved):
lgtm

gweiying:
merging to fix transaction lock bug

PR #123, opened 2021-03-04
refactor/directory content retrieval endpoint

kwajiehao:
## Overview

This PR refines the endpoint for generalized directory content retrieval. With the changes introduced, one should be able to ping this endpoint with a given repo `path` and receive a list of files and sub-directories in that directory.

This PR makes two changes:
- Move the route into a separate `directory` route file, since the endpoint is a generalized one and not specific to just `folders`
- Refactor the endpoint to receive a `path` path param (which takes the form of an encoded URI component) instead of a query param instead

With this new endpoint, we will be able to start work on the nested images/files problem, where our images and files endpoints are unable to retrieve the files in nested sub-directories because our current APIs are limited to search only in specifically-named repo folders.

Note: the endpoint that was refactored was not in use, so the changes made in this PR are not breaking changes.


alexanderleegs (approved):
lgtm

Issue #116, opened 2021-03-01
Add page moving endpoints


alexanderleegs:
Currently, most of our page moving logic is handled by our frontend - the frontend handles the creation and deletion of files to simulate page moving. This is problematic for multiple reasons:

- It makes our frontend code unnecessarily convoluted
- Our rollback protection does not work as it should because multiple endpoints are being called - if something fails halfway, the user's repo will be left in an intermediate state

We should abstract away this functionality to the backend instead, by creating a new endpoint to move each file type.

Issue #120, opened 2021-03-03
Fix Collection.list()


gweiying:
Since the collections data is no longer stored in the `_config.yml` file, the class method `Collection.list()` no longer works. However, to replace the functionality of retrieving all collection names will require extensive changes to all routes and classes that use this. As part of this fix, we should check through which routes and classes we'd actually need to refactor, and which will be made redundant with the move to v4.

PR #124, opened 2021-03-08
Feat/move files

alexanderleegs:
This PR introduces new endpoints to handle the moving of multiple files from one folder to another. To be reviewed in conjunction with PR #[369](https://github.com/isomerpages/isomercms-frontend/pull/369) on the isomercms-frontend repo.

In addition, this PR also updates the existing list() method of collections to work with both the existing format as well as the file format post jekyll 4.0 update.

Resolves issues https://github.com/isomerpages/isomercms-backend/issues/116 and https://github.com/isomerpages/isomercms-backend/issues/120

### DIFF
--- a/routes/collections.js
+++ b/routes/collections.js
@@ -1,6 +1,8 @@
 const express = require('express');
 const router = express.Router();
 const Bluebird = require('bluebird')
+const yaml = require('js-yaml');
+const base64 = require('base-64');
 
 // Import middleware
 const { 
@@ -12,6 +14,8 @@ const {
 const { Collection } = require('../classes/Collection.js');
 const { CollectionConfig } = require('../classes/Config.js');
 const { File, CollectionPageType, PageType } = require('../classes/File');
+
+const { deslugifyCollectionName } = require('../utils/utils')
 
 // List collections
 async function listCollections (req, res, next) {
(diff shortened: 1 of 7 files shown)


routes/collections.js
```diff
@@ -61,9 +64,54 @@ async function renameCollection (req, res, next) {
...
+  const collectionSubfolderName = processedCollectionPathTokens[1]
+  const processedTargetPathTokens = decodeURIComponent(targetPath).split('/')
+  const targetCollectionName = processedTargetPathTokens[0]
+  const targetSubfolderName = processedTargetPathTokens[1]
+
+  const IsomerCollection = new Collection(accessToken, siteName)
+  const collections = await IsomerCollection.list()
+
+  // Check if collection already exists
+  if (!collections.includes(targetCollectionName) && targetCollectionName !== 'pages') {
```
kwajiehao, line 83: Also check that target collection name is not one of the `ISOMER_TEMPLATE_DIRS`

kwajiehao (changes requested):
Left a comment. 

Separately, I wonder if removing and adding to the collection config should be abstracted away into the methods of the `File` type. Will follow up offline

kwajiehao (approved):
lgtm

PR #126, opened 2021-03-09
fix: throw NotFoundError if given path cannot be found on GitHub

kwajiehao:
## Overview

This PR fixes a minor bug in the `Directory` class' `list` method. When the path that we attempt to read from in the GET `/:siteName/files/:path` endpoint does not actually exist on GitHub, we currently throw a `400` error. Instead, we should be re-throwing the `404` error that GitHub responds with.


alexanderleegs (approved):
lgtm

PR #127, opened 2021-03-11
Add linting and formatting tools

kwajiehao:
## Overview

This PR adds **eslint** as a linting tool and **Prettier** as a code formatting tool for the repo. Why do we need both? [Linters help us to catch bugs, and code formatting helps us to maintain consistency in our code](https://prettier.io/docs/en/comparison.html), allowing us to avoid `nit` commits and removing cognitive overhead (here's [another source](https://blog.theodo.com/2019/08/why-you-should-use-eslint-prettier-and-editorconfig-together/) on why you should use eslint and Prettier together). Already, there are several parts of our codebase which are inconsistent and confusing (usage of semi-colons, using tabs vs. spaces, etc.)

Note that the PR **only introduces these tools** to the repo. Actual linting and formatting fixes will be done in a subsequent PR, as that is expected to be a more substantial task which requires careful review.

We will use `eslint-config-prettier` to integrate Prettier into our eslint configuration - what this does is that it will switch off the eslint rules related to code formatting which Prettier takes care of. What this means is that our code linting and formatting process is a two-step one:
1. Run eslint (`npx eslint .`) to surface linting errors, fix bugs caught by the linter
2. Run Prettier to fix code formatting issues (`npx prettier --write .`)

#### Instructions for setting up linting and code formatting tools
1. Install eslint ([link](https://eslint.org/docs/user-guide/getting-started))
        - Initialize eslint for the repo. Select `airbnb-base` as the default theme to extend
        - Add `.eslintIgnore`, which is the same as the `.gitignore`. `.eslintignore` also ignores the GitHub workflow file
2. Install Prettier ([link](https://prettier.io/docs/en/install.html))
        - Add Prettier options into config file
        - Add `.prettierignore`, which is the same as `.eslintIgnore`
3. Integrate Prettier with eslint by installing `eslint-config-prettier` and `eslint-plugin-prettier` 

#### Linter configuration
Our eslint configuration extends the `airbnb-base` config. This is based off AirBnB's Javascript style guide ([link](https://github.com/airbnb/javascript)).

#### Code formatting details
After consulting @alexanderleegs and @gweiying, we agreed on a few aspects of code formatting, which can be summed up in our Prettier options config:

```
semi: false, // do not use semi-colons in our repo
singleQuote: false, // use double quotes for strings
useTabs: false // use spaces instead of tabs for indentation
```

#### Other notes:
- Prettier is updated frequently, so we have followed the Prettier documentation's instructions to use an exact version of Prettier in our `package.json`
- Shortcuts for linting and formatting have been introduced using `npm run lint` (lints the codebase using our eslint configs), `npm run format` (formats the codebase using our Prettier config options), and `npm run check-format` (checks the codebase for whether Prettier has been run).

## Instructions for testing

Since we've established several baselines, to test for whether the linter works, we can start with simple checks for the rules we've agreed on (semi-colons, double quotes, and using spaces instead of tabs) by manually creating our own test cases (for e.g. adding semi-colons to the end of some lines) and checking if the two-step process of linting and then formatting with Prettier gets the code to the desired end state (in the case of the example, removing the semi-colon).

## Background reading

For more information on the topic, the following links might be useful:
- "Why coding style matters", from the author of eslint ([link](https://www.smashingmagazine.com/2012/10/why-coding-style-matters/))
- Why we need code formatting tools - "A Prettier Printer" ([link](https://www.youtube.com/watch?v=hkfBvpEfWdA))
- How code formatting tools work under the hood - "Javascript Code Formatting" ([link](https://www.youtube.com/watch?v=0Q4kUNx85_4))

## Work to be done in subsequent PRs

1. Run the linter and code formatter, fix code (#128)
2. Write a pre-commit hook which helps to lint and format our code before committing (#129)
3. Add a formatting check in our CI/CD flow (running `npx prettier --check .`) (#130)


prestonlimlianjie (approved):
lgtm

PR #132, opened 2021-03-15
Fix/delete collection conflict

kwajiehao:
## Overview

Currently, when we attempt to delete an entire collection/folder using the `delete` method in the `Collection` class, the operation fails and GitHub returns a `409` error response when attempting to delete the files from the collection/folder. This PR fixes the `Collection` class' `delete` method by using updating the root Git trees to delete the entire collection/folder instead of deleting pages one-by-one.

Other changes introduced in this PR:
-  Refactor the `rename` method of the `Collection` class to reuse the `currentCommitSha` and `treeSha` obtained
when using the `attachRollbackRouteHandlerWrapper` function

### Explaining the `409` error response from GitHub

The source of the `409` error responses lies in our attempt to delete files from the collection/folder **concurrently** - when we use `Bluebird.each` instead of `Bluebird.map`, the deletion goes through with no issues. To explain why concurrent deletion is a problem, we need to understand the mechanism for file deletion on Git. 

When a file is deleted from a directory, a new tree object (sans the deleted file) is created, along with a new tree SHA. The parent tree then updates its records for the directory to point away from the existing tree SHA to this new tree SHA. The conflict arises because when we try to delete files concurrently because the outcome, or the resulting tree SHA of the directory, after each of the delete operations is different. Consider the following example:

```
base commit: abcde
- example_directory (tree SHA: 12345)
     - example_page_1
     - example_page_2

directory tree SHA after deleting example_page_1:
base commit: bcdef
- example_directory (tree SHA: 23456)
     - example_page_2

directory tree SHA after deleting example_page_2:
base commit: cdefg
- example_directory (tree SHA: 34567)
     - example_page_1
```

Here, we try to delete both `example_page_1` and `example_page_2` at the same time. The problem is that both deletion operations start from the same base commit, but produce conflicting Git trees. If the deletion of `example_page_1` goes through first, the base commit moves from `abcde` to `bcdef`. But when the deletion of `example_page_2` goes through, we can't go from the new base commit `bcdef` to `cdefg`, because the parent commit of `cdefg` is `abcde`.

### DIFF
--- a/classes/Collection.js
+++ b/classes/Collection.js
@@ -70,12 +70,14 @@ class Collection {
     }
   }
 
-  async delete(collectionName) {
+  async delete(collectionName, currentCommitSha, treeSha) {
     try {
-      // Delete collection config
-      const collectionConfig = new CollectionConfig(this.accessToken, this.siteName, collectionName)
-      const { sha } = await collectionConfig.read()
-      await collectionConfig.delete(sha)
+      const commitMessage = `Delete collection ${collectionName}`
+      const gitTree = await getTree(this.siteName, this.accessToken, treeSha)
+      const newGitTree = gitTree.filter(item => {
+        if (item.path !== `_${collectionName}`) return item
+      })
+      await sendTree(newGitTree, currentCommitSha, this.siteName, this.accessToken, commitMessage)
 
       // Delete collection in nav if it exists
       const nav = new File(this.accessToken, this.siteName)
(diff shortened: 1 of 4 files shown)


alexanderleegs (changes requested):
Good call on attaching the commit and tree sha to the request. One small issue - `deleteCollectionPage` (in `routes/collectionPages`) also calls Collection.delete and needs to be updated.

Side note: the `if` block to delete the collection in `deleteCollectionPage` is currently never entered because the `collection.yml` file always exists, but we can fix that issue in a later PR

kwajiehao:
> Good call on attaching the commit and tree sha to the request. One small issue - `deleteCollectionPage` (in `routes/collectionPages`) also calls Collection.delete and needs to be updated.
> 
> Side note: the `if` block to delete the collection in `deleteCollectionPage` is currently never entered because the `collection.yml` file always exists, but we can fix that issue in a later PR

Fixed with [055ad40](https://github.com/isomerpages/isomercms-backend/pull/132/commits/055ad4091c5f1983770b546b9a6e491d714c42c3)

alexanderleegs (approved):
lgtm

PR #133, opened 2021-03-16
Feat/update pageroutes

gweiying:
This PR updates the `pages` and `collectionPages` `read`, `update`, `delete` and `rename` routes so that all path names are received and parsed as encodedURIComponents, and Base64 encoding happens only on the backend. This PR extends the changes introduced by Alex in #131.

For `collectionPages` `delete` and `rename`, the collection config is also updated accordingly to remove/ rename the modified pages.

This PR also updates the `CollectionConfig`'s `addItemToOrder`, `deleteItemFromOrder`, `updateItemInOrder` and `read` methods to make the class more reusable.


classes/Config.js
```diff
@@ -116,22 +116,28 @@ class CollectionConfig extends Config {
...
-      index = contentObject.collections[collectionName].order.length
+    return { contentObject, sha }
+  }
+
+  async addItemToOrder(item, index) {
+    const collectionName = this.collectionName
+    const { contentObject, sha } = await this.read()
+
+    if (index === undefined) {
+      let index
```
kwajiehao, line 130: Can we use a different name instead of `index`? Since this clashes with the method argument

classes/Config.js
```diff
@@ -141,15 +147,18 @@ class CollectionConfig extends Config {
...
+    const { contentObject, sha } = await this.read()
     const index = contentObject.collections[collectionName].order.indexOf(item);
     contentObject.collections[collectionName].order.splice(index, 1)
     const newContent = base64.encode(yaml.safeDump(contentObject))
     
     await this.update(newContent, sha)
+    return { index, item }
+  }
+
+  async updateItemInOrder(oldItem, newItem) {
```
kwajiehao, line 159: It's nice to compose the `updateItemInOrder` from the other methods, but this results in two API calls for what is essentially one update. Have we considered writing specific "update" code so only one API call is made?

routes/pages.js
```diff
@@ -63,24 +63,6 @@ async function listPages (req, res, next) {
   res.status(200).json({ pages })
 }
 
-// Create new page
```
kwajiehao, line 66: Nit: could we also remove unused imports for this file?

kwajiehao (commented):
looks good for the most part, mostly had a question on how the `updateItemInOrder` method  for the `CollectionConfig` class works

alexanderleegs (approved):
lgtm!

As discussed offline, we might want to move the encoding/decoding into `File` instead, but we'll do this later on when cleaning up

classes/Config.js
kwajiehao: @gweiying brought up the possibility of building out the methods for the `CollectionConfig` class but as that would be a more significant undertaking we will be postponing that for now. As discussed, we will modify the `updateItemInOrder` method so that the the directory config file is read only once, and custom updating logic is applied instead of composing it with the `add` and `delete` methods

classes/Config.js
kwajiehao: As discussed offline, we'll leave it as it is since there aren't any errors

routes/pages.js
gweiying: 500f067

classes/Config.js
gweiying: e4f835a

kwajiehao (approved):
lgtm

PR #134, opened 2021-03-18
Fix/collectionconfig read method

kwajiehao:
This PR fixes a minor bug introduced in PR #133. 

The `Config` class' `read` method used to return an object with attributes `content` and `sha`, but with PR #133, the `CollectionConfig` class had a new `read` method which overrode the parent class' read method - this new `read` method returns an object with attributes `contentObject` and `sha`. 

This API change breaks the codebase in two parts:
1. `Collection` class' `rename` method which attempts to extract `content` after reading the `CollectionConfig`
2. `listAllFolderContent` route function, which does the same

The proposed fix is to return a `content` attribute instead of `contentObject` attribute in the `read` method of `CollectionConfig`

### DIFF
--- a/classes/Collection.js
+++ b/classes/Collection.js
@@ -143,8 +143,7 @@ class Collection {
 
       // Update collection.yml in newCollection with newCollection name
       const collectionConfig = new CollectionConfig(this.accessToken, this.siteName, newCollectionName)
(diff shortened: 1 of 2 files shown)


gweiying (approved):
good catch, lgtm!

PR #135, opened 2021-03-18
fix/CollectionConfig `read` method output

kwajiehao:
PR #134 missed out on instances which attempted to destructure `contentObject` from the `read` method of the `CollectionConfig`, when we've actually changed the method to return `content` attribute. This PR fixes that.


gweiying (approved):
lgtm

PR #141, opened 2021-03-29
Fix: position of newly created files

alexanderleegs:
This PR fixes a small bug when adding files to the order in `collection.yml`. Previously, as the variable `index` was not scoped properly, we always spliced newly created files into the first position in the `collection.yml` file. We have fixed the scoping in this PR.


gweiying (approved):
lgtm

PR #142, opened 2021-04-05
Feat: add staging url and last updated endpoints

alexanderleegs:
This PR adds in 2 new endpoints, to retrieve the url of the staging site and the last updated time of the repo. To be reviewed in conjunction with PR #[394](https://github.com/isomerpages/isomercms-frontend/pull/394) on the isomercms-frontend.

We are adding an additional last updated endpoint as the existing one being used (`getSites`) retrieves the details of every single repo.

For the endpoint being used to retrieve the staging url, we currently retrieve this using the repo description, as the staging url is automatically populated upon creation of a site. However, we intend to replace this, as it is possible for users to modify the initial description. We can consider adding the staging/production urls into the `_config.yml` file, or in the long term, adding these details to a dynamodb for us to more easily reference the links to staging and production sites.

### DIFF
--- a/routes/sites.js
+++ b/routes/sites.js
@@ -118,7 +118,35 @@ async function checkHasAccess (req, res, next) {
   }
 }
 
+/* Gets the link to the staging site for a repo. */
+async function getStagingUrl (req, res, next) {
+  // TODO: reconsider how we can retrieve url - we can store this in _config.yml or a dynamodb
+  const { accessToken } = req
+  const { siteName } = req.params
+  
+  const endpoint = `https://api.github.com/repos/${ISOMER_GITHUB_ORG_NAME}/${siteName}`
+  const resp = await axios.get(endpoint, {
+    headers: {
+      Authorization: `token ${accessToken}`,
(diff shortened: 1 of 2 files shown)


routes/sites.js
```diff
@@ -118,7 +118,52 @@ async function checkHasAccess (req, res, next) {
...
+  })
+
+  const { description } = resp.data
+
+  let stagingUrl
+
+  if (description) {
+    // Retrieve the url from the description - repo descriptions have varying formats, so we look for the first link
+    const descTokens = description.replace('/;/g', ' ').split(' ')
+    stagingUrl = descTokens.find(token => token.includes('http'))
```
gweiying, line 158: would it be good to also check for `token.includes("staging")`?

alexanderleegs: Great point, resolved in https://github.com/isomerpages/isomercms-backend/pull/142/commits/4e208f22549187f36c40f878a1858e30667eafb9

gweiying (approved):
lgtm

Issue #143, opened 2021-04-06
Retrieve netlify.toml from isomer-build


gweiying:
As part of the Isomer Jekyll gem changes, we've decided to host a central netlify.toml file at [isomer-build](https://github.com/opengovsg/isomer-build) which is downloaded and used at build time on Netlify. Hence, we should modify our backend to retrieve the netlify.toml from the centralised repo [isomer-build](https://github.com/opengovsg/isomer-build)

PR #144, opened 2021-04-06
Feat: retrieve netlify.toml from isomer build

alexanderleegs:
This PR updates the retrieval of the netlify.toml file to be taken from our central isomer-build repo, to resolve https://github.com/isomerpages/isomercms-backend/issues/143.

To do before merging: ~~update new env vars on AWS on both staging and prod~~

### DIFF
--- a/.env-example
+++ b/.env-example
@@ -6,6 +6,9 @@ export COOKIE_DOMAIN="localhost"
 export AUTH_TOKEN_EXPIRY_DURATION_IN_MILLISECONDS=3600000
 export JWT_SECRET=mysecretblah
 export FRONTEND_URL='http://localhost:8081'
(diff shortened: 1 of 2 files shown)


.env-example
```diff
@@ -6,6 +6,9 @@ export COOKIE_DOMAIN="localhost"
 export AUTH_TOKEN_EXPIRY_DURATION_IN_MILLISECONDS=3600000
 export JWT_SECRET=mysecretblah
 export FRONTEND_URL='http://localhost:8081'
+export GITHUB_ORG_NAME="isomerpages"
```
gweiying, line 9: Where is this used?

alexanderleegs: It's an existing env var - it just wasn't in env-example previously

gweiying (approved):
lgtm

PR #149, opened 2021-04-12
Fix: move auth out of query param

alexanderleegs:
This PR fixes our retrieval of the github access token, in line with github's deprecation of auth through query params ([link](https://developer.github.com/changes/2020-02-10-deprecating-auth-through-query-param/)). We instead specify the username and password in a separate auth object.


gweiying (approved):
thanks for the quick fix! lgtm

PR #150, opened 2021-04-12
Fix: retrieve folders with collection.yml only

alexanderleegs:
This PR updates the get all folder content endpoint to only retrieve information from folders which contain a `collection.yml` file, while other folders are ignored.


gweiying (approved):
lgtm!

PR #153, opened 2021-04-14
fix: add resourceRoomName to readResourcePage response

gweiying:
This PR sends the `resourceRoomName` as part of the API response when a resource page is retrieved.


alexanderleegs (approved):
lgtm

PR #154, opened 2021-04-15
Fix/image endpoints for nested images

kwajiehao:
## Overview

Currently, our backend endpoints for images cannot handle changes to images in nested directories. That's because the endpoints that we're hitting on GitHub assume that there is only one directory where we store all images: the `images` directory in the root of the repo.

This PR updates the following endpoints:
- POST `/:siteName/images`
- GET `/:siteName/images/:imageName`
- POST `/:siteName/images/:imageName/rename/:newImageName`
So that we can read, create, and rename individual images that are in nested directories. I also updated these endpoints to use the `MediaFile` (renamed from `ImageFile`) class instead of the `File` class that was being used before.

As part of this PR, changes were also made to the `MediaFile` class so that the `ImageType` class used accepts an argument in its constructor which determines the base API endpoint. This allows us to set the base endpoint to be a nested directory instead of the root `images` directory.

### DIFF
--- a/classes/ImageFile.js
+++ b/classes/ImageFile.js
@@ -17,8 +17,8 @@ class ImageFile {
     this.fileType = null
   }
 
-  setFileTypeToImage() {
-    this.fileType = new ImageType()
+  setFileTypeToImage(directory) {
+    this.fileType = new ImageType(directory)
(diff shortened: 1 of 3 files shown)


alexanderleegs (approved):
lgtm

Issue #129, opened 2021-03-11
Pre-commit hook which runs linter and code formatter


kwajiehao:
We should introduce a pre-commit hook which runs the linter and code formatter which prevents someone from committing to the repo if there are linter/formatting errors

PR #155, opened 2021-04-19
chore: add husky and commitlint

prestonlimlianjie:
This PR adds the `husky` library which enables us to run git hooks at various phases (e.g. `pre-commit`, `commit-msg`, `pre-push`). This PR uses `husky` to enforce commit message formats using the `commitlint` library (the format is specified in `commitlint.config.js`).

This PR partially addresses https://github.com/isomerpages/isomercms-backend/issues/129. A subsequent PR will be made to add the linter as well as the linting husky hook.


kwajiehao:
What is the rationale for having hooks at both the `commit-msg` stage as well as `pre-push` since technically all commits to be pushed would already have been linted at the `commit-msg` stage. Reproducing the instructions for installing husky by the commitlint team ([link](https://github.com/conventional-changelog/commitlint)) results in only the `commit-msg` hook.

PR #163, opened 2021-04-23
(hotfix to develop) misc bugs

gweiying:
This PR changes the library used for encoding on the backend, standardising it on the `Base64` library. The other library used, `base-64` does not support encoding foreign language, hence users were experiencing difficulty saving resource files containing foreign language characters, as they were being encoded with `base-64`. 

Error message on backend logs:
Screenshot 2021-04-22 at 6 13 32 PM

Separately, this PR also fixes a bug where page content is not added to files moved to/ out of subfolders, which causes pages to appear blank after moving out of/ to a subfolder.

### DIFF
--- a/classes/Collection.js
+++ b/classes/Collection.js
@@ -1,5 +1,4 @@
 const yaml = require('yaml')
-const base64 = require('base-64')
 const Bluebird = require('bluebird')
 const _ = require('lodash')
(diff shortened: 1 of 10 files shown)


alexanderleegs (approved):
lgtm, good catch

PR #166, opened 2021-04-27
Feat new endpoints for media subfolders

alexanderleegs:
This PR adds new endpoints for creating, deleting, and renaming subfolders in the images folder. To be reviewed in conjunction with PR #[433](https://github.com/isomerpages/isomercms-frontend/pull/433) on the isomercms frontend repo.

The create endpoint currently only creates the placeholder file in the subfolder.
The delete/rename endpoints involve recursively retrieving the current git tree and replacing the appropriate files - this operation is slow for sites which have a large number of files. However, there doesn't seem to be an easy way to retrieve the specific tree in order to make this change, since the depth of image subfolders is unbounded. It would be great if we could find a better way to do this.


gweiying (approved):
lgtm!

PR #169, opened 2021-05-03
Feat: add endpoints for file subfolders

alexanderleegs:
This PR modifies the existing `MediaFile` and `MediaSubfolder` classse for document subfolder creation, deletion and renaming. It also updates the existing documents endpoints to use these classes. To be reviewed in conjunction with PR#[442](https://github.com/isomerpages/isomercms-frontend/pull/442) on the isomercms-frontend repo.

### DIFF
--- a/classes/MediaSubfolder.js
+++ b/classes/MediaSubfolder.js
@@ -1,6 +1,6 @@
 const _ = require('lodash')
 
-const { File, ImageType } = require('./File.js')
+const { File, ImageType, DocumentType } = require('./File.js')
 const { getTree, sendTree } = require('../utils/utils.js')
 
 class MediaSubfolder {
@@ -10,10 +10,17 @@ class MediaSubfolder {
     switch (fileType) {
       case 'images':
         this.fileType = new ImageType()
+        this.mediaFolderName = fileType
+        break
+      case 'documents':
+        this.fileType = new DocumentType()
+        this.mediaFolderName = 'files'
+        break
       default:
         this.fileType = new ImageType()
+        this.mediaFolderName = fileType
+        break
     }
-    this.mediaFolderName = fileType
   }
 
   async create(subfolderPath) {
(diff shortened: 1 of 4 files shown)


classes/MediaSubfolder.js
```diff
@@ -10,10 +10,17 @@ class MediaSubfolder {
...
+        this.mediaFolderName = fileType
+        break
+      case 'documents':
+        this.fileType = new DocumentType()
+        this.mediaFolderName = 'files'
+        break
       default:
         this.fileType = new ImageType()
+        this.mediaFolderName = fileType
+        break
```
gweiying, line 22: nit: is there a reason for defaulting to `ImageType`? Maybe we could throw an error or not have a default case, as this feels like this leaves an area for misuse that might cause downstream errors.

routes/documents.js
```diff
@@ -3,11 +3,29 @@ const router = express.Router();
...
 
+const extractDirectoryAndFileName = (documentName) => {
+  let documentDirectory, documentFileName
+
+  const pathArr = documentName.split('/')
+  if (pathArr.length === 1) {
+    documentDirectory = 'files'
+    documentFileName = documentName
+  } else if (pathArr.length > 1) {
+    documentDirectory = `files/${pathArr.slice(0, -1)}`
```
gweiying, line 20: nit: can we leave a short comment what `pathArr` contains and why we need `pathArr.slice(0,-1)`

gweiying:
Nice job, mostly lgtm, just couple of small nits!

classes/MediaSubfolder.js
alexanderleegs: Good point, I've made it throw an error instead for an unknown type - but have left it as a generic error, since this is mostly for our debugging purposes - https://github.com/isomerpages/isomercms-backend/pull/169/commits/0ae5456007b0866195e7ba2ba16e44e830f3e477

routes/documents.js
alexanderleegs: Added comments in https://github.com/isomerpages/isomercms-backend/pull/169/commits/bf0a6684e8af0543ddc2cde08a9f068b6e0c5261

PR #175, opened 2021-05-12
fix: handles media folders not found error

gweiying:
Currently, if a repo has no images and files folder, a 404 is thrown and the frontend is redirected. This PR catches the 404 and sends an empty directoryContent response to the frontend.


alexanderleegs (approved):
lgtm! Thanks for the quick fix

Issue #177, opened 2021-05-20
[bug] Title in index.html when renaming resource category


alexanderleegs:
Currently, we do not rename the title of the index.html file when renaming resource category, instead on changing the file paths of every file in the resource category to be renamed. However, this results in an inconsistent breadcrumb:
![image](https://user-images.githubusercontent.com/22111124/118951690-e21cd600-b98d-11eb-95bc-abd45a02fcde.png)

We should update the index.html when renaming resource category and resource room.

PR #179, opened 2021-05-21
Fix/modify title in index file for resources

alexanderleegs:
This PR resolves https://github.com/isomerpages/isomercms-backend/issues/177. Currently, newly created resource categories/resource room always have their title set to `Resource Room`, which makes the breadcrumb inconsistent with the title of the page. This PR fixes that by properly updating the title on resource room and resource category creation and renaming.


gweiying (approved):
lgtm!

PR #180, opened 2021-05-21
Fix/Return isPrivate property of repos in sites/route

rlyss:
Allow front end to access isPrivate property of repo when calling sites/
Facilitate checking of whether raw image urls in public repos can be used instead of retrieving blob from backend


prestonlimlianjie (approved):
lgtm!

PR #184, opened 2021-06-01
Fix: remove empty footer settings

alexanderleegs:
This PR fixes an issue with the saving of empty fields. Currently, our template checks for the existence of fields in order to determine whether a certain part should be rendered, e.g. it checks for the existence of `feedback` to determine whether to render the feedback blurb inside contact-us. However, we are currently saving these fields as an empty string, which is interpreted as a valid link by our template. This PR fixes this by removing the relevant fields in the `footer.yml` file completely if they are updated to an empty string. 

Also, this PR additionally returns the value of the linkedin-insights for the settings endpoint.

To be reviewed in conjunction with PR#[496](https://github.com/isomerpages/isomercms-frontend/pull/496) on the isomercms frontend repo. Resolves https://github.com/isomerpages/isomercms-frontend/issues/486.

### DIFF
--- a/classes/Settings.js
+++ b/classes/Settings.js
@@ -145,9 +145,29 @@ class Settings {
     }
 
     if (!_.isEmpty(footerSettings)) {
+      // We want to remove empty footer settings entirely so they don't show up in the actual site
+      const clonedFooterSettings = _.cloneDeep(footerSettings)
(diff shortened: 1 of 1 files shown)


gweiying (approved):
lgtm!

PR #188, opened 2021-06-07
Feat: introduce aliases

kwajiehao:
## Overview

This PR accomplishes two things:
- Introduce aliasing, so that instead of importing `../<folder>/<file name>`, we can import `@<folder>/<file name>`
- Use aliasing to sort imports/requires using eslint

The reason we are using the eslint `eslint/import` plugin instead of a prettier plugin is because the existing prettier plugins for sorting (https://github.com/trivago/prettier-plugin-sort-imports) introduced several bugs which we were unable to diagnose.


kwajiehao:
Closing this PR because the rebase was extremely messy, and it's clearer to re-implement the features in a new branch

PR #194, opened 2021-06-07
style: introduce aliases and import sorting

kwajiehao:
# Overview

This PR accomplishes the following things:
- Introduce aliases, so that local dependency requires are simplified
- Defining eslint configs that sort requires according to specified order
- Using eslint to actually sort the requires in each file according to the defined order

## Main changes
This PR contains lots of code changes, firstly because we replaced the usual relative path requires with alias requires, and secondly because we used eslint to sort the requires in every file. There are 3 main files which contain the key changes:
- `bin/www`: here, we import the `module-alias/register` in the root of the app for aliasing to work
- `package.json`: we install `eslint-import-resolver-alias` as a dev dependency, so that eslint can find the module aliases
- `eslintrc.json`: we add additional configuration for eslint. Note the addition of the `import` plugin, the new rules introduced for use with the `import` plugin, as well as the new settings for the import resolver

## Module aliases

Aliases were defined using the `module-alias` npm package. This allows us to define aliases for commonly accessed directories and paths. In the case of this repo, we defined aliases for all code-containing top-level directories. With aliases, we can now require like this: `@classes/Collections.js` instead of `../classes/Collections.js`. This makes the code easier to read.

## Sorting with eslint

Sorting requires in each file depends on the `eslint-plugin-import` [plugin](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/order.md). We had previously played with implementing require sorting through Prettier, but we faced bugs with the most commonly used library, [@trivago/prettier-plugin-sort-imports](https://github.com/trivago/prettier-plugin-sort-imports). As such, we decided to sort requires using eslint.

In our config, we defined rules for each of our aliases relative to each other, requiring them in the following order: 
- @logger
- @error
- @middleware
- @classes
- @routes
- @utils

Requires would be sort according to the following array, which we specified in our eslint config file: ["builtin", "external", "internal", "parent", "sibling", "index", "object"].  In practice, this means that requires would be sorted in the following order:
- built-in packages 
- external packages
- our aliases (whose relative orders are defined above)
- parent and sibling paths


alexanderleegs (approved):
lgtm

Issue #196, opened 2021-06-08
[bug] List directory content route returns 404 if directory is empty


prestonlimlianjie:
### Description
Calling the API endpoint with the following params (`/v1/sites/e2e-test-repo/files/files`) resulted in a 404 error. The expected behavior should be that the endpoint returns a 200 response along with an empty `directoryContents` array like so
```
{
    "directoryContents": []
}
```

This functionality was fine in `58ddd61` but was broken during the linting commits.

PR #197, opened 2021-06-08
fix: address bug in directory route

gweiying:
This is a temporary fix for #196, wrapping the `IsomerDirectory.list()` method in a try-catch to prevent a 404 error from bubbling up when the directory does not exist. 

This should be fixed in the upcoming refactor in one of the following ways:
1) `IsomerDirectory.list()` should either return an empty array when it does not exist
2) The frontend should not expect to receive an empty array if the directory does not exist


prestonlimlianjie (approved):
lgtm

PR #198, opened 2021-06-08
Feat/add husky pre commit hooks

alexanderleegs:
This PR adds a pre-commit hook to run eslint and prettier before commit. It installs the husky and lint-staged packages to support this behaviour.


alexanderleegs:
Superceded by https://github.com/isomerpages/isomercms-backend/pull/203

Issue #185, opened 2021-06-02
updating third nav folder name does not update third nav title name


lisatjide:
when a user rename their third nav folder on cms, only the folder on github is renamed but the front matter `third_nav_title` on each page did not update. this results in their navigation not being updated. 

**_Folder name on CMS_**
<img width="678" alt="Screenshot 2021-06-02 at 7 41 17 PM" src="https://user-images.githubusercontent.com/23736580/120474479-d9d78880-c3da-11eb-83d9-927b6e21f116.png">

**_Folder name on Github and front matter of the page_**
<img width="239" alt="Screenshot 2021-06-02 at 7 40 55 PM" src="https://user-images.githubusercontent.com/23736580/120474492-dba14c00-c3da-11eb-9d50-41a816958e94.png">

PR #200, opened 2021-06-15
fix: modify third nav title when renaming subfolder

prestonlimlianjie:
This PR fixes https://github.com/isomerpages/isomercms-backend/issues/185.

### Problem
The current `renameSubfolder` route method does the following:
- Rename all files from `${folder}/${previousSubfolder}/${fileName}` to `${folder}/${newSubfolder}/${fileName}` using git trees
- Modify the `collection.yml` file within the collection folder

However, this neglects to modify the `third_nav_title` fields within the collection subfolder files.

### Fix
This PR fixes the abovementioned problem by taking the following approach:
- For each file in the `${folder}/${previousSubfolder}` folder
  - Create the new content for each file by modifying the `third_nav_title` field in each retrieved file
  - Save the new content in the new `${folder}/${newSubfolder}` folder
- Modify the `collection.yml` file within the collection folder

### Note
This fix - while functional - is not ideal because it just adds more business logic to the routes layer. We will clean this up in the upcoming backend refactor.

### DIFF
--- a/routes/folders.js
+++ b/routes/folders.js
@@ -12,7 +12,7 @@ const { Collection } = require("@classes/Collection")
 const { CollectionConfig } = require("@classes/Config")
 const { File, CollectionPageType } = require("@classes/File")
 
-const { getTree, sendTree } = require("@utils/utils.js")
+const { getTree, sendTree, deslugifyCollectionName } = require("@utils/utils.js")
 
 
 const router = express.Router()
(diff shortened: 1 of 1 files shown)


kwajiehao (approved):
Lgtm

As discussed offline, we will need to revert the effects of this change when we eventually modify our templates to use `collections.yml` instead of each individual page's `third_nav_title` front matter attributed. Issue #202 was created to track this

PR #201, opened 2021-06-15
fix: update nav when renaming collection

prestonlimlianjie:
This PR fixes https://github.com/isomerpages/isomercms-frontend/issues/501.

### Problem
When a user renames a collection in the Workspace, they should expect the corresponding navigation menu element to be updated. The curious thing was that our `Collection.rename` method already contained code to perform the nav update.

However, the issue was the sequence in which the code in the method was run. Here is a walkthrough of the existing buggy code:
- The rename method takes in several parameters, including `oldCollectionName`, `newCollectionName,` and `treeSha`. The `treeSha` is important since we use git trees to move files within the collection folder.
- Step 1 - we update the `navigation.yml` file by replacing `oldCollectionName` with `newCollectionName`
- Step 2 - we call the git trees API and modify the name of the subtree using `treeSha`
- Rest of steps...

The problem is that the `treeSha` corresponds to the commit before the `navigation.yml` update occurs. As a result, the following steps occur:
- Step 1 - `navigation.yml` is correctly updated
- Step 2: The git trees API is called - however, this step overrides the change in Step 1 since `treeSha` was obtained before Step 1.
- Rest of steps...

### Fix

The fix is to move the code chunk that updates the `navigation.yml` file to after the git tree modifications are done.

### DIFF
--- a/classes/Collection.js
+++ b/classes/Collection.js
@@ -90,9 +90,7 @@ class Collection {
   async delete(collectionName, currentCommitSha, treeSha) {
     const commitMessage = `Delete collection ${collectionName}`
     const gitTree = await getTree(this.siteName, this.accessToken, treeSha)
-    const newGitTree = gitTree.filter((item) => {
-      return item.path !== `_${collectionName}`
-    })
+    const newGitTree = gitTree.filter((item) => item.path !== `_${collectionName}`)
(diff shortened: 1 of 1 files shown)


kwajiehao (approved):
lgtm!

PR #203, opened 2021-06-16
lint: autofix files staged for commit, while ignoring errors which require manual attention

kwajiehao:
## Overview
This PR introduces the `lint-staged` and `eslint-plugin-only-warn` libraries to help introduce automatic linting and formatting at the pre-commit stage, while allowing us to ignore linting errors, which would otherwise block us from committing to our repo, which we expect to resolve in the course of our refactoring work.

This PR supercedes PR #198, which introduces `lint-staged` to run pre-commit hooks for `eslint` and `prettier`.

## Context
We wish to introduce automatic linting and formatting at the pre-commit stage, so that all commits contain properly linted and formatted code.

The problem is that there are still several linting rules which we currently violate (for example, `max-classes-per-file`) which would be resolved as part of the refactoring process. However, if we introduced an automatic linting and formatting hook, our commits might be rejected due to these linting errors which require manual attention. This means that we need to wait till our refactoring is done before introducing automatic linting and formatting tools - this is not ideal because we might accumulate more linting errors which add cognitive load when reviewing code throughout the refactoring process.

## Solution
The [`lint-staged`](https://github.com/okonet/lint-staged) library helps us to run linting and formatting fixes on our staged files before committing them to our repo. It does so by using a `husky` pre-commit hook to run all our fixes (eslint and prettier, in this case). Configuration is straightforward and can be found on the library's GitHub repo.

The [`eslint-plugin-only-warn`](https://www.npmjs.com/package/eslint-plugin-only-warn) library converts all eslint errors into warnings, so that we can take advantage of automatic lint fixes, without blocking our development since we can now ignore linting errors as they are downgraded to warnings. 

Once our refactoring work is complete, we can remove the `only-warn` plugin since the ideal state would be to resolve all linting or formatting errors (manual or not) before we commit code into our repo. Issue #204 was filed to track this issue.


prestonlimlianjie (approved):
lgtm!

PR #207, opened 2021-06-18
fix: cookie expiry date

kwajiehao:
Previously, our cookies weren't expiring and were set as sessions because we used a string instead of an integer. This PR fixes that.


prestonlimlianjie (approved):
lgtm

PR #217, opened 2021-07-14
Fix/update logout

gweiying:
Along with frontend PR [isomercms-frontend/pull/537](https://github.com/isomerpages/isomercms-frontend/pull/537). See [isomer-product/issues/26](https://github.com/isomerpages/isomer-product/issues/26) for more details.

To add env variable `ENCRYPTION_SECRET` before deploying.


alexanderleegs (changes requested):
There's actually an issue with being unable to access the logout route when a token expires naturally - since the logout endpoint requires a valid token to work. Will revisit this

alexanderleegs (approved):
lgtm!

Issue #232, opened 2021-07-26
[Refactor] Convert backend classes to use dependency injection


alexanderleegs:
Our current classes are difficult to test, due to the large chain of dependencies which make classes difficult to test. As part of the refactor, we intend to convert our new classes to use dependency injections, allowing for more independent behaviour which can be stubbed in tests easily.

seaerchin:
@alexanderleegs do we still have any left? if i remember correctly, it should be all done right

alexanderleegs:
Yep this should be done - closing the issue

PR #234, opened 2021-07-27
Refactor/dependency injection conversion

alexanderleegs:
This PR converts the services and controller involved with collection pages to use dependency injection, to resolve https://github.com/isomerpages/isomercms-backend/issues/232. The main purpose for this change is to allow for easier stubbing and independence, allowing us to create unit tests for each component. The full discussion is recorded in [this document](https://docs.google.com/document/d/1uGzg0XA08xvyfvBF2e_2giNXgO1micmLY6Cz9bEq3Qo/edit#heading=h.3eu9rwalz5km).

Due to this change, the initialisation of each of the services is only done in the server.js file.


services/fileServices/MdPageServices/CollectionPageService.js
```diff
@@ -3,100 +3,114 @@ const {
...
-) => {
-  const parsedCollectionName = `_${collectionName}`
+  async Create(reqDetails, { fileName, collectionName, content, frontMatter }) {
+    const parsedCollectionName = `_${collectionName}`
 
-  await CollectionYmlService.AddItemToOrder(reqDetails, {
-    collectionName,
-    item: fileName,
-  })
+    await this.CollectionYmlService.AddItemToOrder(reqDetails, {
```
prestonlimlianjie, line 15: Could we use `camelCase` for both the **instance** as well as the **method** of a class?
i.e. `this.collectionYmlService.addItemToOrder`

The use of `PascalCase` suggests (at least to me) that it is a class name or constructor. Thoughts?

newroutes/RequestSchema.js
```diff
@@ -5,17 +5,20 @@ const FrontMatterSchema = Joi.object({
   permalink: Joi.string().required(),
 })
```
prestonlimlianjie, line 7: Could we name this file with camelCase to fit in with the convention, i.e. `requestSchema.js` - and could we create a new `validators` folder

services/fileServices/MdPageServices/CollectionPageService.js
prestonlimlianjie: Addressed in https://github.com/isomerpages/isomercms-backend/pull/253

newroutes/collectionPages.js
```diff
@@ -1,134 +1,158 @@
...
+    this.collectionController = collectionController
+
+    const methods = Object.getOwnPropertyNames(Object.getPrototypeOf(this))
+    // Bind all methods
+    methods
+      .filter((method) => method !== "constructor")
+      .forEach((method) => {
+        this[method] = this[method].bind(this)
+      })
+  }
```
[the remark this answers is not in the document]
alexanderleegs: I'm not entirely sure of the reason actually - the `this` object is undefined if we don't explicitly bind the class methods for the router. I suspect this is due to the express router intermediary - since the router methods are not called from the class (e.g. `CollectionPagesRouter.createCollectionPage`), the context of `this` isn't the class in the case of the router components only if we don't explicitly bind it - whereas all other classes have their methods directly invoked.

The auto-bind suggestion looks great, I'll make the change for it!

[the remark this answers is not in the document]
alexanderleegs: Addressed in https://github.com/isomerpages/isomercms-backend/pull/234/commits/1c32873c4970dd9b35b60e10d03cc38b3c7a49d2

prestonlimlianjie (approved):
lgtm

PR #253, opened 2021-08-11
style: convert to camel case

prestonlimlianjie:
This PR renames all class methods using camelCase. 

Currently, the methods of the refactored classes are written in PascalCase - however, this is confusing because PascalCase is more commonly associated with Class names or constructors.


alexanderleegs (approved):
lgtm, thanks for doing the style change!

PR #258, opened 2021-08-13
feat: use convict to manage environment variables

### LINKED ISSUE #145: Enforce/validate environment variables using convict

lamkeewei:
## Problem

Closes #145 

## Solution

**Features**:
- Added `convict` to define and validate schema for configuration values. Early failure for application if required environment variables are not provided.
- Added `dotenv` to load environment variables from `.env` file instead of sourcing file

**Improvements**:
- Added `nodemon` for auto-restart on file change during development

## Tests
- [ ] There should not be any functional changes in the behaviour of the server
- [ ] If a required value that has no default value provided (e.g. `CLIENT_ID`) is not provided, the application should fail on startup
- [ ] Values marked as `required-remote-string` should only be required when running in a non-local development environment (`NODE_ENV !== LOCAL_DEV`).

## Deploy Notes
**New dependencies**:
- `convict` : Configuration management
- `dotenv` : Load environment variables from `.env` files

**New dev dependencies**:
- `nodemon` : Monitor script for development


prestonlimlianjie (approved):
lgtm! good stuff - let's also update the env var file in 1Password

config/config.js
```diff
@@ -0,0 +1,242 @@
+require("dotenv").config()
```
prestonlimlianjie, line 1: A note: dependabot just bumped `dotenv` to `10.0.0` in https://github.com/isomerpages/isomercms-backend/pull/219 so we can now use `import` instead of `require`.

https://github.com/motdotla/dotenv/blob/master/CHANGELOG.md

kishore03109:
Closing this as we already have implementing convict

PR #275, opened 2021-08-23
Refactor/remove controller

alexanderleegs:
This PR removes the intermediary Controller component from our backend refactor.

Due to our proposed introduction of a `CollectionDirectoryService` class, as well as the use of dependency injection to help manage circular dependencies, the use of the `CollectionController` class has become largely unnecessary - only being used to delegate work to the appropriate class, an action which could be handled inside our router class instead. As such, this PR removes the `CollectionController`, shifting its responsibilities to the collectionPages router, and updates tests accordingly.

In addition, this PR also fixes an issue with our existing tests by replacing `mockReturnValue` with `mockResolvedValue` for all async calls - this helps to ensure that we are properly awaiting the result of these async method calls.


prestonlimlianjie (approved):
lgtm

PR #280, opened 2021-08-27
Feat: add ability to edit description in settings

alexanderleegs:
This PR adds functionality to allow the editing of the description field in both the `config.yml` file as well as the homepage. To be reviewed in conjunction with PR [#613](https://github.com/isomerpages/isomercms-frontend/pull/613) on the isomercms frontend repo.

PR #282, opened 2021-09-01
refactor: settings v2 (BREAKING CHANGE)

kwajiehao:
**Note: this PR contains a BREAKING CHANGE.**

## Overview
This PR continues our efforts to refactor the CMS as specified by this design document ([link](https://docs.google.com/document/d/1664ww4X5PIy0mi5qB_q9JIWQaJ0qiLVIuV1v993Dlas/edit)). In this PR, I refactor the existing endpoints for reading and updating our repo settings:
- GET `/{siteName}/settings` - read settings
- POST `/{siteName}/settings` - update settings

The main change is the introduction of the new `SettingsRouter` class. The `SettingsRouter` uses the new `SettingsService` class which contains the bulk of the implementation logic, and uses the newly-created `ConfigYmlService`, `FooterYmlService`, `NavYmlService`, and `HomepagePageService` to simplify the retrieval and updating of settings data from different source files. There is some code duplication between the different service classes, but @alexanderleegs and I decided that any further simplification is out of the scope of this PR.

This PR adds a validation schema for the request body of the `updateSettingsPage` route function, and refactors the response sent from the `getSettingsPage` route function to match this schema. This is a **BREAKING CHANGE** and so this PR is to be reviewed in conjunction with PR #[618](https://github.com/isomerpages/isomercms-frontend/pull/618) on the frontend.

Finally, this PR also includes tests for the newly-introduced components in the same style as the existing tests. Some changes to testing include: 
- the creation of a `fixtures` directory which contains example test data that is reusable across different tests
- using `toHaveBeenLastCalledWith` instead of `toHaveBeenCalledWith`, so that we can make more accurate assertions (this is done in addition to the `jest.clearAllMocks()` invocation that is done in between tests)

### DIFF
--- a/newroutes/settings.js
+++ b/newroutes/settings.js
@@ -1,6 +1,7 @@
 const autoBind = require("auto-bind")
 const Bluebird = require("bluebird")
 const express = require("express")
+const _ = require("lodash")
 
 const { BadRequestError } = require("@errors/BadRequestError")
 
@@ -22,6 +23,14 @@ const extractRequiredConfigFields = (config) => ({
   resources_name: config.resources_name,
   colors: config.colors,
 })
+
+const mergeUpdatedData = (currentData, updatedData) => {
+  const clonedCurrentData = _.cloneDeep(currentData)
+  Object.keys(updatedData).forEach((field) => {
+    clonedCurrentData[field] = updatedData[field]
+  })
+  return clonedCurrentData
+}
 
 class SettingsRouter {
   constructor({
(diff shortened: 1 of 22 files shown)


newroutes/settings.js
```diff
@@ -0,0 +1,189 @@
...
+      const mergedConfigContent = mergeUpdatedData(
+        config.content,
+        updatedConfigContent
+      )
+      await this.configYmlService.update(reqDetails, {
+        fileContent: mergedConfigContent,
+        sha: config.sha,
+      })
+
+      // update homepage title only if necessary
```
alexanderleegs, line 105: We actually should be updating the description in the homepage as well - see https://github.com/isomerpages/isomercms-backend/pull/280 for the changes (has not been merged yet)

alexanderleegs (changes requested):
The use of fixtures is a great idea to help with unit testing, thanks for suggesting this implementation!

The auth middleware is currently missing the new routes - we should also add them in.

Aside from the other comments, I was wondering whether it would make sense to split up some of the logic in the settings route into a service level settings class, since currently other route layer classes only process the input and delegate the actual logic to the service classes called - could we discuss this more offline?

newroutes/settings.js
kwajiehao: Fixed with [9e1e3b8](https://github.com/isomerpages/isomercms-backend/pull/282/commits/9e1e3b8464be1275dce87c557b4e2ab194152f59)

gweiying (approved):
lgtm!

PR #296, opened 2021-09-09
hotfix (master): migrate to platform hooks

prestonlimlianjie:
### Context
We discovered that the staging CMS backend EB environment was unable to receive payloads of >1 MB - this led us to realize that there was an issue with the nginx `client_max_body_size  10M;` configuration in the `.ebextensions/nginx-max-body-size.conf` file.

We suspected that this happened because the Amazon Linux 2 platform uses platform hooks for nginx-specific configurations.

### Fix
This PR does the following:
- migrates the existing .ebextensions nginx configuration to use .platform hooks instead
- adds the `engines` field in the package.json file - was seeing the error `Instance deployment: You didn't specify a Node.js version in the 'package.json' file in your source bundle. The deployment didn't install a specific Node.js version.` ([link](https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/nodejs-platform-dependencies.html))
- updates the CI file to reference `cms-backend-prod` for the prod EB env
- updates the CI file to reference `cms-backend-staging` for the staging EB env
- deletes all references to the DEV branch in the CI file since we no longer use it

### DIFF
--- a/.ebextensions/nginx-max-body-size.config
+++ b/.ebextensions/nginx-max-body-size.config
@@ -1,7 +1,0 @@
-files:
-    "/etc/nginx/conf.d/proxy.conf" :
-        mode: "000755"
-        owner: root
-        group: root
-        content: |
-           client_max_body_size 10M;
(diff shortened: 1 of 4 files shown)


gweiying (approved):
tested on staging, lgtm!

remember to hotfix to develop too

PR #299, opened 2021-09-16
Fix: prevent truncating of page body

alexanderleegs:
This PR fixes an issue where having `---` in the page body would truncate the contents when viewed on the CMS. This was due to our reliance on this string to denote the location of the frontmatter. This PR modifies the check for frontmatter to only search for the first 2 occurrences of this, to prevent this behaviour.


gweiying (approved):
lgtm!

PR #303, opened 2021-09-22
Fix: handle 409 errors when editing files

alexanderleegs:
This PR fixes a bug where we were not handling 409 errors when sending update requests to Github. When users attempt to modify a page with an outdated sha, a 409 error is received, which we previously did not anticipate. This PR fixes this by correctly catching and throwing an appropriate error.

### DIFF
--- a/classes/File.js
+++ b/classes/File.js
@@ -131,6 +131,10 @@ class File {
     } catch (err) {
       const { status } = err.response
       if (status === 404) throw new NotFoundError("File does not exist")
+      if (status === 409)
+        throw new ConflictError(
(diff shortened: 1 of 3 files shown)


prestonlimlianjie (commented):
Could I check if we've run this error message through Lisa/Nat?

gweiying (approved):
lgtm

alexanderleegs:
> Could I check if we've run this error message through Lisa/Nat?

After discussion, as this error message is a placeholder and not used by the frontend directly, we have opted to leave it as is.

PR #305, opened 2021-09-22
Fix: change git tree object format

alexanderleegs:
This PR fixes an issue with git trees being too large for sites with a large number of files. Previously, when we wanted to use git trees to delete/rename directories, we would send the information of every file in the repo as a git tree - this resulted in 502 errors as github was unable to parse such a large amount of files. 

Recently, github has updated their API to allow for specifying that files are deleted via leaving the sha of the target files as `null` (see [link](https://developer.github.com/enterprise/2.17/v3/git/trees/#create-a-tree)). This method is much more space efficient, as the number of files deleted is significantly less than the total number of files in a repo.

We hypothesise that this PR will resolve our issues with unrecognised 502 errors.

### DIFF
--- a/utils/utils.js
+++ b/utils/utils.js
@@ -63,6 +63,7 @@ async function getTree(
 // send the new tree object back to Github and point the latest commit on the staging branch to it
 async function sendTree(
   gitTree,
+  baseTreeSha,
   currentCommitSha,
   repo,
   accessToken,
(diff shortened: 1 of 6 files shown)


utils/utils.js
```diff
@@ -63,6 +63,7 @@ async function getTree(
 // send the new tree object back to Github and point the latest commit on the staging branch to it
 async function sendTree(
```
gweiying, line 64: can change this to an object and use destructuring? else its not very clear when we miss out on a parameter

gweiying: Discussed with @alexanderleegs, since all these `classes/` files and this `utils/getTree`, `utils/sendTree` function will be removed by the refactor, it might not be worth the effort to change all the classes to use the new object parameter. 

In place, we will just test extensively.

gweiying (approved):
lgtm

Issue #308, opened 2021-10-01
[bug] "Cannot read property 'sha' of undefined" backend error


gweiying:
**Describe the bug**
Backend showing the following error:
```
{
    "name": "TypeError",
    "message": "Cannot read property 'sha' of undefined",
    "stack": "TypeError: Cannot read property 'sha' of undefined\n    at MediaFile.read (/var/app/current/classes/MediaFile.js:132:7)\n    at runMicrotasks (<anonymous>)\n    at processTicksAndRejections (internal/process/task_queues.js:95:5)\n    at async readImage (/var/app/current/routes/images.js:81:28)"
}
```

**Reason**
Backend code for `MediaFile.read` attempts to access the `sha` property even if the image does not exist. See line below
```
const imageSha = images.filter((image) => image.fileName === fileName)[0]
      .sha
```

We should change this to check if the image exists before attempting to read the `sha` property.

Issue #309, opened 2021-10-01
[bug] "images.filter is not a function" backend error


gweiying:
**Describe the bug**
Backend showing this error
```
{
    "name": "TypeError",
    "message": "images.filter is not a function",
    "stack": "TypeError: images.filter is not a function\n    at MediaFile.read (/var/app/current/classes/MediaFile.js:131:29)\n    at runMicrotasks (<anonymous>)\n    at processTicksAndRejections (internal/process/task_queues.js:95:5)\n    at async readImage (/var/app/current/routes/images.js:81:28)"
}
```

**Reason for bug**
This happens because `MediaFile.list` returns an object `{}` if the GET endpoint to Github to retrieve the list of media files returns with non-200 status code.  We then attempt to run `images.filter` on this object `{}` which throws this unrecognised error. 

To resolve this, we should actually log and throw an error, rather than returning an object that does not match the format for the next function. 

**Other bugs**
However, we should continue to debug for why Github is returning non-200 statuses for our image requests. 

For context, during the period when we received these backend errors, we noticed that images could not be loaded on our private repo `ogp-covid`. We suspect this may be some kind of Github bug or API rate-limiting, as the images could be loaded later.

PR #310, opened 2021-10-01
Fix/unhandled errors for list directory and read media

alexanderleegs:
This PR resolves issues #308 and #309. Previously, we did not handle cases where we were unable to retrieve directory information, as well as when we were unable to retrieve a file/image. This PR fixes both of these issues.


gweiying (approved):
lgtm!

PR #314, opened 2021-10-20
Refactor/resources refactor

alexanderleegs:
This PR refactors the resources and resource pages flow. Tests will be added in a future PR. It adds several new components:

resource categories, resource pages routers
resource category, resource page directory services

The resource pages router contains the following endpoints:
POST /:siteName/resourceRoom/:resourceRoomName/resources/:resourceCategory/pages - Create a resource page
GET /:siteName/resourceRoom/:resourceRoomName/resources/:resourceCategory/pages/:pageName - Read a resource page)
POST /:siteName/resourceRoom/:resourceRoomName/resources/:resourceCategory/pages/:pageName - Update/Rename a resource page
DELETE /:siteName/resourceRoom/:resourceRoomName/resources/:resourceCategory/pages/:pageName - Delete a resource page
 
The resource categories router contains the following endpoints:
GET /:siteName/resourceRoom/:resourceRoomName/resources - List all resource categories
GET /:siteName/resourceRoom/:resourceRoomName/resources/:resourceCategory - List all resource pages in a category
POST /:siteName/resourceRoom/:resourceRoomName/resources - Create a resource category
POST /:siteName/resourceRoom/:resourceRoomName/resources/:resourceCategory - Rename an existing resource category
DELETE /:siteName/resourceRoom/:resourceRoomName/resources/:resourceCategory - Delete an existing resource category
POST /:siteName/resourceRoom/:resourceRoomName/resources/:resourceCategory/move - Move pages in an existing resource category

This PR also introduces new validation schemas for the new routes. The format accepted for each route can be referred to [here](https://docs.google.com/document/d/1ao6AOtrhYTtHgaTF3KjLhdSG_4zZ65sg1gTv8f8Mnnk/edit#heading=h.vwtvtn8wqwf4).

A new method for moving files has been added to the `BaseDirectoryService` rather than the `MoverService` - I felt that the logic for file movement for all files which do not have special information in the front matter or require special data written to files (i.e. all non-collection/unlinked pages) should go here, as the mechanism for file movement in these cases would be to use the existing git tree methods which are similarly used by other methods in the `BaseDirectoryService`, In contrast, `MoverService` currently only has dependencies on page level services (e.g. `UnlinkedPageService`, `CollectionPageService`), and adding an additional dependency on `BaseDirectoryService` would cause the dependency chain to become significantly more difficult to understand.

The resource room related endpoints will be added in a future PR, after the [settings refactor](https://github.com/isomerpages/isomercms-backend/pull/282). We have elected to place the listing of all resource categories under the `ResourceDirectoryService`, rather than the future `ResourceRoomDirectoryService` - we feel that this more closely matches the currently behaviour of the `CollectionDirectoryService`.

Also, the formation of the `index.html` files on creation of a new resource category is currently contained within the `ResourceDirectoryService`. A possible alternative to this would be to move the `index.html` methods out into its own special service. However, given that this file is not used often (only on creating/renaming of a resource category), I feel this is not necessary.

### DIFF
--- a/middleware/auth.js
+++ b/middleware/auth.js
@@ -375,6 +375,32 @@ auth.post(
   verifyJwt
 )
 
+// Resource Categories
+auth.get(
+  "/v2/sites/:siteName/resourceRoom/:resourceRoomName/resources",
+  verifyJwt
+)
+auth.get(
+  "/v2/sites/:siteName/resourceRoom/:resourceRoomName/resources/:resourceCategory",
+  verifyJwt
+)
+auth.post(
+  "/v2/sites/:siteName/resourceRoom/:resourceRoomName/resources",
+  verifyJwt
+)
+auth.post(
+  "/v2/sites/:siteName/resourceRoom/:resourceRoomName/resources/:resourceCategory",
+  verifyJwt
+)
+auth.delete(
+  "/v2/sites/:siteName/resourceRoom/:resourceRoomName/resources/:resourceCategory",
+  verifyJwt
+)
+auth.post(
+  "/v2/sites/:siteName/resourceRoom/:resourceRoomName/resources/:resourceCategory/move",
+  verifyJwt
+)
+
 auth.use((req, res, next) => {
   if (!req.route) {
     return res.status(404).send("Unauthorised for unknown route")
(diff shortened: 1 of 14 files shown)


gweiying (approved):
lgtm!!!

PR #318, opened 2021-11-16
Refactor/resource room

alexanderleegs:
This PR refactors the resources room flow. It adds several new components:

Resource room router
Resource room directory services

## Overview

The resource room router contains the following endpoints:
- GET /:siteName/resourceRoom - get the name of the resource room
- POST /:siteName/resourceRoom - Create a resource room
- POST /:siteName/resourceRoom/:resourceRoomName - Rename an existing resource room
- DELETE /:siteName/resourceRoom/:resourceRoomName - Delete an existing resource room
- GET /:siteName/resourceRoom/:resourceRoomName - Moved from Resource Categories router, to retrieve the list of all resource categories, to better fit with the structure of Collections/Subcollections

This PR also introduces tests for the newly added routes and service.

## Bug fixes
This PR also fixes the following bugs:
- Empty resource category handling
  - Due to PR #310, some existing functionality for handling empty resource categories broke, as non-existent folders now throw an error, rather than return an empty array, as empty resource categories do not have a placeholder file in `_posts` where files are stored.
- Differentiation of empty resource categories and non-existent resource categories
  - The lack of a `_posts` folder in both cases meant that we were previously unable to distinguish between the two, since we only searched for the `_posts` folder of the requested resource category. We have added an additional check for the existence of the resource category so the results can be distinguished.


gweiying (approved):
lgtm! tacos for a good job

PR #319, opened 2021-11-24
fix: handle empty resource categories in v1

alexanderleegs:
This PR also fixes the following bug:

- Empty resource category handling
  - Due to PR Fix/unhandled errors for list directory and read media #310, some existing functionality for handling empty resource categories broke, as non-existent folders now throw an error, rather than return an empty array, as empty resource categories do not have a placeholder file in _posts where files are stored.
  - We now correctly return an empty array for this scenario.

### DIFF
--- a/classes/Resource.js
+++ b/classes/Resource.js
@@ -3,6 +3,8 @@ const _ = require("lodash")
 const yaml = require("yaml")
 
 // Import classes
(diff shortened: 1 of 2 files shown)


gweiying (approved):
lgtm!

PR #324, opened 2021-12-02
Refactor/media flow

alexanderleegs:
This PR refactors the resources room flow. It adds several new components:

- Media directory and media files router
- Media directory and media file services

**Overview**

- GET /:siteName/media/:directoryName - get the files in a media category
- POST /:siteName/media- Create a media category
- POST /:siteName/media/:directoryName - Rename an existing media category
- DELETE /:siteName/media/:directoryName- Delete an existing media category
- POST /:siteName/media/:directoryName/move - Move files in a media category to another media category

The media files router contains the following endpoints:

- GET /:siteName/media/:directoryName/pages/:fileName - get information about a media file
- POST /:siteName/media/:directoryName/pages - Create a media file
- POST /:siteName/media/:directoryName/pages/:fileName - Rename/Update an existing media file
- DELETE /:siteName/media/:directoryName/pages/:fileName - Delete an existing media file

The `GithubService` has also been updated with new methods to support the identification of private repos and the retrieval of image files.

This PR also introduces tests for the newly added routes and services.

**Bug fixes**
This PR also fixes the following bugs:

- Route handler for update pages
  - Previously, we were erroneously using the `WriteRouteHandlerWrapper` for updating pages, even though multiple github operations were being done. We have corrected this to use the `RollbackRouteHandlerWrapper` instead.
- Conflict check for git tree operations
  - Previously, we did not check for duplicate names with git tree operations, causing duplicate file names to override existing files or directories. We have added in a check for this to throw a `ConflictError` in this scenario.

**Notes**

The `getMedia` implementation is inefficient, and is done by calling a `listDirectory` endpoint, due to fileShas not being able to be directly queried for large (>1mb) files. Our current workaround is for the frontend to cache the results as much as possible, as the sha of each file is obtained when retrieving the list of files in a media directory, and this endpoint should be called as little as possible.

### DIFF
--- a/middleware/auth.js
+++ b/middleware/auth.js
@@ -409,27 +409,18 @@ auth.post(
 )
 
 // Media directories
-auth.get("/v2/sites/:siteName/media/:mediaType/:directoryName", verifyJwt)
-auth.post("/v2/sites/:siteName/media/:mediaType", verifyJwt)
-auth.post("/v2/sites/:siteName/media/:mediaType/:directoryName", verifyJwt)
-auth.delete("/v2/sites/:siteName/media/:mediaType/:directoryName", verifyJwt)
-auth.post("/v2/sites/:siteName/media/:mediaType/:directoryName/move", verifyJwt)
+auth.get("/v2/sites/:siteName/media/:directoryName", verifyJwt)
+auth.post("/v2/sites/:siteName/media", verifyJwt)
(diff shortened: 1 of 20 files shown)


gweiying (approved):
lgtm! Thanks for the work on this, was really fast and easy to work with!

PR #325, opened 2022-01-06
Fix/move pages on old format

alexanderleegs:
This PR fixes an issue when moving files still on the old slugified format (e.g. `page-title.md`). As the move methods make use of the individual create methods of each `PageService`, this causes the movement to fail when the dashes in the slugified file titles were detected. As we have decided not to touch pages on the old formatting, we have implemented a bypass for the file title check, to only be used by the mover service.


gweiying (approved):
lgtm!

PR #328, opened 2022-02-03
Refactor/auth

alexanderleegs:
This PR refactors the auth flow. It adds several new components:

- Auth router and service
- Refactored auth middleware to use dependency injection

## Overview

GET /github-redirect - Redirect user to github on login
GET /- Obtain token from github and redirect back to CMS
DELETE /logout- Logout a user from the CMS
GET /whoami - Retrieve user ID

This PR also introduces tests for the newly added routes and services.

To be merged after #324.


newroutes/auth.js
```diff
@@ -0,0 +1,99 @@
...
+
+const { FRONTEND_URL } = process.env
+
+const AUTH_TOKEN_EXPIRY_MS = parseInt(
+  process.env.AUTH_TOKEN_EXPIRY_DURATION_IN_MILLISECONDS,
+  10
+)
+const CSRF_TOKEN_EXPIRY_MS = 600000
+const CSRF_COOKIE_NAME = "isomer-csrf"
+const COOKIE_NAME = "isomercms"
```
seaerchin, line 15: wdyt about shifting these to a `constants` file? (also ok with leaving it here if it's too much trouble)

newroutes/auth.js
```diff
@@ -0,0 +1,99 @@
...
+    } = await this.authService.getAuthRedirectDetails()
+    const csrfTokenExpiry = new Date()
+    csrfTokenExpiry.setTime(csrfTokenExpiry.getTime() + CSRF_TOKEN_EXPIRY_MS)
+    const cookieSettings = {
+      expires: csrfTokenExpiry,
+      httpOnly: true,
+      secure:
+        process.env.NODE_ENV !== "DEV" &&
+        process.env.NODE_ENV !== "LOCAL_DEV" &&
+        process.env.NODE_ENV !== "test",
```
seaerchin, line 37: if this pattern is going to be repeated alot, we should probably have a helper to keep them in sync

newroutes/auth.js
```diff
@@ -0,0 +1,99 @@
...
+    authTokenExpiry.setTime(authTokenExpiry.getTime() + AUTH_TOKEN_EXPIRY_MS)
+    const cookieSettings = {
+      path: "/",
+      expires: authTokenExpiry,
+      httpOnly: true,
+      sameSite: true,
+      secure:
+        process.env.NODE_ENV !== "DEV" &&
+        process.env.NODE_ENV !== "LOCAL_DEV" &&
+        process.env.NODE_ENV !== "test",
```
seaerchin, line 62: see above wrt helper function

newroutes/auth.js
```diff
@@ -0,0 +1,99 @@
...
+    const csrfState = req.cookies[CSRF_COOKIE_NAME]
+    const { code, state } = req.query
+
+    const token = await this.authService.getGithubAuthToken({
+      csrfState,
+      code,
+      state,
+    })
+    const authTokenExpiry = new Date()
+    authTokenExpiry.setTime(authTokenExpiry.getTime() + AUTH_TOKEN_EXPIRY_MS)
```
seaerchin, line 53: see above wrt timezone

server.js
```diff
@@ -112,7 +115,14 @@ const { ResourcePagesRouter } = require("./newroutes/resourcePages")
...
 const { SettingsRouter } = require("./newroutes/settings")
 const { UnlinkedPagesRouter } = require("./newroutes/unlinkedPages")
+const {
+  AuthMiddlewareService,
+} = require("./services/middlewareServices/authMiddlewareService")
+
+const authMiddlewareService = new AuthMiddlewareService()
+const authMiddleware = new AuthMiddleware({ authMiddlewareService })
 
+const authService = new AuthService()
```
seaerchin, line 125: ._. this is p confusing and we could have a helper called `initAuthMiddleware` that just does this

services/middlewareServices/AuthMiddlewareService.js
```diff
@@ -0,0 +1,102 @@
...
+
+const { E2E_TEST_REPO, E2E_TEST_SECRET, E2E_TEST_GH_TOKEN } = process.env
+const E2E_TEST_USER = "e2e-test"
+const GENERAL_ACCESS_PATHS = ["/v1/sites", "/v1/auth/whoami"]
+
+class AuthMiddlewareService {
+  verifyE2E({ cookies, url }) {
+    const { isomercmsE2E } = cookies
+    const urlTokens = url.split("/") // urls take the form "/v1/sites/<repo>/<path>""
+    let isValidE2E
```
seaerchin, line 19:
```js
const isValidE2E = false
```

services/middlewareServices/AuthMiddlewareService.js
```diff
@@ -0,0 +1,102 @@
...
+        userId = retrievedId
+      } catch (err) {
+        logger.error("Authentication error")
+        if (err.name === "TokenExpiredError") {
+          throw new AuthError("JWT token has expired")
+        }
+        if (err.name === "JsonWebTokenError") {
+          throw new AuthError(err.message)
+        }
+        throw new Error(err)
```
seaerchin, line 70: does 
```js
throw err
```
not work?

services/middlewareServices/AuthMiddlewareService.js
```diff
@@ -0,0 +1,102 @@
...
+  verifyJwt({ cookies, url }) {
+    const { isomercms } = cookies
+    const isValidE2E = this.verifyE2E({ cookies, url })
+
+    let accessToken
+    let userId
+    if (isValidE2E) {
+      accessToken = E2E_TEST_GH_TOKEN
+      userId = E2E_TEST_USER
+    } else {
```
seaerchin, line 54: i think this block should just be an early return

services/middlewareServices/AuthMiddlewareService.js
```diff
@@ -0,0 +1,102 @@
...
+      try {
+        const {
+          access_token: retrievedToken,
+          user_id: retrievedId,
+        } = jwtUtils.verifyToken(isomercms)
+        accessToken = jwtUtils.decryptToken(retrievedToken)
+        userId = retrievedId
+      } catch (err) {
+        logger.error("Authentication error")
```
seaerchin, line 63: i think we should have a more structured logging format to avoid information loss. for this, we could capture the cookie/url under a `params` key perhaps.

services/middlewareServices/AuthMiddlewareService.js
```diff
@@ -0,0 +1,102 @@
...
+
+    return isValidE2E
+  }
+
+  verifyJwt({ cookies, url }) {
+    const { isomercms } = cookies
+    const isValidE2E = this.verifyE2E({ cookies, url })
+
+    let accessToken
+    let userId
```
seaerchin, line 50: we can avoid early declarations and just return in the branches of the `if/else + try/catch`

services/middlewareServices/AuthMiddlewareService.js
```diff
@@ -0,0 +1,102 @@
...
+
+  whoamiAuth({ cookies, url }) {
+    const isValidE2E = this.verifyE2E({ cookies, url })
+
+    let accessToken
+    let userId
+    if (isValidE2E) {
+      accessToken = E2E_TEST_GH_TOKEN
+      userId = E2E_TEST_USER
+    } else {
```
seaerchin, line 84: same as above - early return to avoid indentation

services/middlewareServices/AuthMiddlewareService.js
```diff
@@ -0,0 +1,102 @@
...
+        const { isomercms } = cookies
+        const { access_token: verifiedToken } = jwtUtils.verifyToken(isomercms)
+        retrievedToken = jwtUtils.decryptToken(verifiedToken)
+      } catch (err) {
+        retrievedToken = undefined
+      } finally {
+        accessToken = retrievedToken
+      }
+    }
+    return { accessToken, userId }
```
seaerchin, line 96: `userId` is always undefined if it's not e2e - is this supposed to be the case?

services/middlewareServices/AuthMiddlewareService.js
```diff
@@ -0,0 +1,102 @@
...
+      try {
+        const { isomercms } = cookies
+        const { access_token: verifiedToken } = jwtUtils.verifyToken(isomercms)
+        retrievedToken = jwtUtils.decryptToken(verifiedToken)
+      } catch (err) {
+        retrievedToken = undefined
+      } finally {
+        accessToken = retrievedToken
+      }
+    }
```
seaerchin, line 95: wdyt about this? (allows us to remove the early declarations too)
```js
  try {
    const { isomercms } = cookies
    const { access_token: verifiedToken } = jwtUtils.verifyToken(isomercms)
    return {
      accessToken: jwtUtils.decryptToken(verifiedToken),
      userId: undefined,
    }
  } catch (err) {
    // do some logging maybe
    return {
      accessToken: undefined,
      userId: undefined,
    }
  }
```

services/utilServices/AuthService.js
```diff
@@ -0,0 +1,96 @@
+const axios = require("axios")
+const queryString = require("query-string")
+const uuid = require("uuid/v4")
+
+// Import error
```
seaerchin, line 5: not sure what this means - could you help clarify? thank you!

services/utilServices/AuthService.js
```diff
@@ -0,0 +1,96 @@
...
+          password: CLIENT_SECRET,
+        },
+      }
+    )
+
+    const { access_token: accessToken } = queryString.parse(resp.data)
+    if (!accessToken) throw new AuthError("Access token not found")
+
+    // Retrieve user information to put into access token
+    const endpoint = `https://api.github.com/user`
```
seaerchin, line 51: nit: should this be `SCREAMING_SNAKE_CASE`? not sure on the casing rules but would love a clarification!

services/utilServices/AuthService.js
```diff
@@ -0,0 +1,96 @@
...
+      validateStatus,
+      headers: {
+        Authorization: `token ${accessToken}`,
+        Accept: "application/vnd.github.v3+json",
+        "Content-Type": "application/json",
+      },
+    })
+
+    let userId
+    if (userResp.data) userId = userResp.data.login
```
seaerchin, line 62: ```const userId = userResp.data && userResp.data.login```

services/utilServices/AuthService.js
```diff
@@ -0,0 +1,96 @@
...
+      const resp = await axios.get(endpoint, {
+        headers: {
+          Authorization: `token ${accessToken}`,
+          "Content-Type": "application/json",
+        },
+      })
+      userId = resp.data.login
+    } catch (err) {
+      userId = undefined
+    }
```
seaerchin, line 89: wdyt about this:
``` js
try {
    const resp = await axios.get(endpoint, {
      headers: {
        Authorization: `token ${accessToken}`,
        "Content-Type": "application/json",
      },
    })
    return resp.data.login
  } catch (err) {
    return undefined
  }
```

also, i think we might benefit from logging the userId and error here!

seaerchin (commented):
actually i have a few questions on this PR wrt how the codebase is structured!

1. i'm not sure why `AuthMiddleware` in `newmiddleware/auth.js` is exporting a bunch of routes - i was under the impression that this should be the responsibility of the router
2. what's the difference between a `middlewareService` and a `Service`? 
3. and between a `middleware` and a `middlewareService`?

also, how do we decide which functions get to throw? i think throwing is not a good idea because of how it interrupts control flow and in general, is only suited to exceptions, not errors. i think in this case, only the top most layer should be allowed to throw/interact with io while the rest of our downstream consumers should be pure.

newmiddleware/auth.js
```diff
@@ -0,0 +1,430 @@
...
+      "/v2/sites/:siteName/media/:directoryName/pages/:fileName",
+      this.verifyJwt
+    )
+    auth.post(
+      "/v2/sites/:siteName/media/:directoryName/pages/:fileName",
+      this.verifyJwt
+    )
+    auth.delete(
+      "/v2/sites/:siteName/media/:directoryName/pages/:fileName",
+      this.verifyJwt
```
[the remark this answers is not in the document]
alexanderleegs: The 2 routers solution might be problematic, given that the auth middleware also checks and throws an error if the route requested is not one that we expect - I'll try a bit more to see if I can get it working, otherwise we might have to use the helper method instead

newmiddleware/auth.js
```diff
@@ -0,0 +1,430 @@
...
+    // Collection pages
+    auth.get("/v1/sites/:siteName/collections/:collectionName", this.verifyJwt)
+    auth.get(
+      "/v1/sites/:siteName/collections/:collectionName/pages",
+      this.verifyJwt
+    )
+    auth.post(
+      "/v1/sites/:siteName/collections/:collectionName/pages",
+      this.verifyJwt
+    ) // to remove
```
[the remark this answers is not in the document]
alexanderleegs: https://github.com/isomerpages/isomercms-backend/issues/351

newmiddleware/auth.js
```diff
@@ -0,0 +1,430 @@
...
+    auth.get("/v1/sites/:siteName/tree", this.verifyJwt)
+
+    // Menu
+    auth.get("/v1/sites/:siteName/menus", this.verifyJwt)
+    auth.get("/v1/sites/:siteName/menus/:menuName", this.verifyJwt)
+    auth.post("/v1/sites/:siteName/menus/:menuName", this.verifyJwt)
+
+    // Pages
+    auth.get("/v1/sites/:siteName/pages", this.verifyJwt)
+    auth.post("/v1/sites/:siteName/pages", this.verifyJwt) // to remove
```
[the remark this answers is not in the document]
alexanderleegs: https://github.com/isomerpages/isomercms-backend/issues/351

newmiddleware/auth.js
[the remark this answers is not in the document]
alexanderleegs: As discussed offline, a second router is problematic because our current auth middleware implementation requires all routes to pass through it, and throws an error if the route is unrecognised - we've opted to use the helped method instead, implemented in https://github.com/isomerpages/isomercms-backend/pull/328/commits/84222d9ab39bc2e877a3d7b8503a10fec6004880

newmiddleware/auth.js
```diff
@@ -0,0 +1,430 @@
...
+      this.verifyJwt
+    )
+    auth.delete(
+      "/v2/sites/:siteName/media/:directoryName/pages/:fileName",
+      this.verifyJwt
+    )
+
+    auth.use((req, res, next) => {
+      if (!req.route) {
+        return res.status(404).send("Unauthorised for unknown route")
```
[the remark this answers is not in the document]
alexanderleegs: As discussed offline, since all endpoints are queried purely by our frontend, this should only be accessed if someone is attempting to ping an endpoint directly. I've updated the message sent in this case to `Unknown route requested`: https://github.com/isomerpages/isomercms-backend/pull/328/commits/0e38f6f94908110c56afe62fa657f5a01ac4482b

newroutes/__tests__/Auth.spec.js
```diff
@@ -0,0 +1,115 @@
...
+      mockAuthService.getAuthRedirectDetails.mockResolvedValueOnce({
+        redirectUrl,
+        cookieToken,
+      })
+      const resp = await request(app).get(`/github-redirect`)
+      expect(mockAuthService.getAuthRedirectDetails).toHaveBeenCalledTimes(1)
+      expect(resp.status).toEqual(302)
+      expect(resp.headers.location).toContain(redirectUrl)
+      expect(resp.headers["set-cookie"]).toEqual(
+        expect.arrayContaining([expect.stringContaining(CSRF_COOKIE_NAME)])
```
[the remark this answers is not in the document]
alexanderleegs: https://github.com/isomerpages/isomercms-backend/pull/328/commits/8d7d8a4db885da466a023f1068a18e28beade3b1

newroutes/__tests__/Auth.spec.js
```diff
@@ -0,0 +1,115 @@
...
+      )
+    })
+  })
+
+  describe("githubAuth", () => {
```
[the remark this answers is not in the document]
alexanderleegs: https://github.com/isomerpages/isomercms-backend/pull/328/commits/8d7d8a4db885da466a023f1068a18e28beade3b1

newroutes/__tests__/Auth.spec.js
```diff
@@ -0,0 +1,115 @@
...
+        .expect(200)
+      expect(resp.headers["set-cookie"]).toEqual(
+        expect.arrayContaining([
+          expect.stringContaining(CSRF_COOKIE_NAME),
+          expect.stringContaining(COOKIE_NAME),
+        ])
+      )
+    })
+  })
+
```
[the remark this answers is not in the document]
alexanderleegs: As discussed offline, this was my bad - it's not clear that we're checking for the new set-cookie headers to be blank. I've updated the test to check more explicitly for it for greater clarity in this PR: https://github.com/isomerpages/isomercms-backend/pull/328/commits/8d7d8a4db885da466a023f1068a18e28beade3b1

newroutes/__tests__/Auth.spec.js
```diff
@@ -0,0 +1,115 @@
...
+  app.delete("/logout", attachReadRouteHandlerWrapper(router.logout))
+  app.get("/whoami", attachReadRouteHandlerWrapper(router.whoami))
+  app.use(errorHandler)
+
+  beforeEach(() => {
+    jest.clearAllMocks()
+  })
+
+  describe("authRedirect", () => {
+    const redirectUrl = "redirectUrl"
```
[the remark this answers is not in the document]
alexanderleegs: https://github.com/isomerpages/isomercms-backend/pull/328/commits/8d7d8a4db885da466a023f1068a18e28beade3b1

newroutes/__tests__/Auth.spec.js
```diff
@@ -0,0 +1,115 @@
...
+const { attachReadRouteHandlerWrapper } = require("@middleware/routeHandler")
+
+const { AuthRouter } = require("../auth")
+
+const { FRONTEND_URL } = process.env
+const CSRF_COOKIE_NAME = "isomer-csrf"
+const COOKIE_NAME = "isomercms"
+const csrfState = "csrfState"
+const cookieToken = "cookieToken"
+const accessToken = undefined
```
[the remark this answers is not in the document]
alexanderleegs: As discussed offline, we use `SCREAMING_SNAKE_CASE` for constants, and `camelCase` are used for locally scoped variables. I've modified the constants to be imported from the auth file instead here https://github.com/isomerpages/isomercms-backend/pull/328/commits/596c552abad77de364dcc905ace2ccb51cac3666

newroutes/auth.js
```diff
@@ -0,0 +1,99 @@
...
+    autoBind(this)
+  }
+
+  async authRedirect(req, res) {
+    const {
+      redirectUrl,
+      cookieToken,
+    } = await this.authService.getAuthRedirectDetails()
+    const csrfTokenExpiry = new Date()
+    csrfTokenExpiry.setTime(csrfTokenExpiry.getTime() + CSRF_TOKEN_EXPIRY_MS)
```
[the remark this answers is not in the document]
alexanderleegs: getTime() is timezone agnostic, so this works regardless - but i've added an extra comment about it here https://github.com/isomerpages/isomercms-backend/pull/328/commits/370e105d57b130be7ecda03850e5280a91fc4c70

newroutes/auth.js
alexanderleegs: As discussed offline, we've opted to leave it in this file as it is only used for auth and its corresponding test file. https://github.com/isomerpages/isomercms-backend/pull/328/commits/596c552abad77de364dcc905ace2ccb51cac3666

newroutes/auth.js
alexanderleegs: https://github.com/isomerpages/isomercms-backend/pull/328/commits/370e105d57b130be7ecda03850e5280a91fc4c70

newroutes/auth.js
alexanderleegs: https://github.com/isomerpages/isomercms-backend/pull/328/commits/370e105d57b130be7ecda03850e5280a91fc4c70

newroutes/auth.js
alexanderleegs: see above

server.js
alexanderleegs: As discussed, this is a good point that warrants further team discussion - but will be out of scope of this PR regardless

services/middlewareServices/AuthMiddlewareService.js
alexanderleegs: Refactored to return early in https://github.com/isomerpages/isomercms-backend/pull/328/commits/f67202b7ee8461090aff2746ddcf0652b5c69251

services/middlewareServices/AuthMiddlewareService.js
alexanderleegs: Whoops, that does work better, f67202b

services/middlewareServices/AuthMiddlewareService.js
alexanderleegs: f67202b

services/middlewareServices/AuthMiddlewareService.js
alexanderleegs: Agreed, modified in f67202b, but we should have a further team discussion on modifying our existing error logging

services/middlewareServices/AuthMiddlewareService.js
alexanderleegs: f67202b

services/middlewareServices/AuthMiddlewareService.js
alexanderleegs: f67202b

services/middlewareServices/AuthMiddlewareService.js
alexanderleegs: As discussed, this is currently intended behaviour, but not ideal - we will leave this as is because a typescript migration will resolve this issue

services/middlewareServices/AuthMiddlewareService.js
alexanderleegs: f67202b

services/utilServices/AuthService.js
alexanderleegs: For locally scoped variables we'll continue to use `camelCase`

services/utilServices/AuthService.js
alexanderleegs: This would result in a `false` value for userId though, which I think we want to avoid

services/utilServices/AuthService.js
alexanderleegs: Good point about the early return, changed in b1a4a8a! Given that this is for a generic whoami endpoint though, we would expect errors relatively frequently - I think it might cause quite a bit of bloat

newroutes/__tests__/Auth.spec.js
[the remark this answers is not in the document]
seaerchin: double checking to ensure that i've understood this correctly - app wide constants are `SCREAMING_SNAKE_CASE` but constants that are local to the file (ie, never exported) are `camelCase`?

newroutes/auth.js
seaerchin: wdyt about shifting the helper to utils as it might be used elsewhere in the future? eslint also screams at me about `Expected 'this' to be used by class method 'isSecure'.`, which seems to hint that this shouldn't belong within the class

services/middlewareServices/AuthMiddlewareService.js
seaerchin: i think the new implementation still has no initializer and an assignment in the catch, which is pretty unclear (isn't it `undefined` regardless since the try block will fail as a whole). 

i'm erring on the side of clarity here because we go through q a few layers of indirection in the `finally` branch to return `{accessToken: retrievedToken}` (which is `undefined`)

services/utilServices/AuthService.js
seaerchin: sorry, not q sure what you mean by bloat - could you explain more? i think i might be missing something!

seaerchin (commented):
some comments!

newmiddleware/auth.js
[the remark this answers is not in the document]
alexanderleegs: https://github.com/isomerpages/isomercms-backend/pull/328/commits/9a8f361d039c10f763eabb09019b20e332f22158

newroutes/__tests__/Auth.spec.js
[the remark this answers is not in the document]
alexanderleegs: yep, i think we can go with that

services/utilServices/AuthService.js
alexanderleegs: I was thinking that signtoken might have unexpected behaviour - but it seems fine, https://github.com/isomerpages/isomercms-backend/pull/328/commits/bbd8028e94e4cd88b5e9969fc1903f2184fb0df5

services/utilServices/AuthService.js
alexanderleegs: It would generate a lot more log messages which might not be very helpful, given that we already log userid for each route accessed

newmiddleware/auth.js
```diff
@@ -0,0 +1,351 @@
...
+    this.authMiddlewareService = authMiddlewareService
+    // We need to bind all methods because we don't invoke them from the class directly
+    autoBind(this)
+  }
+
+  noVerify(req, res, next) {
+    next("router")
+  }
+
+  // Create new page in collection
```
gweiying, line 15: remove irrelevant comment

alexanderleegs: Removed in https://github.com/isomerpages/isomercms-backend/pull/328/commits/4e5d20221619aeb00ed410d1f0ee3d739d124e0d

gweiying:
> actually i have a few questions on this PR wrt how the codebase is structured!
> 
> 1. i'm not sure why `AuthMiddleware` in `newmiddleware/auth.js` is exporting a bunch of routes - i was under the impression that this should be the responsibility of the router
> 2. what's the difference between a `middlewareService` and a `Service`?
> 3. and between a `middleware` and a `middlewareService`?
> 
> also, how do we decide which functions get to throw? i think throwing is not a good idea because of how it interrupts control flow and in general, is only suited to exceptions, not errors. i think in this case, only the top most layer should be allowed to throw/interact with io while the rest of our downstream consumers should be pure.

Synced with @alexanderleegs offline, I agree with @seaerchin's point 1. where newmiddleware/auth.js is bloated because it chains the middleware onto routes individually. I think it might be better to inject the middleware onto the existing route controllers e.g. CollectionPageController (see [Forms](https://github.com/opengovsg/FormSG/blob/7969d66b36eaf4ed84ba1ab1238d17cbfcaf0563/src/app/modules/form/admin-form/admin-form.routes.ts#L54), [Go](https://github.com/opengovsg/GoGovSG/blob/efa77265225205c105e6e513b34a3ad1579a1097/src/server/api/index.ts#L48) for example), so that we don't maintain a middleware list of routes that is separate from the controller routes. This way we also don't need to forward the `.get`, `.post` and `.delete` individually, since we're injecting the middleware instead. 


Followups from our conversation, @alexanderleegs will look into seeing if this injected middleware pattern works for Isomer, and refactor this PR in an upcoming PR so that the refactor will not be a blocker for existing flows.

alexanderleegs:
> I think it might be better to inject the middleware onto the existing route controllers e.g. CollectionPageController (see [Forms](https://github.com/opengovsg/FormSG/blob/7969d66b36eaf4ed84ba1ab1238d17cbfcaf0563/src/app/modules/form/admin-form/admin-form.routes.ts#L54), [Go](https://github.com/opengovsg/GoGovSG/blob/efa77265225205c105e6e513b34a3ad1579a1097/src/server/api/index.ts#L48) for example), so that we don't maintain a middleware list of routes that is separate from the controller routes.

This looks like it will work well, and we can also separate our initialisation of middleware, similar to what's being done with the identity pieces. I've created an issue for this (https://github.com/isomerpages/isomercms-backend/issues/375), will tackle it in a future PR!

prestonlimlianjie (approved):
lgtm

PR #329, opened 2022-02-04
Fix: remove footer links if empty

alexanderleegs:
This PR fixes a bug introduced in https://github.com/isomerpages/isomercms-backend/pull/282. In the process of our refactor for the settings flow, we reintroduced a bug in which footer settings which had been deleted (indicated via an empty string for the field sent by the frontend) were being stored as the empty string - this caused an issue with the appearance of icons and links on the footer to empty pages, as our template only checks for the existence of the field before rendering the appropriate icon/link. This PR adds additional checks to remove the appropriate fields in the footer if the user has deleted them.

### DIFF
--- a/services/configServices/SettingsService.js
+++ b/services/configServices/SettingsService.js
@@ -82,7 +82,7 @@ class SettingsService {
     }
 
     if (!_.isEmpty(updatedFooterContent)) {
-      const mergedFooterContent = this.mergeUpdatedData(
+      const mergedFooterContent = this.mergeUpdatedFooterData(
         footer.content,
         updatedFooterContent
       )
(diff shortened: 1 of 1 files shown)


gweiying (approved):
functionality looks fine to me - though - would it be cleaner to use something like the lodash's [pickBy or omitBy function](https://lodash.com/docs/4.17.15#omitBy) to remove empty values?

alexanderleegs:
> functionality looks fine to me - though - would it be cleaner to use something like the lodash's [pickBy or omitBy function](https://lodash.com/docs/4.17.15#omitBy) to remove empty values?

As discussed offline, we concluded that this is less efficient and not significantly cleaner since we still need to recursively perform the omitBy function - since footer data is the only place this is used, we have decided to leave it as is.

PR #331, opened 2022-02-07
Refactor/homepage

alexanderleegs:
Overview
--
This PR adds the refactored homepage router, in line with our recent refactoring efforts. It builds off the existing HomepagePageService that already exists for use by the SettingsRouter. It also adds the relevant tests for the homepage router component.

GET /:siteName/homepage - Retrieve homepage information
POST /:siteName/homepage- Update homepage information

Notes
--
I have opted for a looser schema of the updateHomepage request (e.g. not checking all section details), as several sites have additional details in the front matter of their homepage due to legacy issues, though I am open to suggestions on what we could do if we want to enforce a stricter format.


newroutes/homepage.js
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  // Update homepage index file
+  async updateHomepage(req, res) {
+    const { accessToken } = req
+
+    const { siteName } = req.params
+    const { error } = UpdateHomepageSchema.validate(req.body, {
+      allowUnknown: true,
+    })
```
seaerchin, line 42: i think this is a good pattern that we should be using going forward! we validate unknown data and ensure that it's of a valid shape before passing it onto our services!

newroutes/homepage.js
```diff
@@ -0,0 +1,73 @@
...
+    if (error) throw new BadRequestError(error)
```
seaerchin, line 43: if the frontend uses this to display information back to the user, it might be worth writing more fine-grained error messages. 

joi provides [more fine-grained detail](https://joi.dev/api/?v=17.6.0) on the error occurred so i think in any case, we might wanna return that rather than wrapping.

validators/RequestSchema.js
```diff
@@ -11,6 +11,23 @@ const ItemSchema = FileSchema.keys({
...
+    frontMatter: Joi.object({
+      layout: Joi.string().required(),
+      title: Joi.string().required(),
+      description: Joi.string(),
+      permalink: Joi.string().required(),
+      notification: Joi.string(),
+      image: Joi.string(),
+      sections: Joi.array().required(),
+    }).required(),
+    pageBody: Joi.string().allow(""),
```
seaerchin, line 26: could we leave a comment here as to why we allow empty strings instead of making it not required? it's not entirely clear to me the reason why (maybe FE always defaults this?) and it's different from the other validation here

seaerchin (commented):
lgtm with some minor clarifications

newroutes/__tests__/Homepage.spec.js
```diff
@@ -0,0 +1,161 @@
...
+    const expectedResponse = {
+      sha: mockSha,
+      content: mockContent,
+    }
+    mockHomepagePageService.read.mockResolvedValue(expectedResponse)
+    it("retrieves homepage details", async () => {
+      const resp = await request(app).get(`/${siteName}/homepage`).expect(200)
+      expect(resp.body).toStrictEqual(expectedResponse)
+      expect(mockHomepagePageService.read).toHaveBeenCalledWith(reqDetails)
+    })
```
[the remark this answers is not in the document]
alexanderleegs: https://github.com/isomerpages/isomercms-backend/pull/331/commits/1e8ace4807b66ec153a7fc5633ade7c7fc3d0e1d

newroutes/__tests__/Homepage.spec.js
```diff
@@ -0,0 +1,161 @@
...
+            ],
+          },
+          pageBody: mockContent,
+        },
+        sha: mockSha,
+      }
+      await request(app)
+        .post(`/${siteName}/homepage`)
+        .send(incompleteDetails)
+        .expect(400)
```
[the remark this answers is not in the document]
alexanderleegs: https://github.com/isomerpages/isomercms-backend/pull/331/commits/1e8ace4807b66ec153a7fc5633ade7c7fc3d0e1d

newroutes/__tests__/Homepage.spec.js
```diff
@@ -0,0 +1,161 @@
...
+    })
+
+    it("accepts valid homepage update requests and returns the details of the file updated", async () => {
```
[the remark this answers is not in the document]
alexanderleegs: https://github.com/isomerpages/isomercms-backend/pull/331/commits/1e8ace4807b66ec153a7fc5633ade7c7fc3d0e1d

newroutes/homepage.js
```diff
@@ -0,0 +1,73 @@
+const autoBind = require("auto-bind")
+const express = require("express")
+
+// Import middleware
```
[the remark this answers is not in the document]
alexanderleegs: Good catch, https://github.com/isomerpages/isomercms-backend/pull/331/commits/b918cde48190919b01ef4b787ee2383ec9f48231

newroutes/homepage.js
alexanderleegs: Makes sense, https://github.com/isomerpages/isomercms-backend/pull/331/commits/2610f33a177fd7d804a189afb606e5f0229093da

validators/RequestSchema.js
alexanderleegs: Joi seems to disallow empty strings if you only specify that it is of type string, e.g. `pageBody: Joi.string()` would accept an object with no pageBody but not one with `pageBody: ''` unless explicitly allowed. I've added a comment in https://github.com/isomerpages/isomercms-backend/pull/331/commits/540bba889a9fcddc1d8c5be62901c9ea08d9df6f to explain this!

newroutes/homepage.js
```diff
@@ -0,0 +1,77 @@
...
+    const { error } = UpdateHomepageSchema.validate(req.body, {
+      allowUnknown: true,
+    })
+    if (error) throw new BadRequestError(error.message)
+    const {
+      content: { frontMatter, pageBody },
+      sha,
+    } = req.body
+
+    const updateResp = await this.homepagePageService.update(
```
seaerchin, line 51: could we maybe rename this to `updatedHomepage`? not sure `updateResp` is entirely clear

newroutes/homepage.js
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  getRouter() {
+    const router = express.Router()
+
+    router.use(authMiddleware.verifyJwt)
+
+    router.get(
+      "/:siteName/homepage",
+      attachReadRouteHandlerWrapper(this.readHomepage)
```
seaerchin, line 66: i think once we shift to ts, we can try doing this with decorators to see if it eases the friction!

server.js
```diff
@@ -242,6 +244,7 @@ app.use("/v1/sites", navigationRouter)
 app.use("/v1/sites", netlifyTomlRouter)
 
 app.use("/v2/auth", authV2Router.getRouter())
+app.use("/v2/sites", homepageV2Router.getRouter())
```
seaerchin, line 247: just noticed - should we namespace the routers going forward? (don't need to fix here as out of scope)

validators/RequestSchema.js
```diff
@@ -11,6 +11,23 @@ const ItemSchema = FileSchema.keys({
   children: Joi.array().items(Joi.string()),
 })
 
+// Homepage
+const UpdateHomepageSchema = Joi.object({
+  content: Joi.object({
+    frontMatter: Joi.object({
+      layout: Joi.string().required(),
+      title: Joi.string().required(),
+      description: Joi.string(),
```
seaerchin, line 20: will the FE allow a case like so: `description: ""`?

validators/RequestSchema.js
```diff
@@ -11,6 +11,23 @@ const ItemSchema = FileSchema.keys({
...
+      permalink: Joi.string().required(),
+      notification: Joi.string(),
```
seaerchin, line 22: same as above

seaerchin (commented):
minor comments but looking good overall!

newroutes/homepage.js
alexanderleegs: I've updated the variable name, https://github.com/isomerpages/isomercms-backend/pull/331/commits/65f497e0aa5706b6a0648d0e9a2de6e3a2ac893e

server.js
alexanderleegs: Yup that's a good point, I've raised an issue here https://github.com/isomerpages/isomercms-backend/issues/383

validators/RequestSchema.js
alexanderleegs: Good catch, that should be allowed - fixed in https://github.com/isomerpages/isomercms-backend/pull/331/commits/1b04b913d20b907d1b8d2ad1f6dc8f496017cbe5

validators/RequestSchema.js
alexanderleegs: https://github.com/isomerpages/isomercms-backend/pull/331/commits/1b04b913d20b907d1b8d2ad1f6dc8f496017cbe5

newroutes/authenticatedSites/__tests__/Homepage.spec.js
```diff
@@ -0,0 +1,169 @@
...
+
+  describe("readHomepage", () => {
+    const expectedResponse = {
+      sha: mockSha,
+      content: mockContent,
+    }
+
+    mockHomepagePageService.read.mockResolvedValue(expectedResponse)
+
+    it("retrieves homepage details", async () => {
```
seaerchin, line 54: could we add 1 extra test case illustrating what happens if our reads fail?

newroutes/authenticatedSites/__tests__/Homepage.spec.js
```diff
@@ -0,0 +1,169 @@
...
+                  title: "Infobar title",
+                  subtitle: "Subtitle",
+                },
+              },
+            ],
+          },
+          pageBody: mockContent,
+        },
+        sha: mockSha,
+      }
```
seaerchin, line 122: nit: we can use `_.omit` here for clarity

validators/RequestSchema.js
```diff
@@ -11,6 +11,23 @@ const ItemSchema = FileSchema.keys({
...
+// Homepage
+const UpdateHomepageSchema = Joi.object({
+  content: Joi.object({
+    frontMatter: Joi.object({
+      layout: Joi.string().required(),
+      title: Joi.string().required(),
+      description: Joi.string().allow(""),
+      permalink: Joi.string().required(),
+      notification: Joi.string().allow(""),
+      image: Joi.string(),
```
seaerchin, line 23: could i confirm that the frontend omit this property if empty?

seaerchin (approved):
lgtm once extra test case is added and behaviour of FE clarified

validators/RequestSchema.js
alexanderleegs: Yep, the image isn't modifiable inside EditHomepage - so if the image property is originally missing, the request will not contain image also

newroutes/authenticatedSites/__tests__/Homepage.spec.js
alexanderleegs: Good catch, added in https://github.com/isomerpages/isomercms-backend/pull/331/commits/39c91db6702135cc1f48bae11ed7ce1601e20639

newroutes/authenticatedSites/__tests__/Homepage.spec.js
alexanderleegs: Makes sense, https://github.com/isomerpages/isomercms-backend/pull/331/commits/42e416cd091cad47605e8e264d8d0fc788dc3b2c

seaerchin (approved):
lgtm

PR #336, opened 2022-02-08
fix: update image for homepage if shareicon changes

alexanderleegs:
This PR fixes a bug in which users who changed the shareicon of their site would not have it reflected for the homepage.

Newly created sites start off with a default image used by the homepage which shows up for social media previews - this is stored as the `image` field in the `index.md` files in newly created repos. The repo would draw from this field when determining the image preview for the site, in the absence of a `shareicon` field in the `_config.yml` file. 

In one of our recent PRs, we introduced the ability to modify the `shareicon` of a site, which would override the image specified in the index file for all files which did not themselves specify an image. However, there was no way to modify the image in the original `index.md`, and it would have a higher priority than the general `shareicon` specified for the site for the homepage, so users would never be able to edit the image seen on social media previews for their homepage.

To fix this issue, we now update the image field in the homepage if the `shareicon` has been modified in settings.

As discussed offline, we have decided that this should be a temporary fix until a proper frontend flow has been decided for changing the homepage image, see https://github.com/isomerpages/isomercms-backend/issues/338 for more details.

### DIFF
--- a/services/configServices/SettingsService.js
+++ b/services/configServices/SettingsService.js
@@ -73,6 +73,8 @@ class SettingsService {
         if (updatedConfigContent.description)
           updatedHomepageFrontMatter.description =
             updatedConfigContent.description
+        if (updatedConfigContent.shareicon)
+          updatedHomepageFrontMatter.image = updatedConfigContent.shareicon
         await this.homepagePageService.update(reqDetails, {
           content: homepage.content.pageBody,
           frontMatter: updatedHomepageFrontMatter,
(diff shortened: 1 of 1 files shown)


gweiying:
https://github.com/isomerpages/isomerpages-template/blob/5c265559c75a804b959b9ad50ffb4381a04e578d/_includes/head.html#L26-L32

Pasting reference to [isomerpages/isomerpages-template](https://github.com/isomerpages/isomerpages-template) where the `page.image` field overrides the `site.shareicon` field.

gweiying (approved):
lgtm! 

As discussed offline and documented in the linked issue, this shouldn't be a permanent fix as we should separately allow users to modify their homepage page image and description if they want to have it different from the site shareicon. However, this is sufficient for now as multiple new Isomer users are affected by bug where their `page.image` is overriding their `site.shareicon`.

PR #337, opened 2022-02-08
Fix/handle empty directories

alexanderleegs:
This PR fixes a bug involving empty media folders, where we would return a 404 if the directory requested did not exist - this caused an issue for the base media folders (`/images` and `/files`), especially since repos start off with no `/files` directory. We have changed the behaviour on query to return an empty array instead of a 404 error.

This PR also updates the regex to disallow # and ? in media folder names and fixes a few failing tests.

### DIFF
--- a/services/directoryServices/MediaDirectoryService.js
+++ b/services/directoryServices/MediaDirectoryService.js
@@ -12,15 +12,11 @@ class MediaDirectoryService {
     this.gitHubService = gitHubService
   }
 
-  async listFiles(reqDetails, { directoryName }) {
-    // TODO: file preview handling
-    const { siteName } = reqDetails
-    if (!isMediaPathValid({ path: directoryName }))
-      throw new BadRequestError("Invalid media folder name")
-    const mediaType = directoryName.split("/")[0]
-    const { private: isPrivate } = await this.gitHubService.getRepoInfo(
-      reqDetails
-    )
+  /**
+   * Lists files in directory. Returns empty array if directory does not exist
+   * - useful for base media directories which do not have placeholder files
+   */
+  async listWithDefault(reqDetails, { directoryName }) {
     let files = []
     try {
       const retrievedFiles = await this.baseDirectoryService.list(reqDetails, {
@@ -31,6 +27,20 @@ class MediaDirectoryService {
       // return an empty list if directory does not exist
       if (error.status !== 404) throw error
     }
+    return files
+  }
+
+  async listFiles(reqDetails, { directoryName }) {
+    // TODO: file preview handling
+    const { siteName } = reqDetails
+    if (!isMediaPathValid({ path: directoryName }))
+      throw new BadRequestError("Invalid media folder name")
+    const mediaType = directoryName.split("/")[0]
+    const { private: isPrivate } = await this.gitHubService.getRepoInfo(
+      reqDetails
+    )
+    const files = await this.listWithDefault(reqDetails, { directoryName })
+    console.log(files)
 
     const resp = []
     for (const curr of files) {
--- a/services/directoryServices/MediaDirectoryService.js
+++ b/services/directoryServices/MediaDirectoryService.js
@@ -21,9 +21,17 @@ class MediaDirectoryService {
     const { private: isPrivate } = await this.gitHubService.getRepoInfo(
       reqDetails
     )
-    const files = await this.baseDirectoryService.list(reqDetails, {
-      directoryName,
-    })
+    let files = []
+    try {
+      const retrievedFiles = await this.baseDirectoryService.list(reqDetails, {
+        directoryName,
+      })
+      files = retrievedFiles
+    } catch (error) {
+      // return an empty list if directory does not exist
+      if (error.status !== 404) throw error
+    }
+
     const resp = []
     for (const curr of files) {
       if (curr.type === "dir") {
--- a/services/directoryServices/MediaDirectoryService.js
+++ b/services/directoryServices/MediaDirectoryService.js
@@ -40,7 +40,6 @@ class MediaDirectoryService {
       reqDetails
     )
     const files = await this.listWithDefault(reqDetails, { directoryName })
-    console.log(files)
 
     const resp = []
     for (const curr of files) {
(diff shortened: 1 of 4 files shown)


validators/validators.js
```diff
@@ -2,7 +2,7 @@ const specialCharactersRegexTest = /[~%^*_+\-./\\`;~{}[\]"<>]/
 const dateRegexTest = /^[0-9]{4}-[0-9]{2}-[0-9]{2}$/
 
 const mediaSpecialCharactersRegexTest = /[~%^#*+\./\\`;~{}[\]"<>]/ // Allows dashes
-const mediaSubfolderRegexText = /^(images|files|(images|files)\/[^~%^*+\.\\`;~{}[\]"<>]+)$/
+const mediaSubfolderRegexText = /^(images|files|(images|files)\/[^#?~%^*+\.\\`;~{}[\]"<>]+)$/
```
seaerchin, line 5: could we add a brief explanation on what this regex does? it's pretty undecipherable to me .-.

services/directoryServices/MediaDirectoryService.js
```diff
@@ -21,9 +21,17 @@ class MediaDirectoryService {
...
+    try {
+      const retrievedFiles = await this.baseDirectoryService.list(reqDetails, {
+        directoryName,
+      })
+      files = retrievedFiles
+    } catch (error) {
+      // return an empty list if directory does not exist
+      if (error.status !== 404) throw error
+    }
+
```
seaerchin, line 34: i think `try/catch` should be used sparingly, as they place the burden of understanding on the reader, due to the control flow now having a different path. As this file already has alot of control flow handling (`if/else` and `continue` and `throw`s), i think we should be kind to our future selves and avoid the use of the `try/catch` here and save some poor brain cells from having to work.

i would be more inclined to set this as a private method called `listWithDefault`, which can have this behaviour and this section could be structured as follows: 
```js
const files = await this.listWithDefault(reqDetails, {directoryName})
```

seaerchin: also, i think it's good to note down that we require the defaulting behaviour as there's no default `file` folder being created, which leads to crashes when we access resources/files tab. 

this gives context on why the method is required and devs will take more care when working on the method in the future. (as well as allowing them to check that there's no regression bugs when they do work on it)

validators/validators.js
alexanderleegs: Good point, https://github.com/isomerpages/isomercms-backend/pull/337/commits/efee09e07b57ca30ca147bb0f334e478c4d26a41

services/directoryServices/MediaDirectoryService.js
alexanderleegs: I think that's a good suggestion, i've made the changes here https://github.com/isomerpages/isomercms-backend/pull/337/commits/856847adfad112672bb1625152277bc0f6f6d251

validators/validators.js
seaerchin: i'm assuming that the banned characters are this section of the regex: `[^#?~%^*+\.\\`;~{}[\]"<>]+`? also it seems like it allows `files/<subdir>` so perhaps we should also update the comment accordingly.

sidenote: do we allow `<images/files>/` (note the extra `/`)

alexanderleegs: changed in https://github.com/isomerpages/isomercms-backend/pull/337/commits/e10437f12443782a5d800ca66cf7fc982aeb1cee

We disallow ending with a `/` - we use slashes to denote a new directory only

services/directoryServices/MediaDirectoryService.js
seaerchin: wdyt about this method belonging in the `BaseDirectoryService` as it seems like it might be used by more services in the future! 

also, could we add that it crashes when there's no placeholder files so future devs will take more care in removing this!

alexanderleegs: The crashing is due to the error being unhandled on the frontend - I don't think this information should be stored on the backend side, since it's expected behaviour that querying a non-existent directory for its files should throw a 404 error.

I think we might also want to keep this method in the media directory service specifically - media directories are unique in that they don't need any particular processing for a directory to work, other directories require some sort of additional processing (e.g. editing collection.yml, resource directory setup) and hence should have some distinction between properly created folders and not. I'd actually prefer to keep this behaviour for media folders as well (by the setup of placeholder files), but given that most repos already have been set up with no placeholder files in the base media directories, i think it might be unavoidable for media directory. I'm open to moving to this to base directory service if you feel strongly about this though!

seaerchin: ah ok, thanks for clarifying the expected behaviour of this vs the base directory! in that case i concur and we should store it here.

wrt the error being unhandled on FE - actually i think this merits further thought because i also agree that returning a 404 is the correct behaviour. unfortunately, this routes to the `not-found` page due to the behaviour of `useGetDirectoryHook`.

i think the fix here is good enough for now but it does mean that we have exceptional behaviour here, which is not ideal. should we instead structure a fix like so: 

1. FE requests /GET 
2. BE responds 404 
3. FE requests /POST (to create the directory since it doesn't exist) 
4. BE responds 200
5. FE displays data

or is this too complicated? (since we also have to ensure that the files tab on the sidebar has this behaviour)

alexanderleegs: We've discussed this issue in the past, though i think we might want to revisit it - a possible issue with this is it will result in the creation of a lot of unnecessary folders whenever a user types in the wrong url, and thereafter the folder will start to appear in their list of folders, which can clutter up the list very easily - i think folder creation should be a deliberate action taken by the user.

seaerchin: i don't really feel too strongly about this HAHA so actually i'm fine as-is! thanks for explaining and giving context!

services/directoryServices/MediaDirectoryService.js
```diff
@@ -21,9 +39,9 @@ class MediaDirectoryService {
     const { private: isPrivate } = await this.gitHubService.getRepoInfo(
       reqDetails
     )
-    const files = await this.baseDirectoryService.list(reqDetails, {
-      directoryName,
-    })
+    const files = await this.listWithDefault(reqDetails, { directoryName })
+    console.log(files)
```
gweiying, line 43: Unused `console.log`

gweiying (approved):
lgtm, small nit on unused logging statement

services/directoryServices/MediaDirectoryService.js
alexanderleegs: Removed in https://github.com/isomerpages/isomercms-backend/pull/337/commits/8cb9f79d097d7814c053127c731ef1f7206747b9

PR #341, opened 2022-02-11
Refactor/sites

alexanderleegs:
## Overview

This PR introduces the sites flow refactor. It introduces a newly created SitesRouter and SitesService, as well as updating the existing GithubService to have an additional checkHasAccess method.

**Breaking Changes**

**Features**:

Sites router
- GET / - retrieve list of all sites
- GET /:siteName - check if user has access
- GET /:siteName/lastUpdated - retrieve date of last update
- GET /:siteName/stagingUrl - retrieve staging url

**Notes**:
The SitesService uses both the existing GithubService as well as calling axios itself - this is due to the existing GithubService being tailored specifically for the sites endpoint. I'm open to discussion of whether to initialise another axios instance inside the GithubService specifically to deal with this endpoint, but I currently set it without as this is the only github endpoint with the `orgs/repo` prefix.

### DIFF
--- a/newroutes/__tests__/Sites.spec.js
+++ b/newroutes/__tests__/Sites.spec.js
@@ -17,7 +17,7 @@ const siteName = "siteName"
 
 const reqDetails = { siteName, accessToken }
 
-describe("Unlinked Pages Router", () => {
+describe("Sites Router", () => {
   const mockSitesService = {
     getSites: jest.fn(),
     checkHasAccess: jest.fn(),
@@ -76,7 +76,7 @@ describe("Unlinked Pages Router", () => {
       })
     })
 
-    it("allows if user has no access to a site", async () => {
+    it("allows if user has access to a site", async () => {
       await request(app).get(`/${siteName}`).expect(200)
 
       expect(mockSitesService.checkHasAccess).toHaveBeenCalledWith(reqDetails, {
--- a/newroutes/__tests__/Sites.spec.js
+++ b/newroutes/__tests__/Sites.spec.js
@@ -1,0 +1,104 @@
+const cookieParser = require("cookie-parser")
+const express = require("express")
+const request = require("supertest")
+
+const { NotFoundError } = require("@errors/NotFoundError")
+
+const { errorHandler } = require("@middleware/errorHandler")
+const { attachReadRouteHandlerWrapper } = require("@middleware/routeHandler")
+
+const { SitesRouter } = require("../sites")
+
+// Can't set request fields - will always be undefined
+const userId = undefined
+const accessToken = undefined
+
+const siteName = "siteName"
+
+const reqDetails = { siteName, accessToken }
+
+describe("Unlinked Pages Router", () => {
+  const mockSitesService = {
+    getSites: jest.fn(),
+    checkHasAccess: jest.fn(),
+    getLastUpdated: jest.fn(),
+    getStagingUrl: jest.fn(),
+  }
+
+  const router = new SitesRouter({
+    sitesService: mockSitesService,
+  })
+
+  const app = express()
+  app.use(express.json({ limit: "7mb" }))
+  app.use(express.urlencoded({ extended: false }))
+  app.use(cookieParser())
+
+  // We can use read route handler here because we don't need to lock the repo
+  app.get("/", attachReadRouteHandlerWrapper(router.getSites))
+  app.get("/:siteName", attachReadRouteHandlerWrapper(router.checkHasAccess))
+  app.get(
+    "/:siteName/lastUpdated",
+    attachReadRouteHandlerWrapper(router.getLastUpdated)
+  )
+  app.get(
+    "/:siteName/stagingUrl",
+    attachReadRouteHandlerWrapper(router.getStagingUrl)
+  )
+  app.use(errorHandler)
+
+  beforeEach(() => {
+    jest.clearAllMocks()
+  })
+
+  describe("getSites", () => {
+    it("returns the list of sites accessible to the user", async () => {
+      const sitesResp = ["site1", "site2"]
+      mockSitesService.getSites.mockResolvedValueOnce(sitesResp)
+      const resp = await request(app).get(`/`).expect(200)
+      expect(resp.body).toStrictEqual({ siteNames: sitesResp })
+      expect(mockSitesService.getSites).toHaveBeenCalledWith({ accessToken })
+    })
+  })
+
+  describe("checkHasAccess", () => {
+    it("rejects if user has no access to a site", async () => {
+      mockSitesService.checkHasAccess.mockRejectedValueOnce(
+        new NotFoundError("")
+      )
+      await request(app).get(`/${siteName}`).expect(404)
+      expect(mockSitesService.checkHasAccess).toHaveBeenCalledWith(reqDetails, {
+        userId,
+      })
+    })
+
+    it("allows if user has no access to a site", async () => {
+      await request(app).get(`/${siteName}`).expect(200)
+      expect(mockSitesService.checkHasAccess).toHaveBeenCalledWith(reqDetails, {
+        userId,
+      })
+    })
+  })
+
+  describe("getLastUpdated", () => {
+    it("returns the last updated time", async () => {
+      const lastUpdated = "last-updated"
+      mockSitesService.getLastUpdated.mockResolvedValueOnce(lastUpdated)
+      const resp = await request(app)
+        .get(`/${siteName}/lastUpdated`)
+        .expect(200)
+      expect(resp.body).toStrictEqual({ lastUpdated })
+      expect(mockSitesService.getLastUpdated).toHaveBeenCalledWith(reqDetails)
+    })
+  })
+
+  describe("getStagingUrl", () => {
+    it("returns the last updated time", async () => {
+      const stagingUrl = "staging-url"
+      mockSitesService.getStagingUrl.mockResolvedValueOnce(stagingUrl)
+      const resp = await request(app).get(`/${siteName}/stagingUrl`).expect(200)
+      expect(resp.body).toStrictEqual({ stagingUrl })
+      expect(mockSitesService.getStagingUrl).toHaveBeenCalledWith(reqDetails)
+    })
+  })
+})
--- a/newroutes/__tests__/Sites.spec.js
+++ b/newroutes/__tests__/Sites.spec.js
@@ -55,7 +55,9 @@ describe("Unlinked Pages Router", () => {
     it("returns the list of sites accessible to the user", async () => {
       const sitesResp = ["site1", "site2"]
       mockSitesService.getSites.mockResolvedValueOnce(sitesResp)
+
       const resp = await request(app).get(`/`).expect(200)
+
       expect(resp.body).toStrictEqual({ siteNames: sitesResp })
       expect(mockSitesService.getSites).toHaveBeenCalledWith({ accessToken })
     })
@@ -66,7 +68,9 @@ describe("Unlinked Pages Router", () => {
       mockSitesService.checkHasAccess.mockRejectedValueOnce(
         new NotFoundError("")
       )
+
       await request(app).get(`/${siteName}`).expect(404)
+
       expect(mockSitesService.checkHasAccess).toHaveBeenCalledWith(reqDetails, {
         userId,
       })
@@ -74,6 +78,7 @@ describe("Unlinked Pages Router", () => {
 
     it("allows if user has no access to a site", async () => {
       await request(app).get(`/${siteName}`).expect(200)
+
       expect(mockSitesService.checkHasAccess).toHaveBeenCalledWith(reqDetails, {
         userId,
       })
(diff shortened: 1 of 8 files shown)


services/utilServices/SitesService.js
```diff
@@ -0,0 +1,133 @@
+const axios = require("axios")
+const Bluebird = require("bluebird")
+const _ = require("lodash")
+
+// Import error
```
seaerchin, line 5: nit: delete wrong comment

services/utilServices/SitesService.js
```diff
@@ -0,0 +1,133 @@
...
+
+    // Simultaneously retrieve all isomerpages repos
+    const paramsArr = []
+    for (let i = 0; i < ISOMERPAGES_REPO_PAGE_COUNT; i += 1) {
+      paramsArr.push({
+        per_page: GH_MAX_REPO_COUNT,
+        sort: "full_name",
+        page: i + 1,
+      })
+    }
```
seaerchin, line 60: wdyt about writing this as a map?

```js 
const paramsArr = _.fill(Array(ISOMERPAGES_REPO_PAGE_COUNT), null).map((_, idx) => ({
  per_page: GH_MAX_REPO_COUNT,
  sort: "full_name",
  page: i + 1,
}))```

File: services/utilServices/SitesService.js [hunk 3]
```diff
@@ -0,0 +1,133 @@
...
+
+    const sites = await Bluebird.map(paramsArr, async (params) => {
+      const resp = await axios.get(endpoint, {
+        params,
+        headers: {
+          Authorization: `token ${accessToken}`,
+          "Content-Type": "application/json",
+        },
```
seaerchin, line 68: i think i've seen this repeated a ton and it merits its own helper method. might be out of scope (and not that important yet), so maybe we could file an issue?

services/utilServices/SitesService.js
```diff
@@ -0,0 +1,133 @@
...
+      })
+
+      return resp.data
+        .map((repoData) => {
+          const {
+            updated_at: updatedAt,
+            permissions,
+            name,
+            private: isPrivate,
+          } = repoData
```
seaerchin, line 78: wdyt about destructuring here?

services/utilServices/SitesService.js
```diff
@@ -0,0 +1,133 @@
...
+        })
+        .filter(
+          (repoData) =>
+            repoData.permissions.push === true &&
+            !ISOMER_ADMIN_REPOS.includes(repoData.repoName)
+        )
+    })
+
+    const flattenedSites = _.flatten(sites)
+    return flattenedSites
```
seaerchin, line 95: nit: could we inline this?

services/utilServices/SitesService.js
```diff
@@ -0,0 +1,133 @@
...
+    const numDaysAgo = Math.floor(gapInUpdate / (1000 * 60 * 60 * 24))
+    // return a message for number of days ago repo was last updated
+    switch (numDaysAgo) {
+      case 0:
+        return "Updated today"
+      case 1:
+        return "Updated 1 day ago"
+      default:
+        return `Updated ${numDaysAgo} days ago`
+    }
```
seaerchin, line 46: i'm not sure that this should be done on the backend - i'm of hte opinion that the backend returns purely data and makes no judgement on how that data is consumed. this limits flexibility on the FE side (eg: if another FE component calls the same endpoint but wishes to make a `Updated last week` statement) and loses information.

services/utilServices/SitesService.js
```diff
@@ -0,0 +1,133 @@
...
+    let stagingUrl
+
+    if (description) {
+      // Retrieve the url from the description - repo descriptions have varying formats, so we look for the first link
+      const descTokens = description.replace("/;/g", " ").split(" ")
+      // Staging urls also contain staging in their url
+      stagingUrl = descTokens.find(
+        (token) => token.includes("http") && token.includes("staging")
+      )
+    }
```
seaerchin, line 125: This implies that if users were to change their description, we could potentially have a misbehaving branch! In that case, would it be sensible to log the necessary details here so that we can identify the branch?

services/utilServices/SitesService.js
```diff
@@ -0,0 +1,133 @@
...
+
+    return stagingUrl
```
seaerchin, line 127: this implies that `stagingUrl` can be `undefined`, in which case we should adopt some preventive actions rather than doing nothing. perhaps a good backup here might be to direct them to the 404 page?

services/utilServices/__tests__/SitesService.spec.js
```diff
@@ -0,0 +1,133 @@
...
+    it("Checks when site was last updated", async () => {
+      mockGithubService.getRepoInfo.mockResolvedValue(repoInfo)
+      await expect(service.getLastUpdated(reqDetails)).resolves.toEqual(
+        `Updated ${numDaysAgo} days ago`
+      )
+      expect(mockGithubService.getRepoInfo).toHaveBeenCalledWith(reqDetails)
+    })
+  })
+
+  describe("getStagingUrl", () => {
```
seaerchin, line 108: could we add a test case when both config and description lack staging url?

seaerchin (commented):
seems fine overall, just some minor comments wrt style/odd edge cases in our codebase!

newroutes/__tests__/Sites.spec.js
```diff
@@ -0,0 +1,104 @@
...
+
+// Can't set request fields - will always be undefined
+const userId = undefined
+const accessToken = undefined
+
+const siteName = "siteName"
+
+const reqDetails = { siteName, accessToken }
+
+describe("Unlinked Pages Router", () => {
```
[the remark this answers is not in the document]
alexanderleegs: https://github.com/isomerpages/isomercms-backend/pull/341/commits/8fe141dafd56c2d52886d11a9b89512a492c1b23

newroutes/sites.js
```diff
@@ -0,0 +1,73 @@
...
+
+  async getSites(req, res) {
+    const { accessToken } = req
+    const siteNames = await this.sitesService.getSites({ accessToken })
+    return res.status(200).json({ siteNames })
+  }
+
+  async checkHasAccess(req, res) {
+    const { accessToken, userId } = req
+    const { siteName } = req.params
```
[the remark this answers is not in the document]
alexanderleegs: https://github.com/isomerpages/isomercms-backend/pull/341/commits/132b081696e46097d6ff14ef30e06596530df4d2

newroutes/sites.js
```diff
@@ -0,0 +1,73 @@
...
+    const lastUpdated = await this.sitesService.getLastUpdated({
+      accessToken,
+      siteName,
+    })
+    return res.status(200).json({ lastUpdated })
+  }
+
+  async getStagingUrl(req, res) {
+    const { accessToken } = req
+    const { siteName } = req.params
```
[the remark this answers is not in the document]
alexanderleegs: https://github.com/isomerpages/isomercms-backend/pull/341/commits/132b081696e46097d6ff14ef30e06596530df4d2

services/utilServices/SitesService.js
alexanderleegs: https://github.com/isomerpages/isomercms-backend/pull/341/commits/e64781021882ea0289420e5ee8d68a4015f14341

services/utilServices/SitesService.js
alexanderleegs: Yeah that's probably better, https://github.com/isomerpages/isomercms-backend/pull/341/commits/e64781021882ea0289420e5ee8d68a4015f14341

newroutes/sites.js
```diff
@@ -0,0 +1,73 @@
...
+
+    await this.sitesService.checkHasAccess(
+      {
+        accessToken,
+        siteName,
+      },
+      { userId }
+    )
+    return res.status(200).send("OK")
```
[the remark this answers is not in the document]
alexanderleegs: https://github.com/isomerpages/isomercms-backend/pull/341/commits/e64781021882ea0289420e5ee8d68a4015f14341

services/utilServices/SitesService.js
alexanderleegs: f3b0110

services/utilServices/SitesService.js
alexanderleegs: Great suggestion, I've made it throw an error if no staging url is found: 6d645cf

services/utilServices/SitesService.js
alexanderleegs: See above, 6d645cf

services/utilServices/__tests__/SitesService.spec.js
alexanderleegs: Added in 6d645cf!

services/utilServices/SitesService.js
alexanderleegs: I think that makes a lot of sense, I've changed it in 66c43c8 and made an issue on the frontend so we don't forget when refactoring sites

services/utilServices/SitesService.js
alexanderleegs: e647810

services/utilServices/SitesService.js
alexanderleegs: This is used only in `SitesService`, `AuthService` and `GithubService` actually, since this is only relevant when querying Github - I'm wondering if the solution is to create a more general misc service to query endpoints which don't start with `https://api.github.com/repos/`? It felt out of place for these 2 services to have axios calls when the only other place which uses axios is `GithubService`

newmiddleware/auth.js
```diff
@@ -255,6 +255,12 @@ class AuthMiddleware {
     auth.delete("/v2/auth/logout", this.noVerify)
     auth.get("/v2/auth/whoami", this.whoamiAuth)
 
+    // Sites
+    auth.get("/v2/sites", this.verifyJwt)
+    auth.get("/v2/sites/:siteName", this.verifyJwt)
+    auth.get("/v2/sites/:siteName/lastUpdated", this.verifyJwt)
+    auth.get("/v2/sites/:siteName/stagingUrl", this.verifyJwt)
```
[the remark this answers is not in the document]
alexanderleegs: https://github.com/isomerpages/isomercms-backend/pull/341/commits/4a921fe26e9364b9096fe910a55f3467633f986f

services/utilServices/SitesService.js
seaerchin: actually you're right - maybe we should just defer the methods that require headers to GHA and make our main services agnostic of the data provider (github) wdyt?

seaerchin (commented):
looks good and will approve once the test comment is fixed! as an aside, i think destructuring should take place on the params level but this isn't important as we do destructure later on

services/utilServices/SitesService.js
alexanderleegs: Yup i think that makes sense - i'll file an issue to tackle this in a future PR

seaerchin (approved):
minor test comments but lgtm otherwise! just go ahead n merge when the comment is fixed! 

as an aside, i think we shouldn't rebase _previous_ commits as prior review comments on that commit hash will be lost. also, you might wanna double check if you lost work as the prior nit about not returning `"Ok"` got reverted after being fixed (idt there's anything major though but just err on the side of safety)

newroutes/sites.js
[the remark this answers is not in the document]
seaerchin: ^this commit got lost, unsure if there are others. i'm fine iwth not doing this as it's relatively minor

services/utilServices/__tests__/SitesService.spec.js
```diff
@@ -0,0 +1,143 @@
...
+      }))
+      axios.get.mockImplementationOnce(() => ({
+        data: [],
+      }))
+
+      await expect(service.getSites({ accessToken })).resolves.toMatchObject(
+        expectedResp
+      )
+
+      expect(axios.get).toHaveBeenCalledTimes(3)
```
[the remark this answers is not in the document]
alexanderleegs: Going to leave it as is for now - we'll refactor this out into a general axios call in a future PR!

services/utilServices/__tests__/SitesService.spec.js
```diff
@@ -0,0 +1,143 @@
...
+  })
+
+  describe("getStagingUrl", () => {
+    const stagingUrl = "https://repo-staging.netlify.app"
+    it("Retrieves the staging url for a site from config if available", async () => {
+      mockConfigYmlService.read.mockResolvedValue({
+        content: {
+          staging: stagingUrl,
+        },
+      })
```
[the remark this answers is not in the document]
alexanderleegs: https://github.com/isomerpages/isomercms-backend/pull/341/commits/fb05904b7799e50b5a89a5cce4323773ea49105d

newroutes/sites.js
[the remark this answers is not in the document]
alexanderleegs: The commit was preserved as 9b05ff7 - rebasing seems to overwrite the original commit hash! I did reintroduce the `.send("OK")` in a later commit because one of the tests was failing, and elected to use `.send("OK")` to keep it in line with how we return responses in the rest of our routers

PR #349, opened 2022-02-14
Refactor/contact us

### LINKED ISSUE #343: [backend refactor] Contact us flow

alexanderleegs:
## Problem

This PR refactors the contact us flow. It adds a new contact us router and contact us service, as well as tests for the newly introduced components.

Closes #343 

**Breaking Changes**

**Features**:

The contact us router contains the following endpoints:

GET /:siteName/contactUs - Retrieve contact us page information
POST /:siteName/contactUs- Update contact us page information

The customised contact us endpoint is newly introduced, as previously we were using the unlinked pages endpoints to retrieve this data - we have decided to split this into its own flow as the contact us page is significantly different from other pages, as all of its data is contained inside the front matter.

### DIFF
--- a/fixtures/app.js
+++ b/fixtures/app.js
@@ -1,0 +1,16 @@
+const express = require("express")
+
+const { errorHandler } = require("@middleware/errorHandler")
+
+function generateRouter(router) {
+  const app = express()
+  app.use(express.json({ limit: "7mb" }))
+  app.use(express.urlencoded({ extended: false }))
+  app.use(router)
+  app.use(errorHandler)
+  return app
+}
+
+module.exports = {
+  generateRouter,
+}
--- a/newroutes/__tests__/ContactUs.spec.js
+++ b/newroutes/__tests__/ContactUs.spec.js
@@ -6,6 +6,7 @@ const { errorHandler } = require("@middleware/errorHandler")
 const { attachReadRouteHandlerWrapper } = require("@middleware/routeHandler")
 
 const { contactUsContent, contactUsSha } = require("@fixtures/contactUs")
+const { generateRouter } = require("@root/fixtures/app")
 
 const { ContactUsRouter } = require("../contactUs")
 
@@ -19,20 +20,17 @@ describe("ContactUs Router", () => {
     contactUsPageService: mockContactUsPageService,
   })
 
-  const app = express()
-  app.use(express.json({ limit: "7mb" }))
-  app.use(express.urlencoded({ extended: false }))
-
+  const subrouter = express()
   // We can use read route handler here because we don't need to lock the repo
-  app.get(
+  subrouter.get(
     "/:siteName/contactUs",
     attachReadRouteHandlerWrapper(router.readContactUs)
   )
-  app.post(
+  subrouter.post(
     "/:siteName/contactUs",
     attachReadRouteHandlerWrapper(router.updateContactUs)
   )
-  app.use(errorHandler)
+  const app = generateRouter(subrouter)
 
   const siteName = "test-site"
   const accessToken = undefined // Can't set request fields - will always be undefined
--- a/newroutes/__tests__/ContactUs.spec.js
+++ b/newroutes/__tests__/ContactUs.spec.js
@@ -1,0 +1,117 @@
+const express = require("express")
+const _ = require("lodash")
+const request = require("supertest")
+
+const { errorHandler } = require("@middleware/errorHandler")
+const { attachReadRouteHandlerWrapper } = require("@middleware/routeHandler")
+
+const { contactUsContent, contactUsSha } = require("@fixtures/contactUs")
+
+const { ContactUsRouter } = require("../contactUs")
+
+describe("ContactUs Router", () => {
+  const mockContactUsPageService = {
+    read: jest.fn(),
+    update: jest.fn(),
+  }
+
+  const router = new ContactUsRouter({
+    contactUsPageService: mockContactUsPageService,
+  })
+
+  const app = express()
+  app.use(express.json({ limit: "7mb" }))
+  app.use(express.urlencoded({ extended: false }))
+
+  // We can use read route handler here because we don't need to lock the repo
+  app.get(
+    "/:siteName/contactUs",
+    attachReadRouteHandlerWrapper(router.readContactUs)
+  )
+  app.post(
+    "/:siteName/contactUs",
+    attachReadRouteHandlerWrapper(router.updateContactUs)
+  )
+  app.use(errorHandler)
+
+  const siteName = "test-site"
+  const accessToken = undefined // Can't set request fields - will always be undefined
+
+  const reqDetails = { siteName, accessToken }
+
+  beforeEach(() => {
+    jest.clearAllMocks()
+  })
+
+  describe("readContactUs", () => {
+    const expectedResponse = {
+      sha: contactUsSha,
+      content: contactUsContent,
+    }
+    mockContactUsPageService.read.mockResolvedValue(expectedResponse)
+    it("retrieves contactUs details", async () => {
+      const resp = await request(app).get(`/${siteName}/contactUs`).expect(200)
+      expect(resp.body).toStrictEqual(expectedResponse)
+      expect(mockContactUsPageService.read).toHaveBeenCalledWith(reqDetails)
+    })
+  })
+
+  describe("updateContactUs", () => {
+    it("rejects requests with invalid body", async () => {
+      await request(app).post(`/${siteName}/contactUs`).send({}).expect(400)
+    })
+
+    it("rejects requests with incomplete body", async () => {
+      // Missing layout
+      const incompleteDetails = _.cloneDeep(contactUsContent)
+      const reqBody = {
+        content: incompleteDetails,
+        sha: contactUsSha,
+      }
+      delete incompleteDetails.frontMatter.layout
+      await request(app)
+        .post(`/${siteName}/contactUs`)
+        .send(reqBody)
+        .expect(400)
+    })
+
+    it("accepts valid contactUs update requests and returns the details of the file updated", async () => {
+      const reqBody = {
+        content: contactUsContent,
+        sha: contactUsSha,
+      }
+      const expectedServiceInput = {
+        content: contactUsContent.pageBody,
+        frontMatter: contactUsContent.frontMatter,
+        sha: contactUsSha,
+      }
+      await request(app)
+        .post(`/${siteName}/contactUs`)
+        .send(reqBody)
+        .expect(200)
+      expect(mockContactUsPageService.update).toHaveBeenCalledWith(
+        reqDetails,
+        expectedServiceInput
+      )
+    })
+
+    it("rejects contactUs update requests with additional unspecified fields", async () => {
+      const extraUpdateDetails = _.cloneDeep(contactUsContent)
+      // Add extra unspecified field
+      extraUpdateDetails.frontMatter.extra = ""
+      const reqBody = {
+        content: extraUpdateDetails,
+        sha: contactUsSha,
+      }
+      const expectedServiceInput = {
+        content: extraUpdateDetails.pageBody,
+        frontMatter: extraUpdateDetails.frontMatter,
+        sha: contactUsSha,
+      }
+      await request(app)
+        .post(`/${siteName}/contactUs`)
+        .send(reqBody)
+        .expect(400)
+    })
+  })
+})
--- a/newroutes/__tests__/ContactUs.spec.js
+++ b/newroutes/__tests__/ContactUs.spec.js
@@ -49,8 +49,10 @@ describe("ContactUs Router", () => {
       content: contactUsContent,
     }
     mockContactUsPageService.read.mockResolvedValue(expectedResponse)
+
     it("retrieves contactUs details", async () => {
       const resp = await request(app).get(`/${siteName}/contactUs`).expect(200)
+
       expect(resp.body).toStrictEqual(expectedResponse)
       expect(mockContactUsPageService.read).toHaveBeenCalledWith(reqDetails)
     })
@@ -69,6 +71,7 @@ describe("ContactUs Router", () => {
         sha: contactUsSha,
       }
       delete incompleteDetails.frontMatter.layout
+
       await request(app)
         .post(`/${siteName}/contactUs`)
         .send(reqBody)
@@ -85,10 +88,12 @@ describe("ContactUs Router", () => {
         frontMatter: contactUsContent.frontMatter,
         sha: contactUsSha,
       }
+
       await request(app)
         .post(`/${siteName}/contactUs`)
         .send(reqBody)
         .expect(200)
+
       expect(mockContactUsPageService.update).toHaveBeenCalledWith(
         reqDetails,
         expectedServiceInput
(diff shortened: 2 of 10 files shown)


newroutes/contactUs.js
```diff
@@ -0,0 +1,71 @@
...
+
+class ContactUsRouter {
+  constructor({ contactUsPageService }) {
+    this.contactUsPageService = contactUsPageService
+    // We need to bind all methods because we don't invoke them from the class directly
+    autoBind(this)
+  }
+
+  // Read contactUs file
+  async readContactUs(req, res) {
```
seaerchin, line 22: same comment as prev PR: should we prefer destructuring where possible?

services/fileServices/MdPageServices/ContactUsPageService.js
```diff
@@ -0,0 +1,56 @@
...
+      {
+        fileName: CONTACT_US_FILE_NAME,
+        directoryName: CONTACT_US_DIRECTORY_NAME,
+      }
+    )
+    const { frontMatter, pageContent } = retrieveDataFromMarkdown(rawContent)
+    const { content: footerContent } = await this.footerYmlService.read(
+      reqDetails
+    )
+    frontMatter.feedback = footerContent.feedback
```
seaerchin, line 27: could we add a comment on why we set `frontMatter.feedback` to `footerContent.feedback`? it seems abit unclear to me .-.

also, is it guaranteed that the `frontMatter` has no `feedback` property otherwise this is a destructive update

services/fileServices/MdPageServices/ContactUsPageService.js
```diff
@@ -0,0 +1,56 @@
...
+      fileContent: newContent,
+      sha,
+      fileName: CONTACT_US_FILE_NAME,
+      directoryName: CONTACT_US_DIRECTORY_NAME,
+    })
+    const {
+      content: footerContent,
+      sha: footerSha,
+    } = await this.footerYmlService.read(reqDetails)
+    footerContent.feedback = frontMatter.feedback
```
seaerchin, line 43: on this point, is my understanding that we maintain the `feedback` property in `footerContent` and when something (here, `frontMatter`) requires it, we read from `footerContent` and write to `footerContent` on update?

services/fileServices/MdPageServices/__tests__/ContactUsPageService.spec.js
```diff
@@ -0,0 +1,115 @@
...
+    content: mockFooterContent,
+    sha: mockFooterSha,
+  })
+
+  describe("Read", () => {
+    mockGithubService.read.mockResolvedValue({
+      content: mockRawContactUsContent,
+      sha: mockContactUsSha,
+    })
+    it("Reading the contact us page works correctly", async () => {
```
seaerchin, line 71: could we also have sad path tests - ie, when the `read` from `gitHubService` or `footerYmlService` fails, the error propagated is correct

services/fileServices/MdPageServices/__tests__/ContactUsPageService.spec.js
```diff
@@ -0,0 +1,115 @@
...
+        directoryName: CONTACT_US_DIRECTORY_NAME,
+      })
+      expect(mockFooterYmlService.read).toHaveBeenCalledWith(reqDetails)
+    })
+  })
+
+  describe("Update", () => {
+    const oldSha = "54321"
+    mockGithubService.update.mockResolvedValue({ newSha: mockContactUsSha })
+    it("Updating page content works correctly", async () => {
```
seaerchin, line 90: same as above, we should also add 1 failing test case

validators/RequestSchema.js
```diff
@@ -11,6 +11,51 @@ const ItemSchema = FileSchema.keys({
...
 })
 
+// Contact Us
+const UpdateContactUsSchema = Joi.object({
+  content: Joi.object({
+    frontMatter: Joi.object({
+      layout: Joi.string().required(),
+      title: Joi.string().required(),
+      permalink: Joi.string().required(),
+      feedback: Joi.string(),
```
seaerchin, line 21: based on what you've said previously, what happens here when an empty feedback section is submitted? would there be a property like so: `{feedback: ""}`, thus causing this to fail validation?

seaerchin (commented):
minor comments!

validators/RequestSchema.js
alexanderleegs: Thanks for catching - this would cause a failed validation, I've updated it to allow for empty feedback in https://github.com/isomerpages/isomercms-backend/pull/349/commits/13492c4cc615c662479433ccd1d25b6f342fdc08

services/fileServices/MdPageServices/__tests__/ContactUsPageService.spec.js
alexanderleegs: Good point, I've added it in https://github.com/isomerpages/isomercms-backend/pull/349/commits/78c9173af27ab70898c3129fd88d23490291118f

services/fileServices/MdPageServices/__tests__/ContactUsPageService.spec.js
alexanderleegs: https://github.com/isomerpages/isomercms-backend/pull/349/commits/78c9173af27ab70898c3129fd88d23490291118f

services/fileServices/MdPageServices/ContactUsPageService.js
alexanderleegs: Due to template intricacies, the feedback url is used in the footer and the contact-us page, but only stored in the footer - the contact-us link to the feedback page is taken from the feedback url stored in the `footer.yml` file. A feedback parameter in the contact us front matter would not be used anywhere, so it's okay to remove it even if present. I've added a comment as to why we retrieve footer information in https://github.com/isomerpages/isomercms-backend/pull/349/commits/33bfb3dacb3517079edb9acaba5afacda4b2b9c8

services/fileServices/MdPageServices/ContactUsPageService.js
alexanderleegs: See above!

newroutes/contactUs.js
alexanderleegs: Yup, changed in https://github.com/isomerpages/isomercms-backend/pull/349/commits/d162665f610071571ef5c1eadbdd57fc78eadb50

newroutes/__tests__/ContactUs.spec.js
```diff
@@ -0,0 +1,117 @@
...
+  const siteName = "test-site"
+  const accessToken = undefined // Can't set request fields - will always be undefined
+
+  const reqDetails = { siteName, accessToken }
+
+  beforeEach(() => {
+    jest.clearAllMocks()
+  })
+
+  describe("readContactUs", () => {
```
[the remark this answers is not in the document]
alexanderleegs: https://github.com/isomerpages/isomercms-backend/pull/349/commits/78c9173af27ab70898c3129fd88d23490291118f

newroutes/__tests__/ContactUs.spec.js
```diff
@@ -0,0 +1,117 @@
...
+    update: jest.fn(),
+  }
+
+  const router = new ContactUsRouter({
+    contactUsPageService: mockContactUsPageService,
+  })
+
+  const app = express()
+  app.use(express.json({ limit: "7mb" }))
+  app.use(express.urlencoded({ extended: false }))
```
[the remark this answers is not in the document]
alexanderleegs: That makes sense, I've added a helper function to generate the router here - https://github.com/isomerpages/isomercms-backend/pull/349/commits/2f8cc3c8640f6188c65a9c96fac7880552679483, will replicate this on the other router tests in a future PR

newroutes/__tests__/ContactUs.spec.js
```diff
@@ -0,0 +1,121 @@
...
+    })
+
+    it("rejects requests with incomplete body", async () => {
+      // Missing layout
+      const incompleteDetails = _.cloneDeep(contactUsContent)
+      const reqBody = {
+        content: incompleteDetails,
+        sha: contactUsSha,
+      }
+      delete incompleteDetails.frontMatter.layout
```
seaerchin, line 71: wdyt about using [`_.omit`](https://lodash.com/docs/4.17.15#omit) here?

newroutes/contactUs.js
```diff
@@ -0,0 +1,76 @@
...
+      content: { frontMatter, pageBody },
+      sha,
+    } = req.body
+
+    const updateResp = await this.contactUsPageService.update(
+      { siteName, accessToken },
+      { content: pageBody, frontMatter, sha }
+    )
+
+    return res.status(200).json(updateResp)
```
seaerchin, line 55: wdyt about renaming this to `updatedContactUsPage` for clarity?

services/fileServices/MdPageServices/ContactUsPageService.js
seaerchin: could we add the above explanation to the comment so that it's more fully fledged?

services/fileServices/MdPageServices/ContactUsPageService.js
seaerchin: same - add the explanation in the comment here too!

services/fileServices/MdPageServices/__tests__/ContactUsPageService.spec.js
```diff
@@ -0,0 +1,171 @@
...
+  } = require("@services/fileServices/MdPageServices/ContactUsPageService")
+  const service = new ContactUsPageService({
+    gitHubService: mockGithubService,
+    footerYmlService: mockFooterYmlService,
+  })
+
+  const {
+    retrieveDataFromMarkdown,
+    convertDataToMarkdown,
+  } = require("@utils/markdown-utils")
```
seaerchin, line 57: could we shift the imports up to the top of the file?

services/fileServices/MdPageServices/__tests__/ContactUsPageService.spec.js
```diff
@@ -0,0 +1,171 @@
...
+          fileName: CONTACT_US_FILE_NAME,
+          content: mockContent,
+          frontMatter: mockUpdatedFrontMatter,
+          sha: oldSha,
+        })
+      ).resolves.toMatchObject({
+        content: { frontMatter: mockUpdatedFrontMatter, pageBody: mockContent },
+        oldSha,
+        newSha: mockContactUsSha,
+      })
```
seaerchin, line 126: i think this got abit hard to read; could we perhaps separate the action (`service.update`) from the expectation (`expect(...).resolves.toMatchObject(...)`)

services/fileServices/MdPageServices/__tests__/ContactUsPageService.spec.js
```diff
@@ -0,0 +1,171 @@
...
+      mockGithubService.update.mockRejectedValueOnce(new NotFoundError(""))
+
+      await expect(
+        service.update(reqDetails, {
+          fileName: CONTACT_US_FILE_NAME,
+          content: mockContent,
+          frontMatter: mockFrontMatter,
+          sha: oldSha,
+        })
+      ).rejects.toThrowError(NotFoundError)
```
seaerchin, line 157: same comment as above, i think the readability might be better if we separated the action/expectation!

seaerchin (commented):
seems good, just some minor stylistic fixes/documentation

fixtures/app.js
```diff
@@ -0,0 +1,16 @@
...
+  app.use(express.json({ limit: "7mb" }))
+  app.use(express.urlencoded({ extended: false }))
+  app.use(router)
+  app.use(errorHandler)
+  return app
+}
+
+module.exports = {
+  generateRouter,
+}
```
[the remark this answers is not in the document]
alexanderleegs: Will need to replicate this on the rest of the tests as well! Should probably have done it in a separate PR haha

newroutes/__tests__/ContactUs.spec.js
```diff
@@ -0,0 +1,121 @@
...
+  // We can use read route handler here because we don't need to lock the repo
+  subrouter.get(
+    "/:siteName/contactUs",
+    attachReadRouteHandlerWrapper(router.readContactUs)
+  )
+  subrouter.post(
+    "/:siteName/contactUs",
+    attachReadRouteHandlerWrapper(router.updateContactUs)
+  )
+  const app = generateRouter(subrouter)
```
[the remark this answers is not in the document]
alexanderleegs: Yeah the reason why we're doing this currently is because of the middleware, we'll have to look into this more in future!

fixtures/app.js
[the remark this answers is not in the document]
seaerchin: Raise issue HAHAHAHAH

services/fileServices/MdPageServices/__tests__/ContactUsPageService.spec.js
alexanderleegs: This is a little tricky because the methods need to be imported after the mocks have been established - and the mocks use variables defined only inside the test description. We could move the variable definition upwards as well - but that might cause the imports section to become messy as well. What do you think?

newroutes/__tests__/ContactUs.spec.js
alexanderleegs: Great suggestion, I missed out that functionality - https://github.com/isomerpages/isomercms-backend/pull/349/commits/1d049826623b36a4708e27b6056792652beb24fa

newroutes/contactUs.js
alexanderleegs: Makes sense, https://github.com/isomerpages/isomercms-backend/pull/349/commits/5c5a12643e54b04a652e5a8110d5b93acb86e4e9

services/fileServices/MdPageServices/ContactUsPageService.js
alexanderleegs: https://github.com/isomerpages/isomercms-backend/pull/349/commits/3b6ca17dd13e63e6a283354987a50c0a4fd60c2d

services/fileServices/MdPageServices/ContactUsPageService.js
alexanderleegs: https://github.com/isomerpages/isomercms-backend/pull/349/commits/3b6ca17dd13e63e6a283354987a50c0a4fd60c2d

services/fileServices/MdPageServices/__tests__/ContactUsPageService.spec.js
alexanderleegs: I've split up the req and resp object definitions so we can still use `.resolves/.rejects` https://github.com/isomerpages/isomercms-backend/pull/349/commits/5e6d1148ec52c0c2fec2e9e3cd423be26e34b378

services/fileServices/MdPageServices/__tests__/ContactUsPageService.spec.js
seaerchin: oh i think that won't be an issue. this is because jest will [automatically hoist](https://jestjs.io/docs/manual-mocks#using-with-es-module-imports) ESM import mocks. 

this means that if we convert this to an import, it'll work fine. (not sure if this is the case for `require` but can try also cos TS is compiling down to ESM anyway

seaerchin (approved):
lgtm once import order addressed

services/fileServices/MdPageServices/__tests__/ContactUsPageService.spec.js
alexanderleegs: As discussed offline - this unfortunately causes issues, and we're opting to leave it as is for now as it should be resolved by a typescript migration.

PR #354, opened 2022-02-21
Fix/retrieve last updated time

alexanderleegs:
**Bug Fixes**:

This PR fixes a bug in which we were not retrieving the correct last updated time for repos. Previously, we were using the `updated_at` parameter to retrieve the time of last edit for a repo. However, this is actually only updated when the repository object is updated, e.g. when the description of the repository is updated, and not when a new commit is pushed to the repo. Instead, we should use the `pushed_at` parameter.

### DIFF
--- a/routes/sites.js
+++ b/routes/sites.js
@@ -76,7 +76,7 @@ async function getSites(req, res) {
     return resp.data
       .map((repoData) => {
         const {
-          updated_at: updatedAt,
+          pushed_at: updatedAt,
           permissions,
           name,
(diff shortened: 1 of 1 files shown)


seaerchin (approved):
lgtm - just double checking, does this also require a corresponding FE change?

alexanderleegs:
> lgtm - just double checking, does this also require a corresponding FE change?

No FE changes necessary - we were returning the wrong time in this endpoint

Issue #352, opened 2022-02-15
Error logging


alexanderleegs:
@seaerchin pointed out that we currently lose information when logging errors, as the custom errors we throw do not log all the information received from the original error (e.g. `if (resp.status === 404)
      throw new NotFoundError("Config page does not exist")` does not log the )

We should aim to preserve as much information as possible when logging errors so that it is easier to debug when checking through our logs.

alexanderleegs:
In addition, we should look into handling i/o only at the edges of our application (i.e. only `GithubService` and the routers), and intermediate service layers are kept pure. 

We should look into type-level guarantees about inputs so that inputs conform to some correctness notion.

seaerchin:
something to consider is [tslog](https://tslog.js.org/#/), which even has priority levels hehe

alexanderleegs:
We're currently inconsistent with our error handling from axios as well: we validate 404 errors because these are expected, but none of our other expected github errors are being handled in this way. Ideally, we should be standardising this:

```
const validateStatus = (status) => {
  return (status >= 200 && status < 300) || status >= 400
}
```
This way, we would have error handling in the different layers as follows:
Github service layer: captures the error type and encodes it in the type system
Services: transforms the error type into something more specific (throw isomer specific errors)
Router: throws the actual error to FE

PR #357, opened 2022-02-22
Refactor/netlify toml

alexanderleegs:
## Overview

This PR introduces the refactor for the netlify toml retrieval flow. It introduces a newly created NetlifyTomlRouter and NetlifyTomlService.

**Breaking Changes**

**Features**:

Sites router
- GET /:siteName/netlify-toml - retrieve the netlify-toml

**Notes**:
The NetlifyToml router still uses the site name of the repo, despite retrieving this from the isomer-build repo rather than the individual site - we could alternatively drop the siteName and purely call a general endpoint ( /netlify-toml)


newroutes/__tests__/NetlifyToml.spec.js
```diff
@@ -0,0 +1,45 @@
...
+  const reqDetails = { accessToken }
+
+  beforeEach(() => {
+    jest.clearAllMocks()
+  })
+
+  describe("readNetlifyToml", () => {
+    const netlifyTomlHeaderValues = "netlifyTomlHeaderValues"
+    mockNetlifyTomlService.read.mockResolvedValue(netlifyTomlHeaderValues)
+    it("retrieves netlifyToml details", async () => {
```
seaerchin, line 38: could we have a line spacing here for readability?

newroutes/netlifyToml.js
```diff
@@ -0,0 +1,36 @@
...
+
+class NetlifyTomlRouter {
+  constructor({ netlifyTomlService }) {
+    this.netlifyTomlService = netlifyTomlService
+    // We need to bind all methods because we don't invoke them from the class directly
+    autoBind(this)
+  }
+
+  // Read netlify.toml file
+  async readNetlifyToml(req, res) {
```
seaerchin, line 14: not a comment wrt this in particular but i realised that, in general, we have no idea what our functions throw. ie, all our router is showing is the happy path that the code takes and we rely upon the error handler middleware to transform and return errors.

my concern is that this causes devs to be unaware of possible errors (and how to handle them), instead deferring it to the frontend to format and display the error messages. some preconditions are also not clearly obvious - this requires the jwt to be authenticated (and we do so by having using hte auth middleware) but this isn't clear. 

i think the precondition section can be mitigated somewhat with proper ts typings but i'm not sure on hte error section. it'd be good to get your thoughts on this as well!

newroutes/netlifyToml.js
```diff
@@ -0,0 +1,36 @@
...
+    })
+
+    return res.status(200).json({ netlifyTomlHeaderValues })
+  }
+
+  getRouter() {
+    const router = express.Router()
+
+    router.get(
+      "/netlify-toml",
```
seaerchin, line 28: we should really maintain a mapping of keys to raw values

services/configServices/NetlifyTomlService.js
```diff
@@ -0,0 +1,39 @@
...
+    if (resp.status === 404)
+      throw new NotFoundError("netlify.toml file does not exist")
+
+    const { content } = resp.data
+
+    // Convert to readable form
+    const netlifyTomlReadableContent = toml.parse(Base64.decode(content))
+
+    // Headers is an array of objects, specifying a set of access rules for each specified path
+    // Under our current assumption, we apply the first set of access rules to all paths
```
seaerchin, line 32: i think i might be lacking context - could i clarify what's the current assumption here?

services/configServices/__tests__/NetlifyTomlService.js
```diff
@@ -0,0 +1,54 @@
+const axios = require("axios")
```
seaerchin, line 1: rename the file to include `.spec`

services/configServices/__tests__/NetlifyTomlService.js
```diff
@@ -0,0 +1,54 @@
...
+const {
+  netlifyTomlContent,
+  netlifyTomlHeaderValues,
+} = require("@root/fixtures/netlifyToml")
+
+const { GITHUB_BUILD_ORG_NAME, GITHUB_BUILD_REPO_NAME } = process.env
+
+jest.mock("axios")
+
+describe("Auth Service", () => {
```
seaerchin, line 15: i think you might have meant `NetlifyToml Service`!

services/configServices/__tests__/NetlifyTomlService.js
```diff
@@ -0,0 +1,54 @@
...
+  } = require("@services/configServices/NetlifyTomlService")
+
+  const service = new NetlifyTomlService()
+
+  beforeEach(() => {
+    jest.clearAllMocks()
+  })
+
+  describe("read", () => {
+    it("Reading the netlify.toml file works correctly", async () => {
```
seaerchin, line 31: nit: it `should read successfully when the netlify.toml is valid` (cos i alw read the test description with the "it" ._.)

seaerchin (commented):
seems good generally but some comments and jsut a general remark on the code structure

newroutes/netlifyToml.js
seaerchin: also, i realised we have no proper integration tests HAHA

newroutes/__tests__/NetlifyToml.spec.js
alexanderleegs: [9a6cbb5](https://github.com/isomerpages/isomercms-backend/pull/357/commits/9a6cbb53dc4fdc4de1e57ed72614e48e9f62ba71)

services/configServices/__tests__/NetlifyTomlService.js
alexanderleegs: Good catch, [a6cdec7](https://github.com/isomerpages/isomercms-backend/pull/357/commits/a6cdec7b64e02ef1769b77c35a2cfcfa3b882986)

services/configServices/__tests__/NetlifyTomlService.js
alexanderleegs: [a6cdec7](https://github.com/isomerpages/isomercms-backend/pull/357/commits/a6cdec7b64e02ef1769b77c35a2cfcfa3b882986)

services/configServices/__tests__/NetlifyTomlService.js
alexanderleegs: Makes sense, [a6cdec7](https://github.com/isomerpages/isomercms-backend/pull/357/commits/a6cdec7b64e02ef1769b77c35a2cfcfa3b882986)

services/configServices/NetlifyTomlService.js
alexanderleegs: Our assumption is that the file only contains a single set of access rules, so we take the first set!

newroutes/netlifyToml.js
alexanderleegs: Could you clarify which raw values you're referring to?

newroutes/netlifyToml.js
alexanderleegs: As discussed offline, this will be tackled as part of a larger error handling cleanup in a future PR! Relevant issue: https://github.com/isomerpages/isomercms-backend/issues/352

newroutes/netlifyToml.js
seaerchin: the `/netlify-toml`. idt it's a big issue cos the path should be used minimally? (lmk if i'm wrong)

services/configServices/NetlifyTomlService.js
seaerchin: could we write that down in the comment too?

seaerchin (approved):
lgtm; can be merged once the assumption is documented inside the comment!

services/configServices/NetlifyTomlService.js
alexanderleegs: Sure, done in https://github.com/isomerpages/isomercms-backend/pull/357/commits/6517e43054b62063d128deefc695551a26b78157

newroutes/netlifyToml.js
alexanderleegs: Right, this is only used for this single endpoint - do you think router namespaces will help with this issue?

PR #363, opened 2022-02-23
Refactor/navigation

alexanderleegs:
Overview
--
This PR adds the refactored navigation router, in line with our recent refactoring efforts. It builds off the existing NavigationYmlService that already exists for use by the CollectionsRouter. It also adds the relevant tests for the navigation router component.

The navigation router contains the following endpoints:

GET /:siteName/navigation - Retrieve navigation information
POST /:siteName/navigation- Update navigation information


newroutes/navigation.js
```diff
@@ -0,0 +1,69 @@
...
+    // We need to bind all methods because we don't invoke them from the class directly
+    autoBind(this)
+  }
+
+  // Read navigation file
+  async readNavigation(req, res) {
+    const {
+      accessToken,
+      params: { siteName },
+    } = req
```
seaerchin, line 26: might wanna destructure directly in the response!

newroutes/navigation.js
```diff
@@ -0,0 +1,69 @@
...
+    const { error } = UpdateNavigationRequestSchema.validate(req.body)
+    if (error) throw new BadRequestError(error.message)
+    const { content: fileContent, sha } = req.body
+
+    const updateResp = await this.navigationYmlService.update(
+      { siteName, accessToken },
+      { fileContent, sha }
+    )
+
+    return res.status(200).json(updateResp)
```
seaerchin, line 50: for clarification, this is because we want this to act transparently wrt github responses right?

newroutes/navigation.js
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  // Update navigation index file
+  async updateNavigation(req, res) {
+    const { accessToken } = req
+
+    const { siteName } = req.params
+    const { error } = UpdateNavigationRequestSchema.validate(req.body)
+    if (error) throw new BadRequestError(error.message)
+    const { content: fileContent, sha } = req.body
```
seaerchin, line 43: we might want to destructure directly in the params! the reason for doing so is because 1. it's clearer to tell what's used 2. express has an annoying way of typing `response` that gets easier if we can tell at a glance what's the body/params etc!

validators/RequestSchema.js
```diff
@@ -145,6 +145,35 @@ const DeleteMediaFileRequestSchema = Joi.object().keys({
...
+              sublinks: Joi.array().items(
+                Joi.object().keys({
+                  title: Joi.string().required(),
+                  url: Joi.string().required(),
+                })
+              ),
+            })
+            .oxor("url", "collection", "resource_room")
+            .oxor("sublinks", "collection")
+            .oxor("sublinks", "resource_room")
```
seaerchin, line 169: the joi docs are abit unclear on this - i'm assuming this means that as a whole, they are not required but when specified, only 1 of hte given options is allowed?

validators/RequestSchema.js
```diff
@@ -145,6 +145,35 @@ const DeleteMediaFileRequestSchema = Joi.object().keys({
...
+  content: Joi.object()
+    .keys({
+      logo: Joi.string().allow(""),
+      links: Joi.array()
+        .items(
+          Joi.object()
+            .keys({
+              title: Joi.string().required(),
+              url: Joi.string(),
+              collection: Joi.string(),
```
seaerchin, line 158: double checkin on this - they are optional and when not specified, will be missing from the keys of the object?

seaerchin (commented):
looks good, have some minor questions!

newroutes/__tests__/Navigation.spec.js
```diff
@@ -0,0 +1,128 @@
...
+      await request(app)
+        .post(`/${siteName}/navigation`)
+        .send(reqBody)
+        .expect(400)
+    })
+
+    it("rejects requests with incorrect body", async () => {
+      // Multiple types of menu types
+      const incorrectDetails = _.cloneDeep(navigationContent)
+      incorrectDetails.links[0].resource_room = true
```
seaerchin, line 80: sorry, could you help me to understand this? the validator allows **only** true, which seems to imply that this should be valid

validators/RequestSchema.js
alexanderleegs: Yep - we specify that at least one of them exist via the oxor requirements afterwards, but we normally only expect only a single one of `url`/`collection`/`resource_room`/`url` + `sublinks`

newroutes/__tests__/Navigation.spec.js
alexanderleegs: This test was for checking that only one of `url`/`collection`/`resource_room`/`url` + `sublinks` exist - but this item has both `url` and `resource_room`

newroutes/navigation.js
alexanderleegs: This is so the service class handles the formatting of the return!

newroutes/navigation.js
alexanderleegs: Fixed in 20558a7

newroutes/__tests__/Navigation.spec.js
```diff
@@ -0,0 +1,128 @@
...
+      const reqBody = {
+        content: incorrectDetails,
+        sha: navigationSha,
+      }
+
+      await request(app)
+        .post(`/${siteName}/navigation`)
+        .send(reqBody)
+        .expect(400)
```
seaerchin, line 89: should we also check for the error message?

validators/RequestSchema.js
seaerchin, line 169: i think we might wanna check that joi returns a good error message here or if it doesn't, set the message ourselves using `.message` so that when the FE displays the message to users, they aren't confused

alexanderleegs: As discussed offline, since our schema is used more as a test of correctness, we don't expect this error to be propagated to the frontend in normal scenarios, so we're opting to leave it as is for now

newroutes/__tests__/Navigation.spec.js
alexanderleegs: As discussed offline, since we use the default jest error messages, we're going to leave it as is for now

newroutes/navigation.js
```diff
@@ -0,0 +1,75 @@
...
+    const {
+      accessToken,
+      params: { siteName },
+      body: { content: fileContent, sha },
+    } = req
+
+    const updateResp = await this.navigationYmlService.update(
+      { siteName, accessToken },
+      { fileContent, sha }
+    )
```
seaerchin, line 52: could we rename this to `updateNavigationPage` for clarity?

validators/RequestSchema.js
```diff
@@ -145,6 +145,35 @@ const DeleteMediaFileRequestSchema = Joi.object().keys({
...
+              resource_room: Joi.boolean().valid(true),
```
seaerchin, line 159: just a question, why is `false` not allowed

seaerchin (approved):
lgtm once variable name changed

newroutes/navigation.js
alexanderleegs: Done in https://github.com/isomerpages/isomercms-backend/pull/363/commits/9ad16aeb16d940a6205f2e64452eb5ee7516d0c0

validators/RequestSchema.js
alexanderleegs: A navbar item will only contain one of the following set of params:

- Url
- Collection
- Resource_room
- Url + sublinks

This means that if we see `resource_room` as the params, it should always be `true`, otherwise the param will just not exist

PR #364, opened 2022-02-24
[WIP] feat(identity): verified merged

seaerchin:
## Problem

Closes [insert issue #]

## Solution

**Breaking Changes**

  - Details ...

**Features**:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- Details ...

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Deploy Notes

- `env var` : env var details

**New scripts**:

- `script` : script details

**New dependencies**:

**New dev dependencies**:


database/models/Site.js
```diff
@@ -0,0 +1,39 @@
...
+
+module.exports = (sequelize) => {
+  class Site extends Model {
+    static associate(models) {
+      const { User, SiteMember } = models
+      Site.belongsToMany(User, {
+        through: SiteMember,
+        onUpdate: "CASCADE",
+        onDelete: "CASCADE",
+      })
```
seaerchin, line 11: double check if this means that when a site is deleted/updated, all associated users are deleted/updated

database/models/User.js
```diff
@@ -0,0 +1,55 @@
...
+        },
+      },
+      githubId: {
+        allowNull: false,
+        unique: true,
+        type: DataTypes.TEXT,
+        validate: {
+          notEmpty: true,
+        },
+      },
```
alexanderleegs, line 38: Should we be allowing users with valid githubId but no email?

alexanderleegs: Should we be putting email and githubId in separate tables for easier retrieval?

database/models/Site.js
```diff
@@ -0,0 +1,39 @@
...
+      id: {
+        allowNull: false,
+        autoIncrement: true,
+        primaryKey: true,
+        type: DataTypes.BIGINT,
+      },
+      name: {
+        allowNull: false,
+        type: DataTypes.TEXT,
+      },
```
alexanderleegs, line 26: Should be unique - github repo names are unique anyway

services/identity/AuthService.js
```diff
@@ -0,0 +1,27 @@
...
+      return true
+    } catch (err) {
+      const { status } = err.response
+      if (status === 404 || status === 403) {
+        return false
+      }
+      throw err
+    }
+  }
+}
```
alexanderleegs, line 25: We have an endpoint for this in the upcoming Sites refactor

middleware/auth.js
```diff
@@ -102,18 +114,50 @@ const whoamiAuth = (req, res, next) => {
...
+auth.get("/v1/auth/github-redirect", noVerify)
 auth.delete("/v1/auth/logout", noVerify)
 auth.get("/v1/auth/whoami", whoamiAuth)
 
 // Index
 auth.get("/v1", noVerify)
 
+// Inject site access token if available
+auth.use("/v1/sites/:siteName", verifyJwt, useSiteAccessTokenIfAvailable)
+
```
alexanderleegs, line 160: double check if it always hit any given endpoint which starts with this

Also add into v2?

services/identity/SitesService.js
```diff
@@ -0,0 +1,21 @@
...
+    this.repository = repository
+    this.tokenStore = tokenStore
+  }
+
+  async getBySiteName(siteName) {
+    const site = await this.repository.findOne({
+      where: { name: siteName },
+    })
+    return site
+  }
```
alexanderleegs, line 12: Add comment: returns null if not found

routes/auth.js
```diff
@@ -90,8 +104,11 @@ async function githubAuth(req, res) {
     },
   })
 
-  let userId
-  if (userResp.data) userId = userResp.data.login
+  const githubId = userResp.data && userResp.data.login
+
+  // Create user if does not exists. Set last logged in to current time.
+  const user = await identityServices.usersService.login(githubId)
+  if (!user) throw Error("Failed to create user")
```
alexanderleegs, line 111: To confirm: will githubId ever be false? -> will cause bug if false because we will attempt to create a new user with githubId `false`

middleware/auth.js
alexanderleegs: Update: it does

alexanderleegs (commented):
We should also add a custom error for database operations - right now all default `Error`s are not `BaseIsomerErrors`, and hence will throw a 500 error, which also triggers our alerts

alexanderleegs (approved):
lgtm!!

PR #366, opened 2022-02-28
build(app): adds typescript to backend

seaerchin:
## Problem
Our BE currently is in JS, which is a dynamically and weakly typed language. this could cause potential problems for us down the line, especially when the codebase grows and becomes more complicated 

## Solution
This PR adds a **conservative** `tsconfig`, where i attempt to stick as strictly as possible to the original `jsconfig`. Hence, i didn't opt for stuff like `module: ESNext`. If there is a need to, we can in the future but to avoid potential build issues, this was not done now.

## Tests
1. set `noEmit` in the `tsconfig` to `true`
2. set `outDir` in the `tsconfig` to `/build`
3. run `npx tsc` - this should work

## devDeps


seaerchin:
Closing in favour of having this in #367 - the reason for doing so is that this significantly complicates the workflow. Because this has no testable changes, any change here needs to go through teh following steps: 

1. Change in this PR 
2. `verified-merged` branch rebases off this PR 
3. `add-types` branch rebases off `verified-merged` and checks that the changes work. 

Due to this additional complication and the fact that adding TS is only used at the `add-types` branch, this has been closed in favour of putting TS in that branch

PR #367, opened 2022-03-01
feat(identity): add types

seaerchin:
part of #364 

## Problem

This PR aims to add types to services within identity. Identity previously was written in javascript, which is weakly and dynamically typed. Adding TS allows us to be more confident in our code and ensure that information is encoded within the type system itself. 

## Solution

The PR consists of 2 sections: 

1. Adding TS + eslint support to the backend
2. Adding TS to identity

Note that this PR **does not** aim to write idiomatic typescript (this will be done in a subsequent PR) but instead, is aimed at ensuring correctness/soundness of identity code. 

This PR also does **not** migrate `UserService` and `SitesService` due to a dependency on the db being typed. This will be done off a subsequent PR to avoid review blocks.

This PR also does **not** support `npm run <command>` as `ts-node` is not installed and `mjs` imports conflict with the default `cjs` loader in node.

**New dev dependencies**:

- `typescript` : self-explanatory, together with all required deps
- `eslint`: all associated ts plugins for `eslint`


prestonlimlianjie (approved):
lgtm

PR #369, opened 2022-03-03
feat(identity): unit tests for services

seaerchin:
## Problem

This PR adds unit tests for the services that were migrated to TS in #367.  

## Solution

1. Add unit tests
2. modules which we have an explicit dependency on had mocks created so that we could mock return values and ensure that the calls were being made with correct arguments (see [here](https://jestjs.io/docs/manual-mocks) for more details on module mocking)

**New dev dependencies**:

- `jest-mock-axios` : This package is used for an automatic mock of `axios`.

### DIFF
--- a/services/identity/MailClient.ts
+++ b/services/identity/MailClient.ts
@@ -2,19 +2,17 @@ import axios from "axios"
 
 import logger from "@logger/logger"
 
+const { POSTMAN_API_KEY } = process.env
 const POSTMAN_API_URL = "https://api.postman.gov.sg/v1"
 
 class MailClient {
-  // NOTE: This is set as a private readonly property
-  // rather than a variable within the file for testing.
-  // This is to allow us to test that initialization fails when the API key is empty
-  private readonly POSTMAN_API_KEY: string
+  POSTMAN_API_KEY: string
 
   constructor() {
-    const { POSTMAN_API_KEY } = process.env
     if (!POSTMAN_API_KEY) {
       throw new Error("Postman.gov.sg API key cannot be empty.")
     }
+
     this.POSTMAN_API_KEY = POSTMAN_API_KEY
   }
 
@@ -30,7 +28,7 @@ class MailClient {
     try {
       await axios.post(endpoint, email, {
         headers: {
-          Authorization: `Bearer ${this.POSTMAN_API_KEY}`,
+          Authorization: `Bearer ${POSTMAN_API_KEY}`,
         },
       })
     } catch (err) {
--- a/services/identity/MailClient.ts
+++ b/services/identity/MailClient.ts
@@ -5,6 +5,9 @@ import logger from "@logger/logger"
 const POSTMAN_API_URL = "https://api.postman.gov.sg/v1"
 
 class MailClient {
+  // NOTE: This is set as a private readonly property
+  // rather than a variable within the file for testing.
+  // This is to allow us to test that initialization fails when the API key is empty
   private readonly POSTMAN_API_KEY: string
 
   constructor() {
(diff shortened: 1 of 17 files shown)


services/identity/MailClient.ts
```diff
@@ -2,14 +2,23 @@ import axios from "axios"
...
+
+  constructor() {
+    const { POSTMAN_API_KEY } = process.env
+    if (!POSTMAN_API_KEY) {
       throw new Error("Postman.gov.sg API key cannot be empty.")
+    }
+    this.POSTMAN_API_KEY = POSTMAN_API_KEY
+  }
 
+  async sendMail(recipient: string, body: string): Promise<void> {
```
alexanderleegs, line 21: I feel that changing this structure to support tests for the presence of the env var makes our code less readable, could we look into changing the env var directly in the test instead? 

See https://stackoverflow.com/a/48042799

alexanderleegs (changes requested):
Good find on the jest-mock-axios library - some of our existing tests currently mock axios directly, when it might be easier to use this in future. Left a comment regarding env vars!

services/identity/MailClient.ts
seaerchin: i do think it'll be better but i'm not sure this can be done - typescript complains about this: `An import declaration can only be used at the top level of a namespace or module.`, which would seem to imply that dynamic imports (in the test itself) wouldn't be possible

seaerchin: fixed - typescript has dynamic imports using `const x = await (import(your_path_here))` but take note that this requires that it is either 1. destructured or 2. called using `.default`like so: `const x =  await (import(your_path_here)).default`

services/identity/MailClient.ts
```diff
@@ -2,14 +2,20 @@ import axios from "axios"
 
 import logger from "@logger/logger"
 
-const { POSTMAN_API_KEY } = process.env
 const POSTMAN_API_URL = "https://api.postman.gov.sg/v1"
 
 class MailClient {
-  async sendMail(recipient: string, body: string): Promise<void> {
-    if (!POSTMAN_API_KEY)
+  private readonly POSTMAN_API_KEY: string
```
prestonlimlianjie: Actually this commit is really a refactor commit than a test commit - in future, we might want to split out refactor commits into a separate PR from the test one

services/identity/__tests__/MailClient.spec.ts
```diff
@@ -39,10 +53,15 @@ describe("Mail Client", () => {
     // Store the API key and set it later so that other tests are not affected
     const curApiKey = process.env.POSTMAN_API_KEY
     process.env.POSTMAN_API_KEY = ""
+    // NOTE: This is because of typescript transpiling down to raw js
+    // Export default compiles down to module.exports.default, which is also
+    // done by babel.
+    // Read more here: https://www.typescriptlang.org/tsconfig#allowSyntheticDefaultImports
+    const _MailClientWithoutKey = (await import("../MailClient")).default
```
prestonlimlianjie, line 60: This is ok for now - let's treat this as a work-in-progress. 

I believe the end state we should aim for is to write MailClient in a dependency injectable way, so that we can invoke it more easily in our tests. Could we write this as a TODO comment in the MailClient file?

prestonlimlianjie (commented):
lgtm - left a request to add a comment

services/identity/__tests__/MailClient.spec.ts
seaerchin: if by this you mean that the `POSTMAN_API_KEY` is given as a parameter to the constructor, i can do it here!

prestonlimlianjie: It's ok, we can leave it for now and add that to the various third-party client constructors in a separate PR/issue

PR #374, opened 2022-03-03
refactor: use sequelize-typescript

prestonlimlianjie:
This PR adds `sequelize-typescript` to allow us to
1. type our database models so that these types can be used elsewhere in the codebase, and
2. express our model files in a declarative manner for better clarity

### Dependency changes
- `sequelize-typescript`
- `reflect-metadata` - peer dep of `sequelize-typescript`
- `ts-node`
- `@swc/core` - third party transpiler that we pass to `ts-node`

### Dev-dependency changes
- `@types/node` - peer dep of `sequelize-typescript`
- `@types/validator` - peer dep of `sequelize-typescript`
- `ts-node-dev` - equivalent of `ts-node` but with hot reloading


seaerchin:
we have a dependency cycle in our db models, with each model transitively requiring the other 2 - is this intended and if not, should we fix it?

seaerchin:
i think we need to rollback the changes to `config.js` - `sequelize` only accepts .js or .json extensions. this isn't a problem when initializing the database, as we run this through `sequelize-typescript` but it does mean that when we run migrations, `sequelize-cli` will fail (can't recognise the .ts extension). 

if we run `npx sequelize db:migrate`, 
1. it fails beacuse it can't find `./database/config.js` which is specified in `.sequelizerc` (file is in .ts) 
2. if we change the extension to `.ts`, it fails due to the point mentioned above 
3. we have to rollback to `.js` then in our `.ts` file that consumes the config, **cast** to the correct type (`SequelizeOptions`). `json` isn't a good choice here because we depend on env vars for the db config

prestonlimlianjie:
> ` [55ccb9b](/isomerpages/isomercms-backend/pull/374/commits/55ccb9ba735fea8d790ce2490b246f725605f387)`

I've proposed a way to resolve this issue with these commits:
- [55ccb9b](https://github.com/isomerpages/isomercms-backend/pull/374/commits/55ccb9ba735fea8d790ce2490b246f725605f387)
- [93c37a2](https://github.com/isomerpages/isomercms-backend/pull/374/commits/93c37a2c016093a629dc9719d99aadfa251cbdd6)


The commits
- Revert the change in `.sequelizerc` back to using the ts config file, i.e. `database/config.ts`
- Add an `npm run build` script to compile the entire repo (including `database/config.ts`) into the `build/` folder
- Update `.sequelizerc` to use the compiled config file, i.e. `build/database/config.js`
- Add an `npm run database:migrate` script that calls `npm run build` and then `sequelize db:migrate`

prestonlimlianjie:
> we have a dependency cycle in our db models, with each model transitively requiring the other 2 - is this intended and if not, should we fix it?

Hm I'm not quite sure how to fix it while keeping the declarative way of describing model associations within the model file, and it hasn't really been a problem in the past. 

I suspect sequelize-typescript must be doing some sort of behind-the-scenes magic (perhaps by doing a lazy-load on associations specifically); their [examples](https://github.com/RobinBuschmann/sequelize-typescript/tree/master/examples/simple/lib) seem to recommend a dependency cycle too.

bin/www
```diff
@@ -45,18 +43,16 @@ function onError(error) {
     throw error
   }
 
-  const bind = typeof port === "string" ? "Pipe " + port : "Port " + port
+  const bind = typeof port === "string" ? `Pipe ${port}` : `Port ${port}`
```
seaerchin, line 46: we should file an issue to refactor and get rid of this ._.

database/config.js
```diff
@@ -1,45 +1,43 @@
...
-    // Migrations and seeders options
-    migrationStorage: "sequelize",
-    migrationStorageTableName: "sequelize_migration_data",
-    seederStorage: "sequelize",
-    seederStorageTableName: "sequelize_seed_data",
-  }
-}
-
-module.exports = {
-  ...getDbConfig(),
```
seaerchin, line 44: just double checking - this shouldn't matter right? (the switching from a function call to an object)

database/models/User.ts
```diff
@@ -0,0 +1,68 @@
...
+    type: DataType.BIGINT,
+  })
+  id!: number
+
+  @Column({
+    allowNull: true,
+    unique: true,
+    type: DataType.TEXT,
+    validate: {
+      isEmail: true,
```
seaerchin, line 29: i think this is already validated in `newroutes/users` ([here](https://github.com/isomerpages/isomercms-backend/pull/364/files#diff-f5a20552f742c7a8229e2a99ba078eaee60527809c656c1678f0cbba5458bfebR19-R20)) so it might be helpful to just drop a comment here. (they're the exact same validation underneath as sequelize also uses `validators.js`, see [here](https://sequelize.org/master/manual/validations-and-constraints.html))

database/models/User.ts
```diff
@@ -0,0 +1,68 @@
...
+    validate: {
+      notEmpty: true,
+    },
+  })
+  githubId!: string
+
+  @Column({
+    allowNull: true,
+    type: DataType.STRING(255),
+  })
```
seaerchin, line 47: we should either 1. have the same validator as [here](https://github.com/isomerpages/isomercms-backend/pull/364/files#diff-f5a20552f742c7a8229e2a99ba078eaee60527809c656c1678f0cbba5458bfebR49-R50) or 2. remove validation entirely (we allow everything) 

i think we should just use the validator as per the first point and actually stricten the check in here/the routes to sg only by specifying the [options](https://github.com/validatorjs/validator.js).

database/models/User.ts
```diff
@@ -0,0 +1,68 @@
...
+  contactNumber!: string
+
+  @Column({
+    type: DataType.DATE,
+    defaultValue: DataType.NOW,
```
seaerchin, line 52: i think this might be a little dangerous. we are explicitly setting the value when users login and if we allow creation of users by admins, might lead to an erroneous logged in time. 

should we just remove the default value?

seaerchin (commented):
some comments regarding model definition/validators! also, we should file an issue and handle bootstrapping in a separate file, which will allow us to split up each individual loading function rather than lumping them together in `server.js`. wdyt about this?

bin/www
prestonlimlianjie: Filed an issue here: https://github.com/isomerpages/isomercms-backend/issues/387

database/config.js
prestonlimlianjie: Yeah I think it's fine

database/models/User.ts
prestonlimlianjie: Let's follow up on [this thread](https://github.com/isomerpages/isomercms-backend/pull/376#discussion_r828747469) before returning to this discussion

database/models/User.ts
prestonlimlianjie: Will fix

database/models/User.ts
prestonlimlianjie: Will fix in a separate issue/PR.

database/models/User.ts
prestonlimlianjie: Defer decision till later. File an issue

PR #376, opened 2022-03-08
feat(identity): add types to last 2 services

seaerchin:
**this PR blocks #368**

## Problem
This PR adds types to the remaining 2 identity services, namely `UsersService` and `SitesService`. As this PR is a follow-up, refer to #367 for the associated write-up.

## Notes 
This PR removes the implicit initialization that was present in `server.js` previously. This means that after this PR, #368 will have to be rebased onto this and explicitly pass in and initialize `identity/AuthService` to `AuthMiddlewareService` (see [here](https://github.com/isomerpages/isomercms-backend/pull/368/files#diff-bae30b04856cd9da11361dc2ac4bd5981f86ca6cd142e69e5e66dc888e410462R102) for the line that uses `identity/AuthService`)


services/identity/TotpGenerator.ts
```diff
@@ -8,7 +8,7 @@ interface TotpGeneratorProps {
 class TotpGenerator {
   generator: typeof totp
 
-  expiry: number
+  expiry?: number
```
alexanderleegs, line 11: Why do we make this optional? Could we enforce that the `OTP_EXPIRY` env var is set instead?

alexanderleegs (approved):
Asking for a minor clarification, but otherwise lgtm!

services/identity/TotpGenerator.ts
seaerchin: We don't enforce because the original implementation defaults it to 5! I'm fine either way but I chose to kept it close to the original!

seaerchin: wait no u r right, i made a mistake - this would never be `undefined` as we default it in the constructor... will change

prestonlimlianjie (approved):
Stale review

server.js
```diff
@@ -107,10 +104,9 @@ const {
 const {
   NavYmlService,
 } = require("@services/fileServices/YmlFileServices/NavYmlService")
-const { initializeIdentityServices } = require("@services/identity")
+const { usersService } = require("@services/identity")
```
prestonlimlianjie, line 107: Nit: Use PascalCase to keep consistent with the rest of the services/routers

services/identity/UsersService.ts
```diff
@@ -1,56 +1,83 @@
...
     })
     return user
   }
 
-  async login(githubId) {
+  async login(githubId: string) {
     return sequelize.transaction(async (transaction) => {
-      const user = await this.findOrCreate(githubId, { transaction })
+      // NOTE: The service's findOrCreate is not being used here as this requires an explicit transaction
+      const user = await this.repository.findOrCreate(githubId, { transaction })
```
prestonlimlianjie, line 67: Could we consider removing the `this.findOrCreate()` method since we don't use it anywhere else? It could get quite confusing.

Also it looks like we are not using the transaction in `this.findOrCreate()`.

The new login method could look something like this

```
 async login(githubId: string) {
    return sequelize.transaction(async (transaction) => {
      const [user] = await this.repository.findOrCreate({
        where: { githubId },
        transaction
      })
      user.lastLoggedIn = new Date()
      return user.save({ transaction })
   }
}
```

database/config.js
```diff
@@ -34,10 +35,4 @@ module.exports = {
     min: DB_MIN_POOL ? parseInt(DB_MIN_POOL, 10) : 1,
     max: DB_MAX_POOL ? parseInt(DB_MAX_POOL, 10) : 10,
   },
-
-  // Migrations and seeders options
-  migrationStorage: "sequelize",
```
prestonlimlianjie, line 39: Why remove this? Do we have this elsewhere?

services/identity/UsersService.ts
prestonlimlianjie: Ah I see this was resolved in https://github.com/isomerpages/isomercms-backend/pull/378/commits/593355d275ddbba72f76e198faf6905d03ce093d

prestonlimlianjie: But we could still remove this.findOrCreate() because it isn't used anywhere else

seaerchin: i'm not opposed to removing it but this does mean that users will only ever be created through logging in/manually. this could potentially mean that if we wish to allow agency users to create and manage their own users, it would have to be added back in. 

wdyt about this scenario or should we just remove if first?

database/config.js
seaerchin: no, we don't. this arose as a result of types actually. let me briefly outline the sequence of steps below for clarity: 

1. conversion to `.ts`
2. used `SequelizeOptions` from `sequelize-typescript`
3. type error thrown on the removed properties as they **do not** exist on the `SequelizeOptions` type.

This then implies that these properties don't exist or should not be set, which is why i removed it.

services/identity/UsersService.ts
prestonlimlianjie: Hmm that's a good question. This might be a question for @kathleenkhy and @NatMaeTan.

Do we allow admin users to create other users in our system? In GitHub for example, an admin user cannot add a new user if that new user hasn't created their GitHub account. My first thought is that we might want to follow GitHub's model because we might have gibberish users in the system (i.e. there is no way to validate that gibberish@tech.gov.sg is not a real human).

Given that we aren't entirely sure what to do just yet, I'm in favor of removing this method for now. Thoughts?

server.js
prestonlimlianjie: We use PascalCase for the Classes (e.g. MediaCategoriesRouter) and camelCase for the instances (e.g. mediaDirectoryV2Router). This is the same across routers and services

seaerchin: actually then it should be snakeCased as this (`userService`) is an instance ._.

services/identity/UsersService.ts
kathleenkhy: I'm thinking we may not need to have such a control right now? Have checked with @jacksonOGP that FormSG doesn't check on the validity of emails when adding new collaborators (as long as the email domain ends with gov.sg or a whitelisted domain) and agencies aren't really bothered with it as well.

If our concern is on security or dirty data that gibberish emails are added, : 
1. Security: at the end of the day as long as the email belongs to a valid officer that can log in with the verified OTP, I don't think there should be much of a risk in terms of security (or I don't see a motive for a valid officer to deface/ harm another agency user's site)
2. Dirty data: Agency users should keep their list of team members with access to repos clean and should remove invalid users that have been granted access to the repos (gibberish emails included). We could emphasise on this further after we launch Identity as well but if I'm not wrong, this is an IM8 requirement (though we do have a waiver, agencies should abide to this).

Let me know if this makes sense or if you might have other views as well? @seaerchin @prestonlimlianjie

prestonlimlianjie: Will take a look at FormSG's implementation of this feature and get back to you

Issue #375, opened 2022-03-07
[Refactor] Initialise middleware separately from normal routers


alexanderleegs:
Our refactored auth middleware file is currently bloated because it chains the middleware onto routes individually, meaning routes are repeated twice - once in the actual router definition, then again in this file to chain the middleware onto the appropriate routes.  We can inject the middleware onto the existing route controllers e.g. CollectionPageController (see [Forms](https://github.com/opengovsg/FormSG/blob/7969d66b36eaf4ed84ba1ab1238d17cbfcaf0563/src/app/modules/form/admin-form/admin-form.routes.ts#L54), [Go](https://github.com/opengovsg/GoGovSG/blob/efa77265225205c105e6e513b34a3ad1579a1097/src/server/api/index.ts#L48) for example), so that we don't maintain a middleware list of routes that is separate from the controller routes.

What we could do is to initialise middleware separately, similar to how we've implemented the identity routers, since any supporting middlewareServices are self contained. We could then inject the specific auth middleware into the router directly.

Issue #351, opened 2022-02-15
Remove v1 auth endpoints


alexanderleegs:
To be done after the backend refactor is completed

PR #377, opened 2022-03-09
Refactor/separate auth initialisation

alexanderleegs:
This PR sets up the scaffolding for https://github.com/isomerpages/isomercms-backend/issues/375. Previously, we had a lot of duplicated routes, due to the need to specify each authenticated route again inside our auth middleware router. This PR modifies the auth middleware such that it is initialised in a separate `newmiddleware/index` file, and each individual router imports the middleware to utilise its `verifyJwt` or `whoami` methods. 

As the auth middleware is no longer a router which is always hit before the actual routes, the check for unknown routes has been moved out to `server.js`. In addition, the `noVerify` method has been removed as it is no longer useful, given that routes which do not need verification can now simply be used without middleware - I am open to adding this back if it provides us with greater clarity that those specific routes do not need verification.


newmiddleware/auth.js
```diff
@@ -31,320 +26,7 @@ class AuthMiddleware {
...
-    }
-    auth.verifiedPost = (path) => {
-      auth.post(path, this.verifyJwt)
-    }
-
-    // Health check
-    auth.get("/v2/ping", this.noVerify)
-
-    // V1 endpoints
-    // TODO [#351]: Remove after refactor is done
```
seaerchin, line 53: is #351 still relevant?

routes/index.js
```diff
@@ -1,19 +0,0 @@
-const express = require("express")
-const uuid = require("uuid/v4")
-
-const { CLIENT_ID } = process.env
-const { REDIRECT_URI } = process.env
-
-const router = express.Router()
-
-/* GET home page. */
-function getCmsHomepage(req, res) {
```
seaerchin, line 10: could i check why this is deleted?

seaerchin (approved):
lgtm (i didn't actually check to see that every route in the removed `getRouter` method has the `verifyJwt` method attached but it should be?)

we should also create a path alias for `@newmiddleware` to ease the imports (but i'm ok w/ not changing the existing `@root/newmiddleware/*`

routes/index.js
seaerchin: just saw that it's unused - how are we getting the index now then

newmiddleware/auth.js
alexanderleegs: Yup, will remove the V1 endpoints after all routes have been migrated

routes/index.js
seaerchin: resolved over slack - the above is correct

PR #378, opened 2022-03-10
 test(identity): add unit tests for user/sites service

seaerchin:
the high line count is due to `package-lock.json`, which can be ignored; same goes for `jest.config.js`

## Problem
This PR is a follow up to #376, where additional unit tests are added for the remaining 2 services.

This PR also introduces some functionality changes/typing changes, in order to make the identity feature more rigorous and reliable. 

## Solution
This PR adds a function to obtain `UserService` and refactors `UserService` to take in an instance of `sequelize.` This is because the service requires `sequelize` for a lock using `transaction` - as we initialize `sequelize` outside of server initially, this complicated testing. Using a mock would not work also because of `sequelize-typescript` decorators and hence, `sequelize` was specified as a direct dependency and injected in `server.js` while providing a function to ease dev friction. 

This PR also caught a potential bug in `SitesService`, where the `null` return from `getBySiteName` was not being checked in `getSiteAccessToken`. This check was done in auth middleware at runtime but not in the function, which meant that using `getSiteAccessToken` in another context might lead to breakages. We should investigate using libraries to encode errors into the type system but as that is out of scope, the `null` return is simply bubbled up and more `if` statements were added (to my chagrin) 

### Notes 
1. identity uses a whitelist to specify allowed domains - this defaults to `.gov.sg` which means that `something@gov.sg` will be rejected. this is ok as the `gov.sg` email is rare and defaulting to `gov.sg` also entails accepting `spoof@notgov.sg`
2. specifying a whitelist entails exclusion of `.gov.sg` - we should probably just remove teh defaulting behaviour subsequently and set the env var to be explicitly `.gov.sg`.


alexanderleegs (approved):
lgtm!

services/identity/UsersService.ts
```diff
@@ -29,15 +26,28 @@ class UsersService {
...
     this.smsClient = smsClient
     this.repository = repository
+    this.sequelize = sequelize
+
+    // Allowed domains is a semicolon separate list of domains (e.g. .gov.sg; @agency.com.sg; etc)
+    // that are allowed to login.
+    const { DOMAIN_WHITELIST } = process.env
 
-    this.whitelistDomains = (DOMAIN_WHITELIST || ".gov.sg")
+    this.whitelistDomains = (DOMAIN_WHITELIST || "gov.sg")
```
prestonlimlianjie, line 50: The default value should be `.gov.sg`. Using `gov.sg` could open us to an attack where a malicious actor buys `evilgov.sg` and is able to use our system.

Additionally, we don't foresee users having `@gov.sg` domains since that is reserved for MCI, and MCI uses `@mci.gov.sg`.

prestonlimlianjie, line 50: The default value should be `.gov.sg`. Using `gov.sg` could open us to an attack where a malicious actor buys `evilgov.sg` and is able to use our system.

Additionally, we don't foresee users having `@gov.sg` domains since that is reserved for MCI, and MCI uses `@mci.gov.sg`.

services/identity/UsersService.ts
```diff
@@ -51,7 +61,7 @@ class UsersService {
   }
 
   async updateUserByGitHubId(githubId: string, user: string) {
```
prestonlimlianjie, line 63: `user` should be of type `Partial<User>` in line 63. And then the first param in `this.repository.update()` in line 64 should be `user`

prestonlimlianjie, line 63: `user` should be a Partial User object

services/identity/__tests__/UsersService.spec.ts
```diff
@@ -0,0 +1,168 @@
...
+  it("should return the result of calling the update method by githubId on the db model", async () => {
+    // Arrange
+    const mockUser = "user1"
+
+    // Act
+    await UsersService.updateUserByGitHubId(mockGithubId, mockUser)
+
+    // Assert
+    expect(MockRepository.update).toBeCalledWith(
+      { user: mockUser },
```
prestonlimlianjie, line 95: Should just be `mockUser` instead of `{ user: mockUser }`

prestonlimlianjie, line 95: Should just be `mockUser` instead of `{ user: mockUser }`

database/index.ts
```diff
@@ -10,8 +10,14 @@ import sequelizeConfig from "@database/config.js"
...
  */
-export default (models: Parameters<Sequelize["addModels"]>[0]) => {
-  const sequelize = new Sequelize(sequelizeConfig as SequelizeOptions)
+export default (
+  models: Parameters<Sequelize["addModels"]>[0],
+  dbConfig?: SequelizeOptions
+) => {
+  const sequelize = new Sequelize({
+    ...sequelizeConfig,
+    ...dbConfig,
```
prestonlimlianjie, line 19: Which sequelizeConfig vars do we expect to override with dbConfig? If it is database, user, password, port, host - could we instead modify the env vars for the test env?

prestonlimlianjie, line 19: Which sequelizeConfig vars do we expect to override with dbConfig? If it is database, user, password, port, host - could we instead modify the env vars for the test env?

prestonlimlianjie (changes requested):

services/identity/UsersService.ts
alexanderleegs: It looks like it is fixed in https://github.com/isomerpages/isomercms-backend/pull/380/commits/9cfff3dfda32c06555203db4a289539581acf0c7 in PR #380

services/identity/__tests__/UsersService.spec.ts
seaerchin: i'm not sure this is correct ._. changing it to this causes the test to fail sadly

seaerchin: oops, i think this issue came up in a downstream branch - fixed [here](https://github.com/isomerpages/isomercms-backend/pull/382/commits/a9df6051d379779aa8763b48b9940909af59c284)

database/index.ts
prestonlimlianjie: Yes - I was wondering if we could use a test env file like `.env.test`, but open to either.

prestonlimlianjie (approved):
lgtm

Issue #379, opened 2022-03-14
[REF] Property setting on express request


seaerchin:
## Problem
Express request objects are not meant to have properties set on them (`req.property = some_value`), and the types reflect this (no way to adequately reflect that `req` has a property set). 

## Solution
We could adopt either
1. declaration merging ([like so](https://dev.to/kwabenberko/extend-express-s-request-object-with-typescript-declaration-merging-1nn5))
2. set properties on `res.locals` for all intents identical apart from naming w/ built in support from express. 

The suggested solution is to use `res.locals` as different properties set using declaration merging will all be merged into 1 giant request type, which would be extremely confusing.

PR #380, opened 2022-03-14
feat(identity): add types to user router

seaerchin:
## Problem
See #367 for details on conversion to TS. This is a further extension on that by converting the top level router to typescript too and partially addressing #379.

## Solution
1. Add types to `newroutes/users`
2. refactor so that we set property on `res.locals` instead of `req`. 

**New dependencies**:

- `type-fest` : package for utility types!

### DIFF
--- a/newroutes/users.ts
+++ b/newroutes/users.ts
@@ -1,20 +1,34 @@
-const autoBind = require("auto-bind")
-const express = require("express")
-const validator = require("validator")
+import autoBind from "auto-bind"
+import express, { Request, Response } from "express"
+import validator from "validator"
 
-const logger = require("@logger/logger")
+import logger from "@logger/logger"
 
-const { BadRequestError } = require("@errors/BadRequestError")
+import { BadRequestError } from "@errors/BadRequestError"
 
-const { attachReadRouteHandlerWrapper } = require("@middleware/routeHandler")
+import { attachReadRouteHandlerWrapper } from "@middleware/routeHandler"
 
-class UsersRouter {
-  constructor({ usersService }) {
+import UsersService from "@services/identity/UsersService"
+
+import { isError, RequestHandler } from "../types"
+
+interface UsersRouterProps {
+  usersService: UsersService
+}
+
+// eslint-disable-next-line import/prefer-default-export
+export class UsersRouter {
+  private readonly usersService
+
+  constructor({ usersService }: UsersRouterProps) {
     this.usersService = usersService
     autoBind(this)
   }
 
-  async sendEmailOtp(req, res) {
+  sendEmailOtp: RequestHandler<never, unknown, { email?: string }> = async (
+    req,
+    res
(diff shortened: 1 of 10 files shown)


newroutes/users.ts
```diff
@@ -1,20 +1,34 @@
...
+  constructor({ usersService }: UsersRouterProps) {
     this.usersService = usersService
     autoBind(this)
   }
 
-  async sendEmailOtp(req, res) {
+  sendEmailOtp: RequestHandler<never, unknown, { email?: string }> = async (
+    req,
+    res
+  ) => {
```
alexanderleegs, line 31: Why do we specify never/unknown? Do the remaining 2 unspecified params remain as unknown - we could possibly explicitly define the remaining 2 as unknown if so?

newroutes/users.ts
```diff
@@ -29,22 +43,37 @@ class UsersRouter {
...
+      if (isError(err)) {
+        logger.error(err.message)
+        throw new BadRequestError(err.message)
+      } else {
+        // If we encountered something that isn't an error but still ends up in the error branch,
+        // log this to cloudwatch with the relevant details
+        logger.error(
+          `Encountered unknown error type: ${err} when sendEmailOtp with email: ${email}`
+        )
+      }
```
alexanderleegs, line 55: How would something end up in the error branch without being an error?

newroutes/users.ts
```diff
@@ -54,13 +83,14 @@ class UsersRouter {
     return res.sendStatus(200)
   }
 
-  async verifyMobileNumberOtp(req, res) {
+  async verifyMobileNumberOtp(req: Request, res: Response) {
```
alexanderleegs, line 86: Why do we not need to use the same RequestHandler we do with verifyEmailOtp?

newroutes/users.ts
seaerchin: to clarify this - it's because the first two need to be specified in order to get to specify the concrete type. i specified the first as `never` because it's for the params, which is not set. 

res body is the second param and it's not used but _could be,_ hence the `unknown`.

newroutes/users.ts
seaerchin: good catch! was using this initially until i realised `express` had a `RequestHandler` type; changed it!

newroutes/users.ts
prestonlimlianjie: thanks for the explanation - and for the last 2 params, if we don't specify them, will they remain `unknown` by default?

prestonlimlianjie (approved):
lgtm

alexanderleegs (approved):
lgtm

newroutes/users.ts
seaerchin: since this `RequestHandler` is internal, yes! but the express one defaults it to `any` (hence the need for an internal one)

PR #382, opened 2022-03-16
test(identity): integration tests for user routes

### LINKED ISSUE #361: Epic: Identity phase 1
Details to be updated subsequently

### Frontend [P0]
- [x] Rebase existing frontend PR https://github.com/isomerpages/isomercms-frontend/pull/559 (done)

### Backend
- [x] Rebase existing backend PR to verify that it works [P0]
- [x] Set up Twilio (details TBD) [P0]
- [x] Set up integration with Postman (details TBD) [P0]
~- [ ] Break up existing backend PR into smaller chunks [P1]~ (not done, brief writeup given for overview)
- [x] Set up local dev database [P1]
- [x] Test suite [P1]
    - [x] Unit tests
    - [x] Integration tests
- [x] Integration with API v2 endpoints [P0]

### Infra [P0]
- [x] Set up RDS cluster for staging (done) and production
    - [x] Ensure that each cluster has at least 1 writer and 1 reader instance
    - [x] Ensure that the RDS instances are in private subnets of the respective VPCs
    - [x] Ensure that the RDS instance security groups are only reachable by the backend EB security group and the bastion SSH instance
- [x] Set up bastion SSH instances using AWS EC2 for staging and production
    - [x] Ensure that the bastion SSH instances are in the public subnets of the respective VPCs
    - [x] Ensure that the bastion SSH instance security groups only has SSH ports open to our team members’ IP addresses
- [x] Update infra diagram
- [x] Set up alarms for RDS (notify Slack `isomer-monitoring` channel for both staging and prod alarms, but only OpsGenie for prod alarms)
    - [x] CPU utilization > 60%, treat missing data as bad
    - [x] Freeable memory <=250MB, treat missing data as bad
    - [x] Write latency > 5s, treat missing data as bad
    - [x] Read latency > 5s, treat missing data as bad

seaerchin:
## Problem
This PR adds integration tests, which didn't exist before in the code base. As this is a new mode of testing, a brief outline of the testing approach will be outlined.

integration tests are meant to test how internal modules operate with each other. hence, doing integration tests means that only **external** packages are mocked, and our internal modules are fully used throughout. 

Final part of #361!!!!!!!

## Solution

**infrastructure**
1. sets up a new db instance (for testing) in docker, pointing from host port of 54321 to 5432 internally. 
2. run a global setup/teardown script located in `/tests` + adding path aliases. This allows future integration tests to already have the database configured properly with the appropriate tables.
3. add a `.env.test` file for testing environment

**integration testing**
1. no mocks of internal modules
2. required data is inserted into test db
3. as our application runs through a login cycle and most protected routes require a `userId` object set on `request`, a middleware has been setup to allow this. 
      -  if future unit/integration tests repeat this pattern, we might want to export a express app for testing purposes. 

**New dependencies**:

- `umzug` : this module was installed to programmatically run migrations.


newroutes/__tests__/Users.spec.ts
```diff
@@ -126,8 +131,10 @@ describe("Users Router", () => {
       let otp = ""
       mockAxios.post.mockImplementationOnce((_, email) => {
         otp = extractEmailOtp(email.body)
+        console.log(email.body)
         return email
       })
+      console.log("OTP: ============", otp)
```
alexanderleegs, line 137: Could remove the log statements!

alexanderleegs (approved):
Minor nit, but looks good otherwise!

One minor point about the location of the integration tests - I think we might want to place them in a separate `tests` folder to distinguish between unit and integration tests, since integration tests touch multiple components of our code!

newroutes/__tests__/Users.spec.ts
seaerchin: oops, good catch!!

seaerchin:
> Minor nit, but looks good otherwise!
> 
> One minor point about the location of the integration tests - I think we might want to place them in a separate `tests` folder to distinguish between unit and integration tests, since integration tests touch multiple components of our code!

that is a good idea - should we place it in `tests/integration` or do you think a top level `integration` folder would be clearer

alexanderleegs:
> > Minor nit, but looks good otherwise!
> > One minor point about the location of the integration tests - I think we might want to place them in a separate `tests` folder to distinguish between unit and integration tests, since integration tests touch multiple components of our code!
> 
> that is a good idea - should we place it in `tests/integration` or do you think a top level `integration` folder would be clearer

I think either `tests` or `integration` should be good, since the unit tests should remain in their current location!

PR #392, opened 2022-03-17
fix: order of ping

alexanderleegs:
## Problem

Our auth refactor made the health check endpoint (`/v2/ping`) inaccessible as it was placed after the unknown route check - this PR fixes the issue by rearranging their order.


seaerchin (approved):
lgtm

PR #405, opened 2022-03-29
feat(identity): infrastructure changes

seaerchin:
## Problem
Previously, we were using `ts-node` with elastic beanstalk. this posed some problems as elastic beanstalk had some difficulty with running `ts-node` and typescript.

Also, another issue that occurred was that the `swc` compiler had some issues with transpiling experimental stuff (decorators/metadata), which is troubling as `sequelize-typescript` uses these features. 
 
## Solution
1. remove `swc` as a transpiler
2. add a `build` step before deploying to EB (to transpile from TS -> JS) 
3. changed some commands on `package.json` so that the path aliases are respected.


alexanderleegs (approved):
lgtm!

PR #406, opened 2022-03-29
Fix: modify sequelizerc file to use config.js

alexanderleegs:
Our .sequelizerc file was not specifying the correct config path - this PR fixes the issue.


seaerchin (commented):
lgtm

seaerchin (approved):

PR #408, opened 2022-03-30
Fix: set reply_to param for OTP email

alexanderleegs:
## Problem

This PR fixes the reply email for OTP - previously, it was being sent to a private email rather than an official isomer email. partially fixes https://github.com/isomerpages/isomercms-frontend/issues/826

### DIFF
--- a/services/identity/MailClient.ts
+++ b/services/identity/MailClient.ts
@@ -23,6 +23,7 @@ class MailClient {
       from: "IsomerCMS <donotreply@mail.postman.gov.sg>",
       body,
       recipient,
+      reply_to: "admin@isomer.gov.sg",
     }
 
     try {
--- a/services/identity/MailClient.ts
+++ b/services/identity/MailClient.ts
@@ -23,7 +23,7 @@ class MailClient {
       from: "IsomerCMS <donotreply@mail.postman.gov.sg>",
       body,
       recipient,
-      reply_to: "admin@isomer.gov.sg",
+      reply_to: "support@isomer.gov.sg",
     }
 
     try {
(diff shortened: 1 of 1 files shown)


seaerchin:
sorry to do this but can we change it to the noreply one!! don't want to train our users to ignore cues

prestonlimlianjie:
> sorry to do this but can we change it to the noreply one!! don't want to train our users to ignore cues

Do you mean to set the reply_to email to be `donotreply@mail.postman.gov.sg`? Unfortunately the Postman transactional email API (and also the regular campaign API) requires a reply_to email

seaerchin:
> > sorry to do this but can we change it to the noreply one!! don't want to train our users to ignore cues
> 
> Do you mean to set the reply_to email to be `donotreply@mail.postman.gov.sg`, or do you mean removing the reply_to field altogether?
> 
> The latter doesn't seem possible because unfortunately the Postman transactional email API (and also the regular campaign API) requires a reply_to email

i meant that we should set it to noreply@isomer.gov.sg or something similar so that when they click reply to, they see that it's a non-existent email

alexanderleegs:
> > > sorry to do this but can we change it to the noreply one!! don't want to train our users to ignore cues
> > 
> > 
> > Do you mean to set the reply_to email to be `donotreply@mail.postman.gov.sg`, or do you mean removing the reply_to field altogether?
> > The latter doesn't seem possible because unfortunately the Postman transactional email API (and also the regular campaign API) requires a reply_to email
> 
> i meant that we should set it to [noreply@isomer.gov.sg](mailto:noreply@isomer.gov.sg) or something similar so that when they click reply to, they see that it's a non-existent/functional email

Makes sense, changed in https://github.com/isomerpages/isomercms-backend/pull/408/commits/949cbd569ac5a350bd2df05caaf8810b8b35a647

seaerchin (approved):
lgtm

seaerchin:
i tried sending an email to noreply@isomer.gov.sg - this results in an expected delivery failure as the email wasn't found. this is expected behaviour and is in line with what we want

PR #412, opened 2022-03-30
chore(build): automate jumphost setup

seaerchin:
## Problem
Jumphost set up is tedious and prone to errors. Developer friction could be eased by automating parts of the jumphost setup and making it a fixed command.

## Solution
- Add a `.ssh` folder for storing related information and shift env files + credentials there
- Add command to automate jumphost process to eliminate human error

## Notes
- only the `jump` command has actually been tested; the `db:migrate:<env>` command has not as the corresponding migrations have already been ran


alexanderleegs (approved):
lgtm, thanks for doing this up, it makes it a lot more convenient to run migrations!

Issue #403, opened 2022-03-25
[CLEANUP] Post-identity cleanup for code cleanliness


seaerchin:
tagging @alexanderleegs in case i missed something out

### Problem 
Identity had a bunch of PRs by different authors - this meant that there were some areas of the codebase that are unclear due to the massive scope of work. A best effort exhaustive list will be provided below

- [ ] `fixtures/identity` vs `test/constants` 
- [ ] using a main test router instead of redefining in every test file
- [ ] using `res.locals` instead of setting properties on req

alexanderleegs:
Remaining flows to be refactored on the frontend:

- [ ] https://github.com/isomerpages/isomercms-frontend/issues/749
  - [ ] https://github.com/isomerpages/isomercms-frontend/issues/823
  - [ ] https://github.com/isomerpages/isomercms-frontend/issues/812
- [ ] https://github.com/isomerpages/isomercms-frontend/issues/833
- [ ] https://github.com/isomerpages/isomercms-frontend/issues/834
- [ ] https://github.com/isomerpages/isomercms-frontend/issues/835

rc-davis:
Thanks @alexanderleegs. I copied your tasks to the first comment, because this allows us to track how many are complete (see the top). Items there will be checked automatically when you close the issues. Feel free to delete your later comment from yesterday, since everything is at the top now.

If anyone has more to add, feel free to edit the top again.

seaerchin:
> Remaining flows to be refactored on the frontend:
> 
> * [ ]  [[refactor - frontend] refactor EditHomepage flow isomercms-frontend#749](https://github.com/isomerpages/isomercms-frontend/issues/749)
>   
>   * [ ]  [[CHORE] EditHomepage variable shadowing isomercms-frontend#823](https://github.com/isomerpages/isomercms-frontend/issues/823)
>   * [ ]  [[refactor] rebase EditHomepage with new FormField Components isomercms-frontend#812](https://github.com/isomerpages/isomercms-frontend/issues/812)
> * [ ]  [[refactor] Edit Navigation isomercms-frontend#833](https://github.com/isomerpages/isomercms-frontend/issues/833)
> * [ ]  [[refactor] Swap netlify.toml endpoint to v2 isomercms-frontend#834](https://github.com/isomerpages/isomercms-frontend/issues/834)
> * [ ]  [[refactor] Contact us flow isomercms-frontend#835](https://github.com/isomerpages/isomercms-frontend/issues/835)
> 
[...]

PR #413, opened 2022-03-31
chore(app): replace custom properties on req with res.locals

### LINKED ISSUE #379: [REF] Property setting on express request

seaerchin:
## Problem
Application code previously used to set custom properties on the `req` object of express. This is acceptable in javascript but as express typings doesn't expose such properties on the request object, will prove to be troublesome to type in typescript.

Express instead has a [`res.locals` ](https://expressjs.com/en/api.html#res.locals) property (and a corresponding type parameter in the `RequestHandler` type) that is used for such custom properties. We should instead be using this for ease of transition to typescript.

Part of #403
Closes #379 

## Solution
1. change existing declarations of `req.property = some_value` into `res.locals.property = some_value`
2. update call sites of `req.property` to `res.locals.property`

### DIFF
--- a/newmiddleware/auth.js
+++ b/newmiddleware/auth.js
@@ -34,9 +34,7 @@ class AuthMiddleware {
     const {
       params: { siteName },
     } = req
-    const {
-      locals: { userId, accessToken: userAccessToken },
-    } = res
+    const { userId, accessToken: userAccessToken } = res.locals
 
     const siteAccessToken = await this.authMiddlewareService.retrieveSiteAccessTokenIfAvailable(
       { siteName, userAccessToken, userId }
--- a/newmiddleware/auth.js
+++ b/newmiddleware/auth.js
@@ -32,11 +32,10 @@ class AuthMiddleware {
   // Replace access token with site access token if it is available
   async useSiteAccessTokenIfAvailable(req, res, next) {
     const {
-      accessToken: userAccessToken,
       params: { siteName },
     } = req
     const {
-      locals: { userId },
+      locals: { userId, accessToken: userAccessToken },
     } = res
 
     const siteAccessToken = await this.authMiddlewareService.retrieveSiteAccessTokenIfAvailable(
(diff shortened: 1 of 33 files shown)


alexanderleegs (approved):
lgtm!

alexanderleegs (changes requested):
We missed changing a few endpoints -
- newroutes/authenticated/sites
- newmiddleware/auth

seaerchin:
> We missed changing a few endpoints -
> 
> * newroutes/authenticated/sites
> * newmiddleware/auth

good catch! missed this because i forgot that `.` doesn't include line breaks, which resulted in a faulty regex used for this

newmiddleware/auth.js
```diff
@@ -32,11 +32,10 @@ class AuthMiddleware {
   // Replace access token with site access token if it is available
   async useSiteAccessTokenIfAvailable(req, res, next) {
     const {
-      accessToken: userAccessToken,
       params: { siteName },
     } = req
     const {
-      locals: { userId },
+      locals: { userId, accessToken: userAccessToken },
     } = res
```
alexanderleegs, line 39: Minor nit: could we swap this to
```
const { userId, accessToken: userAccessToken } = res.locals
```
?

Same for newroutes/authenticated/sites!

alexanderleegs (approved):
Minor nit, but lgtm otherwise!

newmiddleware/auth.js
seaerchin: good catch!! i think this helps with readability and i'd be more than happy to make this change!

PR #415, opened 2022-03-31
test(ci): update ci config

seaerchin:
## Problem
Previously, our ci configuration was not working due to a few issues. These are respectively: 
1. requiring docker containers to be setup prior to running tests (for integration testing) 
2. `source` command not existing in github runner shell
3. jest not automatically exiting

## Solution
In order, the fixes are as follows: 
1. run `npm run dev:services` command
2. change `source` to `.` 
3. This is more interesting and more detail will be provided - i think this [issue](https://github.com/sequelize/sequelize/issues/6758) covers it up pretty well and i adopted their solution. Basically all the tests pass but jest wasn't exiting and resulted in an indefinite wait. This problem is due to sequelize having a connection pool and the teardown script doesn't close all the connections, merely _one_ connection. To fix this, there's another script that's ran after every test suite and this script closes sequelize connections. Oddly enough, setting `DB_MAX_POOL = 1` in the env vars don't work for this

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,7 +57,6 @@ jobs:
   test:
     needs: lint
     runs-on: ubuntu-latest
-    environment: staging
     steps:
       - uses: actions/checkout@v2
       - name: Use Node.js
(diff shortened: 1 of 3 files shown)


alexanderleegs (approved):
lgtm, thanks for investigating this!!

PR #418, opened 2022-04-05
chore(app): remove deprecated `bin/www`

### LINKED ISSUE #387: Relook at the use of `./bin/www`
The `./bin/www` file was introduced by the Express generator in 4.x. Express 4.x explicitly introduces this file, whereas the entry point used to be in app.js in Express 3.x.

This makes introducing TS into the codebase a little funky. Had a chat with @LoneRifle and he too feels that this `./bin/www` convention exists for some historical reasons that may not be applicable to us today.

To "modernize" the codebase, we should move the start up code back into an `app.ts` file. The [Express migration docs](https://github.com/expressjs/express/wiki/Migrating-from-3.x-to-4.x) suggest that this is fine:
> If you look at the npm start script in the package.json file, you will notice that the actual command that starts the app is node ./bin/www, which used to be node app.js in Express 3.
>
> Because the app.js file that was generated by the Express 4 generator is now a Node.js module, it can no longer be started independently as an app (unless you modify the code). The module must be loaded in a Node.js file and started via the Node.js file. The Node.js file is ./bin/www in this case.
>
> Neither the bin directory nor the extensionless www file is mandatory for creating an Express app or starting the app. They are just suggestions made by the generator, so feel free to modify them to suit your needs.

seaerchin:
## Problem
See the associated issue (#387) for why this is needed. Also, using `bin/www` meant that our app won't exit despite failing to connect to `sequelize` (not exactly sure why this happens) 

Closes #387 

## Solution
1. rework `bin/www` functions into `bootstrap/index` and add types to them
2. update `build`/`start`/`dev` commands (tested all 3 on local) 
3. update `server.js` so that the `bootstrap` function is only called when the db connection succeeds.

**Breaking Changes**

  - Details ...

## Deploy Notes
Take note that our new entrypoint is from `server.js`. 

Additionally, if AWS has automatic process respawn, this might interfere with how we exit programs. this can be replicated locally by running `npm run dev` and comparing behaviour to `npm run build` then `source .env && npm run start` when the db isn't running.  The former won't exit (the initial process is killed but then it respawns) whereas the former exits immediately.


alexanderleegs (approved):
lgtm

PR #419, opened 2022-04-05
refactor(mailclient): pass api key explicitly

### LINKED ISSUE #395: [REF] Add DI for MailClient
# Problem
In #369, the unit tests for `MailClient` was contorted using `jest.resetCache` to use a fresh copy of the env vars for testing. This is quite awkward and jarring, especially for engineers who do not know what `resetCache` does and why it's needed.

# Solution
Rather than getting the key from the env var and using it directly, we should rewrite `MailClient` to use dependency injection and pass the `POSTMAN_API_KEY` as a parameter. This allows easier testing

seaerchin:
## Problem
see #395 for an associated write-up on the problem. 

Closes #395

## Solution
1. pass the api key explicitly in the constructor \

### DIFF
--- a/services/identity/index.ts
+++ b/services/identity/index.ts
@@ -13,17 +13,24 @@ import TokenStore from "./TokenStore"
 import TotpGenerator from "./TotpGenerator"
 import UsersService from "./UsersService"
 
-const IS_LOCAL_DEV = process.env.NODE_ENV === "LOCAL_DEV"
-const { OTP_EXPIRY, OTP_SECRET } = process.env
+const {
+  OTP_EXPIRY,
+  OTP_SECRET,
+  NODE_ENV,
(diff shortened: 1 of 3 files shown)


services/identity/index.ts
```diff
@@ -33,10 +40,16 @@ const totpGenerator = new TotpGenerator({
   expiry: parseInt(OTP_EXPIRY!, 10) ?? undefined,
 })
 
+if (!POSTMAN_API_KEY) {
+  throw new Error(
+    "Please ensure that you have set POSTMAN_API_KEY in your env vars and that you have sourced them!"
+  )
+}
+
```
alexanderleegs, line 48: local dev runs with a mockMailer - we might want to put this warning only when it's not local_dev

alexanderleegs (approved):
minor nit, but otherwise lgtm

PR #420, opened 2022-04-06
test(resourceDirectory): add check for directory/file call order

### LINKED ISSUE #400: [Tests] Add check for directory and file level operation ordering
For the renaming of resource categories and collections/subcollections, we perform both directory and file level operations. For directory level operations, we use git trees, but since we use a commit hash from the start of the operation, any directory level operations need to be done before file level operations. We should add this into our tests for renaming folders in ResourceDirectoryService and CollectionDirectoryService - we could look into [jest-extended for .toHaveBeenCalledBefore](https://github.com/jest-community/jest-extended#tohavebeencalledbefore).

seaerchin:
## Problem
See #400 for a brief description of the issue.

Closes #400

## Solution
1. add `jest-extended`
2. check that directory level calls occur before file level reads

**New dev dependencies**:

- `jest-extended` : package to extend `jest` and add more matchers

### DIFF
--- a/services/directoryServices/__tests__/ResourceDirectoryService.spec.js
+++ b/services/directoryServices/__tests__/ResourceDirectoryService.spec.js
@@ -254,8 +254,6 @@ describe("Resource Directory Service", () => {
       sha,
     })
     it("Renaming a resource category works correctly", async () => {
-      const renameResourceSpy = jest.spyOn(service, "renameResourceDirectory")
-
       await expect(
         service.renameResourceDirectory(reqDetails, {
           resourceRoomName,
@@ -287,7 +285,9 @@ describe("Resource Directory Service", () => {
         fileName: INDEX_FILE_NAME,
         directoryName: `${resourceRoomName}/${newDirectoryName}`,
       })
-      expect(renameResourceSpy).toHaveBeenCalledBefore(mockGitHubService.update)
+      expect(mockBaseDirectoryService.rename).toHaveBeenCalledBefore(
+        mockGitHubService.update
+      )
     })
 
     it("Renaming a resource category slugifies the name correctly", async () => {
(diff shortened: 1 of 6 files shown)


services/directoryServices/__tests__/ResourceDirectoryService.spec.js
```diff
@@ -284,21 +287,26 @@ describe("Resource Directory Service", () => {
         fileName: INDEX_FILE_NAME,
         directoryName: `${resourceRoomName}/${newDirectoryName}`,
       })
+      expect(getResourceSpy).toHaveBeenCalledBefore(mockGitHubService.update)
```
alexanderleegs, line 290: Not sure how useful it is that the util function is called before the update - I think we might want to check that the `rename` operation from `baseDirectoryService` is called before `update` operation from `GitHubService`

services/directoryServices/__tests__/ResourceDirectoryService.spec.js
```diff
@@ -322,6 +330,7 @@ describe("Resource Directory Service", () => {
         fileName: INDEX_FILE_NAME,
         directoryName: `${resourceRoomName}/${slugifiedResourceCategory}`,
       })
+      expect(getResourceSpy).toHaveBeenCalledBefore(mockGitHubService.update)
```
alexanderleegs, line 333: see above!

alexanderleegs (changes requested):
left comments!

services/directoryServices/__tests__/ResourceDirectoryService.spec.js
seaerchin: thanks for the catch! updated accordingly

alexanderleegs (approved):
lgtm!

Issue #381, opened 2022-03-16
[Identity] Whitelisting of domains


alexanderleegs:
Our current proposed method for whitelisting domains is to store them in an env var - this may be infeasible in the long term because [AWS restricts the length of env vars](https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/environments-cfg-softwaresettings.html#environments-cfg-softwaresettings-console):
```
- Keys can contain up to 128 characters. Values can contain up to 256 characters.

- The combined size of all environment properties cannot exceed 4,096 bytes when stored as strings with the format key=value.
```

As the number of whitelisted domains is expected to grow, we should look into storing them in a table and referencing from it instead of using env vars.

jacksonOGP:
to add on context. New email domains can come up on either any of these scenarios

- agency creating a division (mha.gov.sg created new domain for hta.gov.sg awhile back)
- agency creating several new teams/departments
- organisation creating temporary emails for vendor, even tho it's .gov.sg, there's additional field in the domain, root domain doesn't really change

jacksonOGP:
Update after speaking with Hong, we might have to whitelist even more Vendors for future upcoming migration. Do take note of this flow since I believe we will hit character limits pretty soon

PR #422, opened 2022-04-11
Feat/move whitelist into database

alexanderleegs:
## Problem

This PR modifies our existing method for verifying the domains were whitelisted - we were previously relying on env vars for the list of whitelisted domains, but as the number of sites may increase with the CWP migration, we have opted to move this into a database, which removes the character restriction we previously had on the list of names.

Closes https://github.com/isomerpages/isomercms-backend/issues/381

## Solution

Adds a new table to our database for whitelisted domains.

**Breaking Changes**

- Yes - this PR contains breaking changes
  - Run migrations on staging/prod db before merging

## Fixes
This PR also fixes the User table types - we were previously not correctly specifying that the `email` and `contact_number` could be a null value.

**New scripts**:

- `db:migrate` : modifies this script to use `db:migrate` instead of `npm run db:migrate`

### DIFF
--- a/database/migrations/20220407074127-create-whitelist.js
+++ b/database/migrations/20220407074127-create-whitelist.js
@@ -1,6 +1,6 @@
 module.exports = {
   up: async (queryInterface, Sequelize) => {
-    await queryInterface.createTable("whitelists", {
+    await queryInterface.createTable("whitelist", {
       id: {
         allowNull: false,
         autoIncrement: true,
@@ -31,6 +31,6 @@ module.exports = {
   },
 
   down: async (queryInterface) => {
-    await queryInterface.dropTable("whitelists")
+    await queryInterface.dropTable("whitelist")
   },
 }
--- a/database/migrations/20220407074127-create-whitelist.js
+++ b/database/migrations/20220407074127-create-whitelist.js
@@ -11,14 +11,11 @@ module.exports = {
         allowNull: false,
         unique: true,
         type: Sequelize.TEXT,
-        validate: {
-          isEmail: true,
-        },
       },
       expiry: {
         type: Sequelize.DATE,
         allowNull: true,
-        defaultValue: null,
+        defaultValue: Sequelize.literal("NOW() + INTERVAL '6 months'"),
       },
       created_at: {
         type: Sequelize.DATE,
--- a/database/migrations/20220407074127-create-whitelist.js
+++ b/database/migrations/20220407074127-create-whitelist.js
@@ -1,0 +1,39 @@
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.createTable("whitelists", {
+      id: {
+        allowNull: false,
+        autoIncrement: true,
+        primaryKey: true,
+        type: Sequelize.BIGINT,
+      },
+      email: {
+        allowNull: false,
+        unique: true,
+        type: Sequelize.TEXT,
+        validate: {
+          isEmail: true,
+        },
+      },
+      expiry: {
+        type: Sequelize.DATE,
+        allowNull: true,
+        defaultValue: null,
+      },
+      created_at: {
+        type: Sequelize.DATE,
+        allowNull: false,
+        defaultValue: Sequelize.fn("NOW"),
+      },
+      updated_at: {
+        type: Sequelize.DATE,
+        allowNull: false,
+        defaultValue: Sequelize.fn("NOW"),
+      },
+    })
+  },
+
+  down: async (queryInterface) => {
+    await queryInterface.dropTable("whitelists")
+  },
+}
--- a/database/models/Whitelist.ts
+++ b/database/models/Whitelist.ts
@@ -8,7 +8,7 @@ import {
   UpdatedAt,
 } from "sequelize-typescript"
 
-@Table({ tableName: "whitelists" })
+@Table({ tableName: "whitelist" })
 export class Whitelist extends Model {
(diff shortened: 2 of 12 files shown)


database/migrations/20220407074127-create-whitelist.js
```diff
@@ -0,0 +1,39 @@
...
+        autoIncrement: true,
+        primaryKey: true,
+        type: Sequelize.BIGINT,
+      },
+      email: {
+        allowNull: false,
+        unique: true,
+        type: Sequelize.TEXT,
+        validate: {
+          isEmail: true,
```
seaerchin, line 15: not sure this is going to work - we only validate a suffix, not an email. this change means that organizational emails have to be added individually (eg: for `.edu.sg`, we have to add all indiv emails rather than just the suffix)

database/migrations/20220407074127-create-whitelist.js
```diff
@@ -0,0 +1,39 @@
...
+        },
+      },
+      expiry: {
+        type: Sequelize.DATE,
+        allowNull: true,
+        defaultValue: null,
+      },
```
seaerchin, line 22: could we do expiry of 6 months from now and let the null be the exception? don't foresee us adding many indefinite whitelists as they would be vendors

database/models/Whitelist.ts
```diff
@@ -0,0 +1,42 @@
...
+    },
+  })
+  email!: string
+
+  @Column({
+    allowNull: true,
+    type: DataType.DATE,
+    defaultValue: null,
+  })
+  expiry!: Date
```
seaerchin, line 35: wrong typing - the exclamation mark means that it can **never** be `null`, which is not what's declared

package.json
```diff
@@ -15,7 +15,7 @@
     "format-fix": "npx prettier --write .",
     "prepare": "husky install",
     "version": "auto-changelog -p && git add CHANGELOG.md",
-    "npm run db:migrate": "source .env && npx sequelize-cli db:migrate",
+    "db:migrate": "source .env && npx sequelize-cli db:migrate",
```
seaerchin, line 18: good fix, i can't believe i didn't catch this when i wrote it @___@

services/identity/UsersService.ts
```diff
@@ -88,10 +83,18 @@ class UsersService {
...
+        expiry: {
+          [Op.or]: [{ [Op.is]: null }, { [Op.gt]: new Date() }],
+        },
+      },
+    })
+    const whitelistDomains = whitelistEntries.map((entry) => entry.email)
     const hasMatchDomain =
-      this.whitelistDomains.filter((domain) => email.endsWith(domain)).length >
-      0
+      whitelistDomains.filter((domain) => email.endsWith(domain)).length > 0
```
seaerchin, line 97: this is fine as-is and it might be over-eager from my end, but wdyt about having a `WhitelistService` that 

1. returns valid whitelists 
2. checks if a given email is valid

seaerchin (commented):
main concerns are regarding migration script + the db model, rest are misc comments which i'm fine w/ not addressing

database/migrations/20220407074127-create-whitelist.js
alexanderleegs: You're right, thanks for catching! Fixed in https://github.com/isomerpages/isomercms-backend/pull/422/commits/a8805db30268136fa43a14e40d5929b86ce8544b

database/migrations/20220407074127-create-whitelist.js
alexanderleegs: That's a good point, made the change in https://github.com/isomerpages/isomercms-backend/pull/422/commits/a8805db30268136fa43a14e40d5929b86ce8544b

database/models/Whitelist.ts
alexanderleegs: Good catch, https://github.com/isomerpages/isomercms-backend/pull/422/commits/fa8737f8d288a7a00ad4500172dd9d23b9d31a3b

services/identity/UsersService.ts
alexanderleegs: As discussed offline, we currently don't foresee the whitelist being used anywhere other than login - as such, we'll leave it as is for now

database/models/Whitelist.ts
seaerchin: the type should be `Date | null` i think

database/migrations/20220407074127-create-whitelist.js
```diff
@@ -0,0 +1,36 @@
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.createTable("whitelists", {
```
seaerchin, line 3: nit: could we change the table name to be singular as this is a singular whitelist of emails

database/models/Whitelist.ts
alexanderleegs: Right, fixed in https://github.com/isomerpages/isomercms-backend/pull/422/commits/df92bcbaa153da5cf075fe9e8e0aa43ba2ef18b8

database/migrations/20220407074127-create-whitelist.js
alexanderleegs: Makes sense, changed in https://github.com/isomerpages/isomercms-backend/pull/422/commits/61220d8803326153f6854fd9c85a13caac9ed31d

seaerchin (approved):
lgtm

PR #423, opened 2022-04-12
Fix: check for JsonWebTokenError

alexanderleegs:
## Problem

This PR addresses an issue with our identification of expired tokens - previously, we were expecting a `TokenExpiredError` for expired tokens, however, we were receiving an abnormal number of unhandled server errors. The reason for this is that we also set an expiry for the cookie the token is attached to - this means that after the token/cookie expires, the token we attempt to decrypt is empty, which results in a `JsonWebTokenError` instead. As users getting logged out due to expiring tokens is an expected error, this PR handles this error in the same way, in order to reduce the number of alerts we receive on our monitoring channels.

### DIFF
--- a/services/middlewareServices/AuthMiddlewareService.js
+++ b/services/middlewareServices/AuthMiddlewareService.js
@@ -81,7 +81,9 @@ class AuthMiddlewareService {
         logger.error(`Authentication error: JWT token expired. Url: ${url}`)
         throw new AuthError(`JWT token has expired`)
       } else {
-        logger.error(`Authentication error. Message:${err.message} Url: ${url}`)
+        logger.error(
+          `Authentication error. Message: ${err.message} Url: ${url}`
+        )
       }
       throw err
     }
--- a/services/middlewareServices/AuthMiddlewareService.js
+++ b/services/middlewareServices/AuthMiddlewareService.js
@@ -73,7 +73,10 @@ class AuthMiddlewareService {
         throw new AuthError(
           `Authentication error: user not logged in with email`
         )
-      } else if (err.name === "TokenExpiredError") {
+      } else if (
+        err.name === "TokenExpiredError" ||
+        err.name === "JsonWebTokenError"
+      ) {
         logger.error(`Authentication error: JWT token expired. Url: ${url}`)
         throw new AuthError(`JWT token has expired`)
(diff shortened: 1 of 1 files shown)


seaerchin:
just double checking - if people attempt to change cookie and login etc, will we miss that? main worry is that changing the error type to cover `JsonWebTokenError ` might lead to actual useful logs being omitted.

if you don't feel that this is a big issue, i'm ok with this as-is.

seaerchin (commented):
as above

alexanderleegs:
> JsonWebTokenError

That's a great point actually, thanks for catching! We should still monitor when people are attempting to change their cookie - I've changed this check to just look for an empty cookie early. I've also added additional details to the errors that trigger our alarms by logging `err.message`. https://github.com/isomerpages/isomercms-backend/pull/423/commits/707dc3d0c7b03df47ae981145e98a4553c08a9a4

seaerchin (approved):
the check for `!isomercms` is dangerous - if `isomercms` is an _empty_ object, this will return `false`. i'm not too sure of the type so i'm approving this **only if** it's not an empty object.

services/middlewareServices/AuthMiddlewareService.js
```diff
@@ -77,7 +81,7 @@ class AuthMiddlewareService {
         logger.error(`Authentication error: JWT token expired. Url: ${url}`)
         throw new AuthError(`JWT token has expired`)
       } else {
-        logger.error(`Authentication error. Url: ${url}`)
+        logger.error(`Authentication error. Message:${err.message} Url: ${url}`)
```
seaerchin, line 84: also extremely minor nit: could we add a space like so `Message: ${err.message}`

alexanderleegs: Fixed in https://github.com/isomerpages/isomercms-backend/pull/423/commits/68574b98a088437c7edca179a1930164b22a797c

alexanderleegs:
> the check for `!isomercms` is dangerous - if `isomercms` is an _empty_ object, this will return `false`. i'm not too sure of the type so i'm approving this **only if** it's not an empty object.

Cookies are in a key:value format, so if the cookie is missing, `isomercms` will be undefined!

PR #432, opened 2022-04-20
Fix/migrate script

alexanderleegs:
## Problem

This PR fixes an issue with our previous migration script - we were attempting to run both `npm run jump:staging` and `npx sequelize-cli db:migrate` in the same command. However, due to the sequential order of execution, we would not be able to run `npx sequelize-cli db:migrate` until the jumphost was closed - this defeated the purpose of combining both commands. This PR separates out the commands and updates the README for the new flow.


seaerchin (approved):
good catch, thanks for fixing!

PR #436, opened 2022-04-29
Fix: remove unnecessary resource category check

alexanderleegs:
## Problem

This PR fixes an issue where resource categories with special characters were being rejected - this was because we had an extra unnecessary check for the original resource category title, when we were already slugifying the title. We remove the check in this title to stop titles with special characters from failing.


alexanderleegs:
Not required - we will maintain this behaviour because some processing is done via frontend as well

PR #440, opened 2022-05-10
Refactor/consolidate misc GitHub requests

### LINKED ISSUE #362: [Refactor] move axios calls into separate general github service
Several of our miscellaneous services (Sites, Auth, NetlifyToml) make axios calls to Github, but not to endpoints starting with `https://api.github.com/repos/` - this makes the existing GithubService unusable for this situation since we initialise the axiosInstance with this url. We should consider introducing a separate service for miscellaneous Github calls.

### LINKED ISSUE #438: Cleanup + Monitoring of Github calls
- [ ] Consolidate misc axios calls (#362)
- [ ] Introduce basic monitoring (simple logging when call is made) to actual Github calls

alexanderleegs:
## Problem

This PR consolidates and moves our axios instances used to call Github APIs into a single centralised AxiosInstance file. The reason behind this is the need for a centralised handling of logging and monitoring of rate limits - our existing implementation makes it hard for us to accurately monitor our usage of Github API. This is covered partially in https://github.com/isomerpages/isomercms-backend/issues/362, but we have elected to create axios instances rather than a separate service, since a separate service would be largely a simple wrapper over axios calls. This PR introduces these centralised axios instances, which share a common interceptor for the logging of our Github API usage. 

Our existing tests have also been modified to work with this method - of particular note is that we now setup our mockAxios before running tests.

Resolves #362 and resolves #438.


services/db/AxiosInstance.ts
```diff
@@ -0,0 +1,42 @@
+import { AxiosRequestConfig, AxiosResponse } from "axios"
```
seaerchin, line 1: could we put this in `services/api` as this isn't directly related to the db

services/db/AxiosInstance.ts
```diff
@@ -0,0 +1,42 @@
...
+// Env vars
+const { GITHUB_ORG_NAME } = process.env
+
+const requestFormatter = (config: AxiosRequestConfig) => {
+  logger.info("Making GitHub API call")
+  return {
+    ...config,
+    headers: {
+      ...config.headers,
+      "Content-Type": "application/json",
```
seaerchin, line 16: would this cause errors in teh future if we have different content types? eg: when we upload a file, the MIME type is no longer json but image. 

if `config.headers["Content-Type"]` is conditionally set, we can instead reverse the order of spreading (we specify content type first then spread headers) to avoid overriding.

services/db/AxiosInstance.ts
```diff
@@ -0,0 +1,42 @@
...
+    },
+  }
+}
+
+const respHandler = (response: AxiosResponse) => {
+  // Any status code that lie within the range of 2xx will cause this function to trigger
+  const remainingRequests = response.headers["x-ratelimit-remaining"]
+  if (remainingRequests < 2000) {
```
seaerchin, line 24: nit: would prefer to have a constant specifying the limit (eg: `GITHUB_API_LIMIT = 2500`) and specifying this in terms of the limit (`remainingRequests < GITHUB_API_LIMIT * 0.6`) for clarity

tests/setupMockAxios.ts
```diff
@@ -0,0 +1,26 @@
+import axios from "axios"
```
seaerchin, line 1: rather than doing this, we could instead do 2 exports in `__mocks__/axios.ts` that mirrors what we've done in our code

```ts
mockAxios.interceptors.request.use(jest.fn())
mockAxios.interceptors.request.use(jest.fn())
```

services/db/AxiosInstance.ts
```diff
@@ -0,0 +1,42 @@
...
+
+const gitHubRepoAxiosInstance = axios.create({
+  baseURL: `https://api.github.com/repos/${GITHUB_ORG_NAME}/`,
+})
+gitHubRepoAxiosInstance.interceptors.request.use(requestFormatter)
+gitHubRepoAxiosInstance.interceptors.response.use(respHandler)
+
+const miscGitHubAxiosInstance = axios.create()
+miscGitHubAxiosInstance.interceptors.request.use(requestFormatter)
+miscGitHubAxiosInstance.interceptors.response.use(respHandler)
```
seaerchin, line 40: could the names be changed to something like `isomerRepoAxiosInstance` and `githubAxiosInstance` for clarity? the `githubrepo` vs `miscgithubaxios` is abit unclear

services/db/AxiosInstance.ts
alexanderleegs: Good point, https://github.com/isomerpages/isomercms-backend/pull/440/commits/3148628a401ea4f4675d5729d350de8002d3b4c1

services/db/AxiosInstance.ts
alexanderleegs: That's a good catch - it'll be useful to be able to override this in future, although we currently appear to be uploading even images with `application/json`. https://github.com/isomerpages/isomercms-backend/pull/440/commits/36a138e63138508f6ebef4427a9577409263ae12

services/db/AxiosInstance.ts
alexanderleegs: Makes sense, https://github.com/isomerpages/isomercms-backend/pull/440/commits/ea6d2872ca8347fd0086bb77dc8e7f6a35cc9aa8

services/db/AxiosInstance.ts
alexanderleegs: https://github.com/isomerpages/isomercms-backend/pull/440/commits/ffd74fa7a5ce20cb0ba1d2d09cd71378e4c57b26

tests/setupMockAxios.ts
alexanderleegs: Modified in https://github.com/isomerpages/isomercms-backend/pull/440/commits/204cac45369ab3c81cc7b8b756f9c62aa992e0e2 and changes to tests which use jest-mock-axios reverted in https://github.com/isomerpages/isomercms-backend/pull/440/commits/3ccbc6f28094bb1bddb23c2484b0e11eb90052ee

seaerchin (approved):

Issue #439, opened 2022-05-09
[Identity phase 2] Setup scaffolding for token swapover


alexanderleegs:
- [x] Create token table in database
- [x] Create token selection mechanism
- [x] Setup token retrieval mechanism using axios interceptors
- [x] Setup monitoring
  - [x] Implement response inspection to check remaining available token usage
  - [x] Setup cloudwatch alarms
- [x] More informative commit messages

PR #441, opened 2022-05-11
Feat/add access token table

alexanderleegs:
## Problem

This PR partially addresses #439 - it sets up the model for the access token database table and its associated migration. It also implements a randomised selection of access token, as documented [here](https://docs.google.com/document/d/11YJV75FaSBqxJkpe79qlISC3ET07pg8SS71zqvdhb8A/edit#). I have chosen to implement the randomised selection instead of the round robin solution, as we discussed the possibility of having to insert new tokens. we would need to read from the database when deciding which token to use - the randomised selection removes the need to store a mutable global variable of the current number of tokens available.


seaerchin (approved):
lgtm

PR #449, opened 2022-05-31
feat: added system routes for updating site data.

rc-davis:


newroutes/authenticatedSystem/site.ts
```diff
@@ -0,0 +1,102 @@
...
+    req,
+    res
+  ) => {
+    const { siteName } = req.params
+    logger.info(`System Read Site '${siteName}'`)
+    const site = await this.sitesService.getByRepositoryName(siteName)
+    if (site == null) {
+      // null response means that the record was not found.
+      return res.sendStatus(404)
+    }
```
alexanderleegs, line 66: I think the check for whether the site exists should be done in the service layer, rather than here. Ideally, this should be an error type bubbled up from the service layer, which then is identified by our router and actually thrown as a `NotFoundError` - our current implementation is to directly throw an error in the service layer instead.

We have an outstanding issue regarding rewriting our error handling here: https://github.com/isomerpages/isomercms-backend/issues/352 - either way, we should move the check for null into the service layer.

server.js
```diff
@@ -116,6 +120,8 @@ app.use("/v1", authenticatedSubrouterV1)
 app.use("/v2/auth", authV2Router.getRouter())
 // Endpoints which have siteName, used to inject site access token
 app.use("/v2/sites/:siteName", authenticatedSitesSubrouterV2)
+// Endpoints which have siteName, used to inject site access token
+app.use("/v2/sys", authenticatedSystemSubrouterV2)
```
alexanderleegs, line 124: Nit: update comment

integration/SysAuth.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+import { generateRouter } from "@fixtures/app"
+import { getAuthMiddleware } from "@root/newmiddleware"
+import { RequestHandler } from "@root/types"
+import { SYS_AUTH_HEADER } from "@services/middlewareServices/AuthMiddlewareService"
+
+// Root handler for test
+const hanlder: RequestHandler<never, unknown, void> = async (req, res) =>
+  res.sendStatus(200)
+const rootRouter = express.Router({ mergeParams: true })
+rootRouter.get("/one/two", attachReadRouteHandlerWrapper(hanlder))
```
alexanderleegs, line 15: Could we use the same route params we use in the actual router?

middleware/errorHandler.js
```diff
@@ -23,13 +23,13 @@ function errorHandler(err, req, res, next) {
...
+  if (err.name && err.message && err.status) {
+    logger.info(`Internal server error: ${errMsg}`)
+    return res.status(err.status).json({
       error: {
         name: err.name,
-        code: 413,
+        code: err.status,
         message: err.message,
       },
     })
```
alexanderleegs, line 35: This seems to include Github errors as well, which we should be monitoring - doing this will bypass our monitoring and possibly cause us to miss errors that we should be handling. I think we should log the actual error and status code, but continue to send a 500 error to the frontend, and send the logger message `Unrecognized internal server error: ${errMsg}`, as this is the string we are looking out for.

database/models/Site.ts
```diff
@@ -22,23 +22,88 @@ export class Site extends Model {
...
   @CreatedAt
   createdAt!: Date
 
   @UpdatedAt
   updatedAt!: Date
 
+  @Column({
+    type: DataType.TEXT,
+  })
+  agency!: string | null
```
seaerchin, line 45: this annotation is wrong - if we assert that this is strictly non-null, the type should jsut be `string`

integration/SysSite.spec.ts
```diff
@@ -0,0 +1,744 @@
...
+      const actual = await request(app)
+        .patch(`/${mockRepositoryName}`)
+        .send({ redirectFrom: [42] })
+
+      // Assert
+      expect(actual.statusCode).toBe(expected)
+    })
+  })
+
+  // describe("/email/verifyOtp", () => {
```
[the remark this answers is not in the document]
rc-davis: Oops. Done.

newroutes/authenticatedSystem/site.ts
rc-davis: Ok, I moved the null check into the service layer and throw `NotFoundError`.

middleware/errorHandler.js
rc-davis: I will reverted to the previous version in my revised PR. This isn't important enough to change right now, and it sounds like we should think it through some more.

database/migrations/20220531100155-update-sites.js
```diff
@@ -0,0 +1,195 @@
...
+            autoIncrement: true,
+            primaryKey: true,
+            type: Sequelize.BIGINT,
+          },
+          repository_name: {
+            allowNull: false,
+            unique: true,
+            type: Sequelize.TEXT,
+          },
+          created_by: Sequelize.TEXT,
```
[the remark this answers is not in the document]
rc-davis: I agree that this would be ideal, but it will be a bit dangerous to put in this constraint today. Form SG can't check that the form submitter is in the database. If the submitter isn't in the database, then the database insert will fail after the rest of the site creation has succeeded. The only way to recover would be to extract the site information from the error email and insert it into the database manually. Is this good enough? If not, then we should probably wait to do something like this until the day when we can put this interface in the cms frontend.

database/migrations/20220531100155-update-sites.js
```diff
@@ -0,0 +1,195 @@
...
+            defaultValue: Sequelize.fn("NOW"),
+          },
+          updated_at: {
+            type: Sequelize.DATE,
+            allowNull: false,
+            defaultValue: Sequelize.fn("NOW"),
+          },
+          agency: Sequelize.TEXT,
+          site_name: Sequelize.TEXT,
+          contact: Sequelize.TEXT,
```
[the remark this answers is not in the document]
rc-davis: See above reply.

database/migrations/20220531100155-update-sites.js
```diff
@@ -0,0 +1,195 @@
...
+            type: Sequelize.DATE,
+            allowNull: false,
+            defaultValue: Sequelize.fn("NOW"),
+          },
+          updated_at: {
+            type: Sequelize.DATE,
+            allowNull: false,
+            defaultValue: Sequelize.fn("NOW"),
+          },
+          agency: Sequelize.TEXT,
```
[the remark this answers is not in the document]
rc-davis: I think it's a bad idea. It would require a database migration every time an agency is added. Also see above comment about error handling when a user is not in the database. Those problems would be magnified here, since this field is entered without any check at all.

database/migrations/20220531100155-update-sites.js
```diff
@@ -0,0 +1,195 @@
...
+          repository_url: Sequelize.TEXT,
+          hosting_id: Sequelize.TEXT,
+          staging_url: Sequelize.TEXT,
+          production_url: Sequelize.TEXT,
+          live_domain: Sequelize.TEXT,
+          redirect_from: Sequelize.JSONB,
+          uptime_id: Sequelize.TEXT,
+          uptime_url: Sequelize.TEXT,
+          launched_at: Sequelize.DATE,
+          launched_by: Sequelize.TEXT,
```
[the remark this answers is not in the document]
rc-davis: I thought it would be easier to maintain a system if there were fewer tables. As a system grows, I often find myself needing to grab bits of information that I didn't think I would need. It's harder to make such changes if information is be spread between multiple tables for historical reasons.

database/migrations/20220531100155-update-sites.js
```diff
@@ -0,0 +1,195 @@
...
+        },
+        { transaction }
+      )
+
+      await queryInterface.createTable(
+        "site_members",
```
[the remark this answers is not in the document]
rc-davis: How would that help?

database/models/Site.ts
rc-davis: My style is to use database constraints sparingly...only when failing to observe them would be catastrophic. Policies about how data should be gathered and structured tend to change over time, and if policies are built into database constraints, then changing them may require painfully complicated database operations. In most cases, I think it's better to build constraints into code, because code can be changed very easily. In this case, that means 

1. putting most constraints into the Joi validators, so requests with bad data will be rejected.
2. making as few assumptions as possible when reading the data, in case it's "dirty"

database/migrations/20220531100155-update-sites.js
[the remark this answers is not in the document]
seaerchin: this data model **doesn't** have to be enforced on the database layer - we can have a service method that looks like so: 

```ts
const addCreatedBy = (user: UnverifiedUser | VerifiedUser) => {...}
```

where the type is either 

```ts
interface User {
    contact_number?: number
    email?: Email // If we are adapting this, we need to validate email @ routes
    name: string 
}

interface UnverifiedUser extends User { 
    type: "unverified" 
}

interface VerifiedUser extends User { 
    type: "verified" 
    // additional properties
} 
```

but i have to ask - if the email is in the database, we could use a FK to link to an existing user then add a form submitter property to reference this or just check that the user tagged to the existing email is the same as that of the form submitter (and make it `VerifiedUser` else `UnverifiedUser`)

database/migrations/20220531100155-update-sites.js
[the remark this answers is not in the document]
seaerchin: i think this is q a flawed argument - see this [link](https://exceptionnotfound.net/god-objects-the-daily-software-anti-pattern/); for an extension of your argument, this implies that we could simply lump everything (sites, site users, users etc etc) into a single table and be done with it

database/models/Site.ts
seaerchin: you're right! turns out they use the same exclamation mark ._. just that it's different based on context

however, this is still an issue on our data model - we distinguish sites based on whether they are launched or not, and our data model should reflect that. see [here](https://fsharpforfunandprofit.com/series/designing-with-types/) (in f# but idea is same).

similar to the above, we should aim to have 3 kinds - `Site, LaunchedSite, UnlaunchedSite` where `UnlaunchedSite (and LaunchedSite) extends Site` and tags on extra meta data. this data model fits our reality well and is extensible in the future.

integration/SysAuth.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+
+import { attachReadRouteHandlerWrapper } from "@middleware/routeHandler"
+
+import { generateRouter } from "@fixtures/app"
+import { getAuthMiddleware } from "@root/newmiddleware"
+import { RequestHandler } from "@root/types"
+import { SYS_AUTH_HEADER } from "@services/middlewareServices/AuthMiddlewareService"
+
+// Root handler for test
+const hanlder: RequestHandler<never, unknown, void> = async (req, res) =>
```
seaerchin, line 12: nit: typo

services/identity/SitesService.ts
```diff
@@ -1,43 +1,104 @@
...
 
-    if (!site) {
-      return null
+      if (affectedRows === 0) {
+        return null
+      }
+      if (affectedRows === 1) {
+        return data[0]
+      }
+      throw new Error(`Updated ${affectedRows} sites (expected 1).`)
```
seaerchin, line 93: not sure this works as expected - if we get >1 affected row, will the update go through? ie, data is still modified in the database despite throwing an error

services/identity/SitesService.ts
```diff
@@ -1,43 +1,104 @@
...
+  hostingId?: string
+  stagingUrl?: string
+  productionUrl?: string
+  liveDomain?: string
+  redirectFrom?: string[]
+  uptimeId?: string
+  uptimeUrl?: string
+  launchedAt?: string
+  launchedBy?: string
+}
```
seaerchin, line 29: this interface is huge and has a few un-necessary stuff (we should only require very basic info). this is a direct consequence of our data model in the above and we won't ever need alot of stuff here (eg: prod/staging url because we will assign it when we create ._.)

validators/RequestSchema.js
```diff
@@ -271,6 +271,48 @@ const UpdateSettingsRequestSchema = Joi.object().keys({
   logo: Joi.string().allow(""),
 })
 
+// Do not add validation rules for human-entered fields that cannot be checked in FormSG.
+// We don't want backend requests to fail if form submission succeeds,
+// as there is no easy way for form submitters to recover from failed requests.
+const SysCreateSiteSchema = Joi.object().keys({
```
alexanderleegs, line 277: Do we expect these fields to be available on creation? Quite a few of these seem to be things that can only be added e.g. after site launch

newroutes/authenticatedSystem/site.ts
```diff
@@ -0,0 +1,102 @@
...
+      throw new BadRequestError(error.message)
+    }
+
+    const site = await this.sitesService.updateSite(siteName, req.body)
+    if (site == null) {
+      // null response means that the record was not found.
+      return res.status(404).send()
+    }
+    return res.status(200).send(site)
+  }
```
alexanderleegs, line 91: Likewise as above!

newroutes/authenticatedSystem/site.ts
```diff
@@ -0,0 +1,102 @@
...
+
+    logger.info(`System Update Site '${siteName}'`)
+
+    const { error } = SysUpdateSiteSchema.validate(req.body)
+    if (error) {
+      throw new BadRequestError(error.message)
+    }
+
+    const site = await this.sitesService.updateSite(siteName, req.body)
+    if (site == null) {
```
seaerchin, line 86: js/ts have 2 comparison operators ._. you probably want `===` or more ideally, `!site`

newroutes/authenticatedSystem/site.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  // GET /:siteName
+  readSite: RequestHandler<{ siteName: string }, Site, void> = async (
+    req,
+    res
+  ) => {
+    const { siteName } = req.params
+    logger.info(`System Read Site '${siteName}'`)
+    const site = await this.sitesService.getByRepositoryName(siteName)
+    if (site == null) {
```
seaerchin, line 63: js/ts have 2 comparison operators ._. you probably want `===` or more ideally, `!site`

database/migrations/20220531100155-update-sites.js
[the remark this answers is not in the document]
seaerchin: i think it's just for tracking purposes + soft-deletes still preserves data whereas hard delete doesn't. both r fine but i'm just tending towards soft so that we don't lose any data

database/migrations/20220531100155-update-sites.js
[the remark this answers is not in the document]
seaerchin: that's fair

database/migrations/20220531100155-update-sites.js
```diff
@@ -0,0 +1,195 @@
...
+    /**
+     * Add altering commands here.
+     *
+     * Example:
+     * await queryInterface.createTable('users', { id: Sequelize.INTEGER });
+     */
+    const transaction = await queryInterface.sequelize.transaction()
+    try {
+      await queryInterface.dropTable("site_members", { transaction })
+      await queryInterface.dropTable("sites", { transaction })
```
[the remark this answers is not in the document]
seaerchin: i think this is fine - could we just add a comment to say that this is to avoid reordering columns?

seaerchin:
I think more generally, the `SitesService` class doesn't need to just have a passthrough `create` method - we could add logic there before actually creating the db table (eg: whether the user is verified, saving to a "unlaunched sites" table). 

the service layer should in fact have some logic to ensure that our data (which is parsed on controller) is handled and saved into a shape that we expect/adheres to a logical model.

integration/SysSite.spec.ts
```diff
@@ -0,0 +1,744 @@
...
+import { Op } from "sequelize"
+import request from "supertest"
+
+import { Site } from "@database/models"
+import { generateRouter } from "@fixtures/app"
+import { SiteRouter } from "@root/newroutes/authenticatedSystem/site"
+import { sitesService } from "@services/identity"
+import { CreateSiteProps } from "@services/identity/SitesService"
+
+// const mockValidEmail = "open@up.gov.sg"
```
seaerchin, line 11: could we should delete unused variables?

integration/SysSite.spec.ts
```diff
@@ -0,0 +1,744 @@
...
+const mockRepositoryName2 = "arthur-repo"
+const mockRepositoryName3 = "heart-of-gold-repo"
+const mockContact = "slartibartfast@not.important.name"
+const mockUrl = "https://zaphod.com"
+const mockUrl2 = "https://bebblebrox.com"
+const mockAgency = "The Vogsphere"
+const mockSiteName = "Oh freddled gruntbuggly"
+const mockId = "marvin"
+const mockDateString = "2022-06-04T14:22:12.910Z"
+const mockDateString2 = "2022-06-04T10:00:00.000Z"
```
seaerchin, line 30: +1 for hitchhiker's guide to galaxy reference HAHA

integration/SysSite.spec.ts
```diff
@@ -0,0 +1,744 @@
...
+
+      // Act
+      const actual = await request(app).post("/").send(mockSiteCreateFull)
+
+      // Assert
+      expect(actual.statusCode).toBe(expected)
+      expect(actual.body).toMatchObject(mockSiteCreateFull)
+    })
+
+    it("should return when 400 repository name already exists", async () => {
```
seaerchin, line 88: nit: "should return 400 when the repository name already exists"

integration/SysSite.spec.ts
seaerchin: also, if we never alter this, could we use SCREAMING_SNAKE_CASE

integration/SysSite.spec.ts
```diff
@@ -0,0 +1,744 @@
...
+      // Act
+      const actual = await request(app)
+        .patch(`/${mockRepositoryName3}`)
+        .send({ createdBy: `not-${mockContact}` })
+
+      // Assert
+      expect(actual.statusCode).toBe(expected)
+    })
+
+    //---------------------------------------------------------
```
seaerchin, line 368: nit: could we remove this?

integration/SysSite.spec.ts
seaerchin, line 488: we could use `describe.only` for this behaviour (see [here](https://jestjs.io/docs/api#describeonlyname-fn))

services/identity/SitesService.ts
```diff
@@ -1,43 +1,104 @@
...
+  constructor({ repository }: SitesServiceProps) {
     this.repository = repository
-    this.tokenStore = tokenStore
   }
 
-  async getBySiteName(siteName: string): Promise<Site | null> {
-    const site = await this.repository.findOne({
-      where: { name: siteName },
-    })
+  async createSite(props: CreateSiteProps): Promise<Site> {
```
seaerchin, line 58: do we have tests for this and get/update?

seaerchin:
@rc-davis we should probably close this PR

PR #450, opened 2022-06-02
chore: Upgraded dotenv to enable running tests in WebStorm.

rc-davis:
The newest version allows me to import env vars from .env.test on the command line by setting one env var `DOTENV_CONFIG_PATH=.env.test` and passing jest the command line argument `--require dotenv/config`


alexanderleegs (approved):
lgtm

PR #451, opened 2022-06-10
feat: add site creation tables

prestonlimlianjie:
Do not merge until other PRs are complete.

This PR is part of a multi-part PR to add the site creation endpoint into the Isomer CMS backend. This specific PR defines the database models ([diagram here](https://lucid.app/lucidchart/d72eae8d-031e-44ea-a3c9-790eb5f5cab4/edit?invitationId=inv_bbceab50-47e6-4071-b017-1023d430d1ec&page=0_0#)) and adds the relevant migration files.

### What has been done
- Defined the new models (`repos`, `deployments`)
- Added the new associations - repos-sites, deployments-sites, sites-users (creator)
- Added new `job_status` and `site_status` columns to the `sites` table
- Set the `repos`,`deployments`, `sites`, and `users` tables to be paranoid + add the `deleted_at` column to these tables
- Migration definitions to specify steps to modify the existing tables


database/models/User.ts
```diff
@@ -56,10 +58,16 @@ export class User extends Model {
...
+
   @BelongsToMany(() => User, {
     onUpdate: "CASCADE",
     onDelete: "CASCADE",
     through: () => SiteMember,
   })
   sites!: Site[]
+
+  @HasOne(() => Site)
+  siteCreated?: Site
```
alexanderleegs, line 72: Should this be `HasMany`? A user could have created multiple sites, while a site has only one creator

database/models/Site.ts
```diff
@@ -5,13 +5,21 @@ import {
...
   UpdatedAt,
+  DeletedAt,
   BelongsToMany,
+  HasOne,
+  BelongsTo,
+  ForeignKey,
 } from "sequelize-typescript"
 
+import { SiteStatus, JobStatus } from "../../constants"
+
```
alexanderleegs, line 16: Nit: we could use `@constants/index` instead

database/migrations/20220610074055-associate-user-creator.js
```diff
@@ -0,0 +1,25 @@
...
+        type: Sequelize.BIGINT,
+        allowNull: false,
+        references: {
+          model: "users",
+          key: "id",
+        },
+        onUpdate: "CASCADE",
+        onDelete: "CASCADE",
+      }
+    )
```
alexanderleegs, line 16: Actually, if a user is deleted, I think we might not want to cascade? We should still preserve the record of the site

alexanderleegs (commented):
Left some comments!

database/migrations/20220610074044-add-status-columns-to-sites.js
```diff
@@ -0,0 +1,49 @@
...
+          "jobStatus", // name of column we're adding
+          {
+            type: Sequelize.ENUM,
+            values: ["READY", "RUNNING", "FAILED"],
+            allowNull: false,
+            transaction: t,
+          }
+        ),
+      ])
+    ),
```
alexanderleegs, line 26: Mismatch in the name of columns in `down` - we should set these to `site_status` and `job_status` to be consistent in using snake_case

database/models/User.ts
prestonlimlianjie: Good catch - let me fix it

database/models/Site.ts
prestonlimlianjie: This nit is actually a slightly bigger fix; fixed in 5a4eca3.

### Description of fix

To get this to work, I got into a mini-rabbit hole. 

I was curious as to why our CMS backend codebase doesn't use the aliases in the `database/` initially - then I saw this comment in `tests/database.ts`
```
/**
 * NOTE: This is required because globalSetup/Teardown live separately from
 * the rest of the jest ecosystem.
 *
 * This means that the files defined therein won't honour the aliases defined within
 * our jest config and we have to import using relative/absolute imports.
 */
```

I verified that this was indeed an issue by setting `../../constants` to `@constants/index` and got the following error:
```
Error: Jest: Got error running globalSetup - /Users/prestonlim/Documents/isomercms-backend/tests/setup.ts, reason: Cannot find module '@constants/index'
```

So I did a bit of research and found this [StackOverflow post](https://stackoverflow.com/a/67729228) that suggested an easy fix - which was to import `tsconfig-paths/register` at the top of the setup and teardown files.

database/migrations/20220610074044-add-status-columns-to-sites.js
prestonlimlianjie: This fix lead me down another rabbit hole. Fixed in 470692a.

### What led to this initial (faulty) code segment
I had initially set the column name to camelCase (`siteStatus` and `jobStatus`) because I ran into an error when running `npm run test` more than once on the database when the column name was snake_case. The error was:
```
[MigrationError: Migration 20220610074044-add-status-columns-to-sites.js (up) failed: Original error: type "enum_sites_site_status" already exists]
```

Anyway, it turns out that what I did had not addressed the root cause of the problem.

### A problem with the way `sequelize.dropAllTables()` handles enums with multi-part column names

We currently depend on `sequelize.dropAllTables();` during our test database clean up.

However, there appears to be a bug with Sequelize - or at least Sequelize has made some assumptions about the enum names for tables with multi-part column names.

Specifically, when Sequelize attempts to drop the sites table, it assumes that the enum names for the `site_status` and `job_status` columns are `enum_sites_siteStatus` and `enum_sites_jobStatus` respectively.

Unfortunately, we have named our enums `enum_sites_site_status` and `enum_sites_job_status` instead.

Hence, we need to add the following SQL statements to ensure that the abovementioned enums are dropped properly.
```
'DROP TYPE IF EXISTS "public"."enum_sites_site_status" CASCADE;'
'DROP TYPE IF EXISTS "public"."enum_sites_job_status" CASCADE;'
```

database/models/User.ts
prestonlimlianjie: Fixed in [213be36](https://github.com/isomerpages/isomercms-backend/pull/451/commits/213be369fb65d51f0469fcaf6e841bd6b92328a6)

alexanderleegs (approved):
lgtm - thanks for investigating and documenting the fixes!

PR #456, opened 2022-06-21
feat: site creation flow

prestonlimlianjie:
This PR introduces the Deployments and Repos services that handle the creation of Amplify projects and GitHub repositories respectively.


services/identity/SitesService.ts
```diff
@@ -29,6 +29,22 @@ class SitesService {
...
 
+  async create(
+    createParams: Partial<Site> & {
+      name: Site["name"]
+      apiTokenName: Site["apiTokenName"]
+      creator: Site["creator"]
+    }
+  ) {
+    return this.repository.create(createParams)
+  }
```
alexanderleegs, line 40: I think we might not need `apiTokenName` - it doesn't exist on Sites anymore

services/infra/InfraService.ts
```diff
@@ -0,0 +1,100 @@
...
+        jobStatus: JobStatus.Ready,
+      }
+      await this.sitesService.update(updateSuccessSiteInitParams)
+      logger.info(`Successfully created site on Isomer, site ID: ${newSite.id}`)
+
+      // TODO: Handle success by sending email to user who requested to create site
+    } catch (err) {
+      // TODO: Handle error by sending email to user who requested to create site
+      logger.error(err)
+    }
```
alexanderleegs, line 98: Should we be handling cleanup here?

newroutes/authenticated/sites.js
```diff
@@ -17,6 +21,13 @@ class SitesRouter {
     return res.status(200).json({ siteNames })
   }
 
+  async createSite(req, res) {
+    // TODO: authenticate request from formsg
+    const testParams = {}
+    this.infraService.createSite(testParams)
+    return res.status(200).send("OK")
+  }
+
```
alexanderleegs, line 30: Leaving here for reference: the current state of our authentication middleware will not allow submissions from formsg to enter this router (since it requires the appropriate cookies) - we can insert this route back again in future when site creation requests are done through the CMS, but for now, submissions through formsg will need to live in a separate router that bypasses the authentication middleware

alexanderleegs (approved):
Approving here - issues were addressed in #459

services/identity/SitesService.ts
rc-davis: Yes, this should be removed. I have added this as an item to #461.

services/infra/InfraService.ts
rc-davis: We're handling cleanup elsewhere for now. That should be enough until we give users the ability to add sites.

services/infra/InfraService.ts
```diff
@@ -0,0 +1,100 @@
...
+      }
+
+      // 2. If the user exists, create a new site record in the Sites table
+      //    with the associated user (creator) record
+      const newSiteParams = {
+        name: repoName,
+        apiTokenName: "", // TODO: figure this out
+        creator: foundUser,
+        creatorId: foundUser.id,
+      }
```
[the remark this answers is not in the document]
rc-davis: For some reason, sequelize will not create the row unless we include both creator and creatorId! I have verified this myself, but I haven't found a good explanation for why this is necessary.

Issue #453, opened 2022-06-14
[Tooling] Move site creation into CMS backend


prestonlimlianjie:
### Sequence diagram
The following Plant UML diagram was created on [planttext](https://www.planttext.com/).
![image](https://user-images.githubusercontent.com/19917616/173491749-3689a587-ee87-4821-aa57-fcbcc315855c.png)

### Sequence description
[TBD]

### Plant UML code
```

@startuml
actor "Agency User" as user
box "Isomer CMS (AWS)"
participant "CMS backend (EB)" as eb
database "CMS database (RDS)" as db
end box
participant "GitHub" as gh
participant "Amplify" as amp

group Site creation
  group Init site creation
    user -> eb : 1. User submits Form
    eb -> eb : 2a. Authenticate form submission
    eb -> user : 2b. If invalid, email user with failure message: EXIT
    else Else, proceed to 3
    eb -> db: 3. Create site record in Sites table
  end
  |||
  |||
  group Create GitHub repo
    eb -> gh: 4. Create GitHub repo with staging and master branches
    eb -> db: 5a. If 4 fails, Update site record, then
    eb -> user: 5b. Email user with failure message: EXIT
    else Else, proceed to 6
    eb -> db: 6. Create repo record in Repos table
  end
  |||
  group Create Amplify project
    eb -> amp: 7. Create Amplify project with staging and master deployments
    eb -> db: 8a. If 7 fails, Update site record
    eb -> user: 8b. Email user with failure message: EXIT
    else Else, proceed to 9
    eb -> db: 9. Create deployment record in Deployments table
  end
  |||
  group Complete site creation
    eb -> db: 10. Update site record
    eb -> user: 11. Email user with success message: EXIT
  end
end


@enduml
```

prestonlimlianjie:
Note: edited the comment to remove the following because we will be tracking it in a separate issue
> Add a rollback mechanism (see #460)

PR #458, opened 2022-06-23
Feat/add formsg middleware

alexanderleegs:
## Problem

This PR adds in a generic formsg authentication/decryption middleware, adapted from existing work on site-creation-backend and isomer-tooling. Part of #453.

`attachFormSGHandler` provides an array of middleware that should be inserted before any route which requires handling of submission data - the decrypted submission data can be retrieved via `res.locals.submission`.

TODO:

- [x] Add basic tests


newmiddleware/index.js
```diff
@@ -12,6 +16,25 @@ const getAuthMiddleware = ({ identityAuthService }) => {
...
+      "/",
+      attachFormSGHandler(<formKey>),
+      actualRoute
+    )
+ *
+ * Retrieve form data from res.locals.submission.
+ */
+const attachFormSGHandler = (formKey) =>
+  formSGMiddleware.authenticateAndDecrypt({ formKey })
+
```
seaerchin, line 36: double checking here - why not just inline the `authenticateAndDecrypt` method here? it seems to be just a wrapper over the underlying two methods + calling `express.json`. i don't think i'm understanding the benefit of the middleware

newmiddleware/formsg.ts
```diff
@@ -0,0 +1,24 @@
+import autoBind from "auto-bind"
```
seaerchin, line 1: could i ask why the file is lowercased rather than having it in sentence case as it exports a class

newmiddleware/formsg.ts
```diff
@@ -0,0 +1,24 @@
...
+    this.formSGService = formSGService
+    // We need to bind all methods because we don't invoke them from the class directly
+    autoBind(this)
+  }
+
+  authenticateAndDecrypt = ({
+    formKey,
+  }: {
+    formKey: string
+  }): Array<(req: Request, res: Response, next: NextFunction) => void> => [
```
seaerchin, line 19: this function signature is pretty confusing and we should have type aliases or explain it in detail

services/middlewareServices/FormSGService.ts
```diff
@@ -0,0 +1,57 @@
+import FormSG from "@opengovsg/formsg-sdk"
+import { Request, Response, NextFunction } from "express"
+
+import logger from "@logger/logger"
+
+import { AuthError } from "@root/errors/AuthError"
+
+const formsg = FormSG()
+
+export default class FormSGService {
```
seaerchin, line 10: we should probably rename this to indicate its purpose - something like `FormsVerificationService` as its main function is essentially that?

services/middlewareServices/FormSGService.ts
```diff
@@ -0,0 +1,57 @@
...
+      // If the decryption failed, submission will be `null`.
+      if (submission) {
+        // Continue processing the submission
+        res.locals.submission = submission
+        next()
+      } else {
+        res.status(422).send({ message: "Bad submission" })
+      }
+    } catch (e) {
+      logger?.error(e)
```
seaerchin, line 53: why do we need the conditional here

services/middlewareServices/FormSGService.ts
```diff
@@ -0,0 +1,57 @@
...
+    if (!signature) {
+      throw new AuthError("Signature missing")
+    }
+    const postUri = `https://${req.get("host")}${req.baseUrl}${req.path}`
+    try {
+      formsg.webhooks.authenticate(signature, postUri)
+      // Continue processing the POST body
+      next()
+    } catch (e) {
+      logger?.error(e)
```
seaerchin, line 26: do we need the conditional here?

services/middlewareServices/FormSGService.ts
```diff
@@ -0,0 +1,57 @@
...
+import logger from "@logger/logger"
+
+import { AuthError } from "@root/errors/AuthError"
+
+const formsg = FormSG()
+
+export default class FormSGService {
+  authenticate = () => (
+    req: Request,
+    res: Response,
```
seaerchin, line 13: nit: this should be prefixed iwth an underscore as it is unused

services/middlewareServices/FormSGService.ts
seaerchin: idt the `class` is required; this might be better off being an object

newmiddleware/formsg.ts
```diff
@@ -0,0 +1,24 @@
...
+import FormSGService from "@root/services/middlewareServices/FormSGService"
+
+export default class FormSGMiddleware {
+  formSGService: FormSGService
+
+  constructor({ formSGService }: { formSGService: FormSGService }) {
+    this.formSGService = formSGService
+    // We need to bind all methods because we don't invoke them from the class directly
+    autoBind(this)
+  }
```
seaerchin, line 13: actually i just realised all of this is known ahead of time as the service doesn't take anything at runtime also. we might want to just delete this/shift the service here/???

services/middlewareServices/FormSGService.ts
```diff
@@ -0,0 +1,57 @@
...
+    next: NextFunction
+  ): void => {
+    const signature = req.get("X-FormSG-Signature")
+    if (!signature) {
+      throw new AuthError("Signature missing")
+    }
+    const postUri = `https://${req.get("host")}${req.baseUrl}${req.path}`
```
seaerchin, line 20: could i ask how this is derived? think i'm missing some context here

services/middlewareServices/FormSGService.ts
```diff
@@ -0,0 +1,57 @@
...
+      throw new AuthError("Unauthorized")
+    }
+  }
+
+  decrypt = ({ formKey }: { formKey: string }) => (
+    req: Request,
+    res: Response,
+    next: NextFunction
+  ): void => {
+    try {
```
seaerchin, line 36: the given method supposedly only throws `MissingPublicKeyError`, which seems to be required when instantiating the class. given that we instantiate this above without the key, isn't this guaranteed to throw?

services/middlewareServices/FormSGService.ts
```diff
@@ -0,0 +1,57 @@
...
+      )
+
+      // If the decryption failed, submission will be `null`.
+      if (submission) {
+        // Continue processing the submission
+        res.locals.submission = submission
+        next()
+      } else {
+        res.status(422).send({ message: "Bad submission" })
+      }
```
seaerchin, line 51: nit: we can probably invert the order and save a level of indent

seaerchin (commented):
not sure if i'm missing something here, but we don't seem to be using this at all? also, we should write extremely simple tests  just to verify that we are calling the underlying formsg package

newmiddleware/index.js
alexanderleegs: As discussed offline, the primary reason for the middleware/service structure is consistency with the remainder of our codebase, as well as dependency injection for easier testing. As a compromise, since the formsgMiddleware is currently only serving as a thin wrapper, we are opting to move this into newmiddleware/index instead, as it is not a particularly complicated method.

newmiddleware/formsg.ts
alexanderleegs: Will be removed

services/middlewareServices/FormSGService.ts
alexanderleegs: Renamed in https://github.com/isomerpages/isomercms-backend/pull/458/commits/16ee9e43f759785cfc9f22a767998d851e292077. As discussed offline, we will maintain this format for consistency

services/middlewareServices/FormSGService.ts
alexanderleegs: https://github.com/isomerpages/isomercms-backend/pull/458/commits/292486b5631d525c3d3d9189758957c50c8c5dad

newmiddleware/formsg.ts
alexanderleegs: As above, removing this class!

services/middlewareServices/FormSGService.ts
alexanderleegs: This should be to generate the postURI forms needs to verify the submission - it was in Alwyn's existing code

services/middlewareServices/FormSGService.ts
alexanderleegs: Makes sense, https://github.com/isomerpages/isomercms-backend/pull/458/commits/292486b5631d525c3d3d9189758957c50c8c5dad

seaerchin: is the `!submission` a strong enough check? if submission is an empty object, this might still pass (unless we're ok with an empty submission?) - unsure of details but just wanted to double confirm

services/middlewareServices/FormSGService.ts
seaerchin: nit: could we call this `FormSgProcessingService` or `FormsProcessingService`? Apologies for this but FormSG in a name just looks weird ;--;

services/middlewareServices/FormSGProcessingService.ts
```diff
@@ -0,0 +1,74 @@
...
+      // If the decryption failed, submission will be `null`.
+      if (!submission) {
+        res.status(422).send({ message: "Bad submission" })
+      }
+      // Continue processing the submission
+      res.locals.submission = submission
+      next()
+    } catch (e) {
+      logger.error(e)
+      throw new AuthError("Unauthorized")
```
seaerchin, line 71: just noticed that both errors returning from `decrypt` and `authenticate` are identical - we should aim to distinguish our errors and attach more meaningful information to aid in troubleshooting.

this could be done through extending the `AuthError` to `DecryptionError` and also attach more meta data to it (eg: `formKey`)

services/middlewareServices/FormSGService.ts
alexanderleegs: This should be fine - the documentation states that the submission will be `null` if the decryption failed? But to be safe I can explicitly check for null - 493c454

services/middlewareServices/FormSGService.ts
alexanderleegs: db85651

newmiddleware/index.ts
```diff
@@ -0,0 +1,49 @@
...
+import express, { RequestHandler } from "express"
+
+import FormSGProcessingService from "@root/services/middlewareServices/FormSGProcessingService"
+import AuthService from "@services/identity/AuthService"
+
+const {
+  AuthMiddlewareService,
+} = require("@services/middlewareServices/AuthMiddlewareService")
+
+const { AuthMiddleware } = require("./auth")
```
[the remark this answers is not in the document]
alexanderleegs: https://github.com/isomerpages/isomercms-backend/pull/458/commits/5f514b10b0f49f0c436de7b3d273765a2fcd8121

newmiddleware/index.ts
```diff
@@ -0,0 +1,49 @@
...
+ * To be inserted before routes requiring FormSG response data - e.g. 
+ * router.post(
+      "/",
+      attachFormSGHandler(<formKey>),
+      actualRoute
+    )
+ *
+ * Retrieve form data from res.locals.submission.
+ */
+const attachFormSGHandler = (formKey: string): Array<RequestHandler> => [
```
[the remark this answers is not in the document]
alexanderleegs: https://github.com/isomerpages/isomercms-backend/pull/458/commits/413d210a408e4f12f0f72ded1461ecf406d90103

newmiddleware/index.ts
```diff
@@ -0,0 +1,49 @@
...
+  formSGService.authenticate(),
+  express.json(),
+  formSGService.decrypt({ formKey }),
+]
+
+module.exports = {
+  getAuthMiddleware,
+  attachFormSGHandler,
+}
```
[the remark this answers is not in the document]
alexanderleegs: https://github.com/isomerpages/isomercms-backend/commit/5f514b10b0f49f0c436de7b3d273765a2fcd8121

services/middlewareServices/FormSGProcessingService.ts
```diff
@@ -0,0 +1,74 @@
...
+  DecryptParams,
+  DecryptedContent,
+} from "@opengovsg/formsg-sdk/dist/types"
+import { Request, Response, NextFunction } from "express"
+
+import logger from "@logger/logger"
+
+import { AuthError } from "@root/errors/AuthError"
+
+export interface CanDecryptFormSGPayload {
```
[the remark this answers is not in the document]
alexanderleegs: 25d3ebb

services/middlewareServices/FormSGProcessingService.ts
```diff
@@ -0,0 +1,74 @@
...
+    }
+  }
+
+  decrypt = ({ formKey }: { formKey: string }) => (
+    req: Request,
+    res: Response,
+    next: NextFunction
+  ): void => {
+    try {
+      const submission = this.formsg.crypto.decrypt(
```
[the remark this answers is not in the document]
alexanderleegs: Right that makes sense, didn't catch what you were saying before. Fixed in 9905db2

services/middlewareServices/__tests__/FormSGProcessingService.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+      expect(result).toThrow(AuthError)
+      expect(mockFormsg.webhooks.authenticate).not.toHaveBeenCalled()
+      expect(mockNext).not.toHaveBeenCalled()
+    })
+  })
+
+  describe("Decrypt", () => {
+    const mockFormKey = "mockKey"
+    const mockSubmission = "submission"
+    it("should call decrypt successfully, store submission data and call next() when the call is successful", async () => {
```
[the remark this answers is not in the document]
alexanderleegs: e819ba7

services/middlewareServices/__tests__/FormSGProcessingService.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+      // Assert
+      expect(mockNext).toHaveBeenCalled()
+      expect(mockFormsg.crypto.decrypt).toHaveBeenCalledWith(
+        mockFormKey,
+        mockData
+      )
+      expect(mockRes.locals.submission).toStrictEqual(mockSubmission)
+    })
+
+    it("should not call next handler if decrypt is unauthorised", async () => {
```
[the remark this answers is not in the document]
alexanderleegs: Added in e819ba7

services/middlewareServices/__tests__/FormSGProcessingService.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+      // Act
+      const result = () => decryptMiddleware(mockReq, mockRes, mockNext)
+
+      // Assert
+      expect(result).toThrow(AuthError)
+      expect(mockFormsg.crypto.decrypt).toHaveBeenCalledWith(
+        mockFormKey,
+        mockData
+      )
+      expect(mockNext).not.toHaveBeenCalled()
```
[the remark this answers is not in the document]
alexanderleegs: Makes sense - checked in e819ba7 for the new test case since this one has been removed

services/middlewareServices/FormSGProcessingService.ts
alexanderleegs: I've updated the error messages to include the post URI and additional info in 344cfaa and d8909a7 - but we shouldn't be adding the `formKey` data, since that's supposed to be a secret key

services/middlewareServices/FormSGService.ts
seaerchin: actually checking for `null` is weaker because now it doesn't cover `undefined` HAHA on opening it in vscode, i see that hte type is `DecryptedContent | null`, which meant that your prev check/this check both works.

services/middlewareServices/FormSGProcessingService.ts
seaerchin: sure, that sounds good! didn't realise that `formKey` is secret oops, thanks for letting me know!

seaerchin (approved):
lgtm eskeddit

PR #459, opened 2022-06-29
Feat/site creation integrate

rc-davis:
This PR has 4 main goals:

1. Use new formsg authentication and decryption middleware in a route to be used as a FormSG webhook.
2.  Use the modified mail client to send an email to the form submitter on error or success.
3.  Save updated staging & production urls in _config.yml. (The CMS uses this staging URL when directing users to the staging deployment.
4. Store the deployment's Amplify id (as `hosting_id`) in the deployments table for future Amplify API calls.

To complete this PR, I made several big changes:
- The router was moved from `/newroutes/authenticated/sites.js` to `/newroutes/formsg.ts`. The site creation endpoint was consequently moved from `/v2/sites` to `/formsg/create-site`. I did this, because endpoints in `sites.js` use middleware that checks for GitHub credentials. FormSG can't get such credentials, so this endpoint cannot use the same middleware. It seemed cleanest to move the endpoint to an entirely new router that focuses on FormSG.
- Another consequence of the above was that InfraService and MailClient must be created in `server.js`, rather than `newroutes/authenticated/index.js`
- I also moved the mockMailer out of `services/identity/index.ts` and directly into `MailClient.ts` so it could be accessed by more services. I could have moved this into a new `services/index.js` file, but that didn’t seem necessary to me.


services/identity/ReposService.ts
```diff
@@ -34,14 +35,15 @@ type repoCreateParamsType = Partial<Repo> & {
...
-  private readonly repository: ReposServiceProps["repository"]
+  // NOTE: No need to explicitly specify types if they are assigned from props in constructor.
+  private readonly repository
 
   constructor({ repository }: ReposServiceProps) {
     this.repository = repository
   }
 
+  localRepoPath = (repoName: string) => `/tmp/${repoName}`
+
```
alexanderleegs, line 46: Nit: could we have a more descriptive `getLocalRepoPath`?

services/identity/ReposService.ts
```diff
@@ -76,8 +78,18 @@ export default class ReposService {
...
+    const dir = this.localRepoPath(repoName)
+
+    // 1. Set URLs in local _config.yml
+    const configPath = `${dir}/_config.yml`
+    const configFile = fs.readFileSync(configPath, "utf-8")
+    // Assume the last two lines of config contain outdated staging and prod urls
+    const lines = configFile.split("\n").slice(0, -2)
+    lines.push(`staging: ${stagingUrl}`)
+    lines.push(`prod: ${productionUrl}`)
+    fs.writeFileSync(configPath, lines.join("\n"))
```
alexanderleegs, line 90: If the last two lines are not useful, we could modify the site creation to remove those lines instead?

newroutes/formsg.ts
```diff
@@ -0,0 +1,65 @@
...
+import { attachFormSGHandler } from "@root/newmiddleware"
+import InfraService from "@services/infra/InfraService"
+
+const { SITE_CREATE_FORM_KEY } = process.env
+
+function getProp(responses: FormField[], name: string): string | undefined {
+  const response = responses.find(({ question }) => question === name)
+
+  return response?.answer
+}
```
alexanderleegs, line 18: I think this could be moved to a util file to make this router class cleaner

newroutes/formsg.ts
```diff
@@ -0,0 +1,65 @@
...
+    return res.status(200).send("OK")
+  }
+
+  getRouter() {
+    const router = express.Router({ mergeParams: true })
+
+    if (!SITE_CREATE_FORM_KEY) {
+      logger.error("SITE_CREATE_FORM_KEY environment variable is empty.")
+      // Throw an exception instead if it is vital for this to work.
+    }
```
alexanderleegs, line 56: I think it's useful to throw an exception in future when we move site creation to be doable on the cms by users - for now i guess we can leave it as is because it's used internally only

services/identity/MailClient.ts
```diff
@@ -38,3 +38,21 @@ class MailClient {
...
+  )
+}
+
+const mockMailer = {
+  sendMail: (email: string, subject: string, html: string) =>
+    logger.info(`Mock email sent to <${email}>, subject: ${subject}\n${html}`),
+} as MailClient
+export const mailer = IS_LOCAL_DEV
+  ? mockMailer
+  : new MailClient(POSTMAN_API_KEY!)
```
alexanderleegs, line 58: Could I check why the initialisation was moved from index.ts into MailClient itself? I was thinking the index file would be a better place to handle initialisation, as this is consistent across our database currently

services/infra/InfraService.ts
```diff
@@ -91,9 +134,20 @@ export default class InfraService {
...
       logger.info(`Successfully created site on Isomer, site ID: ${newSite.id}`)
 
-      // TODO: Handle success by sending email to user who requested to create site
+      await this.sendCreateSuccess(
+        requesterEmail,
+        repoName,
+        submissionId,
+        deployment.stagingUrl,
+        deployment.productionUrl
+      )
```
alexanderleegs, line 143: Should this be outside the try block? Otherwise failing at this step will attempt to send a repo creation email instead

alexanderleegs (commented):
Left some comments!

In addition, we don't seem to be handling Site/Job status toggling on failure - are we planning to add it in this PR? On a related note, I don't think we have a way to resolve intermediate failure states at the moment, are we planning to do this manually for now?

services/identity/DeploymentsService.ts
```diff
@@ -39,17 +39,18 @@ class DeploymentsService {
...
-    const amplifyDomain = amplifyResult.map((result) => result.defaultDomain)
-
-    if (amplifyDomain.isErr()) {
-      logger.error(`Amplify set up error: ${amplifyResult}`)
-      throw amplifyDomain.error
+    if (amplifyResult.isErr()) {
+      logger.error(`Amplify set up error: ${amplifyResult.error}`)
+      throw amplifyResult.error
     }
+    const amplifyInfo = amplifyResult.value
```
prestonlimlianjie, line 46: Note to self: do we not need .map?

newroutes/authenticated/sites.js
```diff
@@ -2,15 +2,11 @@ const autoBind = require("auto-bind")
 const express = require("express")
 
 // Import middleware
-const {
-  attachReadRouteHandlerWrapper,
-  attachWriteRouteHandlerWrapper,
-} = require("@middleware/routeHandler")
+const { attachReadRouteHandlerWrapper } = require("@middleware/routeHandler")
 
 class SitesRouter {
```
prestonlimlianjie, line 7: Could I understand the reason for moving the handler into a new FormSG router? And by extension, the API endpoint of `POST /formsg/create-site`?

services/identity/ReposService.ts
rc-davis: That is better. I'll make the change.

services/infra/InfraService.ts
```diff
@@ -35,15 +35,22 @@ export default class InfraService {
...
 
   createSite = async ({
-    email,
+    requesterEmail,
     repoName,
   }: {
-    email: string
-    repoName: string
+    requesterEmail: string
+    repoName: string | undefined
```
prestonlimlianjie, line 76: Is this because FormSG doesn't prevent empty strings? Could we leave a comment if this is the case?

Additionally, we could do a check for namespace collision by looking at the `repos` tables to see if any record with this `repoName` already exists. This could be addressed in a separate issue/PR.

services/identity/ReposService.ts
rc-davis: I'd like to, but we'd have to find another way to add them for Netlify sites. (We could give up on this and officially turn away from Netlify, but I have so far managed to keep our existing process running exactly as it was.)

prestonlimlianjie (commented):
Haven't tested it myself but had a few questions/comments

services/identity/ReposService.ts
rc-davis: Now that I'm thinking it through, we could get away with doing this, as these two lines are currently useless. However, it might be good to set up staging and prod deployments for this base repo and change the URLs instead of getting rid of them. If we did this, then the base repo would be editable like any other isomer site. That would be a nice feature.

newroutes/formsg.ts
rc-davis: I agree on both points.

services/identity/DeploymentsService.ts
rc-davis: We don't need `.map` in this case. `.map` converts one kind of a result to another kind of result. In this case, I want to convert an Ok result to the value it contains (`amplifyResult.value`), and convert an Err result to a thrown error (`amplifyResult.error`). Since a result could contain one or the other (and never both), it does surprise me that you can call both `.value` and `.error` on the same variable. The type inference is smart enough to figure out when it is safe to call one or the other.

But I'll admit here that I find all of this kind of wonky. The code moves back and forth between two error handling approaches: one that throws Errors and another that returns errors in neverthrow results. Code like what you see here will appear wherever neverthrow results are processed by code that expects thrown Errors. This seems inevitable as long as we are mixing approaches.

newroutes/authenticated/sites.js
rc-davis: I modified the top of this PR to explain why. Please let me know if you need more explanation.

services/infra/InfraService.ts
rc-davis: FormSG can prevent empty strings by making a field required, but I thought it best not to depend on that. It will always be possible to modify the form so that the field is no longer required.

> Additionally, we could do a check for namespace collision by looking at the repos tables to see if any record with this repoName already exists. 

Right! I forgot that my previous PR ensured this by adding a `unique: true` constraint to the repository name column. I'll add this back or make a note to add it later.

services/identity/MailClient.ts
rc-davis: I just spoke to @alexanderleegs about this, and we can't think of a better place to put this for now. I'll make a note to address this when we reorganize the service files.

services/infra/InfraService.ts
rc-davis: Yes, that's a bit better. I'll do that.

rc-davis: I cant easily move this outside the try/catch block, because the success message depends on information that is in the block’s scope. I could create mutable variable to hold that information and make it available in a wider scope, but that seems uglier than just leaving it the way it is.

newroutes/formsg.ts
rc-davis: Yes, that's a bit cleaner. I'll do that.

services/infra/InfraService.ts
```diff
@@ -11,41 +12,87 @@ interface InfraServiceProps {
...
     this.reposService = reposService
     this.deploymentsService = deploymentsService
+    this.mailer = mailer
+  }
+
+  sendCreateError = async (
+    email: string,
+    repoName: string | undefined,
+    submissionId: string | undefined,
+    error: string
```
seaerchin, line 47: i'm not sure if there's a way around this but the intent of this `sendCreateError` seems to be to provide a diagnostic for end users. hence, i'm not sure that passing `undefined | string` would be a good idea here. i think a better API here would either be to force devs to specify `repoName/submissionId` or if unable, provide a default argument/alternative text in the function body so that our users still get a useful diagnostic

services/infra/InfraService.ts
```diff
@@ -11,41 +12,87 @@ interface InfraServiceProps {
...
+    const html = `<p>Isomer site ${repoName} was <b>not</b> created successfully. (Form submission id [${submissionId}])</p>
+<p>${error}</p>
+<p>This email was sent from the Isomer CMS backend.</p>`
+    await this.mailer.sendMail(email, subject, html)
+  }
+
+  sendCreateSuccess = async (
+    email: string,
+    repoName: string,
+    submissionId: string | undefined,
```
seaerchin, line 59: could i double check why `submissionId` could be `undefined`?

services/infra/InfraService.ts
rc-davis: I just realized that there is a `name` column in both the `sites` table and `repos` table, and they both appear to contain the same thing. What's the rationale behind storing the repo name in the sites table? Can we use `sites.name` for a human-readable name instead?

services/infra/InfraService.ts
```diff
@@ -11,41 +12,87 @@ interface InfraServiceProps {
...
+    requesterEmail,
     repoName,
+    submissionId,
   }: {
-    email: string
-    repoName: string
+    requesterEmail: string
+    repoName: string | undefined
+    submissionId: string | undefined
   }) => {
```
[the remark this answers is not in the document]
seaerchin: the precondition to this working seems to be that we provide a `repoName` and `submissionId` - we could validate this prior to invoking this method so that this method can focus solely on business logic.

also, `submissionId` seems to be used solely for tracking/debugging purposes - perhaps we could defer the responsibility of sending mail up a level and keep this simple?

seaerchin (commented):
we seem to lack a mechanism for rollbacks when operations fail and this makes our db potentially inconsistent with the ground truth. 

eg: if we experience bad network connectivity and our request fails, our error mechanism only reports the error but doesn't do anything about it and our db state is inconsistent as it will show that the repo is created when in actuality is not. 

i think one way we could fix this is to use `neverthrow` across the services and return a **specific** instance of error. as this entails an `andThen` chain, failures won't cascade and we can match the error type to run the corresponding roillback operation

prestonlimlianjie:
> we seem to lack a mechanism for rollbacks when operations fail and this makes our db potentially inconsistent with the ground truth.
> 
> eg: if we experience bad network connectivity and our request fails, our error mechanism only reports the error but doesn't do anything about it and our db state is inconsistent as it will show that the repo is created when in actuality is not.
> 
> i think one way we could fix this is to use `neverthrow` across the services and return a **specific** instance of error. as this entails an `andThen` chain, failures won't cascade and we can match the error type to run the corresponding roillback operation

Perhaps we could address this as a separate issue altogether. Let's keep this PR simple and allow for some imperfection - we can trigger alerts from these errors and handle them manually for now.

newroutes/authenticated/sites.js
prestonlimlianjie: Copying my response from Slack to this PR:
Thanks for the explanation. I think that `POST /formsg/create-site` endpoint name isn't very intuitive because it's not clear what Formsg has to do with Isomer. 

I recommend keeping it to `POST /v2/sites` because the endpoint name is a better description of what the handler actually does; it creates a new Isomer site. I think it's fine for the `SiteRouter` to have methods with various middlewares.

prestonlimlianjie: @alexanderleegs @seaerchin thoughts?

rc-davis: I just discussed this with @prestonlimlianjie. I think he's about to leave another comment, but I'll attempt to summarize our discussion.

Having endpoints with different middleware in the same router would require some significant reworking of our routing architecture. It's hard to justify such reworking, since these FormSG endpoints may not be around for long. They were meant to be a stopgap measure until we can create sites in the CMS itself, hopefully in the next six months.

services/infra/InfraService.ts
rc-davis: I just discussed this with @prestonlimlianjie . He clarified that `sites.name` is meant to be visible to users, while `repos.name` is meant to contain the GitHub repository name.

newroutes/formsg.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  constructor({ infraService }: FormsgRouterProps) {
+    this.infraService = infraService
+    // We need to bind all methods because we don't invoke them from the class directly
+    autoBind(this)
+  }
+
+  async createSite(req: Request, res: Response) {
+    const { submissionId } = req.body.data
+    const { responses } = res.locals.submission as DecryptedContent
```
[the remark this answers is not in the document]
seaerchin: yeap! actually the types (`Request/Response`) exported by `express` have quite a number of generic parameters indicating the different components of the req/res!

so all we'd need to do here is to declare the types as follows: 
```ts
async createSite(
  req: Request<unknown, unknown, { data: { submissionId: string } }, unknown>,
  res: Response<unknown, { submission: DecryptedContent }>
  ) { ... }
```

we could also go a step further by declaring our own utility types for interfacing with express to smooth this process along in the future. do noite that because we're expecting a `res.locals.submission` here, we should ideally export the middleware together with the router or something bundling them together so that devs know when/where this gets set.

services/identity/ReposService.ts
rc-davis: Ok, my latest changes allow the staging and prod URLs to be anywhere in `_config.yml` and throw an error if either one is not present.

newroutes/formsg.ts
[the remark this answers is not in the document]
rc-davis: Thanks! I made the change. It would be great if we could get these types in place for all of our routes, but I think that will have to wait for another day.

newroutes/authenticated/sites.js
prestonlimlianjie: Thanks for summarizing the discussion - the main reason for splitting out the new `createSite` handler into its own router is because the existing sites router uses the `authenticatedSubrouter` (`authenticatedSubrouter.use("/sites", sitesV2Router.getRouter())`). This `authenticatedSubrouter` requires all routes to use JWT authentication, which the current `createSite` handler does not require, or have access to, since the endpoint is called by FormSG.

We will refactor the `createSite` handler back into the sites router once we move the site creation flow into the CMS, where the handler will have access to the JWT.

services/identity/MailClient.ts
seaerchin: agreed with @rc-davis on the fact that `MailClient` doesn't seem to belong in `services/identity`! in view of time pressure, i think we're filing an issue to shift `MailClient` to `UtilServices` and rename it to `MailService`; this is fine as a band-aid fix **for now**

newroutes/formsg.ts
```diff
@@ -0,0 +1,65 @@
...
+    const requesterEmail = getProp(responses, "Government E-mail")
+    const repoName = getProp(responses, "Repository Name")
+
```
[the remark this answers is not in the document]
rc-davis: Good Idea.

services/infra/InfraService.ts
rc-davis: This is no longer necessary.

services/infra/InfraService.ts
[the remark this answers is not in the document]
rc-davis: I'll use a standard argument list instead of an object.

I'll move other FormSG related stuff from InfraService into the router, which is FormSG specific. That should take care of your second comment.

seaerchin:
> > we seem to lack a mechanism for rollbacks when operations fail and this makes our db potentially inconsistent with the ground truth.
> > eg: if we experience bad network connectivity and our request fails, our error mechanism only reports the error but doesn't do anything about it and our db state is inconsistent as it will show that the repo is created when in actuality is not.
> > i think one way we could fix this is to use `neverthrow` across the services and return a **specific** instance of error. as this entails an `andThen` chain, failures won't cascade and we can match the error type to run the corresponding roillback operation
> 
> Perhaps we could address this as a separate issue altogether. Let's keep this PR simple and allow for some imperfection - we can trigger alerts from these errors and handle them manually for now.

yeap, agreed. that seems to be out of scope for this PR. i've added a new issue #460 to track this specifically

rc-davis:
I have pushed some more changes. Most of these are small changes, but there is a fairly significant refactor of InfraService. I have moved everything related to FormSG out of this service and into `formsgSiteCreation.ts` (the new name for FormSG router, previously named `formsg.ts`) Because of this, InfraService no longer depends on MailClient or UsersService, while the router does, so I had to do some rewiring.

services/identity/MailClient.ts
rc-davis: I added this as a line item to this issue #461, which we could discuss on Monday.

services/infra/InfraService.ts
rc-davis: This problem went away when I moved FormSG-related stuff out of InfraService and into formsgSiteCreation.

newroutes/formsgSiteCreation.ts
```diff
@@ -0,0 +1,154 @@
...
+const REQUESTER_EMAIL_FIELD = "Government E-mail"
+const SITE_NAME_FIELD = "Site Name"
+const REPO_NAME_FIELD = "Repository Name"
+
+export interface FormsgRouterProps {
+  usersService: UsersService
+  infraService: InfraService
+}
+
+export class FormsgRouter {
```
alexanderleegs, line 27: I think the router might be doing a bit too much work - ideally we've tried to keep the router layer as mostly for the parsing of request data and the business logic with services. I think we can tackle this in future with a separate PR!

rc-davis: Agreed. I put this as an item in #461.

PR #480, opened 2022-08-01
Feat/email login flow

alexanderleegs:
## Problem

This PR builds on https://github.com/isomerpages/isomercms-backend/pull/479, as part of our login flow release. To be reviewed in conjunction with PR #[996](https://github.com/isomerpages/isomercms-frontend/pull/996) on the isomercms-frontend repo. It introduces endpoints for the new login flow and modifies certain methods (site retrieval, commit messages) which were previously reliant on having a user's github id.
In addition, this PR also fixes a bug with our existing DB schema - we were previously associating Users to the Users table via the SiteMembers table, which has been corrected to associate the Users table with the Sites table. Aliases have also been added to the relations to allow for querying of related tables.

In summary, the following items have been implemented:

- New endpoints for verification and sending of email otp for login
- Updating of commit message to include user id of the person making the change
- Fixing of our existing database schema
- Modification of the existing users database table to allow for empty `githubId` field
- Separate flow for site retrieval for email and github users
- Updating of user by isomer id rather than github id

TBD before merging:

- Run db migration


seaerchin:
@alexanderleegs i think the otp test seems to be failing - i'll wait for you to doublecheck this before reviewing!

alexanderleegs:
> @alexanderleegs i think the otp test seems to be failing - i'll wait for you to doublecheck this before reviewing!

Thanks for catching, forgot to run tests before pushing! Have corrected the tests + made necessary bug fixes

src/services/utilServices/SitesService.js
```diff
@@ -29,12 +29,33 @@ const ISOMER_ADMIN_REPOS = [
...
+  }) {
     this.githubService = gitHubService
     this.configYmlService = configYmlService
+    this.usersService = usersService
+    this.isomerAdminsService = isomerAdminsService
   }
 
   async getSites(sessionData) {
+    const isEmailUser = sessionData.getIsEmailUser()
+    let emailAccessibleSites = []
```
seaerchin, line 46: nit: could we rename this to `retrievedSitesByEmail`? i got confused by this at first ._.

src/services/utilServices/SitesService.js
```diff
@@ -29,12 +29,33 @@ const ISOMER_ADMIN_REPOS = [
...
+    if (!isAdminUser && isEmailUser) {
+      const user = await this.usersService.findSitesByUserId(userId)
+      if (user) {
+        const { site_members: siteMemberEntries } = user
+        emailAccessibleSites = siteMemberEntries.map((entry) => {
+          const repoData = entry.repo
+          return repoData.name
+        })
+      }
+    }
```
seaerchin, line 58: this seems abit cluttered - wdyt about combining the nested `if` conditions to make this more clear? alternatively, we could shift this and the conditional check below into helper methods (maybe together with the retrieval?) which is fine too

src/services/utilServices/__tests__/AuthService.spec.js
```diff
@@ -101,19 +116,76 @@ describe("Auth Service", () => {
...
+        service.verifyOtp({ email: mockEmail, otp: mockOtp })
+      ).rejects.toThrow(BadRequestError)
+      expect(mockUsersService.verifyOtp).toHaveBeenCalledWith(
+        mockEmail,
+        mockOtp
+      )
+    })
+  })
+
   describe("getUserInfo", () => {
```
seaerchin, line 165: is it possible for github and email to be both unset and if so, should we add a test case checking that this throws an error? 

should we also check that invalid github ids and emails return errors?

src/services/utilServices/__tests__/SitesService.spec.js
```diff
@@ -24,18 +28,140 @@ describe("Resource Page Service", () => {
...
+
+      expect(mockIsomerAdminsService.getByUserId).toHaveBeenCalledWith(
+        mockIsomerUserId
+      )
+      expect(genericGitHubAxiosInstance.get).toHaveBeenCalledTimes(3)
+      process.env.ISOMERPAGES_REPO_PAGE_COUNT = currRepoCount
+      expect(process.env.ISOMERPAGES_REPO_PAGE_COUNT).toBe(currRepoCount)
+    })
+
+    it("Filters accessible sites for email user with no sitescorrectly", async () => {
```
seaerchin, line 131: nit: add spacing between sites and correctly

seaerchin (changes requested):

src/fixtures/sessionData.ts
```diff
@@ -19,5 +19,13 @@ export const mockSessionData = new SessionData({
...
+
+export const mockSessionDataEmailUser = new SessionData({
+  isomerUserId: mockIsomerUserId,
+  email: mockEmail,
+})
+
 mockSessionData.addGithubState(mockGithubState)
 mockSessionData.addSiteName(mockSiteName)
+mockSessionDataEmailUser.addGithubState(mockGithubState)
+mockSessionDataEmailUser.addSiteName(mockSiteName)
```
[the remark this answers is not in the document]
alexanderleegs: Yup, fixed in the refactor

src/routes/v2/authenticated/__tests__/Sites.spec.js
```diff
@@ -59,28 +59,6 @@ describe("Sites Router", () => {
     })
   })
 
-  describe("checkHasAccess", () => {
```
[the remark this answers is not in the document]
alexanderleegs: Forgot to add new tests, thanks for catching! The equivalent tests have been introduced in https://github.com/isomerpages/isomercms-backend/pull/480/commits/a30c37d1f3e9427497fd4968d3509cb7c9477836

src/services/db/__tests__/GitHubService.spec.js
```diff
@@ -596,6 +611,10 @@ describe("Github Service", () => {
       const secondSha = "second-sha"
       const gitTree = "git-tree"
       const message = "message"
+      const finalExpectedMessage = JSON.stringify({
+        message: message || `isomerCMS updated ${siteName} state`,
+        userId,
+      })
```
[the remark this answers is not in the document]
alexanderleegs: You're right, this isn't needed - removed in https://github.com/isomerpages/isomercms-backend/pull/480/commits/0fa4fd60fde264238ddc7ca7c2dfbe651151005d

src/services/identity/UsersService.ts
```diff
@@ -58,12 +58,41 @@ class UsersService {
...
+          required: true,
+          include: [
+            {
+              model: Repo,
+              required: true,
+              where: {
+                name: siteName,
+              },
+            },
+          ],
```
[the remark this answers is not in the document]
alexanderleegs: `Repo` is the database table containing `siteName`, which we need to use for comparison!

src/services/identity/UsersService.ts
```diff
@@ -58,12 +58,41 @@ class UsersService {
...
 
+  async findSitesByUserId(isomerId: string) {
+    return this.repository.findOne({
+      where: { id: isomerId },
+      include: [
+        {
+          model: Site,
+          as: "site_members",
+          required: true,
+          include: [{ model: Repo, required: true }],
```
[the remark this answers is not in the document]
alexanderleegs: See above

src/services/utilServices/AuthService.js
```diff
@@ -90,19 +92,49 @@ class AuthService {
...
-        headers: {
-          Authorization: `token ${accessToken}`,
-          "Content-Type": "application/json",
-        },
-      })
-      const userId = resp.data.login
+      if (sessionData.getIsEmailUser()) {
+        const email = sessionData.getEmail()
+        const { contactNumber } = await this.usersService.findByEmail(email)
+        return { userId: "", email, contactNumber }
```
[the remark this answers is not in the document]
alexanderleegs: Makes sense, https://github.com/isomerpages/isomercms-backend/pull/480/commits/6c0561fc8643e077980d3ab382ef0f538901292f

src/services/utilServices/SitesService.js
alexanderleegs: Sure, https://github.com/isomerpages/isomercms-backend/pull/480/commits/0fdffecba8dfc13ac707441a01c702444935db99

src/services/utilServices/SitesService.js
alexanderleegs: Shifted into a helper method in https://github.com/isomerpages/isomercms-backend/pull/480/commits/88065efa8178db74d582fe792c7242f5780248f7

src/services/utilServices/__tests__/AuthService.spec.js
alexanderleegs: It shouldn't be possible for both to be unset - this should be caught by an earlier middleware.

I think the invalid checks belongs in the called methods - otherwise we're just forcing the called method to throw an error

src/services/utilServices/__tests__/SitesService.spec.js
alexanderleegs: Fixed in https://github.com/isomerpages/isomercms-backend/pull/480/commits/686087e5ee9a7cb687931ed23f0774783f633f9a

src/services/identity/UsersService.ts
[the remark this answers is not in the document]
seaerchin: ah ok, gotcha! thanks for clarifying

src/services/utilServices/AuthService.js
```diff
@@ -9,7 +9,9 @@ const { ForbiddenError } = require("@errors/ForbiddenError")
 const validateStatus = require("@utils/axios-utils")
```
seaerchin, line 9: i just realised this is also `AuthService` :monkas:

src/services/utilServices/SitesService.js
```diff
@@ -29,12 +29,38 @@ const ISOMER_ADMIN_REPOS = [
...
+  async getEmailUserSites(userId) {
+    const user = await this.usersService.findSitesByUserId(userId)
+    if (user) {
+      const { site_members: siteMemberEntries } = user
+      return siteMemberEntries.map((entry) => {
+        const repoData = entry.repo
+        return repoData.name
+      })
+    }
+    return []
```
seaerchin, line 53: we could flip the condition so this reads more smoothly

src/services/utilServices/SitesService.js
```diff
@@ -74,6 +100,12 @@ class SitesService {
             isPrivate,
           }
         })
+        .filter(
+          (repoData) =>
+            isAdminUser ||
+            !isEmailUser ||
+            retrievedSitesByEmail.includes(repoData.repoName)
+        )
```
seaerchin, line 108: could i ask why we're doing a `filter` operation here rather than just an early return above, when `isAdminUser` and `isEmailUser` is known? 

it seems like it'll still return an empty array and it would be better for readability too

src/services/utilServices/SitesService.js
```diff
@@ -29,12 +29,38 @@ const ISOMER_ADMIN_REPOS = [
...
   }
 
   async getSites(sessionData) {
+    const isEmailUser = sessionData.isEmailUser()
+    let retrievedSitesByEmail = []
+    const { isomerUserId: userId } = sessionData
+    const isAdminUser = !!(await this.isomerAdminsService.getByUserId(userId))
+    if (!isAdminUser && isEmailUser)
+      retrievedSitesByEmail = await this.getEmailUserSites(userId)
+
```
seaerchin, line 63: not sure if this is out of scope for this PR but wdyt about refactoring this to 2 methods called `getSitesForUser` and `getSitesForAdmin`? 

this method is p confusing to read through at the moment, especially with the nested `map/filter` operations ._.

src/services/utilServices/__tests__/AuthService.spec.js
```diff
@@ -101,21 +116,77 @@ describe("Auth Service", () => {
...
+        service.verifyOtp({ email: mockEmail, otp: mockOtp })
+      ).resolves.toEqual(signedToken)
+      expect(mockUsersService.verifyOtp).toHaveBeenCalledWith(
+        mockEmail,
+        mockOtp
+      )
+      expect(mockUsersService.loginWithEmail).toHaveBeenCalledWith(mockEmail)
+    })
+
+    it("should throw an error if otp is incorrect", async () => {
```
seaerchin, line 152: hmm, is it possible to add a test case for expired otps? (lmk if this is diff to do)

seaerchin (commented):
not sure if this is done but i saw that the changes to _integration_ tests were q minor and seemed to be compat fixes rather than true integration tests - do we already have integration tests for this and if not, should we do so, considering that this is a p important flow?

src/services/utilServices/SitesService.js
alexanderleegs: https://github.com/isomerpages/isomercms-backend/pull/480/commits/9607bdc51f19e84a9c33cb909573417825a46ad8

src/services/utilServices/SitesService.js
alexanderleegs: Hmm I think this part isn't very clear - the idea was for all sites retrieved to pass through if the user was an admin user, or using their own github token (since that only retrieves repos they have access to). I've refactored the code to make this clearer in https://github.com/isomerpages/isomercms-backend/pull/480/commits/86cdc315049dfd6dead583c715fefe02329822ce

src/services/utilServices/SitesService.js
alexanderleegs: Same as above, I think this part might have been confusing, so I refactored the code - isAdminUser is supposed to simply allow the result of the github api query to return all sites. Changed in https://github.com/isomerpages/isomercms-backend/pull/480/commits/86cdc315049dfd6dead583c715fefe02329822ce, lmk if it's still confusing

src/services/utilServices/__tests__/AuthService.spec.js
alexanderleegs: I actually don't know if there's a special error message for expired otps - I've only been able to receive wrong otp error messages, so I don't have a test case for expired ones

src/services/utilServices/AuthService.js
alexanderleegs: Do you think we should also rename this service? We could file an issue (and there's also an `AuthService` under identity)

src/routes/v2/authenticated/__tests__/Sites.spec.js
```diff
@@ -61,28 +61,6 @@ describe("Sites Router", () => {
...
```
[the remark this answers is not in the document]
alexanderleegs: Yep

src/services/api/AxiosInstance.ts
```diff
@@ -13,7 +13,11 @@ const requestFormatter = async (config: AxiosRequestConfig) => {
   const authMessage = config.headers.Authorization
 
   // If accessToken is missing, authMessage is `token `
-  if (!authMessage || authMessage === "token ") {
+  if (
+    !authMessage ||
+    authMessage === "token " ||
+    authMessage === "token undefined"
+  ) {
```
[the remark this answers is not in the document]
alexanderleegs: I also don't really like this, but we either seem to be having a blank access token or undefined access token and this was the place which required the least code change to fix - i figured we could leave it as this is relatively temporary (until we move away from github login entirely)

alexanderleegs:
> not sure if this is done but i saw that the changes to _integration_ tests were q minor and seemed to be compat fixes rather than true integration tests - do we already have integration tests for this and if not, should we do so, considering that this is a p important flow?

Hmm there isn't actually anything which really queries the database to require an integration test other than the new `getSites` - i've added an integration test for that 86cdc31

src/services/utilServices/SitesService.js
seaerchin: double checking here - the reason why `!isEmailUser` is part of the condition (`isAdminuser || !isEmailUser`) is because we use the user's own github access token to retrieve the sites and github will resolve permissions for us on their end?

src/services/utilServices/SitesService.js
seaerchin: i've made a comment below wrt this - i think it reads better now but the underlying assumption (that if they're github users, then they are using their own token, which will retrieve your owned sites and we can omit the `filter` on our end) should be documented here!

src/services/api/AxiosInstance.ts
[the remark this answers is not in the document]
seaerchin: yea, i think fixing this is beyond hte scope of this PR but just smth to keep in mind :(

src/services/utilServices/AuthService.js
seaerchin: tbh, our naming is p confusing so i feel like we should; i think filing an issue might be kicking the can down the road so if it's convenient here, i'm more inclined to doing it here (TIA HAHAHAH)

src/services/utilServices/__tests__/AuthService.spec.js
seaerchin: i looked through - there isn't; **all** verification is treated equally and we return the same error message even if the underlying error is different...

idt this is part of this PR (probably out of scope) but i think we should be aware of this and i'll file an issue accordingly @__@

src/integration/Sites.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+import { SitesService } from "@root/services/utilServices/SitesService"
+import { getUsersService } from "@services/identity"
+import { sequelize } from "@tests/database"
+
+const mockSite = "mockSite"
+const mockSiteId = "1"
+const mockBlockedSite = "adminOnly"
+const mockUpdatedAt = "now"
+const mockPermissions = { push: true }
+const mockPrivate = true
```
seaerchin, line 23: i think we can omit this :P but if you wanna keep this, could we rename it to `mockIsPrivate`?

src/integration/Sites.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+      ;((genericGitHubAxiosInstance.get as unknown) as jest.Mock).mockImplementationOnce(
+        () => ({
+          data: [],
+        })
+      )
+      ;((genericGitHubAxiosInstance.get as unknown) as jest.Mock).mockImplementationOnce(
+        () => ({
+          data: [],
+        })
+      )
```
seaerchin, line 140: actually i think this works b/c there's a default `axios` mock in `__mocks__` - we should be using that instead HAHA

src/integration/Sites.spec.ts
```diff
@@ -0,0 +1,205 @@
...
```
seaerchin, line 196: should we consider using `mockAxios`?

src/integration/Sites.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+      })
+    })
+    afterEach(async () => {
+      // Clean up so that different tests using
+      // the same mock user don't interfere with each other
+      await IsomerAdmin.destroy({
+        where: { userId: mockIsomerUserId },
+      })
+    })
+    it("should return list of only sites available to user", async () => {
```
seaerchin, line 101: nit: we should specify that this is an email user as github users have slightly diff behaviour

src/integration/Sites.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+subrouter.use(sitesSubrouter)
+const app = generateRouter(subrouter)
+
+describe("Sites Router", () => {
+  afterEach(() => {
+    jest.resetAllMocks()
+    mockAxios.reset()
+  })
+
+  describe("/", () => {
```
seaerchin, line 61: could we have a few more test cases so that these invariants are clear

1. for github users, their access token only gives them access to their site
2. sites w/o `siteMembers` are seen by admins (i think this is so for `blockedSite`? if so, maybe just rename it to make it clearer)
3. public sites can be seen by everyone (<-- not q sure what should be the behaviour here, correct me if i'm wrong)

src/services/utilServices/SitesService.js
```diff
@@ -47,7 +68,7 @@ class SitesService {
       })
     )
 
-    const sites = await Bluebird.map(paramsArr, async (params) => {
+    const allSites = await Bluebird.map(paramsArr, async (params) => {
```
seaerchin, line 71: nit: `allSites` reads slightly weird but as i have no good name, i'm ok w/ just leaving this as is ._.

src/services/utilServices/AuthService.js
alexanderleegs: It would probably involve frontend changes as well (endpoints are currently located in `/auth/...`) - hence why i think this would be better to handle in a separate PR

src/integration/Sites.spec.ts
alexanderleegs: Hmm i was thinking this is useful because we can check that regular users only have access to items in the site members database that include them, while admin users see all by default - I think it might be better to leave in, but I'm not sure if `mockIsPrivate` conveys that idea

src/integration/Sites.spec.ts
alexanderleegs: I tried that earlier actually, but couldn't get it to work - I think it might have something to do with the `AxiosInstance` defined elsewhere which uses both `create` and interceptors. This was the way I got it to work with `genericGithubAxiosInstance`, but do let me know if it can be done with `mockAxios` instead!

src/integration/Sites.spec.ts
alexanderleegs: 1. I left this one out of the integration tests because it's covered by the unit tests - the Github API call is the one which restricts the sites accessible to the user, so I didn't think there was much benefit in including it here - we'd just be mocking the Github results and seeing that we receive the same mocked results.
2. Yup that was the purpose of `blockedSite` - I've renamed it to `mockAdminSite` if that's clearer? in 1a88af5, lmk if there's a better name for it
3. Hmm I think there's no such thing as public sites really - sites are "public" in that anyone can view them, but access to sites are still restricted based on whether someone has _write_ permissions to the site, which is what the getSites method is returning

src/services/utilServices/SitesService.js
alexanderleegs: Makes sense, added comment in 85aa065

src/integration/Sites.spec.ts
alexanderleegs: Fixed in https://github.com/isomerpages/isomercms-backend/pull/480/commits/1a88af565617b2f51cb3932b664c811902078c4c!

CHANGELOG.md
```diff
@@ -4,8 +4,26 @@ All notable changes to this project will be documented in this file. Dates are d
 
 Generated by [`auto-changelog`](https://github.com/CookPete/auto-changelog).
 
+#### [v0.10.1](https://github.com/isomerpages/isomercms-backend/compare/v0.10.0...v0.10.1)
```
seaerchin, line 7: assuming this is beacuse of merge?

src/integration/Sites.spec.ts
seaerchin: yeap sure, 1/2 seems reasonable! so just double checking for 3 - when we call `getSites`, this is restricted based on _write_ access and not _read_ access? this is because all of our sites are public (on github) and when we say write access, in this context what it means is **users who are members of that github organization that own the site**? think the semantics of this needs to be nailed down HAHA

src/integration/Sites.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+          },
+          {
+            lastUpdated: mockUpdatedAt,
+            repoName: mockAdminSite,
+            isPrivate: mockPrivate,
+            permissions: mockPermissions,
+          },
+        ],
+      }
+      ;((genericGitHubAxiosInstance.get as unknown) as jest.Mock).mockImplementation(
```
seaerchin, line 169: this actually impacts readability pretty significantly .-. do we have a better way around this? given that this is a test file, i'm not too concerned by it but if there's a quick way to rewrite this, i think we should

src/services/utilServices/SitesService.js
```diff
@@ -81,7 +102,16 @@ class SitesService {
         )
     })
 
-    return _.flatten(sites)
+    const flattenedAllSites = _.flatten(allSites)
+    // Github users are using their own access token, which already filters sites to only those they have write access to
+    if (isAdminUser || !isEmailUser) return flattenedAllSites
```
seaerchin, line 107: sanity check here - should this be `!isAdminUser || !isEmailUser)` or simply `!isEmailUser`? from the comment, this seems like it should be a non-admin user using their own github access token (which follows the same flow as an admin user)

seaerchin (commented):
some comments, overall looks good!

src/integration/Sites.spec.ts
seaerchin: resolved this - i left a comment in the code for this but the tl;dr is that `SitesService` is hitting the `genericGithubAxiosInstance`, which is a named export. hence, this requires a mock of the named export after which everything will work fine

src/services/utilServices/AuthService.js
seaerchin: sure, this sounds reasonable - could we file a tracking issue?

src/integration/Sites.spec.ts
alexanderleegs: Yup! To clarify - the github repos are all public, and anyone can access them via github directly. On our CMS, only sites that a user is allowed to write to are viewable - for github login users, that's via checking push permissions (`repoData.permissions.push === true`), and for email login, via checking the `site_members` table

src/services/utilServices/SitesService.js
alexanderleegs: Hmm the check is correct - to clarify, the flows are:

1. Email user - uses our centralised access token, so flattenedAllSites returns all sites, so we need to further restrict based on our database
2. Email user (admin) - uses our centralised access token, so flattenedAllSites returns all sites.
3. Github user - uses their own access token, so flattenedAllSites returns only those they have write access to
4. Github user (admin) - uses their own access token, so flattenedAllSites returns only those they have write access to (which is all sites)

I've added an extra comment about admin users in f23cbd2!

src/integration/Sites.spec.ts
seaerchin: ok, thanks for clarifying!

src/services/utilServices/SitesService.js
seaerchin: gotcha i think i misunderstood the conditional as github login is caught by the `!isEmailUser` portion

seaerchin (approved):
lgtm

src/services/utilServices/AuthService.js
alexanderleegs: Yup, issue filed https://github.com/isomerpages/isomercms-backend/issues/498

PR #487, opened 2022-08-12
fix: remove unnecessary update step

alexanderleegs:
## Problem

This PR fixes an issue introduced by https://github.com/isomerpages/isomercms-backend/pull/482 - we had previously removed the description in the repo update step as we believed it was not necessary - however, this caused issues with the site creation form, as we were attempting to call update on the repo with nothing to update. This PR removes the extra step to prevent this bug from occurring.


kishore03109 (approved):
lgtm

PR #490, opened 2022-08-16
fix: remove resources_name and add support for url

### LINKED ISSUE #454: [BUG] Additional parameter being passed to FE
**Description**
see [here](https://github.com/isomerpages/isomercms-frontend/pull/931#discussion_r897647436) for a description. basically, our BE is passing an extra param called `resources_name` back to frontend, which will cause the type-checking to fail on the frontend and leads to extra work to omit this param. 

**Suggested Solution**
delete the parameter

dcshzj:
## Problem

Closes #454.

## Solution

**Breaking Changes**

**Bug Fixes**:

- The `resources_name` parameter has been removed from the response when calling GET `/v2/sites/:siteName/settings`, as it is an extra parameter and can cause issues on the frontend.
- The `url` parameter has been enhanced to support POST `/v2/sites/:siteName/settings`.
  - The parameter only takes in a valid domain name.
  - All invalid inputs will throw a 400 BadRequestError.

## Tests

**Unit tests**:
- Run `npm run tests`

**Smoke tests**:
- Check out the `fix/settings-parameters` branch on the CMS frontend.
- Run `npm run dev`.
- Log in and navigate to any site dashboard, then access the "Site settings" page.
- Modify the value in the "Site URL" input field.
- Verify that the new settings can be saved and correctly displayed upon page refresh.

## Deploy Notes

*None*

### DIFF
--- a/src/fixtures/config.js
+++ b/src/fixtures/config.js
@@ -118,7 +118,7 @@ const configContent = {
 const configSha = "configsha"
 
 const configResponse = {
-  url: configContent.url,
+  url: configContent.url.replace("https://", ""),
   title: configContent.title,
   description: configContent.description,
   favicon: configContent.favicon,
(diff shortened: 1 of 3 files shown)


alexanderleegs (approved):
lgtm

PR #496, opened 2022-08-29
Fix/unexpected footer fields

alexanderleegs:
## Problem

This PR fixes an issue with existing legacy site formats. Previously, if a site has extra footer fields, we would run into an error when saving settings on the frontend, as we specify the exact data structure that we are supposed to receive. This PR constrains the fields provided to only those that we expect. It also fixes an issue with handling empty social media fields.

### DIFF
--- a/src/services/configServices/SettingsService.js
+++ b/src/services/configServices/SettingsService.js
@@ -137,16 +137,17 @@ class SettingsService {
 
   mergeUpdatedFooterData(currentData, updatedData) {
     // Special configuration to remove empty footer settings entirely so they don't show up in the actual site
+    // Any updated data with empty strings are to be deleted from the footer object
     const clonedCurrentData = _.cloneDeep(currentData)
     Object.keys(updatedData).forEach((field) => {
       if (field === "social_media") {
+        // social_media is an object with each social link in a separate property
         // Set social_media field if it doesn't exist
         if (!clonedCurrentData[field]) clonedCurrentData[field] = {}
         const socials = updatedData[field]
         Object.keys(socials).forEach((social) => {
           if (!socials[social]) {
-            if (clonedCurrentData[field][social])
-              delete clonedCurrentData[field][social]
+            delete clonedCurrentData[field][social]
           } else {
             clonedCurrentData[field][social] = updatedData[field][social]
           }
--- a/src/services/configServices/SettingsService.js
+++ b/src/services/configServices/SettingsService.js
@@ -135,33 +135,28 @@ class SettingsService {
     return clonedCurrentData
   }
 
+  cloneDeepNonEmpty(originalObj, updatedObj) {
+    const clonedOriginalObj = originalObj ? { ...originalObj } : {}
+    Object.keys(updatedObj).forEach((field) => {
+      if (typeof updatedObj[field] === "object" && updatedObj[field] !== null) {
+        clonedOriginalObj[field] = this.cloneDeepNonEmpty(
+          originalObj[field],
+          updatedObj[field]
+        )
+      } else if (updatedObj[field] === "") {
+        // Check for empty string because false value exists
+        delete clonedOriginalObj[field]
+      } else {
+        clonedOriginalObj[field] = updatedObj[field]
+      }
+    })
+    return clonedOriginalObj
+  }
+
   mergeUpdatedFooterData(currentData, updatedData) {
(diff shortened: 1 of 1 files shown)


src/services/configServices/SettingsService.js
```diff
@@ -140,12 +140,15 @@ class SettingsService {
...
+        if (!clonedCurrentData[field]) clonedCurrentData[field] = {}
         const socials = updatedData[field]
         Object.keys(socials).forEach((social) => {
           if (!socials[social]) {
-            delete clonedCurrentData[field][social]
+            if (clonedCurrentData[field][social])
+              delete clonedCurrentData[field][social]
           } else {
-            clonedCurrentData[field] = updatedData[field]
+            clonedCurrentData[field][social] = updatedData[field][social]
```
seaerchin, line 151: not sure if out of scope but this has 4 levels of `if/else` and lacks comments so it's q difficult to follow ._. perhaps we should try to make this more readable **if** we have time/it's easy

src/services/configServices/SettingsService.js
```diff
@@ -63,7 +63,7 @@ class SettingsService {
         mergedConfigContent.url !== "" &&
         !mergedConfigContent.url.startsWith("https://")
       ) {
-        mergedConfigContent.url = "https://" + mergedConfigContent.url
+        mergedConfigContent.url = `https://${mergedConfigContent.url}`
```
kishore03109, line 66: Just to clarify, this change was done for your personal preference in code right? Correct me if I am wrong but `"https://" + mergedConfigContent.url`  and  ``https://${mergedConfigContent.url}`` are semantically equivalent right?

src/services/configServices/SettingsService.js
```diff
@@ -140,12 +140,15 @@ class SettingsService {
     const clonedCurrentData = _.cloneDeep(currentData)
     Object.keys(updatedData).forEach((field) => {
       if (field === "social_media") {
+        // Set social_media field if it doesn't exist
+        if (!clonedCurrentData[field]) clonedCurrentData[field] = {}
```
kishore03109, line 144: Nit: Can we add curly braces around this one line if condition for maintainability ya

kishore03109 (changes requested):

src/services/configServices/SettingsService.js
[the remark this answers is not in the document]
dcshzj: I can take this since I introduced this line earlier. The `url` will only be a domain because it is enforced by the validators check:

https://github.com/isomerpages/isomercms-backend/blob/dd5de4a9bfb75f227ce8b40dacda45056126109b/src/validators/RequestSchema.js#L239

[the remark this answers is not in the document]
seaerchin: gotcha, thanks!

alexanderleegs: Yup, this was actually auto modified by our linter I think

src/services/configServices/SettingsService.js
alexanderleegs: Hmm I've added comments and removed one layer of unnecessary conditionals in https://github.com/isomerpages/isomercms-backend/pull/496/commits/2103a6a94c99c7f1586caf9cee969a85ea6cb34c - lmk if it's still unclear!

src/services/configServices/SettingsService.js
alexanderleegs: Sure, added in https://github.com/isomerpages/isomercms-backend/pull/496/commits/7e4c001c0336c36010f40ec427e0d030b007558a

src/services/configServices/SettingsService.js
seaerchin: after reading through this, i feel that we should extract this pattern out into a function, perhaps called `cloneDeepNonEmpty` (pls feel free to suggest a better naming ._.)

this is because what this is trying to do is basically trying to clone non empty fields for both the top level object as well as `socialMediaData`

kishore03109 (approved):

src/services/configServices/SettingsService.js
alexanderleegs: https://github.com/isomerpages/isomercms-backend/pull/496/commits/cc11c5f7d34b260acba0ddfefd7eb10c6ab4ecc2

PR #504, opened 2022-09-27
Chore/update dependencies

alexanderleegs:
This PR runs `npm audit fix` and additionally upgrades some outdated packages to resolve vulnerabilities (`winston-cloudwatch` to upgrade `vm2`, and `moment`)

### DIFF
--- a/src/logger/logger.js
+++ b/src/logger/logger.js
@@ -5,7 +5,7 @@ const Bluebird = require("bluebird")
 const moment = require("moment-timezone")
 // Logging tools
 const winston = require("winston")
-const WinstonCloudWatch = require("winston-cloudwatch")
+const WinstonCloudwatch = require("winston-cloudwatch")
 
 // Env vars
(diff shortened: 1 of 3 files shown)


src/logger/logger.js
```diff
@@ -66,7 +66,7 @@ class CloudWatchLogger {
           handleExceptions: true,
         }
 
-        this._logger.add(new WinstonCloudWatch(cloudwatchConfig))
+        this._logger.add(new WinstonCloudwatch(cloudwatchConfig))
```
dcshzj, line 69: Can I check if we need to add a `cloudWatchLogs: new AWS.CloudWatchLogs()` in the cloudwatchConfig? Otherwise I'm not sure why we need to add `@aws-sdk/client-cloudwatch-logs` into our package.json.

alexanderleegs: Seems like our logs are still fine!

dcshzj (approved):

PR #508, opened 2022-10-03
Feat/notifications

alexanderleegs:
## Problem

This PR adds a partial implementation of the notifications feature, as described in https://www.notion.so/opengov/Comments-Notifications-4332401f397c487e9a6e3bf641b23814#9557a87f679e458ba0b1cfc6c3d11eca. 

It adds a baseline for notification creation, retrieval, and updating, but actual notification creation will be tackled in a future PR, after review requests have been completed. It also adds a notifications table to our database, and modifies our existing site members table to add an id field for easier reference by related tables. To be reviewed in conjunction with PR https://github.com/isomerpages/isomercms-frontend/pull/1092 on the isomercms-frontend repo.

Key changes

- Added new P1 notification endpoints (getRecentNotifications, getAllNotifications, markNotificationsAsRead)
- Added NotificationService to handle above endpoints, as well as generic create endpoint to support notification creation in future
- Added Notification database table
- Modified existing SiteMember database table to add id column and swap primary key

Future work

- Notification creation as side effect of actions

Testing

As notifications cannot be created currently, we have to manually populate the notifications table. To do so:

1. Run db:migrate
2. Add the following notifications to the notifications table
```
1	{your_site_member_id}	{your_site_id}	{your_user_id}	user created a review request.	google.com	username	sent_request	NULL	2	2022-09-30 08:57:20.689+00	2022-10-02 06:27:36.233+00
2	{your_site_member_id}	{your_site_id}	{your_user_id}	different message here to read	google.com	diffuser	sent_request	2022-10-03 04:42:01.784979+00	2	2022-09-30 08:57:20.689+00	2022-10-02 06:27:36.233+00
3	{your_site_member_id}	{your_site_id}	{your_user_id}	third different message here	google.com   undefined     	sent_request	2022-10-03 03:36:40.597+00	2	2022-09-30 08:57:20.689+00	2022-10-03 03:36:40.597+00
```
3. The notification which has no first_read_time should show up as an unread notification on the frontend, and be marked as read once viewed. The notification which has no username defined should have a generic bell icon show up instead of the user's initials (icon to be changed later depending on notification type)


src/services/identity/NotificationsService.ts
```diff
@@ -0,0 +1,258 @@
...
+      // Update existing notification
+      await recentTargetNotification.update({
+        firstReadTime: null,
+        createdAt: new Date(),
+      })
+    } else {
+      // Create new notification
+      await this.repository.create({
+        siteMemberId: siteMember?.id,
+        siteId: siteMember?.siteId,
```
dcshzj, line 242: Nit: What happens if `siteMember` cannot be found? Would it cause `this.repository.create()` to be called with a `null` value, resulting in a database error?

src/routes/v2/authenticatedSites/notifications.ts
```diff
@@ -0,0 +1,92 @@
...
+  constructor({ notificationsService }: NotificationsRouterProps) {
+    this.notificationsService = notificationsService
+    autoBind(this)
+  }
+
+  getRecentNotifications: RequestHandler<
+    never,
+    unknown,
+    unknown,
+    unknown,
```
dcshzj, line 30: Should these be updated to either `never` or the `notifications` type? Same for the rest in this file.

dcshzj (changes requested):

src/services/identity/NotificationsService.ts
alexanderleegs: I modified this endpoint in a [later PR](https://github.com/isomerpages/isomercms-backend/pull/523) to use existing entries - this way we can enforce that the user does actually exist

src/routes/v2/authenticatedSites/notifications.ts
alexanderleegs: https://github.com/isomerpages/isomercms-backend/pull/508/commits/036340a7f81bdec0dd8a95a9bd5fc029725f1736

src/services/identity/NotificationsService.ts
```diff
@@ -0,0 +1,258 @@
...
+              required: true,
+              where: {
+                name: siteName,
+              },
+            },
+          ],
+        },
+      ],
+    })
+    const recentTargetNotification = await this.repository.findOne({
```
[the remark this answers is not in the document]
alexanderleegs: We reusing it because we don't want to create new notifications if the previous one is too recent, i've elaborated more on this in https://github.com/isomerpages/isomercms-backend/pull/508/commits/9c4f6233a930ce03a5f2c668deeef114737e81a5

src/services/identity/NotificationsService.ts
```diff
@@ -0,0 +1,258 @@
...
+              model: Repo,
+              required: true,
+              where: {
+                name: siteName,
+              },
+            },
+          ],
+        },
+      ],
+    })
```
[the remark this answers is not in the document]
alexanderleegs: Makes sense, https://github.com/isomerpages/isomercms-backend/pull/508/commits/9c4f6233a930ce03a5f2c668deeef114737e81a5

src/database/models/Site.ts
```diff
@@ -10,6 +10,7 @@ import {
   HasOne,
   BelongsTo,
   ForeignKey,
+  HasMany,
```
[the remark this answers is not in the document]
alexanderleegs: https://github.com/isomerpages/isomercms-backend/pull/508/commits/54c23db0ed6e5eef45c778e9972ee6967f01e63b

src/database/models/Site.ts
```diff
@@ -19,6 +20,8 @@ import { Repo } from "@database/models/Repo"
 import { SiteMember } from "@database/models/SiteMember"
 import { User } from "@database/models/User"
 
+import { Notification } from "./Notification"
```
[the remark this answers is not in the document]
alexanderleegs: https://github.com/isomerpages/isomercms-backend/pull/508/commits/54c23db0ed6e5eef45c778e9972ee6967f01e63b

src/database/models/User.ts
```diff
@@ -10,6 +10,7 @@ import {
   HasMany,
 } from "sequelize-typescript"
 
+import { Notification } from "@database/models/Notification"
```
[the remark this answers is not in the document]
alexanderleegs: https://github.com/isomerpages/isomercms-backend/pull/508/commits/54c23db0ed6e5eef45c778e9972ee6967f01e63b

src/services/identity/NotificationsService.ts
```diff
@@ -0,0 +1,249 @@
...
+    userId: string
+    findOptions?: FindOptions<Notification>
+  }) {
+    return this.repository.findAll({
+      where: {
+        user_id: userId,
+      },
+      order: [
+        ["first_read_time", "DESC NULLS FIRST"],
+        ["priority", "ASC"],
```
dcshzj, line 65: Sorry one more point, should this priority be descending instead? Meaning that a higher priority should be shown first.

Also, should we also include the notification creation time and sort by descending order? So the sorting order is something like this:
1. Notifications that the user has not seen before will be at the top.
2. Notifications that are more recent are shown first, if the user has not seen it before or multiple notifications have the same first read time.
3. If all else equal, then show notifications with a higher priority first.

dcshzj (changes requested):

src/services/identity/NotificationsService.ts
dcshzj: Hmm I feel that it might be safer to have an additional check here, but it's just a minor point so feel free to skip!

alexanderleegs: Oh as in the method uses an existing `siteMember` as an argument, so there's no situation where they don't exist

src/services/identity/NotificationsService.ts
alexanderleegs: I've add the time the notification is created to the sorting criteria! But 2 and 3 are swapped - when a user first logs in, they should see in order:

1. Unseen notifications
2. High priority notifications
3. Notifications created more recently

The reason for this is that upon login, any high priority unread notifications should always be at the top of the list, and since reminder notifications are going to only be created once a day, it's very likely for a lower priority notification to be created afterwards.

I've also left the priority direction as is for now - I put it as ascending because I was imagining lower numbers to mean higher priority (so P1 > P2), but I'm open to changing this if it's not intuitive, since the higher priority reminder notifications will only be added later!

alexanderleegs: As discussed offline, both conditions actually need to be fulfilled - 

- For unread notifications, priority should be the determining factor
- For read notifications, they can be sorted by creation time.

This behaviour has been fixed in https://github.com/isomerpages/isomercms-backend/pull/508/commits/7b4e42364dbd6d3c523824d96d01ff574040d280!

dcshzj (approved):
LGTM! Just need to add comments to explain the sorting order + lower priority number == higher priority notification

PR #512, opened 2022-10-05
ref(services): migrate SitesService to TypeScript

dcshzj:
## Problem

There are 2 SitesService, one under `utilServices` which is in JavaScript and one under `identity` which is in TypeScript. This can be confusing for developers, and makes it difficult to implement new features for this service. Also, TypeScript is better.

This hinders the development of the site dashboard, as the site info route will need to be implemented in JavaScript without this change, which would make future migrations even more costly.

## Solution

**Breaking Changes**

**Improvements**:

- The SitesService under `utilServices` have been migrated to the SitesService under `identity`. Additionally:
    - The `getEmailUserSites()` function has been renamed to `getSitesForEmailUser()` to be clearer on its purpose, and the implementation has been simplified.
    - The `checkHasAccess()` function has been renamed to `checkHasAccessForGitHubUser()` to differentiate the access checking between GitHub login users and email login users.
- New types, `GitHubRepositoryData` and `RepositoryData`, have been introduced.
- Unit tests for both SiteServices have now been merged with no change in test scopes.

## Tests


src/constants/constants.ts
```diff
@@ -18,3 +18,27 @@ export enum CollaboratorRoles {
 export const E2E_ISOMER_ID = "-1"
 export const E2E_TEST_EMAIL = "test@e2e"
 export const E2E_TEST_CONTACT = "12345678"
+
+export const GH_MAX_REPO_COUNT = 100
+export const ISOMERPAGES_REPO_PAGE_COUNT =
+  (process.env.ISOMERPAGES_REPO_PAGE_COUNT &&
+    parseInt(process.env.ISOMERPAGES_REPO_PAGE_COUNT)) ||
```
seaerchin, line 25: think we missed the radix argument here

src/routes/v2/authenticated/sites.ts
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  getLastUpdated: RequestHandler<
+    never,
+    unknown,
+    never,
+    { siteName: string },
+    { userSessionData: UserWithSiteSessionData }
+  > = async (req, res) => {
+    const { siteName } = req.params
```
seaerchin, line 44: wrong declaration i think - should be first argument. this is confusing because there's 2 kinds of params, path and query.

path parameters are stuff like `/:siteName/something`, where you access by `const {siteName} = req.params` whereas query parameters are `?siteName="abc"`. 

the `RequestHandler` type has `P` (the path param) as its first argument, and this is evidenced here by how you use `req.params` (whereas the type you specified is for `ReqQuery`, which are request parameters)

src/routes/v2/authenticated/sites.ts
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  getStagingUrl: RequestHandler<
+    never,
+    unknown,
+    never,
+    { siteName: string },
+    { userSessionData: UserWithSiteSessionData }
+  > = async (req, res) => {
+    const { siteName } = req.params
```
seaerchin, line 56: see above

src/routes/v2/authenticated/__tests__/Sites.spec.js
```diff
@@ -71,7 +71,7 @@ describe("Sites Router", () => {
 
       expect(resp.body).toStrictEqual({ lastUpdated })
       expect(mockSitesService.getLastUpdated).toHaveBeenCalledWith(
-        mockUserWithSiteSessionData
+        mockUserWithSiteSessionData.siteName
```
seaerchin, line 74: was this passing last time ._.

src/services/identity/SitesService.ts
```diff
@@ -1,46 +1,207 @@
...
   }
 
+  async getSitesForEmailUser(userId: string) {
+    const user = await this.usersService.findSitesByUserId(userId)
+
+    if (!user) {
+      return []
+    }
+
+    return user.site_members.map((site) => site.repo?.name)
```
seaerchin, line 78: this returns undefined if `repo` doesn't exist right? does the method signature make htis clear

src/services/identity/SitesService.ts
```diff
@@ -1,46 +1,207 @@
...
+  }
+
+  async getSites(sessionData: UserSessionData): Promise<RepositoryData[]> {
+    const isEmailUser = sessionData.isEmailUser()
+    const { isomerUserId: userId } = sessionData
+    const isAdminUser = !!(await this.isomerAdminsService.getByUserId(userId))
+
+    const { accessToken } = sessionData
```
seaerchin, line 86: style nit: could we inline this above

src/services/identity/SitesService.ts
```diff
@@ -1,46 +1,207 @@
...
+      (_, idx) => ({
+        per_page: GH_MAX_REPO_COUNT,
+        sort: "full_name",
+        page: idx + 1,
+      })
+    )
+
+    const allSites: RepositoryData[] = await Bluebird.map(
+      paramsArr,
+      async (params) => {
```
seaerchin, line 100: turns out this is typed as `any`... from bluebird [docs](https://github.com/petkaantonov/bluebird), 

> Please use native promises instead if at all possible.

so if it's not a big chore, could we do this using native `map`

src/services/identity/UsersService.ts
```diff
@@ -79,7 +79,11 @@ class UsersService {
     return !!siteMember
   }
 
-  async findSitesByUserId(isomerId: string) {
+  async findSitesByUserId(
+    isomerId: string
+  ): Promise<User & { site_members: Site[] }> {
```
seaerchin, line 84: can i check if you've verified this? the collaborators PR types this as `SiteMembers` (and i found that to be true at runtime) 

separately, since i think we'll be using this alot (eg: collaborators PR), do you think this merits an extra type? 

nit: once retrieved, we should transform this into `siteMembers`

src/services/identity/__tests__/SitesService.spec.ts
```diff
@@ -1,78 +1,355 @@
...
 
 describe("SitesService", () => {
+  beforeAll(() => {
+    // NOTE: Because SitesService uses an axios instance,
+    // we need to mock the axios instance using es5 named exports
+    // to ensure that the calls for .get() on the instance
+    // will actually return a value and not fail.
+    jest.mock("../../../services/api/AxiosInstance.ts", () => ({
+      __esModule: true, // this property makes it work
+      genericGitHubAxiosInstance,
```
seaerchin, line 80: good docs! could i check if we should prefer using `__mocks__/axios.ts`?  can refer to `Sites.spec.ts` for an eg (sorry if troublesome ._.)

seaerchin (commented):
thanks for doing this! not reviewing the methods (as i assume they are ported over as-is) but i do have some comments

src/services/identity/SitesService.ts
seaerchin: ah yea it does, ignore this

src/constants/constants.ts
dcshzj: Whoops TIL about the radix argument. Done in 3658486.

src/routes/v2/authenticated/sites.ts
dcshzj: Hmm yeah I think I got confused by the params and query parts yesterday, adjusted in 3658486.

src/routes/v2/authenticated/__tests__/Sites.spec.js
dcshzj: Reverted this change in 3658486, I misunderstood how the accessToken is obtained for email users.

src/services/identity/UsersService.ts
dcshzj: Hmm I don't think it should be the SiteMembers table, as the return value should have the `repo` property, which we use to obtain the name of the repository (refer to `getSitesForEmailUser()` under SitesService.ts).

I updated this in f7045aa as it should match the type for the `sites` property in User.ts.

As for the collaborators PR I think it's slightly different. This one is from the User.ts perspective, whereas the other one is from the Sites.ts property, so the types are inherently different.

src/services/identity/__tests__/SitesService.spec.ts
dcshzj: Migrated in 8c2275e.

src/__mocks__/axios.ts
```diff
@@ -1,3 +1,24 @@
...
+export const prepareAxiosMock = () => {
+  // NOTE: We need to mock the axios instance using es5 named exports
+  // to ensure that the calls for .get() on the instance will actually
+  // return a value and not fail.
+  jest.mock("../services/api/AxiosInstance.ts", () => ({
+    __esModule: true, // this property makes it work
+    mockAxiosInstance,
+  }))
+}
+
```
seaerchin, line 23: double checking - why do we need this? i think just `import mockAxios from "jest-mock-axios"` should be enough .-.

src/services/identity/UsersService.ts
seaerchin: good point on `User vs Site`. also, i realised (by dint of experimentation and subsequent investigation) that if we have the correct annotations on the respective model, the base model (`User` here) should contain the foreign key **object** (ie, `site_members`).

Hence, this casting might be indicative of a erroneous assocation or lack thereof... **but** if you verified that this is working, we can ignore this (for now) and fix it later (perhaps on friday?)

seaerchin (commented):
minor comments; feel free to ignore the comment about association or just work on it later

src/services/identity/UsersService.ts
dcshzj: I am guessing the error came because of how we used the `as` operator to declare it as `site_members`, but the actual key is `sites` here:

https://github.com/isomerpages/isomercms-backend/blob/398d70f8dc46d97edd672f4d0df04400b580e8f4/src/database/models/User.ts#L64-L70

It's a hypothesis that the key should be renamed to `site_members` so we would not have this kind of props issues, similar to how it is done for Site.ts. I think I can look into this further beyond this PR, to see if simply renaming the prop is possible.

seaerchin (approved):

src/services/identity/UsersService.ts
seaerchin: for future reference, this is correct and the link to the docs is [here](https://github.com/sequelize/sequelize-typescript#type-safe-through-table-instance-access)

PR #513, opened 2022-10-07
feat: introduce a new site info API endpoint

dcshzj:
## Problem

There are currently no existing endpoints suitable for the site dashboard to use to obtain some basic information about the site.

## Solution

**Breaking Changes**

**Features**:

- A new site info API endpoint has been introduced, which presents:
    - The email address of the user that last committed to the staging and master branches
    - The date that the latest commit to the staging and master branches were made
    - The staging URL of the site
    - The live production site URL
- A new `getUrlsOfSite()` function has been created based on a refactoring of the existing `getStagingUrl()` function.
- A new `getCommitAuthorEmail()` function has been created to retrieve the email address of the user that made a particular commit.
- The existing `getStagingUrl()` function has been refactored to utilise the new method of obtaining the staging URL.

## Tests


src/services/identity/SitesService.ts
```diff
@@ -78,6 +84,107 @@ class SitesService {
...
+        staging: "",
+        prod: "",
+      }
+    }
+
+    const deployment = site?.deployment
+    const output: SiteUrls = {
+      staging: deployment?.stagingUrl ? deployment.stagingUrl : "",
+      prod: deployment?.productionUrl ? deployment.productionUrl : "",
+    }
```
seaerchin, line 139: note that `Deployment` has `allowNull: false` set on both of these so we probably shouldn't do this; instead, what we could do is to define a `doesDeploymentExists` variable that determines if we should fetch using `configYmlService` (or just `!deployment`) 

this does make it slightly less future proof (eg: we want to `allowNull`) though... lmk what you think!

src/services/identity/SitesService.ts
```diff
@@ -78,6 +84,107 @@ class SitesService {
...
+  async getUrlsOfSite(
+    sessionData: UserWithSiteSessionData,
+    urlTypes: SiteUrlTypes[]
+  ): Promise<SiteUrls> {
+    // Tries to get the site urls in the following order:
+    // 1. From the deployments database table
+    // 2. From the config.yml file
+    // 3. From the GitHub repository description
+    // Otherwise, returns an empty string for the url
+    const { siteName } = sessionData
```
seaerchin, line 116: i think the code below is pretty janky to read and we should instead write it as a flow of something like

1. obtain data 
2. use util function to assign properties if it exists in data 
3. check if both properties are valid otherwise do 1-3 again

wdyt about this?

src/services/identity/SitesService.ts
```diff
@@ -78,6 +84,107 @@ class SitesService {
...
+        const siteUrl = repoDescTokens.find(
+          (token) => token.includes("http") && token.includes(urlType)
+        )
+        if (siteUrl) {
+          output[urlType] = siteUrl
+        }
+      }
+    }
+
+    return output
```
seaerchin, line 185: note that `output` here is not guaranteed to be well-formed. legacy sites could potentially have misbehaving behaviour and hence, still have undefined `staging | prod`. 

at the end, we could validate properties and return either 1. well formed output, 2. a single output (either `staging | prod` only or 3. error

double check wrt github repo desc @alexanderleegs

src/services/identity/SitesService.ts
```diff
@@ -154,29 +261,13 @@ class SitesService {
...
-    if (description) {
-      // Retrieve the url from the description - repo descriptions have varying formats, so we look for the first link
-      const descTokens = description.replace("/;/g", " ").split(" ")
-      // Staging urls also contain staging in their url
-      const stagingUrl = descTokens.find(
-        (token) => token.includes("http") && token.includes("staging")
+    const { staging } = await this.getUrlsOfSite(sessionData, ["staging"])
+    if (!staging) {
+      throw new NotFoundError(
+        `${sessionData.siteName} does not have a staging url`
```
seaerchin, line 267: shouldn't throw, prefer returning an error and let the router decide what to do

src/services/identity/SitesService.ts
```diff
@@ -194,6 +285,41 @@ class SitesService {
...
+  async getSiteInfo(sessionData: UserWithSiteSessionData): Promise<SiteInfo> {
+    const {
+      staging: stagingUrl,
+      prod: prodUrl,
+    } = await this.getUrlsOfSite(sessionData, ["staging", "prod"])
+
+    const stagingCommit = ((await this.gitHubService.getLatestCommitOfBranch(
+      sessionData,
+      "staging"
+    )) as unknown) as GitHubCommitData
```
seaerchin, line 298: this is a dangerous cast. we should validate the properties and use an assertion.

src/services/identity/SitesService.ts
```diff
@@ -194,6 +285,41 @@ class SitesService {
...
+    const prodCommit = ((await this.gitHubService.getLatestCommitOfBranch(
+      sessionData,
+      "master"
+    )) as unknown) as GitHubCommitData
```
seaerchin, line 302: same

src/routes/v2/authenticated/sites.ts
```diff
@@ -84,6 +84,25 @@ export class SitesRouter {
...
+    never,
+    never,
+    { userSessionData: UserWithSiteSessionData }
+  > = async (req, res) => {
+    const { userSessionData } = res.locals
+    const { siteName } = req.params
+    const userWithSiteSessionData = this.addSiteNameToSessionData(
+      userSessionData,
+      siteName
+    )
```
alexanderleegs, line 99: Actually we should already have siteName in the sessionData here! Via the extra attachSiteHandler middleware for this route

src/services/identity/SitesService.ts
```diff
@@ -78,6 +84,107 @@ class SitesService {
...
+    let configYmlData: ConfigYmlData = {}
+    let repoDescTokens: string[] = []
+
+    const site = await this.siteRepository.findOne({
+      where: { name: siteName },
+      include: {
+        model: Deployment,
+        as: "deployment",
+      },
+    })
```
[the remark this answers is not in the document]
alexanderleegs: Actually, given that this is the SitesService itself, does this suggestion mean that we would need to query SitesService for the existence of the site, before we can call this endpoint? Also, this might cause duplicated code if we want to reuse this method (e.g. if multiple methods call `getUrlsOfSite`, both would need to first query some generic `SitesService.get` method to retrieve a `Site`, before this method could be called) - I'm not sure if this behaviour is desirable

src/fixtures/identity.ts
```diff
@@ -138,3 +140,23 @@ export const mockSiteOrmResponseWithNoCollaborators = {
...
+
+export const mockCommitMessage2 = "Update file: Test.md"
+export const mockCommitFileName2 = "Test.md"
+export const mockGitHubEmailAddress2 = "test2@example.com"
+export const mockGitHubDate2 = "2022-09-28T06:25:14Z"
+export const mockCommitMessageObject2 = {
+  message: mockCommitMessage2,
+  fileName: mockCommitFileName2,
+  userId: mockIsomerUserId,
+}
```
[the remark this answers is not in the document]
dcshzj: Updated in b126c53.

src/routes/v2/authenticated/sites.ts
```diff
@@ -84,6 +84,25 @@ export class SitesRouter {
...
+    const siteInfo = await this.sitesService.getSiteInfo(
+      userWithSiteSessionData
+    )
+    return res.status(200).json(siteInfo)
+  }
```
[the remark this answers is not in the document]
dcshzj: Good point! Added in e9b0112.

src/services/db/GitHubService.js
```diff
@@ -284,6 +284,27 @@ class GitHubService {
     return { treeSha, currentCommitSha }
   }
 
+  async getLatestCommitOfBranch(sessionData, branch) {
+    const { accessToken } = sessionData
+    const { siteName } = sessionData
```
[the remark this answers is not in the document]
dcshzj: Whoops, combined in d335ebf.

src/services/identity/SitesService.ts
```diff
@@ -78,6 +84,107 @@ class SitesService {
     return user.site_members.map((site) => site.repo?.name)
   }
 
+  async getCommitAuthorEmail(commit: GitHubCommitData) {
+    const { message } = commit
+
+    // Commit message created as part of phase 2 identity
+    if (message.startsWith("{") && message.endsWith("}")) {
+      const { userId }: { userId: string } = JSON.parse(message)
```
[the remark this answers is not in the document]
dcshzj: TIL about the JSONParseError, added a `try/catch` in 5ef57a4.

src/services/identity/SitesService.ts
```diff
@@ -78,6 +84,107 @@ class SitesService {
...
+    })
+
+    if (!site) {
+      return {
+        staging: "",
+        prod: "",
+      }
+    }
+
+    const deployment = site?.deployment
```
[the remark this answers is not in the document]
dcshzj: Adjusted in 66defaa.

src/services/identity/SitesService.ts
dcshzj: Changed in ed61f55.

src/services/identity/SitesService.ts
dcshzj: Added an assertion function in 9ca0838.

src/services/identity/SitesService.ts
dcshzj: 9ca0838

src/routes/v2/authenticated/sites.ts
dcshzj: Removed in a12ad20!

src/services/db/GitHubService.js
```diff
@@ -284,6 +284,27 @@ class GitHubService {
     return { treeSha, currentCommitSha }
   }
 
+  async getLatestCommitOfBranch(sessionData, branch) {
```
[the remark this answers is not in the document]
dcshzj: Discussed offline, would prefer to centralise everything so that when refactoring to TypeScript, it is easier to find all the functions that we need to refactor in one place.

src/services/identity/SitesService.ts
```diff
@@ -78,6 +84,107 @@ class SitesService {
...
+    const { message } = commit
+
+    // Commit message created as part of phase 2 identity
+    if (message.startsWith("{") && message.endsWith("}")) {
+      const { userId }: { userId: string } = JSON.parse(message)
+      const user = await this.usersService.findById(userId)
+
+      if (user) {
+        return user.email
+      }
```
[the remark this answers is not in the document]
dcshzj: This branch will not be entered and will be handled later, as the GitHubCommitData is guaranteed to have the email address, which will be returned.

src/services/identity/SitesService.ts
```diff
@@ -78,6 +84,107 @@ class SitesService {
...
+        as: "deployment",
+      },
+    })
+
+    if (!site) {
+      return {
+        staging: "",
+        prod: "",
+      }
+    }
```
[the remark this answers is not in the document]
dcshzj: Adjusted in 4432732.

src/services/identity/SitesService.ts
```diff
@@ -78,6 +84,107 @@ class SitesService {
...
+
+    const deployment = site?.deployment
+    const output: SiteUrls = {
+      staging: deployment?.stagingUrl ? deployment.stagingUrl : "",
+      prod: deployment?.productionUrl ? deployment.productionUrl : "",
+    }
+
+    for (const urlType of urlTypes) {
+      // We only call ConfigYmlService only when necessary
+      if (!output[urlType] && Object.keys(configYmlData).length === 0) {
```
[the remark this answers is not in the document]
dcshzj: Adjusted in 4432732.

src/services/identity/SitesService.ts
```diff
@@ -78,6 +84,107 @@ class SitesService {
...
+      if (!output[urlType] && urlType in configYmlData) {
+        const configSiteUrl = configYmlData[urlType]
+
+        if (configSiteUrl) {
+          output[urlType] = configSiteUrl
+        }
+      }
+
+      // We only call GitHubService only when necessary
+      if (!output[urlType] && repoDescTokens.length == 0) {
```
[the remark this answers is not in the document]
dcshzj: Adjusted in 4432732.

src/services/identity/SitesService.ts
```diff
@@ -78,6 +84,107 @@ class SitesService {
...
+    // Legacy style of commits, or if the user is not found
+    const {
+      author: { email: authorEmail },
+    } = commit
+    return authorEmail
+  }
+
+  async getUrlsOfSite(
+    sessionData: UserWithSiteSessionData,
+    urlTypes: SiteUrlTypes[]
```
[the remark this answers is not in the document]
dcshzj: Adjusted in 4432732.

src/services/identity/SitesService.ts
dcshzj: If deployment (or site) does not exist, then we will just continue which is intended.

src/services/identity/SitesService.ts
dcshzj: Adjusted in 4432732.

src/services/identity/SitesService.ts
dcshzj: Adjusted in 4432732. If all 3 sources of data does not contain the info we need, then we will return an UnprocessableError.

src/services/identity/SitesService.ts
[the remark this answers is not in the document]
dcshzj: Discussed offline. This is a refactor that we will do separately soon<sup>TM</sup>

dcshzj:
Added a fix where the endpoint for getting the latest commit is incorrect.

Also added a stub for getting the email address of the user that last published the site, which should be obtained from the review request table.

src/services/identity/SitesService.ts
```diff
@@ -66,6 +67,89 @@ class SitesService {
...
+    siteUrls = {
+      staging:
+        configYmlData.staging && !siteUrls.staging
+          ? configYmlData.staging
+          : siteUrls.staging,
+      prod:
+        configYmlData.prod && !siteUrls.prod
+          ? configYmlData.prod
+          : siteUrls.prod,
+    }
```
seaerchin, line 110: could we have a helper method for this? (cos also done below but minor)

also, we should create a new object rather than passing `siteUrls` - this modifies `siteUrl` in place and js doesn't pass by value so this is potentially dangerous

src/services/identity/SitesService.ts
```diff
@@ -66,6 +67,89 @@ class SitesService {
...
+
+    const stagingUrlFromDesc = repoDescTokens.find(
+      (token) => token.includes("http") && token.includes("staging")
+    )
+    const prodUrlFromDesc = repoDescTokens.find(
+      (token) => token.includes("http") && token.includes("prod")
+    )
+
+    // Only replace the urls if they are not already present
+    siteUrls = {
```
seaerchin, line 142: see above

src/services/identity/SitesService.ts
```diff
@@ -89,11 +173,15 @@ class SitesService {
...
-      const { userId }: { userId: string } = JSON.parse(message)
-      const user = await this.usersService.findById(userId)
+      try {
+        const { userId }: { userId: string } = JSON.parse(message)
+        const user = await this.usersService.findById(userId)
 
-      if (user) {
-        return user.email
+        if (user) {
+          return user.email
```
seaerchin, line 181: could we only return if `user.email` is a string? this is `string | null | undefined`, and we can do better by returning the commit author's email

src/services/identity/SitesService.ts
```diff
@@ -104,18 +192,32 @@ class SitesService {
...
+    } = commit
+    return authorEmail
+  }
+
   async getUrlsOfSite(
-    sessionData: UserWithSiteSessionData,
-    urlTypes: SiteUrlTypes[]
-  ): Promise<SiteUrls> {
+    sessionData: UserWithSiteSessionData
+  ): Promise<SiteUrls | NotFoundError> {
```
seaerchin, line 214: specialise the error here to give more context

src/services/identity/SitesService.ts
```diff
@@ -125,64 +227,24 @@ class SitesService {
...
-    const deployment = site?.deployment
-    const output: SiteUrls = {
-      staging: deployment?.stagingUrl ? deployment.stagingUrl : "",
-      prod: deployment?.productionUrl ? deployment.productionUrl : "",
+    // Note: site may be null if the site does not exist
+    let siteUrls: SiteUrls = {
+      staging: site?.deployment?.stagingUrl ? site.deployment.stagingUrl : "",
+      prod: site?.deployment?.productionUrl
+        ? site.deployment.productionUrl
+        : "",
```
seaerchin, line 235: we can use `??` or `||` - both easier to read

src/services/identity/SitesService.ts
```diff
@@ -125,64 +227,24 @@ class SitesService {
...
-        }: { description: string } = await this.gitHubService.getRepoInfo(
-          sessionData
-        )
-
-        // Retrieve the url from the description
-        // repo descriptions have varying formats, so we look for the first link
-        repoDescTokens = description.replace("/;/g", " ").split(" ")
-      }
+    siteUrls = await this.insertUrlsFromConfigYml(siteUrls, sessionData)
+    siteUrls = await this.insertUrlsFromGitHubDescription(siteUrls, sessionData)
```
seaerchin, line 239: should aim to merge - lodash has a method for this iirc

src/services/identity/SitesService.ts
```diff
@@ -260,13 +322,18 @@ class SitesService {
...
+  async getStagingUrl(
+    sessionData: UserWithSiteSessionData
+  ): Promise<string | NotFoundError> {
+    const siteUrls = await this.getUrlsOfSite(sessionData)
+    if (siteUrls instanceof NotFoundError) {
+      return new NotFoundError(
         `${sessionData.siteName} does not have a staging url`
       )
     }
+
```
seaerchin, line 334: actually we could do a hierarchy of errors like so: `NotFoundError` -> `Site...` -> `StagingUrl...`

src/routes/v2/authenticated/sites.ts
```diff
@@ -70,17 +60,17 @@ export class SitesRouter {
...
-    )
+    const { userWithSiteSessionData } = res.locals
     const stagingUrl = await this.sitesService.getStagingUrl(
       userWithSiteSessionData
     )
+
+    // Check for error and throw
+    if (stagingUrl instanceof BaseIsomerError) {
+      throw stagingUrl
+    }
```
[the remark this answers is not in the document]
dcshzj: Done in 99d9140.

src/routes/v2/authenticated/sites.ts
```diff
@@ -70,17 +60,17 @@ export class SitesRouter {
...
+    { userWithSiteSessionData: UserWithSiteSessionData }
   > = async (req, res) => {
-    const { userSessionData } = res.locals
-    const { siteName } = req.params
-    const userWithSiteSessionData = this.addSiteNameToSessionData(
-      userSessionData,
-      siteName
-    )
+    const { userWithSiteSessionData } = res.locals
     const stagingUrl = await this.sitesService.getStagingUrl(
```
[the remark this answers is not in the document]
dcshzj: Done in 99d9140.

src/routes/v2/authenticated/sites.ts
```diff
@@ -89,17 +79,18 @@ export class SitesRouter {
...
+    const { userWithSiteSessionData } = res.locals
+
     const siteInfo = await this.sitesService.getSiteInfo(
       userWithSiteSessionData
     )
+
+    // Check for error and throw
+    if (siteInfo instanceof BaseIsomerError) {
+      throw siteInfo
+    }
```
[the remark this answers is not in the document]
dcshzj: 99d9140

src/services/db/__tests__/GitHubService.spec.js
```diff
@@ -557,6 +558,51 @@ describe("Github Service", () => {
...
+        sessionData,
+        "staging"
+      )
+      expect(actual).toEqual(expected)
+      expect(mockAxiosInstance.get).toHaveBeenCalledWith(endpoint, {
+        headers,
+      })
+    })
+
+    it("Getting an invalid branch should throw UnprocessableError", async () => {
```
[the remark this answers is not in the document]
dcshzj: Good point! Added in cdf94ea.

src/services/identity/SitesService.ts
```diff
@@ -66,6 +67,89 @@ class SitesService {
     this.tokenStore = tokenStore
   }
 
+  assertGitHubCommitData(commit: any): commit is GitHubCommitData {
```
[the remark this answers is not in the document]
dcshzj: Adjusted in 1eb32dd.

src/services/identity/SitesService.ts
```diff
@@ -66,6 +67,89 @@ class SitesService {
...
+    if (
+      commit &&
+      (commit as GitHubCommitData).author !== undefined &&
+      (commit as GitHubCommitData).author.name !== undefined &&
+      (commit as GitHubCommitData).author.date !== undefined &&
+      (commit as GitHubCommitData).author.email !== undefined &&
+      (commit as GitHubCommitData).message !== undefined
```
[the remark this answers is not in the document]
dcshzj: Whoops, adjusted in 1eb32dd.

src/services/identity/SitesService.ts
dcshzj: Hmm I think it might be clearer here instead of having a separate helper method?

Also adjusted the name/assignment in 3692b08.

src/services/identity/SitesService.ts
dcshzj: Done in 3692b08.

src/services/identity/SitesService.ts
dcshzj: Good catch! Fixed in fdf07a5.

src/services/identity/SitesService.ts
dcshzj: Improved in 42e73df.

src/services/identity/SitesService.ts
dcshzj: Improved in 42e73df.

seaerchin (approved):
lgtm

PR #518, opened 2022-10-11
feat(rr): add database models

seaerchin:
## Problem
This PR adds the database models which our services use. There is a dual annotation of `ForeignKey` together with an association `Belongs/Has`. This is required for sequelize to know what to use to reference the joined table and return the joined row. 

## Solution
1. Add relevant db models
2. Add `reviewStatus` column onto the `reviewRequest` rather than fetching from github; this is deliberate because github ties the review status (approved/merged etc) to the individual whereas for our data model, this is tied to the review request itself. this, for example, means that when a single person approves, the approval is shared throughout and the review request can be merged


src/database/models/User.ts
```diff
@@ -73,4 +76,7 @@ export class User extends Model {
     as: "sites_created",
   })
   sitesCreated?: Site[]
+
+  @HasMany(() => ReviewRequest)
+  reviewRequests?: ReviewRequest[]
```
dcshzj, line 81: Can I confirm that this is referring to the review requests that this User has **created**, and not the list of review requests that this User has been selected to be a reviewer?

seaerchin:
btw to reviewers - test was fixed upstream prior to rebase ._.

src/database/models/User.ts
seaerchin: actually i think this is ambiguous (didn't try) and it's a good point raised. wdyt if i changed it to `reviewsRequested` and `reviewsCreated` for clarity?

dcshzj: Yup I think that would make things clearer, if it is possible to codify such a relationship in this model.

seaerchin: this might be slightly annoying to do btw... @__@

dcshzj: I think it should be sufficient to get the `reviewsCreated`.

src/database/migrations/20221007124138-create-review-status.js
```diff
@@ -0,0 +1,30 @@
...
+      await queryInterface.addColumn(
+        "review_requests", // name of Source model
+        "review_status", // name of column we're adding
+        {
+          type: Sequelize.ENUM,
+          values: ["OPEN", "MERGED", "CLOSED", "APPROVED"],
+          allowNull: false,
+          defaultValue: "OPEN",
+          transaction: t,
+        }
```
dcshzj, line 13: We should also drop this in `src/tests/teardown.ts` as it is causing issues when running `npm run tests` repeatedly.

dcshzj (changes requested):

seaerchin:
- do database schema changes later - **but before actual rollout**
- update `teardown.ts` and merge

src/database/migrations/20221007124138-create-review-status.js
seaerchin: fixed

PR #519, opened 2022-10-11
feat(rr): services + routes

seaerchin:
## Problem
This PR adds the logic for services and routes for `ReviewRequest`. This consists of CRUD endpoints together with a few utility functions.

## Solution
1. add routes - the routes perform data validation and parsing, to ensure that the preconditions to call the service is met. Such things include ensuring that the user has correct permissions/ensuring that the site exists. 
2. add service - performs business logic on the `ReviewRequest`. Note that apart from the retrieval methods, most other methods existing on the service ask for a well formed `ReviewRequest` so that we can remove the parsing logic away from  the service call.
    - at present, the service is a mix of impure fetching functions (required to inject data from github) and pure business logic. this should be done better but is left as-is due to time constraints :sadge:)


seaerchin:
snyk failed to complete the test on `package.json` - not sure if important ._.

kishore03109 (changes requested):

src/types/github.ts
```diff
@@ -0,0 +1,27 @@
...
+// https://docs.github.com/en/rest/commits/commits#compare-two-commits
+
+export type FileChangeStatus =
+  | "added"
+  | "removed"
+  | "modified"
+  | "renamed"
+  | "copied"
+  | "changed"
+  | "unchanged"
```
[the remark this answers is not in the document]
seaerchin: i actually got it from the response schema (attaching ss so you don't have to trawl through like me) - 
Screenshot 2022-10-13 at 10 44 44 AM

src/types/github.ts
```diff
@@ -0,0 +1,27 @@
...
+  additions: number
+  deletions: number
+  changes: number
+  // eslint-disable-next-line camelcase
+  blob_url: string
+  // eslint-disable-next-line camelcase
+  raw_url: string
+  // eslint-disable-next-line camelcase
+  contents_url: string
+  patch: string
```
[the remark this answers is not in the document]
seaerchin: yeap, i think i'm gonna remove it - added it in initially to see if it wsa gonna be useful for the diff view but i think i'd rather just get the raw and compare

src/services/review/ReviewRequestService.ts
```diff
@@ -7,8 +7,8 @@ import { ReviewMeta } from "@database/models/ReviewMeta"
 import { ReviewRequest } from "@database/models/ReviewRequest"
 import { Site } from "@root/database/models/Site"
 import { User } from "@root/database/models/User"
-import { RawFileChangeInfo } from "@root/types/github"
-import { FileChangeInfo } from "@root/types/review"
+import { EditedItemDto, FileType } from "@root/types/dto /review"
```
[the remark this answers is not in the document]
seaerchin: err i'd prefer not to - the dto is an api contract between teh FE/BE. The fact that this is functionally the same as `EditedItemProps` shouldn't matter because the props can be free to change but when it hits `apiService` should still be `EditedItemDto` to ensure that stuff is alw compatible b/w FE/BE

src/services/review/ReviewRequestService.ts
```diff
@@ -51,34 +51,56 @@ export default class ReviewRequestService {
...
+
+    const mappings = this.computeShaMappings(commits)
+
+    return files.map(({ filename, sha }) => {
+      const fullPath = filename.split("/")
+      return {
+        type: this.computeFileType(filename),
+        // NOTE: The string is guaranteed to be non-empty
+        // and hence this should exist.
+        name: fullPath.pop() || "",
```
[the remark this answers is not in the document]
seaerchin: `pop` returns `T | undefined` but we know (as authors kek) that it can't be empty. this is just to satisfy the TS compiler

src/routes/v2/authenticated/review.ts
```diff
@@ -0,0 +1,106 @@
...
+    if (!hasAccess) {
+      return res.status(400).send()
+    }
+
+    const userWithSiteSessionData = new UserWithSiteSessionData({
+      ...userSessionData,
+      siteName,
+    })
+
+    // Check if reviewers exist
```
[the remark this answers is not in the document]
seaerchin: think this is only on that 1 commit

src/types/github.ts
```diff
@@ -64,10 +64,14 @@ export interface IsomerCommitMessage {
...
-    userId: parsed.userId,
+  try {
+    const parsed = JSON.parse(message)
+    return {
+      message: parsed.message,
+      fileName: parsed.filename,
+      userId: parsed.userId,
+    }
+  } catch {
+    return {}
```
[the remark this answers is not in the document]
seaerchin: github doesn't return in json format. we introduced a message format _that happens to be json encoded_ - this isn't true for all pages (eg: if we update `homepage`) so we catch here and return an empty object

src/routes/v2/authenticated/collaborators.ts
```diff
@@ -71,18 +72,24 @@ export class CollaboratorsRouter {
...
+    { siteName: string },
+    { collaborators: UserDto[] },
     never,
     unknown,
-    never,
-    { siteName: string },
     { userWithSiteSessionData: UserWithSiteSessionData }
   > = async (req, res) => {
     const { siteName } = req.params
+    const x = req.query
```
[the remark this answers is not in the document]
seaerchin: think this is just on that 1 commit

src/services/review/ReviewRequestService.ts
```diff
@@ -281,22 +280,53 @@ export default class ReviewRequestService {
...
+        model: ReviewMeta,
+        as: "reviewMeta",
+        where: {
+          pullRequestNumber,
+        },
+      },
+    })
+
+    if (!reviewRequest) {
+      return new RequestNotFoundError()
```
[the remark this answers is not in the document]
seaerchin: returning the error is more informative than throwing, and it lets the router decide what to do (eg: what error message to show etc etc)

src/services/db/review.ts
```diff
@@ -62,6 +64,22 @@ export const approvePullRequest = (
...
+      headers: {
+        // NOTE: This is currently done because
+        // we have a lock on the master branch
+        // and github requires an approval from
+        // *another* account that is not the creator
+        // of the pull request.
+        // This is a temporary workaround until we
+        // write a migration script to remove the lock on master.
+        // TODO!: Remove this
+        Authorization: `token ${E2E_TEST_GH_TOKEN}`,
```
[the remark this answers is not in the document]
seaerchin: no strong opinion (@alexanderleegs thoughts?) - i went with this route but alex suggested doing migration script so that we can eliminate the need for extra tokens. 

i think our assumption is that the users aren't tech savvy enough to use the repo themselves

src/routes/v2/authenticated/review.ts
```diff
@@ -67,33 +70,80 @@ export class ReviewsRouter {
...
     unknown,
     { userSessionData: UserSessionData }
   > = async (req, res) => {
-    // Step 1: Check that user exists.
+    // Step 1: Check that the site exists
+    const { siteName } = req.params
+    const site = await this.sitesService.getBySiteName(siteName)
+
+    if (!site) {
+      return res.status(404).send({
```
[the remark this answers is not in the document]
seaerchin: actually this is a common scenario - if a user miskeys their repo name, this will happen as we don't validate `siteName` until this point!

that aside, i think this is a good suggestion and i'll probably do this for the remaining functions in this file also

src/services/db/review.ts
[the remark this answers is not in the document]
alexanderleegs: I think we could set it up such that admin accounts can bypass the restrictions? We probably need to look into it later on, but for now i think this is fine

src/services/identity/UsersService.ts
```diff
@@ -59,8 +59,11 @@ class UsersService {
     return this.repository.findOne({ where: { githubId } })
   }
 
-  async hasAccessToSite(userId: string, siteName: string): Promise<boolean> {
-    const siteMember = await this.repository.findOne({
+  async hasAccessToSite(
+    userId: string,
+    siteName: string
+  ): Promise<User | null> {
```
dcshzj, line 65: Hmm I find the return type a little weird. I think we should keep to returning a boolean so that it fits the method name and the purpose of the method (which is to determine if the user has access to the site or not.

src/services/identity/UsersService.ts
```diff
@@ -79,8 +82,28 @@ class UsersService {
         },
       ],
     })
+  }
 
-    return !!siteMember
+  async isSiteAdmin(userId: string, siteName: string) {
```
dcshzj, line 87: Same for this method, should return a boolean instead of a User object.

src/services/review/ReviewRequestService.ts
```diff
@@ -0,0 +1,365 @@
...
+    // Refer here for details; https://docs.github.com/en/rest/commits/commits#compare-two-commits
+    // Note that we need a triple dot (...) between base and head refs
+    const { files, commits } = await this.apiService.getCommitDiff(siteName)
+
+    const mappings = await this.computeShaMappings(commits)
+
+    return files.map(({ filename, contents_url }) => {
+      const fullPath = filename.split("/")
+      const items = contents_url.split("?ref=")
+      // NOTE: We have to compute sha this way rather than
```
dcshzj, line 72: Minor nit: I think this comment is incomplete? Might need more explanation on what this sha is referring to (I got confused here when reviewing hahahaha).

src/services/review/ReviewRequestService.ts
```diff
@@ -0,0 +1,365 @@
...
+    await reviewRequest.save()
+  }
+
+  mergeReviewRequest = async (
+    reviewRequest: ReviewRequest
+  ): Promise<ReviewRequest | RequestNotFoundError> => {
+    const siteName = reviewRequest.site.name
+    const { pullRequestNumber } = reviewRequest.reviewMeta
+
+    await this.apiService.approvePullRequest(siteName, pullRequestNumber)
```
dcshzj, line 359: Question: Why do we only approve here when merging and not when during the approve review request step?

src/services/review/ReviewRequestService.ts
```diff
@@ -0,0 +1,365 @@
...
+      description
+    )
+
+    await Promise.all([savePromise, githubUpdatePromise])
+  }
+
+  // NOTE: The semantics of our reviewing system is slightly different from github.
+  // The approval is tied to the request, rather than the user.
+  approveReviewRequest = async (reviewRequest: ReviewRequest) => {
+    reviewRequest.reviewStatus = "APPROVED"
```
dcshzj, line 340: Should we use an enum instead? Same for the other `reviewStatus` below.

src/routes/v2/authenticated/review.ts
```diff
@@ -0,0 +1,700 @@
...
+        message:
+          "User atttempted to create review request with invalid permissions",
+        method: "createReviewRequest",
+        meta: {
+          userId: userWithSiteSessionData.isomerUserId,
+          email: userWithSiteSessionData.email,
+          siteName,
+        },
+      })
+      return res.status(400).send({
```
dcshzj, line 130: Nit: This should be 403 Forbidden.

src/routes/v2/authenticated/review.ts
```diff
@@ -0,0 +1,700 @@
...
+    const { siteName } = req.params
+
+    // Check if they have access to site
+    const hasAccess = this.identityUsersService.hasAccessToSite(
+      userWithSiteSessionData.isomerUserId,
+      siteName
+    )
+
+    if (!hasAccess) {
+      return res.status(400).send()
```
dcshzj, line 72: Nit: This should be 403 Forbidden.

src/routes/v2/authenticated/review.ts
```diff
@@ -0,0 +1,700 @@
...
+      logger.error({
+        message: "Insufficient permissions to view review request",
+        method: "listReviews",
+        meta: {
+          userId: userWithSiteSessionData.isomerUserId,
+          email: userWithSiteSessionData.email,
+          siteName,
+        },
+      })
+      return res.status(400).send({
```
dcshzj, line 239: Nit: This should be 403 Forbidden.

src/routes/v2/authenticated/review.ts
```diff
@@ -0,0 +1,700 @@
...
+        message: "Insufficient permissoins to retrieve review request",
+        method: "getReviewRequest",
+        meta: {
+          userId: userWithSiteSessionData.isomerUserId,
+          email: userWithSiteSessionData.email,
+          siteName,
+          requestId,
+        },
+      })
+      return res.status(400).send({
```
dcshzj, line 303: Nit: This should be 403 Forbidden.

src/routes/v2/authenticated/review.ts
```diff
@@ -0,0 +1,700 @@
...
+        message: "Insufficient to update review request",
+        method: "updateReviewRequest",
+        meta: {
+          userId: userWithSiteSessionData.isomerUserId,
+          email: userWithSiteSessionData.email,
+          siteName,
+          requestId,
+        },
+      })
+      return res.status(401).json({
```
dcshzj, line 394: Nit: This should be 403 Forbidden. 401 is when the user is not authenticated and the response would change if the user were to be authenticated.

src/routes/v2/authenticated/review.ts
```diff
@@ -0,0 +1,700 @@
...
+        message: "Insufficient permissions to merge review request",
+        method: "mergeReviewRequest",
+        meta: {
+          userId: userSessionData.isomerUserId,
+          email: userSessionData.email,
+          siteName,
+          requestId,
+        },
+      })
+      return res.status(400).send({
```
dcshzj, line 480: Nit: This should be 403 Forbidden.

src/routes/v2/authenticated/review.ts
```diff
@@ -0,0 +1,700 @@
...
+        message: "Insufficient permissions to close review request",
+        method: "closeReviewRequest",
+        meta: {
+          userId: userWithSiteSessionData.isomerUserId,
+          email: userWithSiteSessionData.email,
+          siteName,
+          requestId,
+        },
+      })
+      return res.status(401).json({
```
dcshzj, line 655: Nit: This should be 403 Forbidden.

src/routes/v2/authenticated/review.ts
```diff
@@ -0,0 +1,700 @@
...
+        meta: {
+          userId: userWithSiteSessionData.isomerUserId,
+          email: userWithSiteSessionData.email,
+          siteName,
+        },
+      })
+      return res.status(404).send({
+        message: "Please ensure that the site exists!",
+      })
+    }
```
dcshzj, line 582: Nit: I think this response is not correct, should be 403 Forbidden with details that the user is not a reviewer authorised to approve the review request.

src/routes/v2/authenticated/review.ts
```diff
@@ -0,0 +1,700 @@
...
+    // as otherwise, they would be rejected by our middleware
+    // Check if they are a collaborator
+    const role = await this.collaboratorsService.getRole(
+      siteName,
+      userWithSiteSessionData.isomerUserId
+    )
+
+    if (!role) {
+      logger.error({
+        message: "Insufficient permissoins to retrieve review request",
```
dcshzj, line 294:
```suggestion
        message: "Insufficient permissions to retrieve review request",
```

dcshzj (changes requested):

src/services/identity/UsersService.ts
seaerchin: See above

src/services/review/ReviewRequestService.ts
seaerchin: Oh oof I think the comment got deleted through rebase

src/services/review/ReviewRequestService.ts
seaerchin: No strong opinion here, as functionality is the same. I can change it to enum!

src/routes/v2/authenticated/review.ts
seaerchin: Hmm I think @alexanderleegs told me something about how GitHub presents stuff that you don't have access to as 400s (or was it 404s) rather than 403s, which is why I did this. 

Do we still want to enforce this behaviour?

src/routes/v2/authenticated/review.ts
seaerchin: ^

src/routes/v2/authenticated/review.ts
seaerchin: ^

src/routes/v2/authenticated/review.ts
alexanderleegs: Yeah Github returns 404s for stuff you don't have access to - I think generally the choice for whether to return 404 or 403 works either way - 404s give potential attackers less information on whether a particular resource exists or not. I'm leaning towards 404 mostly because we already use this in a few places

src/services/review/ReviewRequestService.ts
seaerchin: because our semantics are different from github - github's approval is _per user_, whereas ours is _per review_.

hence, this means that using github's approval is more annoying to work with (eg: base PR states are `open | closed` and we have to traverse the data structure + reconcile to see if the PR is approved)

src/routes/v2/authenticated/review.ts
seaerchin: ok - for errors originating from insufficient perms, i'll return a 404? wdyt @dcshzj

dcshzj: Yup 404 is fine too.

src/services/review/ReviewRequestService.ts
dcshzj: Yup functionally is the same, but I think having an enum helps to enforce correctness a little bit better.

src/services/review/ReviewRequestService.ts
dcshzj: I see, that makes sense. Thanks for clarifying!

src/services/identity/UsersService.ts
dcshzj: Hmm in that case can I suggest switching the method name to be something like `getUserIfUserHasAccessToSite` or something like that. I see the merits of saving on boilerplate but I feel based on the naming of the method, I expect to be returned a boolean rather than a User object.

src/routes/v2/authenticated/review.ts
seaerchin: updaetd to 404

src/routes/v2/authenticated/review.ts
seaerchin: 404

src/routes/v2/authenticated/review.ts
seaerchin: 404

src/routes/v2/authenticated/review.ts
seaerchin: oops, thanks for catching!

src/routes/v2/authenticated/review.ts
seaerchin: updated to 403

src/services/identity/UsersService.ts
seaerchin: i think this is fair but lemme just check w/ @alexanderleegs if he has a downstream dependency here

dcshzj (approved):
LGTM, only the naming of the method left.

src/services/identity/UsersService.ts
seaerchin: thoughts on calling this `getSiteUser` or `getSiteMember`

dcshzj: `getSiteMember` sounds good

seaerchin: updated method + call sites

src/services/identity/UsersService.ts
seaerchin: turns out this method isn't used at all? more inclined to delete but renamed for now

PR #520, opened 2022-10-11
feat: add collaborators statistics API endpoint

dcshzj:
## Problem

The site dashboard requires an endpoint that provides statistics on the site collaborators, but there is none currently in the backend. This results in the site dashboard being unable to retrieve such data to be displayed.

## Solution

**Breaking Changes**

**Features**:

- A new `/sites/:siteName/collaborators/statistics` API endpoint has been introduced to display the statistics of the list of collaborators for the given site.

## Tests

## Deploy Notes

*None*


src/routes/v2/authenticated/__tests__/collaborators.spec.ts
```diff
@@ -176,4 +183,28 @@ describe("Collaborator Router", () => {
       )
     })
   })
+
+  describe("get collaborators statistics", () => {
+    it("should get collaborators statistics", async () => {
```
seaerchin, line 188: can add 1 more test for failure case?

src/routes/v2/authenticated/collaborators.ts
```diff
@@ -103,6 +103,23 @@ export class CollaboratorsRouter {
...
+    never,
+    { userWithSiteSessionData: UserWithSiteSessionData }
+  > = async (req, res) => {
+    const { siteName } = req.params
+    const statistics = await this.collaboratorsService.getStatistics(siteName)
+
+    // Check for error and throw
+    if (statistics instanceof BaseIsomerError) {
+      throw statistics
+    }
```
seaerchin, line 119: we shouldn't throw for _expected_ errors

src/services/identity/CollaboratorsService.ts
```diff
@@ -235,6 +238,39 @@ class CollaboratorsService {
...
+      ],
+    })
+
+    const collaborators = site?.site_members ?? []
+    const totalCount = collaborators.length
+
+    if (totalCount === 0) {
+      // Every site must have at least one collaborator
+      return new NotFoundError(`Site does not exist`)
+    }
```
seaerchin, line 263: wouldn't the correct condition be `!site`?

seaerchin (approved):
minor nits - i'm fine w/ just merging as is but the code style i want to see is that engineers are thoughtful about errors and edge cases so could i trouble you to fix this!!

src/routes/v2/authenticated/collaborators.ts
dcshzj: Fixed in 4481e22.

src/routes/v2/authenticated/__tests__/collaborators.spec.ts
dcshzj: Added in 1d4e43e.

src/services/identity/CollaboratorsService.ts
dcshzj: Hmm this condition is wider because earlier I checked if `site` and `site.site_members` exists. Beyond that, in the unlikely event that the site exists but there are no collaborators, this condition still catches it correctly.

PR #521, opened 2022-10-12
feat: add model and migration for review_request_views table

dcshzj:
## Problem

There is currently no database tables suitable for storing a record of the view status of a user for a particular review request. This makes it impossible for the site dashboard to display a differentiated look for new review requests that the user has not seen.

## Solution

**Breaking Changes**

**Features**:

- A new model and database migration file has been added to add the `review_request_views` table into the database. This table will keep a record of the first time that a user has seen the review request on the dashboard (using the `created_at` column), and also the latest time that the user has clicked into the review request itself (using the `last_viewed_at` column).

## Tests

**Smoke test**:
1. Perform the DB migration using `npm run db:migrate`.
2. Observe that the new `review_request_views` table has been created in the database.

## Deploy Notes

The database migration will need to be performed first during deployment.


src/database/migrations/20221012064037-create-review-request-views.js
```diff
@@ -0,0 +1,59 @@
...
+        primaryKey: true,
+        type: Sequelize.BIGINT,
+        references: {
+          model: "review_requests",
+          key: "id",
+        },
+        onUpdate: "CASCADE",
+        onDelete: "CASCADE",
+      },
+      site_id: {
```
seaerchin, line 16: actually idt we need to store this cos the review request holds both site and user id

seaerchin (approved):
minor comment

src/database/migrations/20221012064037-create-review-request-views.js
seaerchin: update: the rr table holds siteId and _requestor_ id (so we might still need the user id)

dcshzj: Yeah we will still need the user id as it can be someone that is completely not involved in the review request.

PR #523, opened 2022-10-13
Feat/generic notification creation

alexanderleegs:
This PR builds on https://github.com/isomerpages/isomercms-backend/pull/508 - it introduces notification creation for the following actions:

- Creating a review request
- Cancelling a review request
- Approving a review request
- Editing a repo while a review request is open

This PR should be reviewed from commit [750fdfc](https://github.com/isomerpages/isomercms-backend/pull/523/commits/750fdfc3c571adf03efc8db3886420f4161a2d5f) - the remainder are due to merge issues which will be resolved once we merge the relevant PRs.

Key Features
This PR adds a dependency of the reviews router to NotificationsService. It also adds a new middleware which is used for all endpoints which modify repo info - this is introduced in the `authenticatedSites` router, and the notifications router has been moved out of this router to facilitate this.

Other Misc changes

- Modify routeHandlers to allow for `next` to be passed to wrapped routes
- Notifications router moved to `authenticated` subrouter instead

### DIFF
--- a/src/middleware/notificationOnEditHandler.ts
+++ b/src/middleware/notificationOnEditHandler.ts
@@ -55,15 +55,18 @@ export class NotificationOnEditHandler {
     if (reviewRequests.length === 0) return
     // For now, we only have 1 active review request
     const reviewRequest = reviewRequests[0]
-    users.forEach(async (user: User & { SiteMember: SiteMember }) => {
-      if (user.id.toString() === userId) return // Don't create notification for the source user
-      const { SiteMember: siteMember } = user
-      await this.notificationsService.create({
-        siteMember,
-        link: "TODO",
-        notificationType: "updated_request",
-        notificationSourceUsername: email,
+
+    await Promise.all(
+      users.map(async (user: User & { SiteMember: SiteMember }) => {
+        if (user.id.toString() === userId) return // Don't create notification for the source user
+        const { SiteMember: siteMember } = user
+        await this.notificationsService.create({
+          siteMember,
+          link: "TODO",
+          notificationType: "updated_request",
+          notificationSourceUsername: email,
+        })
       })
-    })
+    )
   }
 }
--- a/src/middleware/notificationOnEditHandler.ts
+++ b/src/middleware/notificationOnEditHandler.ts
@@ -1,0 +1,69 @@
+import autoBind from "auto-bind"
+
+import UserWithSiteSessionData from "@root/classes/UserWithSiteSessionData"
+import { SiteMember, User } from "@root/database/models"
+import CollaboratorsService from "@root/services/identity/CollaboratorsService"
+import NotificationsService from "@root/services/identity/NotificationsService"
+import SitesService from "@root/services/identity/SitesService"
+import ReviewRequestService from "@root/services/review/ReviewRequestService"
+import { RequestHandler } from "@root/types"
+
+export class NotificationOnEditHandler {
+  private readonly reviewRequestService: ReviewRequestService
+
+  private readonly sitesService: SitesService
+
+  private readonly collaboratorsService: CollaboratorsService
+
+  private readonly notificationsService: NotificationsService
+
+  constructor({
+    reviewRequestService,
+    sitesService,
+    collaboratorsService,
+    notificationsService,
+  }: {
+    reviewRequestService: ReviewRequestService
+    sitesService: SitesService
+    collaboratorsService: CollaboratorsService
+    notificationsService: NotificationsService
+  }) {
+    this.reviewRequestService = reviewRequestService
+    this.sitesService = sitesService
+    this.collaboratorsService = collaboratorsService
+    this.notificationsService = notificationsService
+    // We need to bind all methods because we don't invoke them from the class directly
+    autoBind(this)
+  }
+
+  createNotification: RequestHandler<
+    never,
+    unknown,
+    unknown,
+    never,
+    { userWithSiteSessionData: UserWithSiteSessionData }
+  > = async (req, res, next) => {
+    const { userWithSiteSessionData } = res.locals
+    const { siteName, isomerUserId: userId, email } = userWithSiteSessionData
+    const site = await this.sitesService.getBySiteName(siteName)
+    const users = await this.collaboratorsService.list(siteName, userId)
+    if (!site) throw new Error("Site should always exist")
+    const reviewRequests = await this.reviewRequestService.listReviewRequest(
+      userWithSiteSessionData,
+      site
+    )
+    if (reviewRequests.length === 0) return
+    // For now, we only have 1 active review request
+    const reviewRequest = reviewRequests[0]
+    users.forEach(async (user: User & { SiteMember: SiteMember }) => {
+      if (user.id.toString() === userId) return // Don't create notification for the source user
+      const { SiteMember: siteMember } = user
+      await this.notificationsService.create({
+        siteMember,
+        link: "TODO",
+        notificationType: "updated_request",
+        notificationSourceUsername: email,
+      })
+    })
+  }
+}
--- a/src/middleware/notificationOnEditHandler.ts
+++ b/src/middleware/notificationOnEditHandler.ts
@@ -56,6 +56,8 @@ export class NotificationOnEditHandler {
       site
     )
     if (reviewRequests.length === 0) return
+    // For now, we only have 1 active review request
+    const reviewRequest = reviewRequests[0]
 
     await Promise.all(
       users.map(async (user: User & { SiteMember: SiteMember }) => {
@@ -63,7 +65,7 @@ export class NotificationOnEditHandler {
         const { SiteMember: siteMember } = user
         await this.notificationsService.create({
           siteMember,
-          link: "TODO",
+          link: `/sites/${siteName}/review/${reviewRequest.id}`,
           notificationType: "updated_request",
           notificationSourceUsername: email,
         })
--- a/src/middleware/notificationOnEditHandler.ts
+++ b/src/middleware/notificationOnEditHandler.ts
@@ -36,6 +36,9 @@ export class NotificationOnEditHandler {
     autoBind(this)
   }
 
+  /**
+   * Creates a notification. Requires attachSiteHandler as a precondition
+   */
   createNotification: RequestHandler<
     never,
     unknown,
@@ -53,8 +56,6 @@ export class NotificationOnEditHandler {
       site
     )
     if (reviewRequests.length === 0) return
-    // For now, we only have 1 active review request
-    const reviewRequest = reviewRequests[0]
 
     await Promise.all(
       users.map(async (user: User & { SiteMember: SiteMember }) => {
--- a/src/routes/v2/authenticated/__tests__/Notifications.spec.ts
+++ b/src/routes/v2/authenticated/__tests__/Notifications.spec.ts
@@ -3,10 +3,9 @@ import request from "supertest"
 
 import { attachReadRouteHandlerWrapper } from "@middleware/routeHandler"
 
-import { NotificationsRouter as _NotificationsRouter } from "@routes/v2/authenticatedSites/notifications"
-
 import { generateRouter } from "@fixtures/app"
 import { mockSiteName, mockIsomerUserId } from "@fixtures/sessionData"
+import { NotificationsRouter as _NotificationsRouter } from "@root/routes/v2/authenticated/notifications"
 import NotificationsService from "@services/identity/NotificationsService"
 
 describe("Notifications Router", () => {
--- a/src/routes/v2/authenticated/__tests__/Notifications.spec.ts
+++ b/src/routes/v2/authenticated/__tests__/Notifications.spec.ts
@@ -5,6 +5,7 @@ import { attachReadRouteHandlerWrapper } from "@middleware/routeHandler"
 
 import { generateRouter } from "@fixtures/app"
 import { mockSiteName, mockIsomerUserId } from "@fixtures/sessionData"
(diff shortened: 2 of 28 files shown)


src/middleware/notificationOnEditHandler.ts
```diff
@@ -0,0 +1,72 @@
...
+    // For now, we only have 1 active review request
+    const reviewRequest = reviewRequests[0]
+
+    await Promise.all(
+      users.map(async (user: User & { SiteMember: SiteMember }) => {
+        if (user.id.toString() === userId) return // Don't create notification for the source user
+        const { SiteMember: siteMember } = user
+        await this.notificationsService.create({
+          siteMember,
+          link: "TODO",
```
seaerchin, line 65: could i just clarify what the `link` property means? if it's a link to the comment itself, could i check why the url can't be generated at this point?

src/middleware/notificationOnEditHandler.ts
```diff
@@ -0,0 +1,72 @@
...
+  }) {
+    this.reviewRequestService = reviewRequestService
+    this.sitesService = sitesService
+    this.collaboratorsService = collaboratorsService
+    this.notificationsService = notificationsService
+    // We need to bind all methods because we don't invoke them from the class directly
+    autoBind(this)
+  }
+
+  createNotification: RequestHandler<
```
seaerchin, line 39: this doesn't seem to be used ._.

src/routes/v2/authenticated/__tests__/Notifications.spec.ts
```diff
@@ -0,0 +1,113 @@
+import express from "express"
+import request from "supertest"
+
+import { attachReadRouteHandlerWrapper } from "@middleware/routeHandler"
```
seaerchin, line 4: skipped - assumed to be same as that of tests PR

src/routes/v2/authenticated/review.ts
```diff
@@ -185,6 +191,21 @@ export class ReviewsRouter {
...
+      const notificationType = reviewersMap[user.email || ""]
+        ? "sent_request"
+        : "request_created"
+      await this.notificationsService.create({
+        siteMember: user.SiteMember,
+        link: "TODO",
+        notificationType,
+        notificationSourceUsername: userWithSiteSessionData.email,
+      })
+    })
```
seaerchin, line 207: could we use `Promise.all(collaborators.map(...))` instead? see [here](https://gist.github.com/joeytwiddle/37d2085425c049629b80956d3c618971) for how `forEach` behaves in async loops (badly)

src/routes/v2/authenticated/review.ts
```diff
@@ -185,6 +191,21 @@ export class ReviewsRouter {
       description
     )
 
+    // Step 5: Create notifications
+    collaborators.forEach(async (user: User & { SiteMember: SiteMember }) => {
+      // Don't send notification to self
+      if (user.id.toString() === userWithSiteSessionData.isomerUserId) return
+      const notificationType = reviewersMap[user.email || ""]
```
seaerchin, line 198: should we even create if the user is a github user?

src/routes/v2/authenticated/review.ts
```diff
@@ -185,6 +191,21 @@ export class ReviewsRouter {
...
+        ? "sent_request"
+        : "request_created"
+      await this.notificationsService.create({
+        siteMember: user.SiteMember,
+        link: "TODO",
```
seaerchin, line 203: see above wrt link generation

src/routes/v2/authenticated/review.ts
```diff
@@ -685,7 +706,43 @@ export class ReviewsRouter {
...
+
+    // Step 5: Approve review request
+    // NOTE: We are not checking for existence of PR
+    // as the underlying Github API returns 404 if
+    // the requested review could not be found.
+    await this.reviewRequestService.approveReviewRequest(possibleReviewRequest)
+
+    // Step 6: Create notifications
+    const collaborators = await this.collaboratorsService.list(siteName)
+    collaborators.forEach(async (user: User & { SiteMember: SiteMember }) => {
```
seaerchin, line 734: see above comment wrt `forEach` and async loop

src/routes/v2/authenticated/review.ts
```diff
@@ -685,7 +706,43 @@ export class ReviewsRouter {
...
+      // Don't send notification to self
+      if (user.id.toString() === userWithSiteSessionData.isomerUserId) return
+      await this.notificationsService.create({
+        siteMember: user.SiteMember,
+        link: "TODO",
```
seaerchin, line 739: see above wrt link gen

src/routes/v2/authenticated/review.ts
```diff
@@ -766,6 +823,18 @@ export class ReviewsRouter {
     // The error is discarded as we are guaranteed to have a review request
     await this.reviewRequestService.deleteAllReviewRequestViews(site, requestId)
 
+    // Step 7: Create notifications
+    const collaborators = await this.collaboratorsService.list(siteName)
+    collaborators.forEach(async (user: User & { SiteMember: SiteMember }) => {
```
seaerchin, line 828: same 2 comments

src/routes/v2/authenticatedSites/index.js
```diff
@@ -225,10 +222,7 @@ const getAuthenticatedSitesSubrouter = ({
   authenticatedSitesSubrouter.use("/contactUs", contactUsV2Router.getRouter())
   authenticatedSitesSubrouter.use("/homepage", homepageV2Router.getRouter())
   authenticatedSitesSubrouter.use("/settings", settingsV2Router.getRouter())
-  authenticatedSitesSubrouter.use(
-    "/notifications",
-    notificationsRouter.getRouter()
-  )
+  authenticatedSitesSubrouter.use(notificationOnEditHandler.createNotification)
```
seaerchin, line 225: double checking - we added a `next` parameter because we want control to come here and create a new notification, correct?

src/services/identity/NotificationsService.ts
```diff
@@ -218,31 +196,30 @@ class NotificationsService {
...
-        createdAt: new Date(),
-      })
+      // createdAt is a special column which must be flagged as changed
+      recentTargetNotification.changed("createdAt", true)
+      await recentTargetNotification.update(
+        {
+          firstReadTime: null,
+          createdAt: new Date(),
+        },
+        { raw: true }
```
seaerchin, line 215: why do we require this flag?

src/services/identity/NotificationsService.ts
```diff
@@ -218,31 +196,30 @@ class NotificationsService {
...
-      await recentTargetNotification.update({
-        firstReadTime: null,
-        createdAt: new Date(),
-      })
+      // createdAt is a special column which must be flagged as changed
+      recentTargetNotification.changed("createdAt", true)
+      await recentTargetNotification.update(
+        {
+          firstReadTime: null,
+          createdAt: new Date(),
```
seaerchin, line 213: could i get some clarity on this? i'm not entirely sure why we are erasing these properties and setting `createdAt` to now

src/services/identity/__tests__/NotificationsService.spec.ts
```diff
@@ -0,0 +1,211 @@
+import { ModelStatic } from "sequelize/types"
```
seaerchin, line 1: skipped

src/server.js
```diff
@@ -192,10 +201,10 @@ app.use("/v1/sites/:siteName", authenticatedSitesSubrouterV1)
 app.use("/v1", authenticatedSubrouterV1)
 
 app.use("/v2/auth", authV2Router.getRouter())
-// Endpoints which have siteName, used to inject site access token
-app.use("/v2/sites/:siteName", authenticatedSitesSubrouterV2)
 // Endpoints which have require login, but not site access token
 app.use("/v2", authenticatedSubrouterV2)
+// Endpoints which modify the github repo, used to inject site access token
+app.use("/v2/sites/:siteName", authenticatedSitesSubrouterV2)
```
seaerchin, line 207: no strong opinion, just curious why do here rather than the init method in `identity/index`

src/services/identity/NotificationsService.ts
alexanderleegs: `createdAt` seems to be a special param because it's automatically generated - it can't seem to be edited normally, so we use raw to allow the normal setter to be ignored

See https://sequelize.org/v3/api/instance/#setkey-value-options!

src/services/identity/NotificationsService.ts
alexanderleegs: We're updating an existing notification so that we don't flood the user with notifications if many of the same type are created within a short period of time - the desired behaviour is for the past notification to be updated instead, and we simulate this by updating the `createdAt` time, which is used to indicate the time of notification

src/server.js
alexanderleegs: This was what was already done prior to this PR - I only rearranged the position so that it ends up after `/v2` endpoints! The reason for that was because of the new notifications middleware, which would cause issues for `/v2/site/:siteName` endpoints which belong in `authenticatedSubrouter` instead (such as collaborators and review requests)

I think we had this just because everything in that subrouter uses the same prefix!

src/middleware/notificationOnEditHandler.ts
```diff
@@ -0,0 +1,72 @@
...
+    never,
+    unknown,
+    unknown,
+    never,
+    { userWithSiteSessionData: UserWithSiteSessionData }
```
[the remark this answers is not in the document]
alexanderleegs: Sure, added in https://github.com/isomerpages/isomercms-backend/pull/523/commits/df487349cb1914985191ad3998101c6c427b3582

src/middleware/notificationOnEditHandler.ts
```diff
@@ -0,0 +1,72 @@
...
+    const site = await this.sitesService.getBySiteName(siteName)
+    const users = await this.collaboratorsService.list(siteName, userId)
+    if (!site) throw new Error("Site should always exist")
+    const reviewRequests = await this.reviewRequestService.listReviewRequest(
+      userWithSiteSessionData,
+      site
+    )
+    if (reviewRequests.length === 0) return
+    // For now, we only have 1 active review request
+    const reviewRequest = reviewRequests[0]
```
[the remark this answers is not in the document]
alexanderleegs: Whoops was supposed to add this in after confirming the frontend url for review requests - it's been added in https://github.com/isomerpages/isomercms-backend/pull/523/commits/0dda45e2ae01477e5ea524b8e4733fb778070744

src/middleware/notificationOnEditHandler.ts
alexanderleegs: Was waiting for the frontend url and forgot about it afterwards! Updated in https://github.com/isomerpages/isomercms-backend/pull/523/commits/0dda45e2ae01477e5ea524b8e4733fb778070744

src/routes/v2/authenticated/review.ts
alexanderleegs: Good catch, I forgot to swap these out - fixed in b9449d4

src/routes/v2/authenticated/review.ts
alexanderleegs: It shouldn't, but this route is only accessible to email users right?

src/routes/v2/authenticated/review.ts
alexanderleegs: b9449d4

src/routes/v2/authenticated/review.ts
alexanderleegs: b9449d4

src/routes/v2/authenticated/review.ts
alexanderleegs: b9449d4

src/routes/v2/authenticated/review.ts
alexanderleegs: b9449d4

src/middleware/notificationOnEditHandler.ts
```diff
@@ -0,0 +1,75 @@
...
+    unknown,
+    unknown,
+    never,
+    { userWithSiteSessionData: UserWithSiteSessionData }
+  > = async (req, res, next) => {
+    const { userWithSiteSessionData } = res.locals
+    const { siteName, isomerUserId: userId, email } = userWithSiteSessionData
+    const site = await this.sitesService.getBySiteName(siteName)
+    const users = await this.collaboratorsService.list(siteName, userId)
+    if (!site) throw new Error("Site should always exist")
```
seaerchin, line 53: out of curiosity, how does this work? if the earlier function did its job, it would have done a `res.status(200).send()`. 

however, this will cause a short circuit to `errorHandler` isn't it and cause a conflicting status code

seaerchin (approved):

src/services/identity/NotificationsService.ts
seaerchin: actually i'm still not very sure of this HAHA

it seems that we're finding a recent notification targeted at the current user and setting it to be unread? but wouldn't similar notifs (but with different messages) then be hidden?

src/middleware/notificationOnEditHandler.ts
alexanderleegs: hmm i actually haven't tested this! That said this should be a theoretically unreachable state

src/services/identity/NotificationsService.ts
alexanderleegs: We're also looking for a notif of the same type - currently those all share the same message, but with the only difference being the username in the message, which should be okay to overwrite, so losing the content of the original message shouldn't be a problem

PR #532, opened 2022-10-13
feat: add review request views API endpoint and functions

dcshzj:
## Problem

The current backend does not support recording a user's viewing status of a particular review request, which makes it impossible for the site dashboard to differentiate between a new/existing review request that the user has/has not seen before.

## Solution

**Breaking Changes**

**Features**:

- A new API endpoint `POST /v2/sites/:siteName/review/viewed` has been added, which does not take in any input and will automatically record all open/approved review requests for that particular site as viewed.
- New functions have been added to ReviewRequestService to mark all review requests as viewed, to update the lastViewedAt attribute for a particular review request, and to delete any existing review request views.

## Tests

No tests currently. 😭 

## Deploy Notes

*None*

### DIFF
--- a/src/services/review/ReviewRequestService.ts
+++ b/src/services/review/ReviewRequestService.ts
@@ -283,13 +283,16 @@ export default class ReviewRequestService {
     pullRequestNumber: number
   ): Promise<void | RequestNotFoundError> => {
     const { isomerUserId: userId } = sessionData
-    const reviewRequest = await this.getReviewRequest(site, pullRequestNumber)
+    const possibleReviewRequest = await this.getReviewRequest(
+      site,
+      pullRequestNumber
+    )
 
-    if (isIsomerError(reviewRequest)) {
-      return reviewRequest
+    if (isIsomerError(possibleReviewRequest)) {
+      return possibleReviewRequest
     }
 
-    const { id: reviewRequestId } = reviewRequest
+    const { id: reviewRequestId } = possibleReviewRequest
 
     await this.reviewRequestView.update(
       {
@@ -309,13 +312,16 @@ export default class ReviewRequestService {
     site: Site,
     pullRequestNumber: number
   ): Promise<void | RequestNotFoundError> => {
-    const reviewRequest = await this.getReviewRequest(site, pullRequestNumber)
+    const possibleReviewRequest = await this.getReviewRequest(
+      site,
+      pullRequestNumber
+    )
 
-    if (isIsomerError(reviewRequest)) {
-      return reviewRequest
+    if (isIsomerError(possibleReviewRequest)) {
+      return possibleReviewRequest
     }
 
-    const { id: reviewRequestId } = reviewRequest
+    const { id: reviewRequestId } = possibleReviewRequest
 
     await this.reviewRequestView.destroy({
       where: {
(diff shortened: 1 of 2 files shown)


src/services/review/ReviewRequestService.ts
```diff
@@ -196,13 +215,103 @@ export default class ReviewRequestService {
...
+    const requestIdsViewed = requestsViewed.map(
+      (request) => request.reviewRequestId
+    )
+    const allActiveRequestIds = allActiveRequests.map((request) => request.id)
+    const requestIdsToMarkAsViewed = _.difference(
+      allActiveRequestIds,
+      requestIdsViewed
+    )
+
+    requestIdsToMarkAsViewed.forEach(async (requestId) => {
```
seaerchin, line 255: this might actually be blocking sequentially (see [here](https://stackoverflow.com/questions/37576685/using-async-await-with-a-foreach-loop)). even though semantically, `forEach` is more indicative of a side effect, we might wanna `Promise.all` and `map` here to ensure greater throughput (if it matters. if this endpoint is relatively fast/we have a reasonable bound for array size, fel free to ignore)

src/services/review/ReviewRequestService.ts
```diff
@@ -196,13 +215,103 @@ export default class ReviewRequestService {
...
+    })
+  }
+
+  updateReviewRequestLastViewedAt = async (
+    sessionData: UserWithSiteSessionData,
+    site: Site,
+    pullRequestNumber: number
+  ): Promise<void | RequestNotFoundError> => {
+    const { isomerUserId: userId } = sessionData
+    const reviewRequest = await this.getReviewRequest(site, pullRequestNumber)
```
seaerchin, line 273: change to `possibleReviewRequest` cos it could be an error at this point

src/services/review/ReviewRequestService.ts
```diff
@@ -196,13 +215,103 @@ export default class ReviewRequestService {
...
+        },
+      }
+    )
+  }
+
+  deleteAllReviewRequestViews = async (
+    site: Site,
+    pullRequestNumber: number
+  ): Promise<void | RequestNotFoundError> => {
+    const reviewRequest = await this.getReviewRequest(site, pullRequestNumber)
```
seaerchin, line 299: see above wrt naming

src/services/review/ReviewRequestService.ts
```diff
@@ -196,13 +215,103 @@ export default class ReviewRequestService {
...
+        where: {
+          reviewRequestId,
+          siteId: site.id,
+          userId,
+        },
+      }
+    )
+  }
+
+  deleteAllReviewRequestViews = async (
```
seaerchin, line 295: just curious, why do we need to delete the views? is this to avoid paying storage costs?

seaerchin (commented):
should change the naming so it's obvious that it's a potential error

src/services/review/ReviewRequestService.ts
dcshzj: Yup that's correct. Plus it is in case we want to allow reopening of closed RRs, then we want to show it as a new RR for everyone.

src/services/review/ReviewRequestService.ts
dcshzj: Good point! I forgot about this, adjusted in 5ff10c4.

src/services/review/ReviewRequestService.ts
dcshzj: Adjusted in d9ca3d8.

src/services/review/ReviewRequestService.ts
dcshzj: Adjusted in d9ca3d8.

PR #533, opened 2022-10-14
refactor: remove all usages of TokenStore and tests

dcshzj:
## Problem

The TokenStore is currently not used.

## Solution

**Breaking Changes**

**Improvements**:

- The TokenStore and all usages/tests have been removed.

## Tests

## Deploy Notes

Any usages on Amazon's secrets manager should also be removed just in case.

Need to confirm with @alexanderleegs whether the api_token_name column in the sites table can be dropped as well?


seaerchin (approved):
lgtm

seaerchin:
btw we might wanna drop related cols/tables in our db too

dcshzj:
As discussed, will remove the `api_token_name` column in the future.

PR #534, opened 2022-10-18
Feat/comments

alexanderleegs:
## Problem

This PR adds comments endpoints. It introduces the ability to retrieve and create comments, and also renames an existing endpoint to be used to mark the time at which users read comments. More information on the endpoints can be found in the [design doc](https://www.notion.so/opengov/Comments-Notifications-4332401f397c487e9a6e3bf641b23814).

## Testing
For testing of notification endpoints, the appropriate review request entries in `ReviewRequests`, `ReviewRequestMeta`, and `ReviewRequestView` should be populated, including an active pull request on the desired repo. The creation of a notification will populate the appropriate comment as a github comment on the actual pull request.

### DIFF
--- a/src/routes/v2/authenticated/review.ts
+++ b/src/routes/v2/authenticated/review.ts
@@ -19,6 +19,7 @@ import UsersService from "@root/services/identity/UsersService"
 import { isIsomerError, RequestHandler } from "@root/types"
 import { ResponseErrorBody } from "@root/types/dto/error"
 import {
+  CommentItem,
   DashboardReviewRequestDto,
   EditedItemDto,
   UpdateReviewRequestDto,
@@ -634,6 +635,52 @@ export class ReviewsRouter {
     return res.status(200).send()
   }
 
+  getComments: RequestHandler<
+    { siteName: string; requestId: number },
+    CommentItem[] | ResponseErrorBody,
+    never,
+    unknown,
+    { userWithSiteSessionData: UserWithSiteSessionData }
+  > = async (req, res) => {
+    const { siteName, requestId } = req.params
+    const { userWithSiteSessionData } = res.locals
+    // Step 1: Check that the site exists
+    const site = await this.sitesService.getBySiteName(siteName)
+    if (!site) {
+      return res.status(404).send({
+        message: "Please ensure that the site exists!",
+      })
+    }
+
+    // Step 2: Retrieve comments
+    const comments = await this.reviewRequestService.getComments(
+      userWithSiteSessionData,
+      site,
+      requestId
+    )
+
+    return res.status(200).json(comments)
+  }
+
+  createComment: RequestHandler<
+    { siteName: string; requestId: number },
+    string | ResponseErrorBody,
+    { message: string },
+    unknown,
+    { userWithSiteSessionData: UserWithSiteSessionData }
+  > = async (req, res) => {
+    const { requestId } = req.params
+    const { message } = req.body
+    const { userWithSiteSessionData } = res.locals
+    await this.reviewRequestService.createComment(
+      userWithSiteSessionData,
+      requestId,
+      message
+    )
+
+    return res.status(200).send("OK")
+  }
+
   markReviewRequestCommentsAsViewed: RequestHandler<
     { siteName: string; requestId: number },
     string | ResponseErrorBody,
@@ -794,8 +841,16 @@ export class ReviewsRouter {
       "/:requestId/approve",
       attachReadRouteHandlerWrapper(this.approveReviewRequest)
     )
+    router.get(
+      "/:requestId/comments",
+      attachWriteRouteHandlerWrapper(this.getComments)
+    )
     router.post(
-      "/:requestId/viewedComments",
+      "/:requestId/comments",
+      attachWriteRouteHandlerWrapper(this.createComment)
+    )
+    router.post(
+      "/:requestId/comments/viewedComments",
       attachWriteRouteHandlerWrapper(this.markReviewRequestCommentsAsViewed)
     )
     router.post(
--- a/src/routes/v2/authenticated/review.ts
+++ b/src/routes/v2/authenticated/review.ts
@@ -637,7 +637,7 @@ export class ReviewsRouter {
 
   getComments: RequestHandler<
     { siteName: string; requestId: number },
-    CommentItem[] | ResponseErrorBody,
+    CommentItem[],
     never,
     unknown,
     { userWithSiteSessionData: UserWithSiteSessionData }
--- a/src/routes/v2/authenticated/review.ts
+++ b/src/routes/v2/authenticated/review.ts
@@ -647,6 +647,15 @@ export class ReviewsRouter {
     // Step 1: Check that the site exists
     const site = await this.sitesService.getBySiteName(siteName)
     if (!site) {
+      logger.error({
+        message: "Invalid site requested",
+        method: "getComments",
+        meta: {
+          userId: userWithSiteSessionData.isomerUserId,
+          email: userWithSiteSessionData.email,
+          siteName,
+        },
+      })
       return res.status(404).send({
         message: "Please ensure that the site exists!",
       })
--- a/src/routes/v2/authenticated/review.ts
+++ b/src/routes/v2/authenticated/review.ts
@@ -637,7 +637,7 @@ export class ReviewsRouter {
 
   getComments: RequestHandler<
     { siteName: string; requestId: number },
-    CommentItem[],
+    CommentItem[] | ResponseErrorBody,
     never,
     unknown,
     { userWithSiteSessionData: UserWithSiteSessionData }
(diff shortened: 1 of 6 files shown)


src/routes/v2/authenticated/review.ts
```diff
@@ -634,6 +635,52 @@ export class ReviewsRouter {
...
+      site,
+      requestId
+    )
+
+    return res.status(200).json(comments)
+  }
+
+  createComment: RequestHandler<
+    { siteName: string; requestId: number },
+    string | ResponseErrorBody,
```
seaerchin, line 667: doesn't seem to return `ResponseErrorBody` .-.

src/services/db/review.ts
```diff
@@ -83,3 +90,45 @@ export const approvePullRequest = (
...
+  siteName: string,
+  pullRequestNumber: number
+) => {
+  const rawComments = await axiosInstance
+    .get<RawComment[]>(`${siteName}/issues/${pullRequestNumber}/comments`)
+    .then(({ data }) => data)
+  return _.compact(
+    rawComments.map((rawComment) => {
+      try {
+        const commentData = JSON.parse(rawComment.body)
```
seaerchin, line 104: think i have a type + util for this - could be in `review`! maybe we should use that instead?

src/services/db/review.ts
```diff
@@ -83,3 +90,45 @@ export const approvePullRequest = (
...
+        const { user, message } = commentData
+        return {
+          user,
+          message,
+          createdAt: rawComment.created_at,
+        }
+      } catch (e) {
+        // Not properly formatted comment, ignore
+        return null
```
seaerchin, line 113: i checked in `ReviewRequestService` and this doesn't seem to be reflected, which means that this is a dangerous return. could we annotate the return type on the function explicitly

src/services/db/review.ts
```diff
@@ -83,3 +90,45 @@ export const approvePullRequest = (
...
+  message: string
+) => {
+  const stringifiedMessage = JSON.stringify({
+    user,
+    message,
+  })
+  return axiosInstance.post<void>(
+    `${siteName}/issues/${pullRequestNumber}/comments`,
+    // NOTE: only create body if a valid description is given
+    { body: stringifiedMessage }
```
seaerchin, line 132: this comment seems to be erroneous - could we remove it?

src/services/review/ReviewRequestService.ts
```diff
@@ -455,4 +456,67 @@ export default class ReviewRequestService {
...
+  }
+
+  getComments = async (
+    sessionData: UserWithSiteSessionData,
+    site: Site,
+    pullRequestNumber: number
+  ): Promise<CommentItem[]> => {
+    const { siteName, isomerUserId: userId } = sessionData
+
+    // Find all review requests associated with the site
```
seaerchin, line 482: nit: erroneous comment

seaerchin (commented):
minor comments

src/routes/v2/authenticated/review.ts
```diff
@@ -634,6 +635,52 @@ export class ReviewsRouter {
...
+    const site = await this.sitesService.getBySiteName(siteName)
+    if (!site) {
+      return res.status(404).send({
+        message: "Please ensure that the site exists!",
+      })
+    }
+
+    // Step 2: Retrieve comments
+    const comments = await this.reviewRequestService.getComments(
+      userWithSiteSessionData,
```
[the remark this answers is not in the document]
alexanderleegs: Yup this is true! Just logging into the CMS causes a User to be created - so we should never have the case where someone has a valid jwt token but isn't in our db

src/services/review/ReviewRequestService.ts
alexanderleegs: https://github.com/isomerpages/isomercms-backend/pull/534/commits/886deb1e5d59886e0a4942270a9b7b632612a453

src/services/db/review.ts
alexanderleegs: https://github.com/isomerpages/isomercms-backend/pull/534/commits/886deb1e5d59886e0a4942270a9b7b632612a453

src/routes/v2/authenticated/review.ts
alexanderleegs: Removed in https://github.com/isomerpages/isomercms-backend/pull/534/commits/8f8bbb7e163fe141a0d84ec20d433dfb5c99e0a1

src/routes/v2/authenticated/review.ts
```diff
@@ -634,6 +635,52 @@ export class ReviewsRouter {
...
+    never,
+    unknown,
+    { userWithSiteSessionData: UserWithSiteSessionData }
+  > = async (req, res) => {
+    const { siteName, requestId } = req.params
+    const { userWithSiteSessionData } = res.locals
+    // Step 1: Check that the site exists
+    const site = await this.sitesService.getBySiteName(siteName)
+    if (!site) {
+      return res.status(404).send({
```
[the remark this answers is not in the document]
alexanderleegs: Added in fe7463b

src/services/db/review.ts
alexanderleegs: Swapped in 2667ea5 - Zhong Jun also raised that we shouldn't be putting users' emails directly in the comment, so that has been modified as well

src/services/db/review.ts
alexanderleegs: The nulls are actually removed by `_.compact`! I've added comments to make it clearer

src/routes/v2/authenticated/review.ts
```diff
@@ -634,6 +635,61 @@ export class ReviewsRouter {
     return res.status(200).send()
   }
 
+  getComments: RequestHandler<
+    { siteName: string; requestId: number },
+    CommentItem[],
```
seaerchin, line 640: type needs to be `CommentItem[] | ResponseErrorBody` so that `{message: "<some_message>"}` passes type check

src/routes/v2/authenticated/review.ts
seaerchin: think you removed the wrong one HAHA

src/services/db/review.ts
```diff
@@ -83,3 +91,41 @@ export const approvePullRequest = (
...
+      const commentData = fromGithubCommitMessage(rawComment.body)
+      if (_.isEmpty(commentData)) return null // Will be filtered out by _.compact
+      const { userId, message } = commentData
+      if (!userId || !message) return null // Will be filtered out by _.compact
+      return {
+        userId,
+        message,
+        createdAt: rawComment.created_at,
+      }
+    })
```
seaerchin, line 113: nit: might be better off as `map -> filter -> map` but this is ok as is too

src/services/review/ReviewRequestService.ts
```diff
@@ -125,6 +126,29 @@ export default class ReviewRequestService {
     return mappings
   }
 
+  computeCommentDataMappings = async (
+    comments: {
+      userId: string
+      message: string
+      createdAt: string
+    }[],
```
seaerchin, line 134: could we make a type for this in `types/review`? possibly through reusing `CommentItem`? eg: `Pick<CommentItem, 'message' | 'createdAt'> & {userId: string}`

src/services/review/ReviewRequestService.ts
```diff
@@ -125,6 +126,29 @@ export default class ReviewRequestService {
     return mappings
   }
 
+  computeCommentDataMappings = async (
```
seaerchin, line 129: wdyt about renaming this to `computeCommentData` as it returns an array, not a map of x to y (eg: object/map)

seaerchin (commented):
minor comments, fixing type more impt so the build works

src/routes/v2/authenticated/review.ts
alexanderleegs: whoops, fixed in https://github.com/isomerpages/isomercms-backend/pull/534/commits/d60c399848f15ddc5465da95e048103f03902053

src/routes/v2/authenticated/review.ts
alexanderleegs: oops, fixed in https://github.com/isomerpages/isomercms-backend/pull/534/commits/d60c399848f15ddc5465da95e048103f03902053

src/services/review/ReviewRequestService.ts
alexanderleegs: Sure, added in bd98768 - I've made it a new type because the `createdAt` from github is a string!

src/services/review/ReviewRequestService.ts
alexanderleegs: Sure, bd98768

dcshzj:
Whoops, should not have merged #549 in, so there are some additional lines in ReviewRequestService.ts.

dcshzj (approved):
LGTM

PR #535, opened 2022-10-18
feat: add new siteUrl API endpoint to get the production URL of site

dcshzj:
## Problem

There is currently no existing API endpoints for getting just the production URL of a particular site. There is an existing site info endpoint but it is too expensive (extra calls to get commits information), and the settings endpoint gets a user-defined value of the production URL, which may not exist or may be empty.

## Solution

**Breaking Changes**

**Improvements**:

- Introduce a new `/sites/:siteName/siteUrl` API endpoint to get the production URL of the given site.
- Also fixed a regression where the key for the `stagingUrl` has changed.

## Tests

**Unit tests**: Run `npm run tests`

## Deploy Notes

*None*

### DIFF
--- a/src/routes/v2/authenticated/sites.ts
+++ b/src/routes/v2/authenticated/sites.ts
@@ -71,7 +71,26 @@ export class SitesRouter {
     if (possibleStagingUrl instanceof BaseIsomerError) {
       return res.status(404).json({ message: possibleStagingUrl.message })
     }
-    return res.status(200).json({ possibleStagingUrl })
+    return res.status(200).json({ stagingUrl: possibleStagingUrl })
+  }
+
+  getSiteUrl: RequestHandler<
+    { siteName: string },
+    unknown,
+    never,
+    never,
+    { userWithSiteSessionData: UserWithSiteSessionData }
+  > = async (req, res) => {
+    const { userWithSiteSessionData } = res.locals
+    const possibleSiteUrl = await this.sitesService.getSiteUrl(
+      userWithSiteSessionData
+    )
+
+    // Check for error and throw
+    if (possibleSiteUrl instanceof BaseIsomerError) {
+      return res.status(404).json({ message: possibleSiteUrl.message })
+    }
+    return res.status(200).json({ siteUrl: possibleSiteUrl })
   }
 
   getSiteInfo: RequestHandler<
@@ -107,6 +126,11 @@ export class SitesRouter {
       "/:siteName/stagingUrl",
       attachSiteHandler,
       attachReadRouteHandlerWrapper(this.getStagingUrl)
+    )
+    router.get(
+      "/:siteName/siteUrl",
+      attachSiteHandler,
+      attachReadRouteHandlerWrapper(this.getSiteUrl)
     )
     router.get(
       "/:siteName/info",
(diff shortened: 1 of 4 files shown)


src/routes/v2/authenticated/sites.ts
```diff
@@ -71,7 +71,26 @@ export class SitesRouter {
...
+  > = async (req, res) => {
+    const { userWithSiteSessionData } = res.locals
+    const possibleSiteUrl = await this.sitesService.getSiteUrl(
+      userWithSiteSessionData
+    )
+
+    // Check for error and throw
+    if (possibleSiteUrl instanceof BaseIsomerError) {
+      return res.status(404).json({ message: possibleSiteUrl.message })
+    }
```
alexanderleegs, line 92: Could we throw a `NotFoundError` instead so it's easier to standardise our errors?

seaerchin: Sadge, I'd actually prefer not to

dcshzj: Yeah @seaerchin [mentioned before](https://github.com/isomerpages/isomercms-backend/pull/513#discussion_r994350033) that the error handling middleware should be used for handling unknown exceptions, and we should be able to return the appropriate response for known exceptions.

I don't have too strong opinions but I think it makes sense that we handle errors that we know of here in the router level. It also makes it clear here on what is known to be returned to the user (along with the HTTP status codes). I would see it more like router level handles 4xx errors whereas middleware handles 5xx errors.

seaerchin: to add onto this point, we need to separate _exceptions_ from _errors_ . exceptions are things that our application cannot handle and we are forced to rely on a generic error message to show to the user. These are handled by our middleware (and i've introduced a new error type for this on the FE on the collaborators refactor PR).

Conversely, errors are things that we know about and we can return a well-formed message to the user, rather than just falling back on a uninformative message. In this case, we can return a known shape which the frontend can consume readily and provide targeted feedback to the user (eg: no access rights). 

I think a good next step would be to add a util function on the frontend to soothe over this process (i think i added one, but needs to be refined), so that if we get an _expected error_, we use the backend's error message and otherwise, fallback on our generic error messages (ie, your internet slow)

src/services/identity/__tests__/SitesService.spec.ts
```diff
@@ -508,6 +508,51 @@ describe("SitesService", () => {
     })
   })
 
+  describe("getSiteUrl", () => {
+    const stagingUrl = "https://repo-staging.netlify.app"
+    const productionUrl = "https://repo-prod.netlify.app"
+
+    it("should return the site URL if it is available", async () => {
+      // Arrange
```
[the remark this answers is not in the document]
dcshzj: > iirc we had a convo about whether we should merge or use from one source right? (ie, if db has staging but not prod -> use both from config)

Hmm I don't recall this convo hahaha but I didn't intend to have this. I was thinking in the longer term, we should be moving towards having a single source of truth for the URLs, which means that the deployments DB should be updated if the URLs are not available there and retrieved from GitHub. Only if it is not available do we retrieve from the next best alternative (so users get the best possible results for both).

> 2. if db has staging but not prod, and we query config service, staging will be from db but prod will be from config

Yup, currently this is the setup. The tests are done separately on the `getUrlsOfSite` function within SitesService. There is no specific test for this case where one is in db but not the other, but my observation is that such scenarios are very uncommon.

seaerchin (approved):

PR #539, opened 2022-10-20
feat(rr): allow updating of admins

seaerchin:
## Problem
See [#1108](https://github.com/isomerpages/isomercms-frontend/issues/1108) for an overview 

Closes [#1108](https://github.com/isomerpages/isomercms-frontend/issues/1108)

## Solution
1. Removed updating title/desc (currently no option to use it) 
2. updated to use `$set` to update the field. this is given by `sequelize-typescript` as `sequelize` generates special methods for associations but TS will complain. see [here](https://github.com/sequelize/sequelize-typescript#type-safe-usage-of-auto-generated-functions) for more details


alexanderleegs (approved):
lgtm

PR #546, opened 2022-10-21
Feat: redirection lambda

kishore03109:
##Problem
Create redirection lambdas, this would create a commit in our redirection server


src/services/infra/InfraService.ts
```diff
@@ -121,6 +125,15 @@ export default class InfraService {
...
   }
 
+  isRootDomain = (primaryDomain: string) => {
+    // method to differentiate root domains with 4th level domains
+    if ((primaryDomain.match(/./g) || []).length < 3) {
+      // eg. blah.gov.sg
+      return true
+    }
+    return false
+  }
```
dcshzj, line 135: This might not be right. Elaborating on my thought process:

We are checking if the given domain is a root domain because we want to make sure that we do not create a CNAME record for a domain that may have other entries (e.g. MX records, SOA record, etc). If a CNAME record exists for a particular domain, all other records may be ignored (using the word "may" here because it is explicitly disallowed in the RFC but implementations seem to differ). [Reference](https://www.freecodecamp.org/news/why-cant-a-domain-s-root-be-a-cname-8cbab38e5f5c/).

Having 2nd level domains is straightforward (e.g. `isomer.sg`), because that is the minimum number you can go, and we know for sure it has to be a root domain. For this case, this check is valid. For a root domain, we should create an A record to ensure compliance with the RFC (side note: The `type` in line 241 may have to be adjusted).

Having 3rd and above level domains is less straightforward (e.g. `isomer.gov.sg` or `bendemeersec.moe.edu.sg`). Based on this check, `isomer.gov.sg` would not be a root domain and a CNAME record would be created. However, this would conflict with the existing MX records present for the domain (which allows `admin@isomer.gov.sg` to exist). The same applies for some schools, like `temasekjc.moe.edu.sg` which is a 4th level domain, but there are MX records defined for it. Creating a CNAME record for this domain may cause the MX records to be ignored, which is the behaviour that the RFC requires.

Basically, this means that it is possible for different domains on the same level to be treated differently. For instance, `www.isomer.gov.sg` and `temasekjc.moe.edu.sg` are both 4th level domains, but `www.isomer.gov.sg` is not a root domain whereas `temasekjc.moe.edu.sg` is a root domain.

I can think of 2 possible solutions:
1. Easier solution: Utilise the inputs given in the site launch form to determine whether to use the A or CNAME record. The primary domain is the one that will use the CNAME record, and the redirection domain is the one that will use the A record plus any additional records like MX record, etc.
2. Tougher solution: Perform a DNS lookup query to see if there are any existing records for the given domain. If there is, then it has to be a root domain (i.e. cannot use CNAME record here, have to use an A record). Otherwise, it is not a root domain and CNAME record is fine.

Sorry for the long post! What do you think about this?

microservices/site-launch/lambda-functions/step-functions-trigger/index.ts
```diff
@@ -1,15 +1,17 @@
 /* eslint-disable import/prefer-default-export */ // todo remove this and use prefer-default-export
 
-import AWS, { Lambda } from "aws-sdk"
-
 import { MessageBody } from "@root/services/identity/QueueService"
+import AWS, { Lambda } from "aws-sdk"
 
 export const stepFunctionsTrigger = async (event: MessageBody) => {
+  // console.log("in step functions trigger")
```
seaerchin, line 7: should probably remove this

microservices/site-launch/lambda-functions/redirection-domain-validation/index.ts
```diff
@@ -0,0 +1,62 @@
...
+        name: "isomeradmin",
+        email: "isomeradmin@open.gov.sg",
+      },
+      content: Buffer.from(template, "binary").toString("base64"),
+      branch: "test/redirectionLambdaTest",
+    }
+  )
+  console.log(
+    `status of redirecion commit for ${primaryDomainSource}:\n ${response}`
+  )
```
seaerchin, line 61: should probably remove this if not required or use `logger.info` if we wanna preserve

microservices/site-launch/lambda-functions/step-functions-trigger/index.ts
```diff
@@ -18,17 +20,17 @@ export const stepFunctionsTrigger = async (event: MessageBody) => {
       input: JSON.stringify(event),
     }
 
-    stepFunctions.startExecution(params, (res, err) => {
-      console.log(`Your state machine ${stateMachineArn} executed successfully`)
+    stepFunctions.startExecution(params, (err, res) => {
       if (err) {
         throw err
       }
+      console.log(`Your state machine ${stateMachineArn} executed successfully`)
```
seaerchin, line 27: if we wanna keep this, we probably wanna change it to `logger.info`!

serverless.yml
```diff
@@ -13,35 +13,45 @@ plugins:
...
 stepFunctions:
   stateMachines:
     siteLaunch:
       definition: ${file(./microservices/site-launch/step-function-workflows/site-launch.asl.yaml)}
+# resources:
+#   Resources:
+#     outgoingQueue:
+#       Type: "AWS::SQS::Queue"
+#       Properties:
+#         QueueName: "outgoingQueueStaging"
```
seaerchin, line 57: could we remove this?

src/services/identity/LaunchesService.ts
```diff
@@ -142,6 +143,7 @@ export class LaunchesService {
       subDomainSettings
     )
 
+    console.log(launchAppOptions)
```
seaerchin, line 146: could we remove this?

src/services/identity/LaunchesService.ts
```diff
@@ -155,6 +157,7 @@ export class LaunchesService {
         logger.info(`Successfully published '${domainAssociation}'`)
         return domainAssociation
       })
+    // .catch((err) => console.log(err))
```
seaerchin, line 160: could we remove this?

src/services/infra/InfraService.ts
```diff
@@ -193,17 +197,18 @@ export default class InfraService {
...
         siteId,
         primaryDomainSource: primaryDomain,
         primaryDomainTarget,
         domainValidationSource,
         domainValidationTarget,
-        redirectionDomainSource,
+      }
+
+      if (redirectionDomainList?.length) {
+        newLaunchParams.redirectionDomainSource = `${redirectionDomainList[0].subDomainSetting?.prefix}.${primaryDomain}`
```
seaerchin, line 211: could i get some context on why we're using the [0]th index element? this would be `www` right? if so, i think it'd be clearer to just have `www.${primaryDomain}`

seaerchin (commented):
minor comments, mostly related to cleaning up

src/services/infra/InfraService.ts
kishore03109: thats fair - 3773fbf

src/services/identity/LaunchesService.ts
kishore03109: 3773fbf

src/services/identity/LaunchesService.ts
kishore03109: 3773fbf

serverless.yml
kishore03109: 3773fbf

microservices/site-launch/lambda-functions/step-functions-trigger/index.ts
kishore03109: 3773fbf

microservices/site-launch/lambda-functions/redirection-domain-validation/index.ts
kishore03109: 3773fbf

microservices/site-launch/lambda-functions/step-functions-trigger/index.ts
kishore03109: 3773fbf

microservices/site-launch/lambda-functions/step-functions-trigger/index.ts
```diff
@@ -8,9 +8,11 @@ import logger from "../../shared/logger"
 export const stepFunctionsTrigger = async (event: MessageBody) => {
   const { AWS_REGION, AWS_ACCOUNT_NUMBER, NODE_ENV } = process.env
   const stepFunctionsParams =
-    NODE_ENV === "LOCAL_DEV" ? { endpoint: "http://localhost:8083" } : {}
+    NODE_ENV === "LOCAL_DEV" ? { endpoint: "http://localhost:8003" } : {}
   try {
-    const stepFunctions = new AWS.StepFunctions(stepFunctionsParams)
+    const stepFunctions = new AWS.StepFunctions({
+      endpoint: "http://localhost:8083",
```
seaerchin, line 14: just a sanity check - si this supposed to be 8083 or 8003?

seaerchin (approved):
lgtm w sanity check

microservices/site-launch/lambda-functions/step-functions-trigger/index.ts
kishore03109: will be removed downstream completely as we migrate to aws infra directly

PR #549, opened 2022-10-31
fix: compute the number of new comments to show

dcshzj:
## Problem

The current dashboard summary does not display a count of the number of new comments that the user has not seen.

## Solution

**Breaking Changes**

**Improvements**:

- The existing site dashboard summary endpoint has been enhanced to use the new list comments feature, which provides a boolean for each comment on whether it is the first time the user is seeing the comment. A count is taken and returned to the user.

## Tests

There are no tests 😭 

## Deploy Notes

*None*

### DIFF
--- a/src/services/review/ReviewRequestService.ts
+++ b/src/services/review/ReviewRequestService.ts
@@ -243,14 +243,16 @@ export default class ReviewRequestService {
         // 3. The review request views table contains a record in the
         //    lastViewedAt entry, and the comment has a timestamp greater
         //    than the one stored in the database.
-        const newComments = await this.getComments(
+        const allComments = await this.getComments(
           sessionData,
           site,
           pullRequestNumber
         )
         const countNewComments = await Promise.all(
-          newComments.map(async (value) => value.isRead)
-        ).then((arr) => arr.filter((value) => value).length)
+          allComments.map(async (value) => value.isRead)
+        ).then(
+          (arr) => arr.filter((isCommentRead) => isCommentRead === true).length
+        )
 
         return {
           id: pullRequestNumber,
--- a/src/services/review/ReviewRequestService.ts
+++ b/src/services/review/ReviewRequestService.ts
@@ -250,9 +250,10 @@ export default class ReviewRequestService {
         )
         const countNewComments = await Promise.all(
           allComments.map(async (value) => value.isRead)
-        ).then(
-          (arr) => arr.filter((isCommentRead) => isCommentRead === true).length
-        )
+        ).then((arr) => {
+          const readComments = arr.filter((isRead) => !!isRead)
+          return readComments.length
+        })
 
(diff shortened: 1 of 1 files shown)


src/services/review/ReviewRequestService.ts
```diff
@@ -234,6 +234,24 @@ export default class ReviewRequestService {
...
+        //    user and review request, but the lastViewedAt entry is NULL.
+        // 3. The review request views table contains a record in the
+        //    lastViewedAt entry, and the comment has a timestamp greater
+        //    than the one stored in the database.
+        const newComments = await this.getComments(
+          sessionData,
+          site,
+          pullRequestNumber
+        )
+        const countNewComments = await Promise.all(
```
seaerchin, line 251: could we rename the variables to `allComments` and `newComments` - the current `newComments` just seems to be retrieving all comments of a PR

src/services/review/ReviewRequestService.ts
```diff
@@ -234,6 +234,24 @@ export default class ReviewRequestService {
...
+          newComments.map(async (value) => value.isRead)
+        ).then((arr) => arr.filter((value) => value).length)
```
seaerchin, line 253: could we rework the `arr.filter((value) => value).length` to be on 2 lines? the nested parenthesis and ordering makes it difficult to understand

src/services/review/ReviewRequestService.ts
dcshzj: Oops, I forgot to adjust this name, done in db3dce3.

src/services/review/ReviewRequestService.ts
dcshzj: Reworked in db3dce3.

src/services/review/ReviewRequestService.ts
```diff
@@ -234,6 +234,26 @@ export default class ReviewRequestService {
...
+        //    than the one stored in the database.
+        const allComments = await this.getComments(
+          sessionData,
+          site,
+          pullRequestNumber
+        )
+        const countNewComments = await Promise.all(
+          allComments.map(async (value) => value.isRead)
+        ).then(
+          (arr) => arr.filter((isCommentRead) => isCommentRead === true).length
```
seaerchin, line 254: sorry to be a pedant but wdyt about this?

```ts
(arr) => {
const readComments = arr.filter(isRead => !!isRead) 
return readComments.length
}
```

seaerchin (commented):
comment for readability

src/services/review/ReviewRequestService.ts
dcshzj: Ah I think I misunderstood your original intent, adjusted in e315e2a.

seaerchin (approved):
lgtm

PR #550, opened 2022-10-31
feat: add endpoint to mark a review request as viewed

dcshzj:
## Problem

Currently, there is no specific endpoint for marking a review request as viewed. This would mean that a user that accesses a review request directly using the URL and goes into the site dashboard will incorrectly see the review request as new.

## Solution

**Breaking Changes**

**Features**:

- A new API POST endpoint has been added to allow marking a particular review request as viewed.

## Tests

There are no tests 😭 

## Deploy Notes

*None*

### DIFF
--- a/src/services/review/ReviewRequestService.ts
+++ b/src/services/review/ReviewRequestService.ts
@@ -285,18 +285,12 @@ export default class ReviewRequestService {
     const { isomerUserId: userId } = sessionData
     const { id: reviewRequestId } = reviewRequest
 
-    await this.reviewRequestView.update(
-      {
-        lastViewedAt: new Date(),
-      },
-      {
-        where: {
-          reviewRequestId,
-          siteId: site.id,
-          userId,
-        },
-      }
-    )
+    await this.reviewRequestView.upsert({
+      reviewRequestId,
+      siteId: site.id,
+      userId,
+      lastViewedAt: new Date(),
+    })
   }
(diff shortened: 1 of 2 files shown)


src/services/review/ReviewRequestService.ts
```diff
@@ -299,6 +299,35 @@ export default class ReviewRequestService {
...
   }
 
+  markReviewRequestAsViewed = async (
+    sessionData: UserWithSiteSessionData,
+    site: Site,
+    requestId: number
+  ): Promise<void> => {
+    const { isomerUserId: userId } = sessionData
+
+    const reviewRequestView = await this.reviewRequestView.findOne({
```
seaerchin, line 309: nit: could we try doing an `upsert` instead? see [here](https://sequelize.org/docs/v6/other-topics/upgrade/#modelupsert) for more details.

seaerchin (approved):
lgtm, minor nit about `upsert` rather than update, check and create

src/services/review/ReviewRequestService.ts
dcshzj: Hmm I think it might not be possible, because we only want to insert if it is not found. We do not want to perform any updates to the existing entry.

However, I think using `upsert` would be very useful for `updateReviewRequestLastViewedAt`, updated that in 927116d.

PR #552, opened 2022-11-01
fix: use review request requestor email for production commit email

dcshzj:
## Problem

Currently, the email address of the user that published the site shows the email address stored in the GitHub commit to the master branch. However, as we are using a common access token to interface with GitHub, this email address will always be the same.

## Solution

**Breaking Changes**

**Bug Fixes**:

- The email address of the user that last published the site is now obtained from the review requests table, provided that the commit starts with `isomergithub`, which is the user accounts for Isomer's common access tokens.

## Tests

There are no tests 😭 

## Deploy Notes

*None*

### DIFF
--- a/src/services/identity/SitesService.ts
+++ b/src/services/identity/SitesService.ts
@@ -11,6 +11,7 @@ import {
   ISOMER_ADMIN_REPOS,
 } from "@root/constants"
 import { NotFoundError } from "@root/errors/NotFoundError"
+import RequestNotFoundError from "@root/errors/RequestNotFoundError"
 import { UnprocessableError } from "@root/errors/UnprocessableError"
 import { genericGitHubAxiosInstance } from "@root/services/api/AxiosInstance"
 import { GitHubCommitData } from "@root/types/commitData"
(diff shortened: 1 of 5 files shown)


src/services/identity/SitesService.ts
```diff
@@ -175,27 +189,49 @@ class SitesService {
...
+      return this.extractAuthorEmail(commit)
+    }
 
     // Commit was made by our common identity GitHub user
-    if (authorName.startsWith("isomergithub")) {
-      // TODO: Retrieve email address of user from review request table
+    const site = await this.getBySiteName(siteName)
+    if (!site) {
+      return this.extractAuthorEmail(commit)
+    }
```
seaerchin, line 213: could you go into more detail here? it seems like this conditional is dealing with sites that are not onboarded onto identity right?

src/services/identity/SitesService.ts
```diff
@@ -175,27 +189,49 @@ class SitesService {
...
+  ) {
     const {
       author: { name: authorName },
     } = commit
+    const { siteName } = sessionData
+
+    if (!authorName.startsWith("isomergithub")) {
+      // Legacy style of commits, or if the user is not found
+      return this.extractAuthorEmail(commit)
+    }
```
seaerchin, line 207: should we attempt to parse the message rather than checking the prefix of the author name?

seaerchin (approved):

src/services/identity/SitesService.ts
dcshzj: Yup, I reorganised this part of the code to avoid an arrowhead style of code. But basically, if there are any issues that are encountered along the way, we will fall back to retrieving the email address from the commit itself (i.e. the email address of the git commit author).

src/services/identity/SitesService.ts
dcshzj: Hmm in this case, there would not be a message. It is likely that it might be something like "Merged pull request #... from staging". So we would want to extract the email address of the commit author instead.

PR #554, opened 2022-11-02
fix: include full model in order by due to presence of alias

dcshzj:
## Problem

The order by syntax is incorrect for getting the latest merged review request, which has a cascading effect of causing the whole site info endpoint to not work.

## Solution

**Breaking Changes**

**Bug Fixes**:

- The full ReviewMeta model is now included in the order attribute, as suggested by [this StackOverflow answer](https://stackoverflow.com/a/58524200)

## Tests

There are no tests.

## Deploy Notes

*None*

### DIFF
--- a/src/services/review/ReviewRequestService.ts
+++ b/src/services/review/ReviewRequestService.ts
@@ -443,7 +443,16 @@ export default class ReviewRequestService {
           model: Site,
         },
       ],
-      order: [[ReviewMeta, "pullRequestNumber", "DESC"]],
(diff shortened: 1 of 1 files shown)


seaerchin (approved):

seaerchin:
did we have tests for this endpoint ah

dcshzj:
> did we have tests for this endpoint ah

Nope, the whole ReviewRequestService lacks tests currently :(

PR #559, opened 2022-11-07
Feat/site launch microservice

kishore03109:
todo: add desc


README.md
```diff
@@ -56,3 +56,11 @@ Finally, we want to run the migration script.
 - Open another terminal window.
 - Run `source .env`
 - Run `npx sequelize-cli db:migrate`
+
+## Lambda Functions and Step Function Workflows
+
+### Running locally
+
+1. Run `npm run dev:services` to spin up the local [AWS Step Functions emulator](https://docs.aws.amazon.com/step-functions/latest/dg/sfn-local-docker.html) provided by AWS.
```
dcshzj, line 64: Just checking: Does this docker image need to be added to the `docker-compose.yml` file?

serverless.yml
```diff
@@ -0,0 +1,39 @@
...
+  runtime: nodejs14.x
+
+custom:
+  stepFunctionsLocal:
+    accountId: 101010101010
+    region: us-east-1
+    lambdaEndpoint: http://localhost:3002
+    TaskResourceMapping:
+      Hello: arn:aws:lambda:us-east-1:101010101010:function:hello
+      Hithere: arn:aws:lambda:us-east-1:101010101010:function:hithere
```
dcshzj, line 24: I'm probably not understanding this but does this need to match the one in the `step-functions` folder? The region is different.

Also, are we able to reuse the definitions in line 19 and 20? Something similar to `${self:accountId}`, not sure if this is possible?

serverless.yml
```diff
@@ -0,0 +1,39 @@
...
+    eventBridgeEvents:
+      enabled: true
+      endpoint: http://localhost:4010
+  sqsUrl: http://localhost:4566/000000000000/outgoing-queue-dev
+
+functions:
+  hello:
+    handler: handler.hello
```
dcshzj, line 32: Wanted to clarify: Do we need the `handler.` prefix for this?

dcshzj (changes requested):

serverless.yml
kishore03109: Hey - these are 'fake' functions that were created. `Hello` and `Hithere` were just functions we created to test the functionality of local stack. These are to be removed in the later commits using the variables that you have described!

README.md
kishore03109: yes, surprised isnt added here but it is added in the [future PRs](https://github.com/isomerpages/isomercms-backend/blob/feat/integrate-aws-with-site-launch-microservice/docker-compose.yml)!

serverless.yml
kishore03109: yup! it just dictates where the logic for the lambda function resides. ie  what the lambda function should do when invoked!

dcshzj (approved):
I _think_ this is ok, no actual impact on production. From what I see most of it were also improved in subsequent PRs.

PR #562, opened 2022-11-07
Feat/site launch db

kishore03109:
## Problem

DB models for site launch 

##Deploy notes

need to run DB migrations


src/database/migrations/20220920054800-create-launches.js
```diff
@@ -0,0 +1,47 @@
...
+      },
+      created_at: {
+        allowNull: false,
+        type: Sequelize.DATE,
+      },
+      updated_at: {
+        allowNull: false,
+        type: Sequelize.DATE,
+      },
+    })
```
dcshzj, line 42: I think this is missing a `deleted_at`? The `paranoid` flag is set to `true` in the model.

src/database/migrations/20220920084207-create-redirects.js
```diff
@@ -0,0 +1,39 @@
...
+      },
+      created_at: {
+        allowNull: false,
+        type: Sequelize.DATE,
+      },
+      updated_at: {
+        allowNull: false,
+        type: Sequelize.DATE,
+      },
+    })
```
dcshzj, line 34: Same as above on missing `deleted_at`.

src/database/migrations/20220920054800-create-launches.js
```diff
@@ -0,0 +1,47 @@
...
+        type: Sequelize.INTEGER,
+      },
+      user_id: {
+        allowNull: false,
+        type: Sequelize.INTEGER,
+      },
+      site_id: {
+        type: Sequelize.INTEGER,
+        allowNull: false,
+      },
```
dcshzj, line 17: I think we might need to add a `references` key for both the `user_id` and `site_id` to indicate the foreign key reference to the users and sites table.

src/database/migrations/20220920084207-create-redirects.js
```diff
@@ -0,0 +1,39 @@
...
+        type: Sequelize.INTEGER,
+      },
+      launch_id: {
+        allowNull: false,
+        type: Sequelize.INTEGER,
+      },
+      type: {
+        allowNull: false,
+        type: Sequelize.STRING,
+      },
```
dcshzj, line 17: I think maybe we should encode the ENUM into the database as well.

src/database/migrations/20220920084207-create-redirects.js
```diff
@@ -0,0 +1,39 @@
...
+      id: {
+        allowNull: false,
+        autoIncrement: true,
+        primaryKey: true,
+        type: Sequelize.INTEGER,
+      },
+      launch_id: {
+        allowNull: false,
+        type: Sequelize.INTEGER,
+      },
```
dcshzj, line 13: I think this should also have the `references` key as well, same as above.

src/database/migrations/20220920084207-create-redirects.js
```diff
@@ -0,0 +1,39 @@
...
+        type: Sequelize.STRING,
+      },
+      created_at: {
+        allowNull: false,
+        type: Sequelize.DATE,
+      },
+      updated_at: {
+        allowNull: false,
+        type: Sequelize.DATE,
+      },
```
dcshzj, line 33: Minor nit: I think there might be some value in adding a `defaultValue: Sequelize.fn("NOW")` so that we can automatically populate these columns when creating a new entry.

dcshzj (changes requested):
I haven't checked the other PRs to see if this was done, but I think we should also add the export to `models/index.ts`, plus add the model in `src/tests/database.ts`.

src/database/migrations/20220920054800-create-launches.js
dcshzj: Whoops, this was done in #558, can ignore!

src/database/migrations/20220920084207-create-redirects.js
dcshzj: Done in #558, can ignore!

src/database/migrations/20220920084207-create-redirects.js
dcshzj: Done in #558, can ignore!

src/database/migrations/20220920054800-create-launches.js
```diff
@@ -0,0 +1,47 @@
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    await queryInterface.createTable("launches", {
+      id: {
+        allowNull: false,
+        autoIncrement: true,
+        primaryKey: true,
+        type: Sequelize.INTEGER,
```
seaerchin, line 8: not a nit, just a side-track: turns out there's a diff between `INTEGER` and `BIGINT`! 

from seq docs: 
> The JavaScript [number](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number) type can represent ints ranging from [-9007199254740991](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MIN_SAFE_INTEGER) to [9007199254740991](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER).
>
> If your SQL type supports integer values outside this range, we recommend using [bigint](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) or [string](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String) to represent your integers.

don't have to change cos we'll never have that many users anyway... (hopefully i'm wrong HAHAHAH)

src/database/migrations/20220920054800-create-launches.js
```diff
@@ -0,0 +1,47 @@
...
+        type: Sequelize.STRING,
+      },
+      domain_validation_source: {
+        allowNull: false,
+        type: Sequelize.STRING,
+      },
+      domain_validation_target: {
+        allowNull: false,
+        type: Sequelize.STRING,
+      },
```
seaerchin, line 33: i don't think i have enough context here! could i get your help in explaining what `domain_validation_<source | target>` means?

i noticed your sequence diagram in the design doc has a domain validation lambda, but i couldn't match source/target to what the lambda does as the lambda should only require the agency's desired domain, no?

src/database/migrations/20220920054800-create-launches.js
```diff
@@ -0,0 +1,47 @@
...
+        allowNull: false,
+      },
+      primary_domain_source: {
+        allowNull: false,
+        type: Sequelize.STRING,
+      },
+      primary_domain_target: {
+        allowNull: false,
+        type: Sequelize.STRING,
+      },
```
seaerchin, line 25: correct me if i'm wrong but do source/target here refer respectively to the `<some_random_letters>.amplify.com` and `<agency_domain>.com` respectively?

src/database/models/Launches.ts
```diff
@@ -0,0 +1,61 @@
...
+  })
+  id!: number
+
+  @ForeignKey(() => User)
+  @Column
+  userId!: string
+
+  @ForeignKey(() => Site)
+  @Column
+  siteId!: string
```
seaerchin, line 30: should this be `BIGINT` or `INTEGER`?

seaerchin, line 30: hmm, if we want to set this up so that sequelize retrieves the associated model, we might have to use [model association](https://github.com/sequelize/sequelize-typescript#model-association)

src/database/models/Redirections.ts
```diff
@@ -0,0 +1,52 @@
...
+    autoIncrement: true,
+    primaryKey: true,
+    allowNull: false,
+    type: DataType.BIGINT,
+  })
+  id!: number
+
+  @ForeignKey(() => Launches)
+  @Column
+  launchId!: number
```
seaerchin, line 27: see above comment wrt [model associations](https://github.com/sequelize/sequelize-typescript#model-association)

src/database/migrations/20220920084207-create-redirects.js
kishore03109: 41c0ea3

src/database/migrations/20220920084207-create-redirects.js
kishore03109: fixed: 6721422

src/database/migrations/20220920054800-create-launches.js
kishore03109: fixed: 6721422

src/database/migrations/20220920054800-create-launches.js
kishore03109: sure! for every record (A, CNAME) that we store, we would have a source and a target. For clarity, in our system, we have 
www.blah.gov.sg CNAME gibberish.acm.aws

As such, our 
`domain_validation_source` = `www.blah.gov.sg` 
`domain_validation_target` = `gibberish.acm.aws`

For both the **_primary domains_** and the _**domain validations**_, we dont have to store the _type_ of record as it is guaranteed to be a CNAME record based on our architecture. 
The lambdas check to see if such a record has propagated

src/database/migrations/20220920054800-create-launches.js
kishore03109: for clarity:
source: `<agency_domain>.com`
target: `<some_random_letters>.amplify.com`

src/database/models/Launches.ts
kishore03109: good catch! 0b807bd

kishore03109: @seaerchin hmm, wouldnt the `@ForeignKey(() => User)` already allow us to associate a relationship between launches and users? Exactly what am I missing ya?

src/database/migrations/20220920054800-create-launches.js
```diff
@@ -0,0 +1,57 @@
...
+        references: {
+          model: "users",
+          key: "id",
+        },
+        allowNull: false,
+        type: Sequelize.INTEGER,
+      },
+      site_id: {
+        references: {
+          model: "users",
```
dcshzj, line 20: I think this should be `sites`, not `users`?

dcshzj (changes requested):

src/database/migrations/20220920054800-create-launches.js
kishore03109: 25e5c88 oh dear, thanks for catching this!

dcshzj (approved):
LGTM, but this is contingent on #558 being merged due to some comments that were addressed there.

src/database/models/Launches.ts
seaerchin: hmm i'm actually not sure here! i can't recall off the top of my head if it does so could i trouble you to test this?

src/database/migrations/20220920054800-create-launches.js
seaerchin: yeap it does! why'd we need to store the `source` if we're pinging amplify for the status? is it just to make sure that the target -> source resolution is working?

src/database/migrations/20220920054800-create-launches.js
seaerchin: thanks for clarifying!

src/database/models/Launches.ts
kishore03109: b806ef7

src/database/migrations/20220920054800-create-launches.js
kishore03109: Arh - good question! This is because Amplify replies to us if the website is read to serve, but this does not guarantee that there is indeed a source -> target CNAME record that has actually propagated.

src/database/models/Redirections.ts
dcshzj: Wait no hmm, it should be separate, something like this:

```ts
@ForeignKey(() => Launches)
@Column
launchId!: number

@BelongsTo(() => Launches, {
  onUpdate: "CASCADE",
  onDelete: "CASCADE",
})
launch!: Launches
```

Side note: Just realised we should also use singular for the models, but plural for the actual database name.

kishore03109: cc438e2

dcshzj: Looks good now, thanks!

src/database/models/Launches.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  @ForeignKey(() => Site)
+  @Column
+  siteId!: number
+
+  @BelongsTo(() => Site, {
+    onUpdate: "CASCADE",
+    onDelete: "CASCADE",
+  })
+  site!: Site
```
seaerchin, line 43: i think doing this only exposes the model on the source (`Launches`), which is fine if our target model (`Site`) wouldn't need to know about it (see [here](https://sequelize.org/docs/v6/core-concepts/assocs/#why-associations-are-defined-in-pairs) for details). 

otherwise, wdyt about adding it on the target model too?

src/database/migrations/20220920054800-create-launches.js
seaerchin: ah ok so this is used later on in the services for verification of the source -> target CNAME right? if so, please feel free to mark this as resolved and thanks for answering!

src/database/models/Launches.ts
kishore03109: 8c4b838

seaerchin: would it also be useful to associate a user with their associated launches?

kishore03109: yup -  d5789ce

seaerchin (approved):

PR #566, opened 2022-11-09
tests: add remaining tests for identity features in SitesService

dcshzj:
## Problem

The new identity features do not have sufficient unit test coverage in SitesService, which can be dangerous.

## Solution

**Breaking Changes**

**Improvements**:

- The remaining unit tests for SitesService has been added for the new identity features.
- The SiteUrl type has been moved into the repoInfo.ts types file, so that it can also be used in the unit tests.

## Tests

**Unit tests**: Run `npm run test` on this branch to verify that all the new unit tests pass

## Deploy Notes

*None*

### DIFF
--- a/src/services/identity/__tests__/SitesService.spec.ts
+++ b/src/services/identity/__tests__/SitesService.spec.ts
@@ -84,6 +84,10 @@ const SitesService = new _SitesService({
   reviewRequestService: (MockReviewRequestService as unknown) as ReviewRequestService,
 })
 
+const SpySitesService = {
+  extractAuthorEmail: jest.spyOn(SitesService, "extractAuthorEmail"),
+}
+
 const mockSiteName = "some site name"
 const mockSite = ({
   name: "i m a site",
@@ -587,6 +591,7 @@ describe("SitesService", () => {
       // Assert
       expect(actual).toBe(expected)
       expect(MockUsersService.findById).toHaveBeenCalledWith(mockIsomerUserId)
+      expect(SpySitesService.extractAuthorEmail).not.toHaveBeenCalled()
     })
 
     it("should return the email of the commit author who is a GitHub login user", async () => {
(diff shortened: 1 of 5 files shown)


src/services/identity/__tests__/SitesService.spec.ts
```diff
@@ -149,26 +610,176 @@ describe("SitesService", () => {
...
+      }
+
+      // Act
+      const actual = await SitesService.getMergeAuthorEmail(
+        commit,
+        mockUserWithSiteSessionData
+      )
+
+      // Assert
+      expect(actual).toEqual(expected)
```
seaerchin, line 633: should we assert that `extractAuthorEmail` was called with the commit? (similarly for below tests)

seaerchin (approved):

src/services/identity/__tests__/SitesService.spec.ts
dcshzj: Wait is that possible? My understanding was that the `expect().toHaveBeenCalled()` only works for a mocked function, but `extractAuthorEmail` is an actual function that is undergoing testing as part of this test file.

seaerchin: yeap, what we need to do here is to construct a [spy](https://jestjs.io/docs/jest-object#jestspyonobject-methodname). this would then allow us to call assertions on the spy transparently.

seaerchin: yeap, what we need to do is to construct a [spy](https://jestjs.io/docs/jest-object#jestspyonobject-methodname), which we can then use to make assertions on transparently

dcshzj: TIL, thanks for sharing! Added in c333d8c.

PR #568, opened 2022-11-11
chore(collaborators): update typings

seaerchin:
## Problem
The FE had outdated typings for collaborators, which was fixed by the PR [here](https://github.com/isomerpages/isomercms-frontend/pull/1152); this PR adds the relevant properties to the return value for the `/collaborators` route 

## Solution
1. use `Jsonify` type to indicate properties
2. omit `SiteMembers` model and parse it to `role`

### DIFF
--- a/src/types/dto/review.ts
+++ b/src/types/dto/review.ts
@@ -1,4 +1,8 @@
+import { Jsonify } from "type-fest"
+
 import { CollaboratorRoles } from "@constants/constants"
+
+import { User } from "@root/database/models"
 
 export type ReviewRequestStatus = "OPEN" | "APPROVED" | "MERGED" | "CLOSED"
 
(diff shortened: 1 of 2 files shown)


src/types/dto/review.ts
```diff
@@ -13,7 +17,7 @@ export interface EditedItemDto {
   lastEditedTime: number
 }
 
-export interface UserDto {
+export interface UserDto extends Jsonify<User> {
   email: string
   role: CollaboratorRoles
 }
```
alexanderleegs, line 23: `UserDto` still doesn't seem to recognise `User` fields - is this the intended behaviour?

seaerchin: double checking - what do you mean by this

seaerchin: resolved offline - meant that intellisense wasn't showing the properties. 

as to why this happens, i suspect this is because of the `class` keyword. using `ConditionalExcept<User, Function>` will extract the properties but then we'd need to hint the type of `toJSON` properly also.

went with just specifying props manually as it seemed like less effort ._.

alexanderleegs (approved):
lgtm

Issue #537, opened 2022-10-19
[TESTS] Tests for review request


seaerchin:
## Problem
Review requests have no tests atm and this is suspicious behaviour. 

## Suggested solution
1. add unit tests for service 
2. add integration tests for routes
3. (maybe try adding tests on models?)

PR #569, opened 2022-11-15
tests: add unit tests for ReviewRequestService

dcshzj:
## Problem

Part of #537.

There are currently no unit test coverage for the ReviewRequestService, which can be dangerous as there is no guarantee on the behaviour of individual functions.

## Solution

**Breaking Changes**

**Improvements**:

- Unit tests has been added for ReviewRequestService
- A bug has also been identified and fixed where the listReviewRequest is returning the number of **read** comments, but should instead display the number of **unread** comments

## Tests

## Deploy Notes

*None*


src/services/review/__tests__/ReviewRequestService.spec.ts
```diff
@@ -0,0 +1,1151 @@
...
+      )
+      expect(MockReviewApi.getComments).toHaveBeenCalledWith(
+        mockUserWithSiteSessionData.siteName,
+        MOCK_REVIEW_REQUEST_ONE.reviewMeta.pullRequestNumber
+      )
+      expect(SpyReviewRequestService.computeCommentData).toHaveBeenCalled()
+      expect(MockUsersRepository.findByPk).toHaveBeenCalledTimes(2)
+    })
+
+    it("should return an array of basic review request objects with one new comment", async () => {
```
seaerchin, line 566: could i ask why the `one new comment` is required here

seaerchin (approved):
minor comment

src/services/review/__tests__/ReviewRequestService.spec.ts
dcshzj: Adjusted in a705ca2.

PR #572, opened 2022-11-16
Feat: add generic notification handler tests

alexanderleegs:
This PR adds tests for the notification middleware introduced in #514.

Of special note is the format of the router setup for the test - our actual setup involves invoking the notification creation method only after the response has been sent, but jest ends the test once the response has been received. As such, the test setup involves creating a custom wrapper over the existing notification creation handler to add in the response sent.

### DIFF
--- a/src/integration/NotificationOnEditHandler.spec.ts
+++ b/src/integration/NotificationOnEditHandler.spec.ts
@@ -13,7 +13,7 @@ import {
   User,
   Whitelist,
 } from "@database/models"
-import { generateRouter } from "@fixtures/app"
+import { generateRouterForUserWithSite } from "@fixtures/app"
 import UserSessionData from "@root/classes/UserSessionData"
 import {
   mockEmail,
@@ -71,17 +71,6 @@ const notificationsHandler = new NotificationOnEditHandler({
 
 // Set up express with defaults and use the router under test
 const subrouter = express()
-// As we set certain properties on res.locals when the user signs in using github
-// In order to do integration testing, we must expose a middleware
-// that allows us to set this properties also
-subrouter.use((req, res, next) => {
-  const userSessionData = new UserSessionData({
-    isomerUserId: mockIsomerUserId,
-    email: mockEmail,
-  })
-  res.locals.userSessionData = userSessionData
-  next()
-})
 const subSubrouter = express()
 subSubrouter.get("/:siteName/test", async (req, res, next) =>
   // Dummy subrouter
@@ -96,7 +85,15 @@ subrouter.use(async (req, res, next) => {
   await notificationsHandler.createNotification(req as any, res as any, next)
   res.status(200).send(200)
 })
-const app = generateRouter(subrouter)
+const userSessionData = new UserSessionData({
+  isomerUserId: mockIsomerUserId,
+  email: mockEmail,
+})
+const app = generateRouterForUserWithSite(
+  subrouter,
+  userSessionData,
+  mockSiteName
+)
 
 describe("Notifications Router", () => {
   const mockAdditionalUserId = "2"
@@ -164,6 +161,14 @@ describe("Notifications Router", () => {
       siteId: mockAdditionalSiteId,
     })
   })
+
+  afterAll(async () => {
+    await SiteMember.sync({ force: true })
+    await Site.sync({ force: true })
+    await User.sync({ force: true })
+    await Repo.sync({ force: true })
+  })
+
   describe("createNotification handler", () => {
     afterEach(async () => {
       // Clean up so that different tests using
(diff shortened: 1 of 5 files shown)


kishore03109 (changes requested):
Hey, not sure if you can replicate this, but the tests fail in my local machine :( 
Screenshot 2023-01-04 at 3 00 01 PM

alexanderleegs:
> Hey, not sure if you can replicate this, but the tests fail in my local machine :( Screenshot 2023-01-04 at 3 00 01 PM

Was able to replicate this, and fixed in https://github.com/isomerpages/isomercms-backend/pull/572/commits/893d255f0381e849b23be6e97f8575024c57816c! The issue was because of the unique constraint on the `Site` table - this was causing the `add` method to fail

src/integration/NotificationOnEditHandler.spec.ts
```diff
@@ -0,0 +1,221 @@
...
+import UserSessionData from "@root/classes/UserSessionData"
+import {
+  mockEmail,
+  mockIsomerUserId,
+  mockSiteName,
+} from "@root/fixtures/sessionData"
+import { NotificationOnEditHandler } from "@root/middleware/notificationOnEditHandler"
+import { GitHubService } from "@root/services/db/GitHubService"
+import { ConfigYmlService } from "@root/services/fileServices/YmlFileServices/ConfigYmlService"
+import CollaboratorsService from "@root/services/identity/CollaboratorsService"
```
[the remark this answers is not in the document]
alexanderleegs: Changed in https://github.com/isomerpages/isomercms-backend/pull/572/commits/a312155119e01da5fa108c602ea02295baa1a2cf

src/integration/NotificationOnEditHandler.spec.ts
```diff
@@ -0,0 +1,241 @@
...
+// Set up express with defaults and use the router under test
+const subrouter = express()
+const subSubrouter = express()
+subSubrouter.get("/:siteName/test", async (req, res, next) =>
+  // Dummy subrouter
+  next()
+)
+subrouter.use(subSubrouter)
+
+// We handle the test slightly diferently - jest interprets the end of the test as when the response is sent,
```
kishore03109, line 84:
```suggestion
// We handle the test slightly differently - jest interprets the end of the test as when the response is sent,
```

src/integration/NotificationOnEditHandler.spec.ts
```diff
@@ -0,0 +1,241 @@
...
+// but we normally create a notification after this response, due to the position of the middleware
+// the solution to get tests working is to send a response only after the notification middleware
+subrouter.use(async (req, res, next) => {
+  await notificationsHandler.createNotification(req as any, res as any, next)
```
kishore03109, line 88: nit: await has no effect

src/integration/NotificationOnEditHandler.spec.ts
```diff
@@ -0,0 +1,241 @@
...
+const notificationsHandler = new NotificationOnEditHandler({
+  reviewRequestService,
+  collaboratorsService,
+  sitesService,
+  notificationsService,
+})
+
+// Set up express with defaults and use the router under test
+const subrouter = express()
+const subSubrouter = express()
```
kishore03109, line 77: Hey, is the creation of two routers intended?

kishore03109 (changes requested):

src/integration/NotificationOnEditHandler.spec.ts
alexanderleegs: Yup - this is to simulate how this notification handler functions as closely as possible. The notification handler (represented by `subSubrouter`) is inserted into our `authenticatedSites` subrouter (represented by `subrouter` here), placed after any regular routes (represented by the `/:siteName/test` endpoint here) - this simulates our existing setup as closely as possible.

kishore03109: haha could we rename this to `notificationHandlerRouter` and `authenticatedSitesRouter` instead?

alexanderleegs: Oops correction on the above explanation - `subSubrouter` represents a router that the request is routed to and not the notification handler - the notification handler is just inserted in afterwards. As discussed offline - the names of the routers have been changed to make this clearer!

src/integration/NotificationOnEditHandler.spec.ts
alexanderleegs: Hmm it seems to be required in the test, probably because jest seems to exit once the status code is received - I've left a comment explaining this in d6c1b3ea

kishore03109 (approved):

PR #574, opened 2022-11-17
ref(fixtures): allow generating router for different user types

dcshzj:
## Problem

The current test fixtures do not take into account different possible user types, as it is fixed to be only for a single mock user.

## Solution

**Breaking Changes**

**Improvements**:

- The fixtures have been adjusted to allow a UserSessionData/UserWithSiteSessionData object to be specified when generating a router. This change lays the foundation for future tests to be able to test routes using different personas.
- The `app.js` file has been ported over to TypeScript and the old router generation method has been deprecated.
- Utility methods for adding and removing the site in the session data has been added.

## Tests

**Unit tests**: Run `npm run test` (existing tests should not break)

## Deploy Notes

*None*


src/fixtures/app.ts
```diff
@@ -0,0 +1,118 @@
...
+    userSessionData: UserSessionData
+    userWithSiteSessionData: UserWithSiteSessionData
+    githubSessionData: GithubSessionData
+  }
+> = (req, res, next) => {
+  res.locals.userSessionData = mockUserSessionData
+  res.locals.userWithSiteSessionData = mockUserWithSiteSessionData
+  res.locals.githubSessionData = mockGithubSessionData
+  next()
+}
```
alexanderleegs, line 37: Just a thought - I think it might still be useful to keep this around as a generic option for router tests that don't need to test different user profiles.

As discussed offline, we might consider making one with `SCREAMING_SNAKE_CASE` (to indicate that these are constants) for future use and deprecate this one

alexanderleegs (approved):
lgtm, minor comment!

src/fixtures/app.ts
dcshzj: Added in 960bd2e!

PR #575, opened 2022-11-17
feat(rr): blob diffs

seaerchin:
## Problem
We don't have a way at present to retrieve raw file contents from github. this PR adds that ability in.

## Solution
1. add route + service.

**notes**
as mentioned in the fe PR, this is keyed by `:reviewId`, which might be problematic in the future because the rr is an artificial construct and what we really want is a point in time (by the commit hash) of the file contents...

### DIFF
--- a/src/routes/v2/authenticated/review.ts
+++ b/src/routes/v2/authenticated/review.ts
@@ -1042,6 +1042,9 @@ export class ReviewsRouter {
       })
     }
 
+    // NOTE: Currently, Isomer only allows comparisons between staging and production.
+    // This might change in the future and in that case, the `getBlob` method call below
+    // should have the corresponding ref (`master` or `staging`) changed.
     const prodPromise = await this.reviewRequestService.getBlob(
       siteName,
       path,
--- a/src/routes/v2/authenticated/review.ts
+++ b/src/routes/v2/authenticated/review.ts
@@ -24,6 +24,7 @@ import {
   EditedItemDto,
   UpdateReviewRequestDto,
   ReviewRequestDto,
+  BlobDiffDto,
 } from "@root/types/dto/review"
 import ReviewRequestService from "@services/review/ReviewRequestService"
 // eslint-disable-next-line import/prefer-default-export
@@ -947,7 +948,8 @@ export class ReviewsRouter {
 
     if (!isReviewer) {
       logger.error({
-        message: "",
+        message:
+          "User with insufficient permissions attempted to delete approval",
         method: "deleteReviewRequestApproval",
         meta: {
           userId: userWithSiteSessionData.isomerUserId,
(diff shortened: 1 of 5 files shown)


alexanderleegs (commented):
Minor comment, but otherwise lgtm

alexanderleegs (approved):

src/services/db/review.ts
```diff
@@ -129,3 +129,16 @@ export const createComment = async (
...
+
+export const getBlob = async (
+  repo: string,
+  path: string,
+  ref: string
+): Promise<string> =>
+  axiosInstance
+    .get<string>(`${repo}/contents/${path}?ref=${ref}`, {
+      headers: {
+        Accept: "application/vnd.github.VERSION.raw",
```
dcshzj, line 141: Should we remove the `VERSION` part of this header? I was not able to find it [in the docs](https://docs.github.com/en/rest/overview/media-types?apiVersion=2022-11-28).

src/types/dto/review.ts
```diff
@@ -58,3 +58,8 @@ export interface GithubCommentData {
   message: string
   createdAt: string
 }
+
+export interface BlobDiffDto {
+  old: string
+  new: string
```
dcshzj, line 64: Hmm I think we should avoid using `new` as the key as it is reserved.

src/routes/v2/authenticated/review.ts
```diff
@@ -967,6 +969,101 @@ export class ReviewsRouter {
...
+      path,
+      "master"
+    )
+    const stagingPromise = await this.reviewRequestService.getBlob(
+      siteName,
+      path,
+      "staging"
+    )
+
+    const data = await Promise.all([prodPromise, stagingPromise])
```
dcshzj, line 1059: Can I check: My understanding is that `prodPromise` and `stagingPromise` already has an `await`, which means that by this step, both values should have been resolved. Is this line still required?

dcshzj (changes requested):

src/services/db/review.ts
seaerchin: huh that's weird - you're right, idk where i saw this in the docs but since it's not in the latest one, i'll remove it. thanks for checking!

src/types/dto/review.ts
seaerchin: yeap, thanks for pointing this out; slipped my mind and ts didn't show this to be an error @__@ will change to `oldValue/newValue`!

src/routes/v2/authenticated/review.ts
seaerchin: ah yes - i missed this. should've removed the `await` from earlier on, thanks for catching!

dcshzj (approved):
LGTM for now! There are a number of improvements raised in the frontend PR but that can be done in a separate PR.

PR #577, opened 2022-11-23
Fix/reference sitename from repo

alexanderleegs:
## Problem

This PR fixes an issue with our existing implementation of CollaboratorsService and SitesService - we were attempting to search for entries matching the `name` param inside a `Site` entry. However, this is actually meant to be the human readable title of the repo (e.g. `Example Repo`), while the value we should be referencing is the title of the github repo stored in the `name` param of a `Repo` entry (e.g. `example-repo`).

We should also fix the entries in our local db to bring our local test db closer to our prod db and ensure this issue can be caught more easily in future.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "start": "node -r ts-node/register/transpile-only -r tsconfig-paths/register build/server.js",
     "dev:services": "docker compose up -d",
     "dev": "source .env && ts-node-dev --respawn src/server.js",
-    "test": "source .env.test && jest",
+    "test": "source .env.test && jest --runInBand",
     "release": "npm version $npm_config_isomer_update && git push --tags",
     "lint": "npx eslint .",
     "lint-fix": "eslint --ignore-path .gitignore . --fix",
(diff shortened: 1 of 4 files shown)


dcshzj (approved):
LGTM! Minor nit: The `required: true` part can be removed since sequelize automatically adds that if the `where` clause is present in the `include` ([reference](https://sequelize.org/docs/v6/advanced-association-concepts/eager-loading/#eager-loading-filtered-at-the-associated-model-level)):

> This is the case because, when the `where` option is used inside an `include`, Sequelize automatically sets the `required` option to `true`.

alexanderleegs:
> LGTM! Minor nit: The `required: true` part can be removed since sequelize automatically adds that if the `where` clause is present in the `include` ([reference](https://sequelize.org/docs/v6/advanced-association-concepts/eager-loading/#eager-loading-filtered-at-the-associated-model-level)):
> 
> > This is the case because, when the `where` option is used inside an `include`, Sequelize automatically sets the `required` option to `true`.

Oh good point, I've removed it in https://github.com/isomerpages/isomercms-backend/pull/577/commits/930a2c6332576faec60ab4433fa48774622fe2a7!

PR #584, opened 2022-12-07
fix: hide contactNumber

prestonlimlianjie:
Resolve https://www.notion.so/opengov/VAPT-Remediation-c604c62d822e4023be266c5a293d3a1e low 1 and 2


src/services/identity/CollaboratorsService.ts
```diff
@@ -91,6 +91,10 @@ class CollaboratorsService {
         {
           model: User,
           as: "site_members",
+          attributes: {
+            // Hide PII such as contactNumber
+            exclude: ["contactNumber"],
+          },
```
seaerchin, line 97: i don't feel too strongly about putting it here - however, is it possible that we need the `contactNumber` for internal usage?

if so, perhaps we might wanna consider writing a method that calls `_.omit` just to remove sensitive info (coupled w/ a new type) and shift this out of here.

dcshzj: On a related point, would it be better if we separate out PII to a different database table that has a 1-to-1 relationship with the User table? That way we reduce the risk of accidentally querying for PII and makes requesting for that information more explicit (i.e. getting the contact number is an explicit step to be taken rather than implied when getting the User table).

prestonlimlianjie: At the moment I don't think we need to retrieve the `contactNumber` via the list endpoint just yet - though, I do like @dcshzj's suggestion a lot! Created issue [here](https://github.com/isomerpages/isomercms-backend/issues/593)

PR #585, opened 2022-12-07
Feat/buttons

seaerchin:
**tests coming in a follow up PR yeet**

## Problem
Previously, there was no api providing a translation from filename to staging url. This PR adds this service in. 
Moreover, as part of providing this service, other tangential services have also been mildly reworked to accept a result-esque api. 

Though this is rudimentary at present, we should refine this more in the future so that our APIs are safe.

also short writeup [here](https://www.notion.so/opengov/Page-mapping-f4186d4f6524419497ca715265e2d3c5) for reasoning behind how service operates

## Solution
1. create a `PageService` that parses filenames into their respective types
    - done through recursive descent so that we minimise # of API calls needed as we fetch the full blob...
    - determination the type of a resource item still requires a full network call because the resource can be named differently from the page on github. This is most clear for file resources - the github file acts as a pointer to the actual content by means of `file_url`. For files/ext links, this should be fast b/c the actual file in the resource folder contains only metadata. 
2. Shift to `ResultAsync` based APIs rather than `Promise` based. 
    - only done for `Site/ReviewRequestService` and only for some methods. The way this is done now is very crude and we should aim to return better errors rather than `NotFoundError`. This is just done as a base so that we can change the error type to be a more specific one in the future
    - the `withError` method in `PageService` should be extracted out in the future and used for interfacing with both `.js` files that call github APIs (either directly or indirectly through `GithubService`) and `.ts` files that call some external endpoints
   
## Next steps
- `PageService` should also aim to parse the page type (or defer it to some other `FileService`) as we have config files etc that are not tangentially related to what the user wants to see
- Provide a method call to filter out files that are not directly related to the user (or otherwise provide an interface) + expensive filetypes (files/images). 
    - expensive filetypes is to prevent **meaninglesss** blob comparisons

### DIFF
--- a/src/errors/EmptyStringError.ts
+++ b/src/errors/EmptyStringError.ts
@@ -1,0 +1,12 @@
+import { BaseIsomerError } from "./BaseError"
+
+export default class EmptyStringError extends BaseIsomerError {
+  constructor(
+    message = "An empty string was provided for a method that requires a non-empty string"
+  ) {
+    super()
+    Error.captureStackTrace(this, this.constructor)
+    this.name = this.constructor.name
+    this.message = message
+  }
+}
--- a/src/errors/EmptyStringError.ts
+++ b/src/errors/EmptyStringError.ts
@@ -4,9 +4,6 @@ export default class EmptyStringError extends BaseIsomerError {
   constructor(
     message = "An empty string was provided for a method that requires a non-empty string"
   ) {
-    super()
-    Error.captureStackTrace(this, this.constructor)
-    this.name = this.constructor.name
-    this.message = message
+    super(500, message)
   }
 }
--- a/src/routes/v2/authenticated/review.ts
+++ b/src/routes/v2/authenticated/review.ts
@@ -1,6 +1,7 @@
 import autoBind from "auto-bind"
 import express from "express"
 import _ from "lodash"
+import { ResultAsync } from "neverthrow"
 
 import logger from "@logger/logger"
 
@@ -101,7 +102,7 @@ export class ReviewsRouter {
       userWithSiteSessionData
     )
 
-    if (!site && doesUnmigratedSiteExist)
+    if (site.isErr() && doesUnmigratedSiteExist)
       return res.status(200).json({ message: "Unmigrated site" })
 
     const siteMembers = await this.collaboratorsService.list(siteName)
@@ -121,11 +122,17 @@ export class ReviewsRouter {
       return res.status(404).json({ message: "No site members found" })
     }
 
-    const files = await this.reviewRequestService.compareDiff(
-      userWithSiteSessionData
-    )
-
-    return res.status(200).json({ items: files })
+    return this.sitesService
+      .getStagingUrl(userWithSiteSessionData)
+      .andThen((stagingLink) =>
+        ResultAsync.fromSafePromise(
+          this.reviewRequestService.compareDiff(
+            userWithSiteSessionData,
+            stagingLink
+          )
+        )
+      )
+      .map((items) => res.status(200).json({ items }))
   }
 
   createReviewRequest: RequestHandler<
@@ -140,7 +147,7 @@ export class ReviewsRouter {
     const site = await this.sitesService.getBySiteName(siteName)
     const { userWithSiteSessionData } = res.locals
 
-    if (!site) {
+    if (site.isErr()) {
       logger.error({
         message: "Invalid site requested",
         method: "createReviewRequest",
@@ -229,7 +236,7 @@ export class ReviewsRouter {
       // and assert that the user is an admin of said site.
       // This guarantees that the user exists in our database.
       admin!,
-      site,
+      site.value,
       title,
       description
     )
@@ -271,7 +278,7 @@ export class ReviewsRouter {
       userWithSiteSessionData
     )
 
-    if (!site && doesUnmigratedSiteExist)
+    if (site.isErr() && doesUnmigratedSiteExist)
       return res.status(200).json({ message: "Unmigrated site" })
 
     const siteMembers = await this.collaboratorsService.list(siteName)
@@ -281,7 +288,7 @@ export class ReviewsRouter {
       return res.status(200).json({ message: "Unmigrated users" })
     }
 
-    if (!site) {
+    if (site.isErr()) {
       logger.error({
         message: "Invalid site requested",
         method: "listReviews",
@@ -323,7 +330,7 @@ export class ReviewsRouter {
     // Step 3: Fetch data and return
     const reviews = await this.reviewRequestService.listReviewRequest(
       userWithSiteSessionData,
-      site
+      site.value
     )
 
     return res.status(200).json({
@@ -342,7 +349,7 @@ export class ReviewsRouter {
     const { siteName } = req.params
     const site = await this.sitesService.getBySiteName(siteName)
 
-    if (!site) {
+    if (site.isErr()) {
       return res.status(404).send({
         message: "Please ensure that the site exists!",
       })
@@ -368,7 +375,7 @@ export class ReviewsRouter {
     // Step 3: Update all review requests for the site as viewed
     await this.reviewRequestService.markAllReviewRequestsAsViewed(
       userWithSiteSessionData,
-      site
+      site.value
     )
 
     return res.status(200).send()
@@ -385,7 +392,7 @@ export class ReviewsRouter {
     const { siteName, requestId: prNumber } = req.params
     const site = await this.sitesService.getBySiteName(siteName)
 
-    if (!site) {
+    if (site.isErr()) {
       return res.status(404).send({
         message: "Please ensure that the site exists!",
       })
@@ -410,7 +417,7 @@ export class ReviewsRouter {
 
     // Step 3: Retrieve review request
     const possibleReviewRequest = await this.reviewRequestService.getReviewRequest(
-      site,
+      site.value,
       prNumber
     )
 
@@ -431,7 +438,7 @@ export class ReviewsRouter {
     // Step 4: Mark review request as viewed
     await this.reviewRequestService.markReviewRequestAsViewed(
       userWithSiteSessionData,
-      site,
+      site.value,
       possibleReviewRequest.id
     )
 
@@ -453,7 +460,7 @@ export class ReviewsRouter {
       userWithSiteSessionData
     )
 
-    if (!site && doesUnmigratedSiteExist) {
+    if (site.isErr() && doesUnmigratedSiteExist) {
       return res.status(200).json({ message: "Unmigrated site" })
     }
 
@@ -464,7 +471,7 @@ export class ReviewsRouter {
       return res.status(200).json({ message: "Unmigrated users" })
     }
 
-    if (!site) {
+    if (site.isErr()) {
       logger.error({
         message: "Invalid site requested",
         method: "getReviewRequest",
@@ -505,13 +512,21 @@ export class ReviewsRouter {
       })
     }
 
-    const possibleReviewRequest = await this.reviewRequestService.getFullReviewRequest(
-      userWithSiteSessionData,
-      site,
-      requestId
-    )
+    const possibleReviewRequest = await this.sitesService
+      .getStagingUrl(userWithSiteSessionData)
+      .andThen((stagingLink) =>
+        this.reviewRequestService.getFullReviewRequest(
+          userWithSiteSessionData,
+          site.value,
+          requestId,
+          stagingLink
+        )
+      )
 
-    if (isIsomerError(possibleReviewRequest)) {
+    // NOTE: This method call (`getFullReviewRequest`) should be
+    // in the result also but as it returns a promise at the moment,
+    // we just do a simple check
+    if (possibleReviewRequest.isErr()) {
       logger.error({
         message: "Invalid review request requested",
         method: "getReviewRequest",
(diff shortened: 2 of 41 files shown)


src/types/pages.ts
```diff
@@ -0,0 +1,118 @@
...
+  kind: "ResourceCategoryPage"
+}
+
+export type PageName =
+  | SubcollectionPageName
+  | CollectionPageName
+  | ContactUsPageName
+  | UnlinkedPageName
+  | HomepageName
+  | ResourceCategoryPageName
```
kishore03109, line 58: Just for clarification, the usefulness of this banding of types will be amplified only when we add some level of validation checks when we actually do the casting right (though I understand this would be out of scope of this pr)?
Out of curiosity, is there a reason why you didnt chose to just add enums to differentiate the various types?

src/services/fileServices/MdPageServices/PageService.ts
```diff
@@ -0,0 +1,379 @@
...
+   *
+   * This is done to avoid expensive calls to fetch a page's raw blob data and
+   * parsing it thereafter which, for big files, might be expensive.
+   * @param pageName A **valid and fully qualified** file path
+   * @param sessionData session credentials of the user
+   */
+  parsePageName = (
+    pageName: string,
+    sessionData: UserSessionData
+  ): ResultAsync<PageName, NotFoundError | MissingResourceRoomError> =>
```
kishore03109, line 116: could we just use `NotFoundError` rather than `NotFoundError | MissingResourceRoomError` as `MissingResourceRoomError` is a type of `NotFoundError`?

src/services/fileServices/MdPageServices/PageService.ts
```diff
@@ -0,0 +1,379 @@
...
+// this should return a sum type of the possible errors.
+// This is (at a glance), either a failure to read from github
+// or the parsing fails (we have a sentinel value of an empty string)
+const withError = (promise: Promise<PageInfo>) =>
+  ResultAsync.fromPromise(promise, () => new BaseIsomerError()).andThen(
+    (data) =>
+      // NOTE: This is a fail-safe check as `yaml.parse`
+      // doesn't guarantee existence.
+      data?.content?.frontMatter?.permalink
+        ? ok(data)
```
kishore03109, line 47: Hey, am confused on whats going on in this method call. we only run this method iff there is "either a a failure to read from github or the parsing fails (we have a sentinel value of an empty string)"... and if this is the case, why are we resolving it to an ok with `ok(data)` here ah?

src/services/review/ReviewRequestService.ts
```diff
@@ -416,7 +398,7 @@ export default class ReviewRequestService {
 
     const { id: reviewRequestId } = possibleReviewRequest
 
-    await this.reviewRequestView.destroy({
+    this.reviewRequestView.destroy({
```
kishore03109, line 401: just checking, is this removal of `await` intended?

src/routes/v2/authenticated/review.ts
```diff
@@ -691,6 +709,11 @@ export class ReviewsRouter {
       })
     }
 
+    // Check state here to ensure it's open
+    if (possibleReviewRequest.reviewStatus !== ReviewRequestStatus.Open) {
+      // return 400
```
kishore03109, line 714: unintended comment?

kishore03109 (changes requested):

src/types/pages.ts
seaerchin: i'm not entirely sure i understand the first question here @__@ the usefulness here is that it allows us to enforce a stronger API contract on subsequent methods going forward (eg: a method that operates on `Homepage` wouldn't accept an `UnlinkedPage`. Do take note that for the casting, it's only required for base types (string/number etc) as there's **no way to construct** a `string & {__kind: "blah"}`.

wrt why not use enums - because ts is structurally typed, we do need something to distinguish the types, esp if they're structurally identical. i just used a string (`kind`) cos i've ran into some issues w/ the implementation of enums in ts before wrt typing

src/services/fileServices/MdPageServices/PageService.ts
seaerchin: no, that's actually discouraged. we want to return the **most specific** type that we can. Note that in this case, for eg, by returning `NotFoundError | MissingResourceRoomError`, the caller can determine whether they care about the _type_ of error. 

If we were to return a blanket `NotFoundError`, the caller has to perform validation checks to see if the error is indeed something that they care about. If the caller does **not** care about it, they can always just use the properties etc and we don't lose anything if the subsequent method call requires a `NotFoundError` (cos the interface is the same).

src/services/fileServices/MdPageServices/PageService.ts
```diff
@@ -0,0 +1,379 @@
...
+    }
+
+    const fullPath = pageName.split("/")
+    // NOTE: Name is guaranteed to exist
+    // as this method only accepts a string
+    // and we've validated that the string is not empty
+    const name = fullPath.pop()!
+
+    if (fullPath.length === 0) {
+      return ok({
```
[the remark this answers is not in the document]
seaerchin: the path is `err` but the overall is `ok`. this is because it's possible for files (eg: homepage) to have no path

src/services/fileServices/MdPageServices/PageService.ts
seaerchin: think you've a misunderstanding here. the comment is saying that the possible errors to come from here is either a failure to read from github, or a parse error (key doesn't exist for eg).

if there's a failure to read from github, 
`ResultAsync.fromPromise(promise, () => new BaseIsomerError())` resolves this because the promise will reject. This will turn into `ErrAsync<never, BaseIsomerError>` for the type, which is essentially `errAsync(new BaseIsomerError())` for the actual value. 

the check of `data?.content?.frontMatter?.permalink` is, as the comment implies, a fail-safe check to ensure that the property we care about (`permalink`) exists, as `yaml.parse` returns `any` and we want to make some guarantees on the shape of the object. 

this method runs on the `andThen`, which is only ran in the success case (ie, the github api call is successful), so this subsequent check is just to ensure the property exists

src/routes/v2/authenticated/review.ts
seaerchin: ohyea good catch - this are part of VAPT fixes and shouldn't be here oops

src/services/review/ReviewRequestService.ts
seaerchin: oops, i'll revert this! thanks for spot

src/types/pages.ts
```diff
@@ -0,0 +1,118 @@
...
+export type UnlinkedPage = PageInfo & {
+  fileName: PageBrand<UnlinkedPageName>
+}
+
+export type StagingPermalink = Brand<string, "staging">
+export type ProdPermalink = Brand<string, "prod">
+
+// NOTE: This is not `frontMatter.permalink` as this
+// also includes the respective base url in front.
+export type Permalink = StagingPermalink | ProdPermalink
```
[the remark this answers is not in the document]
seaerchin: sure, will do

src/errors/EmptyStringError.ts
```diff
@@ -0,0 +1,12 @@
+import { BaseIsomerError } from "./BaseError"
+
+export default class EmptyStringError extends BaseIsomerError {
+  constructor(
+    message = "An empty string was provided for a method that requires a non-empty string"
+  ) {
+    super()
+    Error.captureStackTrace(this, this.constructor)
+    this.name = this.constructor.name
+    this.message = message
```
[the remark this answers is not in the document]
seaerchin: i think we'd still need to preserve `this.name = this.constructor.name` because if it's at `super`, the constructor will be `BaseIsomerError`. not entirely sure on this, wdyt?

src/services/review/ReviewRequestService.ts
```diff
@@ -75,35 +87,61 @@ export default class ReviewRequestService {
...
+    baseUrl: string
+  ): Promise<string | undefined> => {
+    // NOTE: frontMatter is typed as `any` but its actual type is string
+    const {
+      content: { frontMatter: _frontMatter },
+    } = await this.pageService.read(sessionData, {
+      fileName,
+    })
+    const frontMatter = _frontMatter as string
+    const permalink = frontMatter
```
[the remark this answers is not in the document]
seaerchin: Think this method doesn't exist anymore - got subsumed into `PageService`; the relevant parsing methods have comments describing the structure already but lmk if it's unclear

src/services/fileServices/MdPageServices/PageService.ts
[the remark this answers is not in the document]
kishore03109: right, got it!

src/services/fileServices/MdPageServices/PageService.ts
kishore03109: ok, this makes sense, but shall we name the method call to something like `parseYamlFile` rather than just `withError`?

src/services/fileServices/MdPageServices/PageService.ts
seaerchin: hmm how would the caller determine if it's indeed a missing resource room? the exact nature of the error would be hidden behind a general interface of `NotFoundError`, no? Esp if, for eg (as per this endpoint), the call can return many various kinds of not found error

src/errors/EmptyStringError.ts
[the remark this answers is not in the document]
kishore03109: woah my hunch was that you were right, but turns out no sia 
The way I validated this was i commented out 
`// this.name = this.constructor.name` and placed a `console.log(this.name)` right below, and it magically printed out `EmptyStringError` rather than `BaseIsomerError` 

At the same time, `this.message = message` in line 10 should still exist!

src/types/pages.ts
kishore03109: arh, I am curious, feel free to ignore this if you dont have concrete examples at hand, do you have examples of issues that you faced w/ implementation of enums in ts?

src/services/fileServices/MdPageServices/PageService.ts
kishore03109: uhm ok wait to be more specific, my question would be wouldnt the cast 
`NotFoundError | MissingResourceRoomError` be semantically equivalent to `NotFoundError` as `MissingResourceRoomError` _extends_ from `NotFoundError`?

seaerchin: Actually because ts uses structural typing, all the errors without differing properties will be identical. This means, for example, that we can even assign a totally unrelated error (`AuthError`) to a method/variable that requires `MissingSiteError`/`NotFoundError`.

what we want to think about here is whether the error itself is useful - we can return a `NotFoundError` but we could be more specific and return a `MissingResourceRoomError`. This allows the caller to tell what errors are returned from thsi function.

as maintainers, we can also know the specific `NotFound` errors that can be returned (eg, if the file wasn't actually found on github, if the contact-us page is missing etc), which helps with ease of maintainability

src/services/fileServices/MdPageServices/PageService.ts
seaerchin: actually this method is meant to act as a error handler over the raw promise. i was thinking about matching on the type of error returned from github and turning it into an actual internal error rather than just returning `BaseIsomerError`, which is not useful. 

i'm fine w/ renaming it but bc of ^, idt `parseYamlFile` is the right choice (as it also doesn't do parsing)

src/types/pages.ts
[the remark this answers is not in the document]
seaerchin: wait this actually contains more info than the one in frontmatter: this is `{url}/{frontMatter.permalink}` - idt we should name this `BasePermalink`; wdyt about `FullPermalink`?

src/errors/EmptyStringError.ts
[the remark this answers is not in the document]
seaerchin: huh cool! so to double check - we don't need the `this.name = this.constructor.name`?

src/types/pages.ts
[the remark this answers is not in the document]
kishore03109: Yup, makes sense!

src/errors/EmptyStringError.ts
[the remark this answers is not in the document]
kishore03109: yes sir

src/services/fileServices/MdPageServices/PageService.ts
kishore03109: arh ok this line of thought makes sense, thanks for the explanation!

src/services/fileServices/MdPageServices/PageService.ts
kishore03109: got it, shall we rename this to `errorHandler` then?

src/types/pages.ts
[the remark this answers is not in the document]
seaerchin: done, thank,s for raising!

src/services/fileServices/MdPageServices/PageService.ts
seaerchin: sure! i called it `withErrorHandler` as it injects it into the promise (as per what's described above!)

src/errors/EmptyStringError.ts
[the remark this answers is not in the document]
seaerchin: done!

kishore03109 (approved):

PR #587, opened 2022-12-09
test(app): updated specs to fit with new API  

seaerchin:
## Problem
Part of #585, where the specs failed due to the API changes. This PR fixes the failing tests; most of the fixes are mostly cosmetic but here are a few meaningful changes described below.

## Solution
1. rework tests to unwrap value
2. for `SitesService.spec`, the spec where `it("should return UnprocessableError when the GitHub commit is not found")` has been altered to only assert that the service is called once. This is because we now short circuit when either the prod/staging commit could not be fetched, rather than continuing on with the code flow. This is because both `staging/prod` commit contain info that we use to derive properties without which the flow isn't meaningful anymore.

### DIFF
--- a/src/errors/DatabaseError.ts
+++ b/src/errors/DatabaseError.ts
@@ -2,9 +2,6 @@ import { BaseIsomerError } from "./BaseError"
 
 export default class DatabaseError extends BaseIsomerError {
   constructor(message = "Unable to retrieve data from database") {
-    super()
-    Error.captureStackTrace(this, this.constructor)
-    this.name = this.constructor.name
-    this.message = message
+    super(500, message)
   }
 }
--- a/src/errors/DatabaseError.ts
+++ b/src/errors/DatabaseError.ts
@@ -1,0 +1,10 @@
+import { BaseIsomerError } from "./BaseError"
+
+export default class DatabaseError extends BaseIsomerError {
+  constructor(message = "Unable to retrieve data from database") {
+    super()
+    Error.captureStackTrace(this, this.constructor)
+    this.name = this.constructor.name
+    this.message = message
+  }
+}
--- a/src/integration/Sites.spec.ts
+++ b/src/integration/Sites.spec.ts
@@ -19,8 +19,20 @@ import UserSessionData from "@root/classes/UserSessionData"
 import { mockEmail, mockIsomerUserId } from "@root/fixtures/sessionData"
 import { getAuthorizationMiddleware } from "@root/middleware"
 import { SitesRouter as _SitesRouter } from "@root/routes/v2/authenticated/sites"
+import { isomerRepoAxiosInstance } from "@root/services/api/AxiosInstance"
 import { GitHubService } from "@root/services/db/GitHubService"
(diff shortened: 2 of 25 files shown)


src/services/identity/SitesService.ts
```diff
@@ -429,6 +434,7 @@ class SitesService {
       if (this.isGitHubCommitData(possibleCommit)) {
         return okAsync(possibleCommit)
       }
+      console.log("=======", possibleCommit)
```
kishore03109: unintended console log?

src/services/identity/__tests__/SitesService.spec.ts
```diff
@@ -863,7 +865,7 @@ describe("SitesService", () => {
       )
 
       // Assert
-      expect(actual).toBeInstanceOf(NotFoundError)
+      expect(actual).toEqual(err(new MissingSiteError()))
```
kishore03109, line 868: nit: should we be using `toBeInstanceOf(NotFoundError)` as in the future if we wish to add metadata to the error object, this assert will fail?

src/errors/DatabaseError.ts
```diff
@@ -0,0 +1,10 @@
+import { BaseIsomerError } from "./BaseError"
+
+export default class DatabaseError extends BaseIsomerError {
+  constructor(message = "Unable to retrieve data from database") {
+    super()
+    Error.captureStackTrace(this, this.constructor)
+    this.name = this.constructor.name
```
kishore03109, line 7: not needed!

src/integration/Sites.spec.ts
```diff
@@ -156,6 +156,10 @@ describe("Sites Router", () => {
 
   describe("/", () => {
     beforeAll(async () => {
+      await User.sync({ force: true })
```
kishore03109, line 159: just to clarify, I thought we prevented the DB clash by removing concurrency using the flag 
`--runInBand`? if this solves the concurrency issue, should we remove the flag `--runInBand` to allow for concurrency again?

kishore03109 (changes requested):

src/services/identity/SitesService.ts
seaerchin: think this got removed already by [this](https://github.com/isomerpages/isomercms-backend/pull/587/commits/e83fe99607259b71315ee026b98d7c2fa83a04da) commit

src/integration/Sites.spec.ts
seaerchin: actually i'm not too very sure on the specifics (cc @dcshzj for clarification); my understanding is that `--runInBand` serializes it and we have `.sync` to prevent db pollution between tests

src/errors/DatabaseError.ts
seaerchin: removed, thanks for catch!

src/integration/Sites.spec.ts
kishore03109: okie

kishore03109 (approved):

PR #588, opened 2022-12-12
test(pageservice): add tests for page service

seaerchin:
## Problem
previously, `PageService` had no associated tests; this PR add the tests in. 

## Solution
1. write unit tests for `pageService`
    - the unit tests are centered around `parsePageName` together with `retrieveStagingPermalink`; the invariants that we want to guarantee will be briefly outlined below
    - for static pages (pages whose name are **always fixed**), maintain the invariant that only that specified filepath can be parsed into that name
    - guarantee that for resource rooms, we will only get permalink for posts 
    - recursive descent (ie, the order in which we attempt to parse is fixed) 
    - note that filepaths terminating in `/` will **not** have a name property - consistent w/ how directories behave; we could extend our model to also parse directories but this is not done at present. this should also not be possible if we are receiving from the fe

### DIFF
--- a/src/constants/pages.ts
+++ b/src/constants/pages.ts
@@ -1,0 +1,1 @@
+export const HOMEPAGE_NAME = "index.md"
--- a/src/constants/pages.ts
+++ b/src/constants/pages.ts
@@ -1,1 +1,3 @@
 export const HOMEPAGE_NAME = "index.md"
+
+export const CONTACT_US_FILENAME = "contact-us.md"
--- a/src/constants/pages.ts
+++ b/src/constants/pages.ts
@@ -1,3 +1,3 @@
-export const HOMEPAGE_NAME = "index.md"
+export const HOMEPAGE_FILENAME = "index.md"
 
 export const CONTACT_US_FILENAME = "contact-us.md"
--- a/src/services/fileServices/MdPageServices/PageService.ts
+++ b/src/services/fileServices/MdPageServices/PageService.ts
@@ -141,10 +141,6 @@ export class PageService {
           CollectionPageName | SubcollectionPageName,
           NotFoundError
         >((rawPath) => {
-          // NOTE: Only 2 levels of nesting
-          if (rawPath.length > 2) {
-            return errAsync(new NotFoundError())
-          }
           if (rawPath.length === 1 && !!rawPath[0]) {
             return okAsync({
               name: Brand.fromString(name),
--- a/src/services/fileServices/MdPageServices/PageService.ts
+++ b/src/services/fileServices/MdPageServices/PageService.ts
@@ -145,19 +145,22 @@ export class PageService {
           if (rawPath.length > 2) {
             return errAsync(new NotFoundError())
           }
-          if (rawPath.length === 1) {
+          if (rawPath.length === 1 && !!rawPath[0]) {
             return okAsync({
               name: Brand.fromString(name),
               collection: rawPath[0],
               kind: "CollectionPage",
             })
           }
-          return okAsync({
-            name: Brand.fromString(name),
-            collection: rawPath[0],
-            subCollection: rawPath[1],
-            kind: "SubcollectionPage",
-          })
+          if (rawPath.length === 2 && !!rawPath[0] && !!rawPath[1]) {
+            return okAsync({
+              name: Brand.fromString(name),
+              collection: rawPath[0],
+              subCollection: rawPath[1],
+              kind: "SubcollectionPage",
+            })
+          }
+          return errAsync(new NotFoundError())
         })
       )
       .mapErr(() => new NotFoundError())
--- a/src/services/fileServices/MdPageServices/PageService.ts
+++ b/src/services/fileServices/MdPageServices/PageService.ts
@@ -1,7 +1,7 @@
 import { ok, err, Result, ResultAsync, okAsync, errAsync } from "neverthrow"
 
 import UserSessionData from "@root/classes/UserSessionData"
-import { CONTACT_US_FILENAME, HOMEPAGE_NAME } from "@root/constants"
+import { CONTACT_US_FILENAME, HOMEPAGE_FILENAME } from "@root/constants"
 import { BaseIsomerError } from "@root/errors/BaseError"
 import EmptyStringError from "@root/errors/EmptyStringError"
 import MissingResourceRoomError from "@root/errors/MissingResourceRoomError"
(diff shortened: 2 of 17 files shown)


src/constants/pages.ts
```diff
@@ -0,0 +1,3 @@
+export const HOMEPAGE_NAME = "index.md"
```
dcshzj, line 1: Minor nit: Should we standardise to call it FILENAME?

src/services/fileServices/MdPageServices/PageService.ts
```diff
@@ -142,19 +142,22 @@ export class PageService {
           if (rawPath.length > 2) {
             return errAsync(new NotFoundError())
           }
```
dcshzj, line 144: Should we remove this part since we still return a NotFoundError later in line 160 if the two if-branches are not satisfied?

src/services/fileServices/MdPageServices/__tests__/PageService.spec.ts
```diff
@@ -0,0 +1,940 @@
...
+  ResourceCategoryPageName,
+  SubcollectionPageName,
+  UnlinkedPageName,
+} from "@root/types/pages"
+import { Brand } from "@root/types/util"
+
+import { HomepagePageService } from "../HomepagePageService"
+import { PageService } from "../PageService"
+import { ResourcePageService } from "../ResourcePageService"
+import { SubcollectionPageService } from "../SubcollectionPageService"
```
dcshzj, line 27: Minor nit: I think we should standardise the way the imports are done using `@services`.

src/services/fileServices/MdPageServices/__tests__/PageService.spec.ts
```diff
@@ -0,0 +1,940 @@
...
+  sha: "",
+})
+
+describe("PageService", () => {
+  describe("parsePageName", () => {
+    it("should only accept 'index.md' as the homepage name", async () => {
+      // Arrange
+      const expected = ok({
+        name: HOMEPAGE_NAME,
+        kind: "Homepage",
```
dcshzj, line 86: I think this concerns mainly the original PR, but should we change this to use an enum instead?

src/services/fileServices/MdPageServices/__tests__/PageService.spec.ts
```diff
@@ -0,0 +1,940 @@
...
+      // Assert
+      expect(actual).toEqual(expected)
+      expect(
+        mockResourceRoomDirectoryService.getResourceRoomDirectoryName
+      ).toBeCalledWith(MOCK_USER_SESSION_DATA_ONE)
+    })
+  })
+
+  describe("extractPathInfo", () => {
+    it("should return a `PathInfo` with a valid path when the string provided is a valid filepath", async () => {
```
dcshzj, line 393: Nit: The `async` can be removed.

dcshzj (changes requested):
Briefly looked through the test cases, seems ok to me.

src/services/fileServices/MdPageServices/__tests__/PageService.spec.ts
seaerchin: actually i ran into troubles with using enums as types in the past and hence, i didn't use it here. i think the original PR has a comment about this raised by kishore also

src/constants/pages.ts
seaerchin: sure!

src/services/fileServices/MdPageServices/PageService.ts
seaerchin: actually i left it in just in case we wanted to return a more specific error in the future but yea, i htink this is more of code bloat for now so i'll just remove it

src/services/fileServices/MdPageServices/__tests__/PageService.spec.ts
seaerchin: i usually don't care if it's in the same/parent directory cos it's only one level of indirection; would we be ok with keeping relative imports only for `../page` and `./page`?

dcshzj: Yup that is fine too, but we should standardise with lines 10-13.

src/services/fileServices/MdPageServices/PageService.ts
```diff
@@ -138,10 +138,6 @@ export class PageService {
           CollectionPageName | SubcollectionPageName,
           NotFoundError
         >((rawPath) => {
-          // NOTE: Only 2 levels of nesting
-          if (rawPath.length > 2) {
-            return errAsync(new NotFoundError())
-          }
```
kishore03109, line 144: hey, just checking in ya, are we guaranteed that the `length` will be <= 2?

seaerchin: this depends on whether the site is actually on isomercms - if they are, then then folders can only be 2 levels deep (pages can either be in a folder or in a subfolder and they can't create another folder within the subfolder)

src/services/fileServices/MdPageServices/__tests__/PageService.spec.ts
seaerchin: standardised, less the directory service.

src/services/fileServices/MdPageServices/PageService.ts
kishore03109: Hmm, if that is the case should we still throw an error if this condition is the `true`? eg. 
```
    // NOTE: Only 2 levels of nesting
    if (rawPath.length > 2) {
       return errAsync(new BaseIsomerError(`Folder/page >2 levels deep, unsupported by IsomerCMS`))
    }
```
rather than not handling it at all?

seaerchin: i don't think there's an error thrown here; could i just clarify that what you meant here is to return a more specific error type with a relevant message rather than returning a generic `NotFoundError` which might not be useful?

if that's the case, then yeap, fully agreed (got lazy and didn't do it HAHA) and i'll make the relevant changes so that it returns a `LegacyError`!

kishore03109: Hmm correct me if I am wrong, but there is an error thrown `if rawPath.length > 2` at line 160 right? 
To be honest, I am ok this as is (I think returning `NotFoundError()`), but perhaps a quick change that we could do is add some meaningful information about the error? 

Concretely, what do you think about the suggestion:
`return errAsync(new NotFoundError(`Error when parsing `rawPath: ${rawPath}`))` in line 160
for easier debugging?

seaerchin: no, there's no error _thrown_. there's a result type of the error variant  _returned_ (we don't short circuit, control flow still as per normal). 

also, this is deleted and the updated code only retains the `errAsync` at the end! wrt debugging - actually we discard the error totally because we call `unwrapOr` at the use site, so whatever we put won't actually be seen. i'll still add it in but just an fyi!

what's happening atm is that i'm extending our data model to be more descriptive so rather than just parsing pages, we are also able to discern for eg, config/images/files, which might be more useful for devs as we can make more static guarantees rather than debugging post errors

kishore03109 (approved):

PR #594, opened 2022-12-21
fix(usersservice): remove explicit `findOrCreate` calls

### LINKED ISSUE #592: [BUG] `findOrCreate` fails to work properly and throws
**Describe the bug**
Occasionally, `findOrCreate` fails to do its job properly and throws an error in our codebase. This leads to an internal server error, when we fail to handle the error gracefully. 

At present, this occurs most frequently during verification of a new user (more specifically, their email).

See [here](https://github.com/sequelize/sequelize/issues/14266#issuecomment-1075282345) for more info 

**Expected behaviour**
1. `findOrCreate` works properly and no error is thrown.

**Suggested fix**
- rewrite to use separate `find` and `create` operations, while wrapping in trxn to ensure A/C.

**Definition of done**
- we should track this for 1-2 wks post release to determine if it's still an issue

seaerchin:
## Problem
See #592 for an overview of the issue

Closes #592

## Solution
1. remove explicit calls of `findOrCreate` on our db model. instead, this has been replaced with explicit calls to both `findOne` and `create`, which are both wrapped into a single txn to ensure atomicity + consistency
2. test cases updated to ensure that we call explicit methods + use the txn

### DIFF
--- a/src/services/identity/UsersService.ts
+++ b/src/services/identity/UsersService.ts
@@ -63,20 +63,37 @@ class UsersService {
     await this.repository.update(user, { where: { githubId } })
   }
 
-  async findOrCreate(githubId: string | undefined) {
-    const [user] = await this.repository.findOrCreate({
-      where: { githubId },
+  async findOrCreate(githubId: string | undefined): Promise<User> {
+    return this.sequelize.transaction<User>(async (transaction) => {
+      const existingUser = await this.repository.findOne({
+        where: { githubId },
+        transaction,
+      })
+
+      if (existingUser) return existingUser
+
+      return this.repository.create({
+        githubId,
+        transaction,
+      })
     })
-    return user
   }
 
   async login(githubId: string): Promise<User> {
     return this.sequelize.transaction<User>(async (transaction) => {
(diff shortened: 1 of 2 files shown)


src/services/identity/UsersService.ts
```diff
@@ -63,20 +63,37 @@ class UsersService {
     await this.repository.update(user, { where: { githubId } })
   }
 
-  async findOrCreate(githubId: string | undefined) {
-    const [user] = await this.repository.findOrCreate({
-      where: { githubId },
+  async findOrCreate(githubId: string | undefined): Promise<User> {
```
kishore03109, line 66: Woah, I just realised we arent using this function anywhere in the codebase apart from our tests ._. 
If you can double confirm this is true, could we remove this method as a whole? This method does not handle undefined githubID gracefully as well, seems to be doing what the `login` function is doing anyways

kishore03109 (changes requested):

src/services/identity/UsersService.ts
seaerchin: you're right; i'll delete this method and the associated tests. odd that this got added w/o any call sites but ok

kishore03109 (approved):

PR #595, opened 2022-12-22
chore(mediafileservice): prevent users from changing file ext

seaerchin:
## Problem
See the notion [link](https://www.notion.so/opengov/VAPT-Remediation-c604c62d822e4023be266c5a293d3a1e) for an overview 

## Solution
1. prevent users from renaming unless the old extension and the new extension are the same. because we disallow creation of files unless they have certain extensions (`file-upload-utils` defines `ALLOWED_FILE_EXTENSIONS`), this means that existing files already have valid extensions. now, we force the old/new ext to be the same, which implies that we only ever go from existing valid ext -> new valid ext.

### DIFF
--- a/src/services/fileServices/MdPageServices/MediaFileService.js
+++ b/src/services/fileServices/MdPageServices/MediaFileService.js
@@ -1,6 +1,8 @@
 const { BadRequestError } = require("@errors/BadRequestError")
 const { MediaTypeError } = require("@errors/MediaTypeError")
 
+const { getFileExt } = require("@root/utils/files")
+
 const { GITHUB_ORG_NAME } = process.env
 
 const { validateAndSanitizeFileUpload } = require("@utils/file-upload-utils")
@@ -109,6 +111,14 @@ class MediaFileService {
   async rename(reqDetails, { oldFileName, newFileName, directoryName, sha }) {
     this.mediaNameChecks({ directoryName, fileName: oldFileName })
     this.mediaNameChecks({ directoryName, fileName: newFileName })
+    const oldExt = getFileExt(oldFileName)
+    const newExt = getFileExt(newFileName)
+
+    if (oldExt !== newExt) {
+      throw new BadRequestError(
+        "Please ensure that the file extension stays the same when renaming!"
+      )
+    }
 
     const gitTree = await this.gitHubService.getTree(reqDetails, {
       isRecursive: true,
--- a/src/services/fileServices/MdPageServices/MediaFileService.js
+++ b/src/services/fileServices/MdPageServices/MediaFileService.js
@@ -5,7 +5,10 @@ const { getFileExt } = require("@root/utils/files")
 
 const { GITHUB_ORG_NAME } = process.env
 
-const { validateAndSanitizeFileUpload } = require("@utils/file-upload-utils")
+const {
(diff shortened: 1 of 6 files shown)


src/services/fileServices/MdPageServices/MediaFileService.js
```diff
@@ -120,6 +123,12 @@ class MediaFileService {
       )
     }
 
+    if (!ALLOWED_FILE_EXTENSIONS.includes(oldExt)) {
```
kishore03109, line 126: +1 for extra validation! 
Do you think we could also do a similar parsing validation in the backend of unallowed characters? (similar to what we already do in our frontend?) 

I do think this this is a 'nice to have' rather than a 'must have', feel free to resolve this comment if you feel this is not needed!

src/utils/file-upload-utils.js
```diff
@@ -6,7 +6,7 @@ const { JSDOM } = require("jsdom")
 const { window } = new JSDOM("")
 const DOMPurify = createDOMPurify(window)
 
-const ALLOWED_FILE_EXTENSIONS = [
+export const ALLOWED_FILE_EXTENSIONS = [
```
kishore03109, line 9: hmm I feel effy about combining es6 and commonjs in the same file, could we standardise this at a file level?

src/utils/files.ts
```diff
@@ -0,0 +1,2 @@
+export const getFileExt = (fileName: string): string | undefined =>
```
kishore03109, line 1: correct me if I am wrong, but as long as fileName is of type `string`, fileName.split(".").shift() will **_not_** gibe us an undefined value right?

kishore03109 (changes requested):

src/services/fileServices/MdPageServices/MediaFileService.js
```diff
@@ -1,6 +1,8 @@
 const { BadRequestError } = require("@errors/BadRequestError")
 const { MediaTypeError } = require("@errors/MediaTypeError")
 
+const { getFileExt } = require("@root/utils/files")
```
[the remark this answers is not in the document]
seaerchin: sure, resolved!

src/utils/files.ts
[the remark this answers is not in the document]
seaerchin: i think default exports make sense when the file is descriptive of the contents (eg: `ForbiddenError.ts` exports `class ForbiddenError`). In this case, because this is a utility file, it's likely to get more exported functions in the future so idt i want to change it. wdyt?

src/services/fileServices/MdPageServices/MediaFileService.js
seaerchin: this is already done by `this.mediaNameChecks`

src/utils/file-upload-utils.js
seaerchin: sure, fixed!

src/utils/files.ts
seaerchin: good catch, changed the return type!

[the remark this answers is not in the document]
kishore03109: sure, am ok with this rationale

kishore03109 (approved):

PR #610, opened 2023-01-18
feat(rate-limit): adds rate limiting to auth endpoints

seaerchin:
## Problem
Closes VAPT issue for brute forcing.

## Solution
This was done through installing an external library for rate limiting. 
    - the library was chosen cos it was in ts + seems easy to use 
    - limit of 100/window (otp space = 10 ** 6 so the chance is 100/10^6 << 1%)
    - limit of 100 is applied to only `auth` endpoint (no strong opinion here, can apply to all if needed) as `auth` endpoint was the one flagged. didn't want it to be the most specific one as auth as a whole is p sensitive so would rather err on the safe side.

### DIFF
--- a/src/services/fileServices/MdPageServices/MediaFileService.js
+++ b/src/services/fileServices/MdPageServices/MediaFileService.js
@@ -1,4 +1,4 @@
-import logger from "@logger/logger"
+const logger = require("@logger/logger")
 
 const { BadRequestError } = require("@errors/BadRequestError")
 const { MediaTypeError } = require("@errors/MediaTypeError")
@@ -38,10 +38,7 @@ class MediaFileService {
     }
 
     // Sanitize and validate file
-    const sanitizedContent = await validateAndSanitizeFileUpload(
-      fileContent,
-      fileBuffer
-    )
+    const sanitizedContent = await validateAndSanitizeFileUpload(content)
     if (!sanitizedContent) {
       throw new MediaTypeError(`File extension is not within the approved list`)
     }
--- a/src/services/fileServices/MdPageServices/__tests__/MediaFileService.spec.js
+++ b/src/services/fileServices/MdPageServices/__tests__/MediaFileService.spec.js
@@ -8,7 +8,7 @@ describe("Media File Service", () => {
   const imageName = "test image.png"
   const fileName = "test file.pdf"
   const directoryName = "images/subfolder"
-  const mockContent = "test"
+  const mockContent = "schema, test"
   const mockSanitizedContent = "sanitized-test"
   const sha = "12345"
   const mockGithubSessionData = "githubData"
@@ -32,6 +32,7 @@ describe("Media File Service", () => {
     validateAndSanitizeFileUpload: jest
       .fn()
       .mockReturnValue(mockSanitizedContent),
+    scanFileForVirus: jest.fn().mockReturnValue({ CleanResult: true }),
   }))
   const {
     MediaFileService,
--- a/src/utils/file-upload-utils.js
+++ b/src/utils/file-upload-utils.js
@@ -33,7 +33,7 @@ const scanFileForVirus = (fileBuffer) =>
       throw new BaseIsomerError(500, "Internal Server Error")
     }
 
-    apiInstance.scanFile(fileBuffer, (error, data, response) => {
+    apiInstance.scanFile(fileBuffer, (error, data) => {
       if (error) {
         logger.error(
           `Error when calling Cloudmersive Virus Scan API: ${error.message}`
@@ -46,7 +46,9 @@ const scanFileForVirus = (fileBuffer) =>
     })
   })
 
-const validateAndSanitizeFileUpload = async (content, fileBuffer) => {
+const validateAndSanitizeFileUpload = async (data) => {
+  const [, content] = data.split(",")
+  const fileBuffer = Buffer.from(content, "base64")
   const detectedFileType = await FileType.fromBuffer(fileBuffer)
 
   if (isSvg(fileBuffer)) {


kishore03109 (changes requested):
Hey, according to the [docs](https://github.com/express-rate-limit/express-rate-limit), since we are behind cloudflare, the IP requests would be the IP of the load balancer. I believe we would need to add a `app.set('trust proxy', numberOfProxies)` in order to make this a non-global rate limit

src/routes/v2/auth.js
```diff
@@ -115,7 +116,7 @@ class AuthRouter {
 
   getRouter() {
     const router = express.Router()
-
+    router.use(this.rateLimiter)
```
alexanderleegs, line 119: Thoughts on using the rate limiter purely on the `/verify` endpoint, since this is the brute-forceable component? The others can't really be brute forced, outside of sending email otps repeatedly via the `/login` endpoint, and we should always allow `/logout` and `/whoami`

src/services/utilServices/RateLimiter.ts
```diff
@@ -0,0 +1,16 @@
...
+// https://github.com/express-rate-limit/express-rate-limit
+// Also, note that our production environment has 2 instances
+// and the rate limiter uses an in memory store,
+// so our effective limit is 100 * 2.
+// This also implies that a client can hit the limit on 1 server
+// but not on the other, leading to inconsistent behaviour.
+// eslint-disable-next-line import/prefer-default-export
+export const rateLimiter = rateLimit({
+  windowMs: 15 * 60 * 1000, // 15 minutes
+  max: 100, // Limit each IP to 100 requests per `window` (here, per 15 minutes)
```
alexanderleegs, line 13: We can sync up the window with the time window for otps, so that they reset at the same intervals.

Also can consider dropping the max  number of requests (maybe to 5? which should be more than enough) if we go with restricting only the actual verification

alexanderleegs (changes requested):
Left some comments on endpoints to restrict!

seaerchin:
> Hey, according to the [docs](https://github.com/express-rate-limit/express-rate-limit), since we are behind cloudflare, the IP requests would be the IP of the load balancer. I believe we would need to add a `app.set('trust proxy', numberOfProxies)` in order to make this a non-global rate limit

yea you're entirely right and i missed this! just to confirm there are actually **2** proxies right? (aws load balancer as well as cloudflare reverse proxy)

src/routes/v2/auth.js
seaerchin: no strong opinion here, i can do this also! only concern is if the testers say something about spam (cos if we don't restrict, any malicious actor can spam to ask for OTP).

lmk if you think this is important enough to fix or can we leave as is

src/services/utilServices/RateLimiter.ts
seaerchin: i'm gonna change it to env var but holding off on `max` (see my point above)

src/services/utilServices/RateLimiter.ts
```diff
@@ -9,7 +11,13 @@ import rateLimit from "express-rate-limit"
 // but not on the other, leading to inconsistent behaviour.
 // eslint-disable-next-line import/prefer-default-export
 export const rateLimiter = rateLimit({
-  windowMs: 15 * 60 * 1000, // 15 minutes
+  windowMs:
+    parseInt(
+      process.env.AUTH_TOKEN_EXPIRY_DURATION_IN_MILLISECONDS ||
+        DEFAULT_AUTH_TOKEN_EXPIRY_MILLISECONDS.toString(),
+      10
```
kishore03109, line 18: nit: Hey, rather than defining an int, parsing it to string and then parsing it back into an int, could we just use the int directly?

src/constants/infra.ts
```diff
@@ -0,0 +1,2 @@
+// NOTE: 1 for aws load balancer and another for cloudflare reverse proxy.
+export const NUM_PROXIES = 2
```
kishore03109, line 2: minor nit: shall we use default export?

kishore03109 (approved):
Hey, mostly lgtm with nits.
non nit: may I please trouble you to test this on staging before pushing this to develop? I think a trivial manual testing is also fine!

src/services/utilServices/RateLimiter.ts
seaerchin: we actually can't cos `parseInt` requires a string (despite being able to return `NaN`). 

another alternative is to do `parseInt(... || "", 10) ??  DEFAULT_AUTH_TOKEN_EXPIRY_MILLISECONDS`, which also seems jank. lmk which one you'd rather have

seaerchin:
> Hey, mostly lgtm with nits. non nit: may I please trouble you to test this on staging before pushing this to develop? I think a trivial manual testing is also fine!

this isn't gonna be merged into `develop` but `vapt`! also, writing load testing here would involve either writing a script or changing the given `max` parameter. would the provided test cases suffice?

src/constants/infra.ts
seaerchin: think i'd rather not here - we should default export for `NamedFiles`, which contain a primary export and potentially other secondary exports (eg: our routers/services in be or our components/pages in fe) but for constants, they should all be equal (+ we might add more here in future) so would rather not

src/services/utilServices/RateLimiter.ts
kishore03109: Arh, ok. Nah, am ok with this

kishore03109 (approved):
I think the unit tests are sufficient. 
What I thought would be safer was to test if the number of proxies is _**indeed**_ 2. 
Concretely, if we can just verify that our otp rate limits work in staging, ie, the limit is a per IP rather than a global one, I think this is fine. 

My concern is that this is something that we cannot have unit tests for, but it will cause issues in production!

seaerchin:
> I think the unit tests are sufficient. What I thought would be safer was to test if the number of proxies is _**indeed**_ 2. Concretely, if we can just verify that our otp rate limits work in staging, ie, the limit is a per IP rather than a global one, I think this is fine.
> 
> My concern is that this is something that we cannot have unit tests for, but it will cause issues in production!

sure, i was gonna let horangi check this AHHA but i'll test it out on `staging` later today or tomorrow!

src/routes/v2/auth.js
alexanderleegs: I think it's important to remove the limits on `logout` and `whoami` - the rest i agree though!

kishore03109:
> > I think the unit tests are sufficient. What I thought would be safer was to test if the number of proxies is _**indeed**_ 2. Concretely, if we can just verify that our otp rate limits work in staging, ie, the limit is a per IP rather than a global one, I think this is fine.
> > My concern is that this is something that we cannot have unit tests for, but it will cause issues in production!
> 
> sure, i was gonna let horangi check this AHHA but i'll test it out on `staging` later today or tomorrow!

HAHAHAH sounds good, we paid for this anyways LOL

PR #616, opened 2023-01-19
fix: nocache

prestonlimlianjie:
## Problem

This PR adds a cache-control: no-cache, no-store setting to the HTTP responses from the API server. This was a low-finding from our VAPT report.

## Solution

We used the `nocache` [library](https://github.com/helmetjs/nocache) from the HelmetJS team.

## Before & After Screenshots

**BEFORE**:
The HTTP response from the API server does not have any `Cache-Control` HTTP header.

Screenshot 2023-01-19 at 11 03 56 AM

**AFTER**:
The HTTP response from the API server now has a `Cache-Control` HTTP header with a `no-cache, no-store` configuration.
Screenshot 2023-01-19 at 11 04 32 AM

## Tests

Inspect the HTTP response headers either via an API client or the browser network tool; all responses should now contain the new `Cache-Control` header.

**New dependencies**:

- `nocache` : A [HelmetJS package](https://github.com/helmetjs/nocache) that disables browser caching out of the box


seaerchin (approved):
Lgtm but having both headers means that we never cache anything, which might potentially degrade performance. Not a big issue as we should solve real problems first but just in case load times get long, we might want to see if we only want the headers set for certain routes etc

PR #621, opened 2023-01-26
fix(server): shift api logger to be initialised within each subrouter

### LINKED ISSUE #620: [Bug] Logging error
Currently, user's name is undefined.

seaerchin:
## Problem
Previously, our logs weren't capturing the github username of the isomer user. This led to the github id being logged as `undefined` in our logs.

Closes #620 

## Solution
Change `apiLogger` to be initialised after `verifyJwt`; this was done through making it a dependency of the required routers (rather than importing) to keep in line w/ existing code conventions

### DIFF
--- a/src/routes/v1/authenticated/index.js
+++ b/src/routes/v1/authenticated/index.js
@@ -3,14 +3,20 @@ const express = require("express")
 const sitesRouter = require("@routes/v1/authenticated/sites")
 const { UsersRouter } = require("@routes/v2/authenticated/users")
 
-const getAuthenticatedSubrouter = ({ authMiddleware, usersService }) => {
+const getAuthenticatedSubrouter = ({
+  authMiddleware,
+  usersService,
+  apiLogger,
+}) => {
   // Workaround - no v1 users router exists
   const usersRouter = new UsersRouter({ usersService })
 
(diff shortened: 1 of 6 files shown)


alexanderleegs (commented):
Hmm this implementation looks good for fixing the bug, but I think we lose the logging over auth endpoints (e.g. `login` especially, which we might want to monitor for repeated attempts from the same ip), do you think it's worth adding the logger to the `/v2/auth` router as well?

seaerchin:
> Hmm this implementation looks good for fixing the bug, but I think we lose the logging over auth endpoints (e.g. `login` especially, which we might want to monitor for repeated attempts from the same ip), do you think it's worth adding the logger to the `/v2/auth` router as well?

oh yea this is a good point, will add it back in!

alexanderleegs:
> > Hmm this implementation looks good for fixing the bug, but I think we lose the logging over auth endpoints (e.g. `login` especially, which we might want to monitor for repeated attempts from the same ip), do you think it's worth adding the logger to the `/v2/auth` router as well?
> 
> oh yea this is a good point, will add it back in!

whoops this should also be added to `/v1/auth/` for now!

alexanderleegs (approved):
lgtm

PR #622, opened 2023-01-26
docs(readme): update readme instructions for running app

seaerchin:
## Problem
app requires more setup to run. this isn't well documented so this pr adds the required steps

## Solution
1. delete unused commands
2. add set up steps 
3. combine required commands


package.json
```diff
@@ -112,7 +111,6 @@
     "lint-staged": "^11.1.2",
     "prettier": "2.2.1",
     "regenerator-runtime": "^0.13.9",
-    "serverless": "^3.22.0",
     "serverless-offline": "^11.0.1",
     "serverless-offline-lambda": "^1.0.6",
     "serverless-plugin-typescript": "^2.1.2",
```
seaerchin, line 116: @kishore03109 double checking - all the `serverless-xxx` stuff can be removed?

package-lock.json
```diff
@@ -4,16 +4,6 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
-    "2-thenable": {
```
alexanderleegs, line 7: Just to double check, are these dependencies all unused?

seaerchin: not entirely sure how these came to be - if they're in our `deps` and we can't find a relevant import in our codebase, then i'm assuming they're unused (subpackages should specify their own deps)

package.json
kishore03109: yup, can be removed!

kishore03109 (approved):

alexanderleegs (approved):

package-lock.json
seaerchin: removed all extra dpes

PR #623, opened 2023-02-01
feat: add webp to allowed images

alexanderleegs:
This PR adds in webp as an allowed format for images that can be uploaded in Isomer. To be reviewed in conjunction with PR [#1180](https://github.com/isomerpages/isomercms-frontend/pull/1180) on the isomercms-frontend repo.

### DIFF
--- a/src/utils/file-upload-utils.js
+++ b/src/utils/file-upload-utils.js
@@ -14,6 +14,7 @@ const ALLOWED_FILE_EXTENSIONS = [
   "tif",
   "bmp",
   "ico",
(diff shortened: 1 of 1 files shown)


seaerchin (approved):
lgtm for now but take note that `webp` is not backwards compat (see [here](https://developers.google.com/speed/webp/faq)) and google suggests we use a legacy format otherwise. 

this might be troublesome for us (need to mandate legacy + `webp`) so perhaps we might wanna just have a placeholder in place for this?

PR #624, opened 2023-02-02
Fix/make login response indistinguishable

alexanderleegs:
This PR modifies the handling of our otp endpoint to always return a 200 status - this serves to hide the validity of an email address that was previously exposed in the case of blacklisted emails, when an error was thrown. Additional logging on email failure has also been added in this PR.

### DIFF
--- a/src/routes/v2/auth.js
+++ b/src/routes/v2/auth.js
@@ -81,7 +81,7 @@ class AuthRouter {
       await this.authService.sendOtp(email)
     } catch (err) {
       // Log, but don't return so responses are indistinguishable
-      logger.log(
+      logger.error(
         `Error occurred when attempting to login user ${email}: ${err}`
       )
     }
--- a/src/routes/v2/auth.js
+++ b/src/routes/v2/auth.js
@@ -7,6 +7,8 @@ const { attachReadRouteHandlerWrapper } = require("@middleware/routeHandler")
 const { FRONTEND_URL } = process.env
 const { isSecure } = require("@utils/auth-utils")
 
+const logger = require("@root/logger/logger")
+
 const AUTH_TOKEN_EXPIRY_MS = parseInt(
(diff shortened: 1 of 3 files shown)


src/routes/v2/auth.js
```diff
@@ -75,7 +77,14 @@ class AuthRouter {
   async login(req, res) {
     const { email: rawEmail } = req.body
     const email = rawEmail.toLowerCase()
-    await this.authService.sendOtp(email)
+    try {
+      await this.authService.sendOtp(email)
+    } catch (err) {
+      // Log, but don't return so responses are indistinguishable
+      logger.log(
+        `Error occurred when attempting to login user ${email}: ${err}`
```
seaerchin, line 85: we should probably have structured logs so it makes querying easier in the future. 

also, could we use `logger.error` here?

seaerchin (commented):
did we have a reason for not using `logger.error`? this is internal to us so perhaps we might wanna highlight them for troubleshooting purposes. 

also, should probably define a simple object for our initial logging format (see [here](https://github.com/winstonjs/winston#logging))

src/routes/v2/auth.js
alexanderleegs: Swapped to logger.error in https://github.com/isomerpages/isomercms-backend/pull/624/commits/e2f81734a80b758ec845f84403c7457850aaceaa!

Currently our logger is initialised with no params, and our custom error middleware adds in the timestamp for each message - what kind of format were you thinking of adding?

seaerchin: was hoping for a object of something like
```ts
message: string,  
meta: Record<string, int>
timestamp: number
``` 

or something

seaerchin (approved):
lgtm for now; we can do the shift to structured logging sometime later

PR #630, opened 2023-02-15
Fix: query for site through repo

alexanderleegs:
This PR fixes an issue where we were querying for the github repo name of a site through the `Site` table (which contains the human readable version of the name) instead of through the `Repo` table (which contains the desired string). All existing queries referencing the `Site` table for the github sitename have been changed to reference the `Repo` table instead.


seaerchin (approved):
lgtm but could you remind me again what `Site.name` references?

PR #631, opened 2023-02-16
feat: Added virus scan functionality for file upload (VAPT finding)

QiluXie:
## Problem
Fixed the  [VAPT finding](https://docs.google.com/document/d/17HYo9aJTdu74e6rv3I8RonGj6XoXDCTa3RjazNiKGPk/edit#heading=h.9jtqo5blys6) - 03: Unrestricted File Upload (Medium)

## Solution

Implemented virus scan feature when user uploads file or image using the cloudmersive API - [API reference](https://api.cloudmersive.com/docs/virus.asp)

**Breaking Changes**

## Tests

Currently due to another issue that adds an extra dot to the file name of newly uploaded file, the file uploading feature is not working in VAPT branch. The issues is addressed in [a separate PR](https://github.com/isomerpages/isomercms-frontend/pull/1184/files) in the frontend codebase.
To test this PR, please comment out the lines of media name check (line 23 & line 112) as a temporary measure as shown in the following screenshot. 
Screenshot 2023-02-16 at 4 38 18 PM
Screenshot 2023-02-16 at 4 38 27 PM

- `CM_API_KEY` : API Key for calling Cloudmersive API 

**New dependencies**:

- `cloudmersive-virus-api-client` : cloudmersive client sdk


.env-example
```diff
@@ -50,3 +50,6 @@ AWS_REGION="ap-southeast-1"
 export DD_ENV="local"
 export DD_SERVICE="isomer"
 export DD_TAGS="service:isomer"
+
+#Cloudmersive
```
seaerchin, line 54: super ultra nit: can add an extra space? 

also, do take note that we need to update the 1PW

src/services/fileServices/MdPageServices/MediaFileService.js
```diff
@@ -21,7 +21,18 @@ class MediaFileService {
 
   async create(sessionData, { fileName, directoryName, content }) {
     this.mediaNameChecks({ directoryName, fileName })
-    const sanitizedContent = await validateAndSanitizeFileUpload(content)
+
+    const [schema, fileContent] = content.split(",")
```
seaerchin, line 25: could we add a comment on why this is required and what the `schema` is? separately, this destructuring assignment always takes the second element as the `fileContent` - are we guaranteed that this is so?

src/services/fileServices/MdPageServices/MediaFileService.js
```diff
@@ -21,7 +21,18 @@ class MediaFileService {
...
+    const fileBuffer = Buffer.from(fileContent, "base64")
+
+    //Scan file for virus - cloudmersive API
+    const virusScanRes = await scanFileForVirus(fileBuffer)
```
seaerchin, line 29: could i ask why we split this from `validateAndSanitizeFileUpload`? placing it there seems to make sense actually

src/utils/file-upload-utils.js
```diff
@@ -11,10 +14,28 @@ const ALLOWED_FILE_EXTENSIONS = [
   "bmp",
   "ico",
 ]
+const defaultClient = CloudmersiveVirusApiClient.ApiClient.instance;
 
-const validateAndSanitizeFileUpload = async (data) => {
-  const [schema, content] = data.split(",")
-  const fileBuffer = Buffer.from(content, "base64")
+// Configure API key authorization: Apikey
+var Apikey = defaultClient.authentications['Apikey']
```
seaerchin, line 20: `var` should **never** be used because it's in the global scope :monkas: (see [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/var) for details)

also, since `ApiKey` seems to be an instance, could we have it in `camelCase`? (see [here](https://www.robinwieruch.de/javascript-naming-conventions/) for casing conventions)

src/utils/file-upload-utils.js
```diff
@@ -11,10 +14,28 @@ const ALLOWED_FILE_EXTENSIONS = [
...
+    apiInstance.scanFile(fileBuffer, function(error, data, response) {
+      if (error) {
+        logger.error('Error when calling cloudmersive API')
+        failure(error)
+      } else {
+        logger.info('Virus Scan API called successfully')
+        success(data)
+      }
+    })
+})}
```
seaerchin, line 36: we probably wanna check the environment and if it's `dev || test`, just return a clean result to avoid calling the actual endpoint.

also, should we have an abstraction for the initialisation of the client (such as a function) so that we can possibly throw if initialisation failed or return a mock client in `dev` or `test` environments

seaerchin (commented):
some comments! sidenote, i think the formatter might be broken for you because comments should automatically have a space after the double slash (`//`) so do take note of that

.env-example
QiluXie: Hey good catch, I did the right spacing in my local .env but forgot to enforce it in .env-example. Also fixed all comment space after double slash issue. 
Fixed it in 6ebc12a9cfe977f4a309329fc571ae04322bb93c

src/services/fileServices/MdPageServices/MediaFileService.js
QiluXie: Hey @seaerchin , this line is actually not written by me. It was in the original code, since my file virus scan also needs the fileContent, so I abstracted this line out of the `validateAndSanitizeFileUpload`  function to avoid duplicate code. 
My guess is it splits the incoming content stream by a comma, and the second part would always be the actual file content (encoded base64 string) since it's a static split. 
Not sure what the `schema` is, might be the algorithm used to encode the file. @alexanderleegs do you know what the `schema` is?

src/utils/file-upload-utils.js
QiluXie: You are absolutely right. I took the sample code from [CloudMersive API reference](https://api.cloudmersive.com/docs/virus.asp) and forgot to change this line before committing. Also changed `authentications['Apikey']` to a dot reference `authentications.Apikey` 
Fixed in 6ebc12a9cfe977f4a309329fc571ae04322bb93c

src/services/fileServices/MdPageServices/MediaFileService.js
QiluXie: Sure, thanks for asking. I'll explain my thoughts in two aspects here: 

1. The functionalities of these two file checks are fundamentally different. `validateAndSanitizeFileUpload` is to validate the file type & extension, and sanitize the file content. `scanFileForVirus` is to perform a post call to [Cloudmersive API endpoint ](https://api.cloudmersive.com/docs/virus.asp)to scan the file for virus. Depending on the use case, these two functions can be used independently and separately. For example, if a file uploading feature doesn't care about the file type & extension, but only want to perform virus scan, it can invoke the scanFileForVirus function and complete the job. Therefore I think logically it's better to split them into two different re-usable functions than combining them together. 
2. The way those methods were implemented are also different. `validateAndSanitizeFileUpload` is doing synchronous local check & purify work for the file content, while `scanFileForVirus` is doing a remote API call to [Cloudmersive ](https://api.cloudmersive.com/docs/virus.asp)using its sdk which requires a callback function to get the results. Therefore I need to use a Promise and await the result from the caller function. Combining them might make it challenging to enforce the order of the checks, regardless of what order we prefer. Hope that makes sense.

src/utils/file-upload-utils.js
QiluXie: Do we want to avoid calling the third-party API in dev & test environment? 
I assumed at least in dev environment, we should have a working environment where the API call is actually working. So if it's not working, we are able to catch it. I understand that we want to prevent dev env from tempering data store in other envs. But this call is a post only call to Cloudmersive and doesn't really affect the data integrity in our backend. With Cloudmersive premium account, we have 20k call per month and the limit shouldn't be an issue. I think we can consider performing this call in dev & test env. 

For the initialization of` apiInstance`, it's actually a local class instance construction rather than a remote initialisation. Check the reference from Cloudmersive:
`The_Cloudmersive_Virus_Scan_API_lets_you_scan_files_and_content_for_viruses_and_identify_security_issues_with_content_.<br>`
   `* The <code>index</code> module provides access to constructors for all the classes which comprise the public API.`
So in this case, the construction of apiInstance should be a possible failing point, `apiInstance.scanFile` is calling the remote API and could fail, which we are already catching the error in the callback. Hope that makes sense.

QiluXie:
@seaerchin thanks for all your comments, very helpful! I fixed some of them in a new commit and provided explanation to the rest, please take a look and let me know. Thanks!

src/services/fileServices/MdPageServices/MediaFileService.js
```diff
@@ -21,7 +21,18 @@ class MediaFileService {
 
   async create(sessionData, { fileName, directoryName, content }) {
     this.mediaNameChecks({ directoryName, fileName })
-    const sanitizedContent = await validateAndSanitizeFileUpload(content)
+
+    const [schema, fileContent] = content.split(",")
+    const fileBuffer = Buffer.from(fileContent, "base64")
+
+    // Scan file for virus - cloudmersive API
+    const virusScanRes = await scanFileForVirus(fileBuffer)
```
harishv7, line 29: should we log intermediate results for backtracing later since this is a "new" feature

src/utils/file-upload-utils.js
```diff
@@ -1,3 +1,7 @@
+import logger from "@logger/logger"
+
+const { CM_API_KEY } = process.env
```
harishv7, line 3: Curious, what happens if this key does not exist. just thinking if it will break the program. do we have a "env" checker function when our server starts up.

That will give us confidence we can directly unwrap the env vars

src/services/fileServices/MdPageServices/MediaFileService.js
harishv7: I mean to log the return data from Cloudmersive api call where the file is ok or not

src/utils/file-upload-utils.js
```diff
@@ -11,10 +15,27 @@ const ALLOWED_FILE_EXTENSIONS = [
...
+// Configure API key authorization: Apikey
+const apikey = defaultClient.authentications.Apikey
+apikey.apiKey = CM_API_KEY
+
+const apiInstance = new CloudmersiveVirusApiClient.ScanApi()
+
+const scanFileForVirus = (fileBuffer) => new Promise((success, failure) => {
+    apiInstance.scanFile(fileBuffer, (error, data, response) => {
+      if (error) {
+        logger.error('Error when calling cloudmersive API')
```
harishv7, line 29: nit: "Cloudmersive API"

src/utils/file-upload-utils.js
```diff
@@ -11,10 +15,27 @@ const ALLOWED_FILE_EXTENSIONS = [
...
+        failure(error)
+      } else {
+        logger.info('Virus Scan API called successfully')
```
harishv7, line 32: nit: "Cloudmersive API called successfully"

src/utils/file-upload-utils.js
harishv7: We should also log the error, example: ``logger.error('Error when calling Cloudmersive API : ${error.message}`)`

src/utils/file-upload-utils.js
```diff
@@ -11,10 +15,27 @@ const ALLOWED_FILE_EXTENSIONS = [
   "bmp",
   "ico",
 ]
+const defaultClient = CloudmersiveVirusApiClient.ApiClient.instance;
```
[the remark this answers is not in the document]
QiluXie: Sounds good! Fixed it in fe67cacf0628aea53ec3f4813423e1616f791c2e

src/utils/file-upload-utils.js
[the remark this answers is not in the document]
QiluXie: Sounds good! Fixed it in fe67cacf0628aea53ec3f4813423e1616f791c2e

QiluXie:
> Could we also keep the lockfile version at 1 for now? Otherwise it'll be difficult to resolve the merge into dev later

@alexanderleegs downgraded the lockfile version to 1 in package-lock.json in fe67cacf0628aea53ec3f4813423e1616f791c2e

src/utils/file-upload-utils.js
QiluXie: Sure, @kishore03109 added a similar comment and I have already added the necessary check for the required env (API key), if the env variable is not present, an error will be thrown and the API won't be called (if we call it, it will return 401 unauthorized and it's a waste of time and resource). Please check the fix in fe67cacf0628aea53ec3f4813423e1616f791c2e

src/services/fileServices/MdPageServices/MediaFileService.js
QiluXie: Sure, that sounds good. The unused declaration should be removed. I fixed it in 47bba00269aa4b540de793007d6952bf621b2d1b

@alexanderleegs fileContent is the intended one and scheme is never used.

src/services/fileServices/MdPageServices/MediaFileService.js
QiluXie: I absolutely agree with you for the type script part. But I think these two functions should be in the same `file-upload-utils` file since they are all related and used by file uploading functionality. This file, together with all the other `js` files under `utils` folder should all be migrated to `ts` file in the future for better type checking/enforcement. 
But it would likely to be a separate effort to refactor the codebase which is beyond the scope of this PR. We will plan for a proper `Typescript` migration for the entire codebase (or at least the `util` folder).  What do you think?

seaerchin: sure, if we're planning for a one off approach then i guess this is ok; but we do need to ensure that we have proper time allocated for said migration so that we don't have small + incremental pile ups. i'm just worried that the migration won't happen so soon but we keep making allowances like this, which eventually makes the migration have a bigger scope than expectedc

src/utils/file-upload-utils.js
seaerchin: that works but i think the linting is abit off @__@ do you have autoformat on save configured @QiluXie?

src/services/fileServices/MdPageServices/MediaFileService.js
QiluXie: Yeah I understand your concern and thanks for sharing it. That's why we need to build a proper backlog to keep track of these issues and plan our capacity to do the migration. We will work on it and plan it together in the near future.

src/utils/file-upload-utils.js
QiluXie: Thanks for the help on setting up the auto-format by prettier. The link/format issues are fixed in the latest commit here: d9994ddf40a99b8c5e573a9a49a37c961068424c

QiluXie:
@seaerchin @harishv7 @alexanderleegs @kishore03109 All the current comments are addressed. Please help to approve the PR if there's no additional comment. Thanks!

harishv7 (approved):

alexanderleegs:
> > Could we also keep the lockfile version at 1 for now? Otherwise it'll be difficult to resolve the merge into dev later
> 
> @alexanderleegs downgraded the lockfile version to 1 in package-lock.json in [fe67cac](https://github.com/isomerpages/isomercms-backend/commit/fe67cacf0628aea53ec3f4813423e1616f791c2e)

actually i think we might need to do a reinstall with a lower node version because the formatting changes will be annoying to resolve later - for some reason it seems like `npm i` takes the system default version of node rather than the one specified by nvm, I had to explicitly use `which npm` to find the path of the node 14 installation and use that path explicitly to install instead (`<PATH> install`)

Hopefully all of this is resolved once we upgrade our node version!

alexanderleegs (approved):

PR #636, opened 2023-02-22
Feat/otp mechanism

harishv7:
## Problem

Fixes the VAPT finding for OTP not being invalidated due to TOTP.

## Solution

We now use a random OTP generator with creation of a OTP table that stores the user email/mobile number to hashed OTP mapping.

We have introduced an OTP service and OTP repository to facilitate the flow.

**Breaking Changes**

  - Details ...

## Tests

Please run npm run test Users.spec.ts --runInBand to confirm functionality

## Deploy Notes

Under .env:

- MAX_NUM_OTP_ATTEMPTS=5
- OTP_EXPIRY=900000

**Migrations**
Please ensure database migration is done when merging this PR. This adds a new table called Otps.

**New dependencies**:

- bcrypt : for salting + hashing of otp for DB storage

**New dev dependencies**:

- `@types/bcrypt` : facilitates types for bcrypt usage


src/services/identity/UsersService.ts
```diff
@@ -2,24 +2,40 @@ import { Op, ModelStatic } from "sequelize"
...
 import MailClient from "@services/utilServices/MailClient"
 
+import OtpService from "./OtpService"
+
+const { OTP_EXPIRY } = process.env
+
+const PARSED_EXPIRY = parseInt(OTP_EXPIRY!, 10) ?? undefined
+
+const MAX_NUM_OTP_ATTEMPTS =
+  parseInt(process.env.MAX_NUM_OTP_ATTEMPTS!, 10) ?? 5
```
seaerchin, line 19: nit: could we inline the `process.env.MAX_NUM_OTP_ATTEMPTS` to the `process.env` destructuring above?

seaerchin, line 19: the ts complain here is correct actually cos these are env vars and are not guaranteed to exist... 

we'd have to write ugly code `parseInt(OTP_EXPIRY || "", 10) ?? undefined` but i think it's ok

src/services/identity/UsersService.ts
```diff
@@ -188,26 +210,106 @@ class UsersService {
...
+    if (!otpEntry) {
+      // create new entry
+      await this.createOtpEntry(email, OtpType.Email, hashedOtp)
+    } else {
+      await otpEntry?.update({
+        hashedOtp,
+        attempts: 0,
+        expiresAt: this.getOtpExpiry(),
+      })
+    }
```
seaerchin, line 226: feels like this should belong inside the `OtpService` for better separation of concerns; wdyt?

src/services/identity/UsersService.ts
```diff
@@ -188,26 +210,106 @@ class UsersService {
...
-    return this.otp.verify(value, otp)
+  async verifyOtp(key: string, keyType: OtpType, otp: string) {
+    let otpEntry
+    if (keyType === OtpType.Email) {
+      otpEntry = await this.otpRepository.findOne({ where: { email: key } })
+    } else {
+      otpEntry = await this.otpRepository.findOne({
+        where: { mobileNumber: key },
+      })
+    }
```
seaerchin, line 264: could we do 
```ts
const otpFindParams = keyType === OtpType.Email ? {email: key} : {mobileNumber: key}
const otpEntry = await this.otpRepository.findOne(otpFindParams)
```

src/services/identity/UsersService.ts
```diff
@@ -188,26 +210,106 @@ class UsersService {
...
+
+    if (!otpEntry?.hashedOtp) {
+      // TODO: Change to use AuthError after FE fix
+      throw new BadRequestError("Hashed OTP not found")
+    }
+
+    if (
+      otpEntry.attempts !== null &&
```
seaerchin, line 272: think this might be extra cos we don't allow `null` on the `otpEntry.attempts`

src/services/identity/UsersService.ts
```diff
@@ -188,26 +210,106 @@ class UsersService {
...
+      otpEntry = await this.otpRepository.findOne({ where: { email: key } })
+    } else {
+      otpEntry = await this.otpRepository.findOne({
+        where: { mobileNumber: key },
+      })
+    }
+
+    if (!otpEntry?.hashedOtp) {
+      // TODO: Change to use AuthError after FE fix
+      throw new BadRequestError("Hashed OTP not found")
```
seaerchin, line 268: i'm more in favour of returning errors rather than throwing them in services - the router can then map these errors into the appropriate error codes/messages to our users so the control flow is clear; wdyt about this?

src/services/identity/UsersService.ts
```diff
@@ -188,26 +210,106 @@ class UsersService {
...
+        hashedOtp,
+        expiresAt: this.getOtpExpiry(),
+      })
+    } else {
+      await this.otpRepository.create({
+        mobileNumber: key,
+        hashedOtp,
+        expiresAt: this.getOtpExpiry(),
+      })
+    }
```
seaerchin, line 312: wdyt about shifting to `OtpService`? also, side note - i feel like providing a `verifyEmailOtp` + `verifyMobileOtp` (and similarly for the create) method instead of the current approach might be clearer; wdyt?

src/services/identity/UsersService.ts
```diff
@@ -188,26 +210,106 @@ class UsersService {
...
+    // increment attempts
+    await otpEntry.update({ attempts: otpEntry.attempts + 1 })
+
+    const isValidOtp = await this.otpService.verifyOtp(otp, otpEntry.hashedOtp)
+    if (!isValidOtp) {
+      // TODO: Change to use AuthError after FE fix
+      throw new BadRequestError("OTP is not valid")
+    }
+
+    return true
```
seaerchin, line 288: we still seem to maintain the entry? so if the user logs in successfully -> token expires -> tries to log in again, they won't be able to cos the entry is still present w/ unique constraint

src/database/migrations/20230214055456-create-otps.js
```diff
@@ -0,0 +1,49 @@
...
+      email: {
+        allowNull: true,
+        unique: true,
+        type: Sequelize.STRING,
+      },
+      mobile_number: {
+        allowNull: true,
+        unique: true,
+        type: Sequelize.STRING,
+      },
```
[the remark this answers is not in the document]
harishv7: I felt that both email and mobile_number will be unique for our use case. Some qns:

1. What would be the overhead here?
2. We are tracking the attempts by email and mobile num. Hence, per-otp, we should have them as unique right? I can't think of a case where it is not unique (e.g. where we have multiple same emails with otps or multiple same mobile numbers with otps)

src/database/models/Otp.ts
```diff
@@ -0,0 +1,61 @@
...
+  Column,
+  CreatedAt,
+  DataType,
+  Model,
+  Table,
+  UpdatedAt,
+} from "sequelize-typescript"
+
+@Table({ tableName: "otps" })
+export class Otp extends Model {
```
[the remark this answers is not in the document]
harishv7: table is a collection of otp(s) => hence `otps` (lowercased for naming convention)
Model is a representation of a Single OTP => hence `Otp`

src/database/migrations/20230214055456-create-otps.js
```diff
@@ -0,0 +1,49 @@
+/** @type {import('sequelize-cli').Migration} */
```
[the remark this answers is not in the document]
harishv7: ah I wasn't aware of this requirement.

@alexanderleegs was this requirement present before as well? should we modify to reset after the `last_sent + REQ_WAIT_TIME`?
If it is within this buffer time, we should reject reset requests. Do we send a BadReqError for that?

src/integration/Users.spec.ts
```diff
@@ -2,7 +2,7 @@ import express from "express"
 import mockAxios from "jest-mock-axios"
```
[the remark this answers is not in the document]
harishv7: Okay will add those.

For point 2,
- We only have 1 OTP at a time. Why do we have >=1 otp?

src/services/identity/UsersService.ts
harishv7: okay sure!

harishv7: Yea, and this brings me back to the other point in the virus scan PR - we need to ensure env vars are all present at time of startup (in index.ts) so that we can confidently use `!` here

I will add the ugly code suggestion for now haha

src/services/identity/UsersService.ts
harishv7: Hmm, it was intentional to keep OtpService agnostic of the biz logic involved here - e.g. to find by email and then performing a reset. This operation can be summarised as "reset the attempts for an user" which is under the UsersService's purview

The OtpService was modelled to provide abstractions over the Otp operations - generating and verifying Otps. If you realise OtpService does not interact with OtpRepository. It merely provides operations required for the creations/verifications.

src/services/identity/UsersService.ts
harishv7: good suggestion, will update

src/services/identity/UsersService.ts
harishv7: I felt is better to be more defensive where possible, but you're right. I dont think this case will ever occur. can remove the null check

src/services/identity/UsersService.ts
harishv7: Yea, but this is a style change across all layers. It should be a paradigm that we follow. In fact, going further, it should be an AuthError that is returned (not a BadReq). 

I am not sure if we should tackle it in this PR though. This should be a codebase-wide change where errors are an "expected" return type across the layers of services

src/services/identity/UsersService.ts
harishv7: i think similar explanation as above about the intentions of the OtpService

Regarding the methods for email and mobile - I felt enums are a more scalable way. But thinking of it, we only verify email and mobile mostly. There is unlikely to be a 3rd use case. So yea, agree with you that methods might be clearer.

So to clarify, are you suggesting to remove the `OtpType` enum and just directly have separate methods?

src/services/identity/UsersService.ts
harishv7: everytime we generate a new otp, we update otp entry in the DB to have new Otp + reset attempts to 0

harishv7:
```
// Reset attempts to login
    const otpEntry = await this.otpRepository.findOne({ where: { email } })
    if (!otpEntry) {
      // create new entry
      await this.createOtpEntry(email, OtpType.Email, hashedOtp)
    } else {
      await otpEntry?.update({
        hashedOtp,
        attempts: 0,
        expiresAt: this.getOtpExpiry(),
      })
    }
```

src/services/identity/UsersService.ts
```diff
@@ -188,26 +210,98 @@ class UsersService {
...
+
+    if (!otpEntry?.hashedOtp) {
+      // TODO: Change to use AuthError after FE fix
+      throw new BadRequestError("Hashed OTP not found")
+    }
+
+    if (otpEntry.attempts >= PARSED_MAX_NUM_OTP_ATTEMPTS) {
+      // TODO: Change to use AuthError after FE fix
+      throw new BadRequestError("Max number of attempts reached")
+    }
```
alexanderleegs, line 269: I realised we're not verifying that the OTP has yet to expire - I think we should be checking against `otpEntry.expiresAt > new Date()`?

src/services/identity/UsersService.ts
```diff
@@ -188,26 +210,98 @@ class UsersService {
...
+    // increment attempts
+    await otpEntry.update({ attempts: otpEntry.attempts + 1 })
+
+    const isValidOtp = await this.otpService.verifyOtp(otp, otpEntry.hashedOtp)
+    if (!isValidOtp) {
+      // TODO: Change to use AuthError after FE fix
+      throw new BadRequestError("OTP is not valid")
+    }
+
+    return true
```
alexanderleegs, line 280: We should also be invalidating the entry after a successful login, otherwise an attacker use the same otp if they hit our verify endpoint directly

src/services/identity/UsersService.ts
seaerchin: think we should use `convict` for env vars tbh and we have a (super) old PR up for this; i think we suffer from having an abundance of env vars, and we don't sanitise them, both of which might be solved w/ `convict`

src/database/models/Otp.ts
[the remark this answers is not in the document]
seaerchin: sure that makes sense! lets go with this convention in the future hehe

src/database/migrations/20230214055456-create-otps.js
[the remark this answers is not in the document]
seaerchin: the overhead here is having to manage the `unique` constraint on the services! i was thinking that we might wanna allow for logins from multiple devices but i realised cos we reset on new otp creation so it's alr possible

src/services/identity/UsersService.ts
seaerchin: i understand where you're coming from but seeing `UsersService` call `otpService.generateLoginOtpWithHash` and then `otpRepository.findOne` feels kinda weird @__@

i'm more inclined towards the NestJS way of gating the repository behind its service but wanted to hear your thouhghts on this

src/services/identity/UsersService.ts
seaerchin: sure, we can leave it till later since it's a small change (for now tm)

src/services/identity/UsersService.ts
seaerchin: yeap! it also helps to make the method name more descriptive and keep the method call itself simpler

src/integration/Users.spec.ts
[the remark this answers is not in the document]
seaerchin: if they call reset otp and we give another otp; we could simply overwrite the existing row per what you've done also

src/services/identity/UsersService.ts
seaerchin: yeap you're right! should just `destroy` the otp entry when login is successful tbh cos otherwise it'll still be valid

src/database/migrations/20230214055456-create-otps.js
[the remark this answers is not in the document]
alexanderleegs: Ah right this was something I missed as well - right now there's a restriction on the frontend for a re-request to only be sent after 1 minute, but calling the endpoint directly to request for a new otp repeatedly is still possible.

I'm wondering if we should leave this as is though, since the new rate limiter on this endpoint stops the user from requesting too often, and we also might not want to block a user who accidentally navigates away from the page after requesting for the otp from requesting a new one. What are your thoughts @seaerchin?

src/services/identity/UsersService.ts
kishore03109: nit: since we are currently not using `convict` or some code to ensure all env vars are loaded up in `index.js` could we add throw error if such a env var doesnt exist?

src/services/identity/UsersService.ts
```diff
@@ -188,26 +210,104 @@ class UsersService {
...
+    if (otpEntry.attempts && otpEntry.attempts >= MAX_NUM_OTP_ATTEMPTS) {
+      // TODO: Change to use AuthError after FE fix
+      throw new BadRequestError("Max number of attempts reached")
+    }
+
+    // increment attempts
+    await otpEntry.update({ attempts: otpEntry.attempts + 1 })
+
+    const isValidOtp = await this.otpService.verifyOtp(otp, otpEntry.hashedOtp)
+    // return isValidOtp
```
kishore03109: nit: unused code

src/services/identity/OtpService.ts
```diff
@@ -19,8 +19,10 @@ class OtpService {
     return { otp, hashedOtp }
   }
 
-  verifyOtp = async (otp: string, hashedOtp: string): Promise<boolean> =>
-    bcrypt.compare(otp, hashedOtp)
+  verifyOtp = async (otp: string, hashedOtp: string): Promise<boolean> => {
+    if (!otp || !hashedOtp) return false
+    return bcrypt.compare(otp, hashedOtp)
```
kishore03109, line 24: the documentation states `A promise to be either resolved with the comparison result salt or rejected with an Error`

should we add a try catch, and return `false` if there was any error thrown?

.env.test
```diff
@@ -11,6 +11,8 @@ export GITHUB_ORG_NAME="isomerpages"
 export GITHUB_BUILD_ORG_NAME="opengovsg"
 export GITHUB_BUILD_REPO_NAME="isomer-build"
 export ISOMERPAGES_REPO_PAGE_COUNT=3
+export MAX_NUM_OTP_ATTEMPTS=5
```
kishore03109, line 14: ooof I thought we removed tracking of `.env.test` HAHA

kishore03109 (changes requested):

kishore03109:
Oh and could we add the `runInBand` within package.json as well? 
I remember this was fixed somewhere but not sure why it isnt present here :(

.env.test
harishv7: apparently needed this for ci-tests to run

kishore03109: as in `.env.test` git ignored [here](bd95543b749d185f66d9e9a0c9a5cee839df716d) haha -> could we add this file in 1pw?

src/database/migrations/20230214055456-create-otps.js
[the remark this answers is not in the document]
seaerchin: i think the intent of gating the otp requests is to prevent potential abuse (try -> fail -> request new otp -> start again); i think as long as we prevent this specific form of abuse (w/ a frontend gate so lawful users still adhere to a reasonable limit), it's ok if we remove this particular one (as the rate limiter still gates it)

src/database/migrations/20230214055456-create-otps.js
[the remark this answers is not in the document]
harishv7: @seaerchin Yea you're right. You can still login as we reset the entry

@kishore03109 Why would it take 30 mins? Also, we support multiple machines just not like concurrently (like at the same time someone wants to login into 2 machines). I don't think that scenario is likely. The usual UX is to login machine by machine which this flow will be supporting

src/integration/Users.spec.ts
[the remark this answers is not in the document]
harishv7: @seaerchin added the tests

src/database/migrations/20230214055456-create-otps.js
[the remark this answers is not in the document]
harishv7: Okay, excluding this since we have rate limiter in place to prevent brute force or abuse

@seaerchin @alexanderleegs what's the rate limiter config though? Where can i find it?

src/services/identity/UsersService.ts
harishv7: @kishore03109 Am going to add the over arching solution - to check env vars at startup real soon as its pretty important. Will do up a PR next week

src/services/identity/UsersService.ts
harishv7: I think we need a revamp of the structure tbh.

- routers -> handlers -> services -> repositories -> models

That should be the flow.

Also, regarding: 

> seeing UsersService call otpService.generateLoginOtpWithHash and then otpRepository.findOne feels kinda weird

We should see services and repositories as like lego blocks that other services and repos can tap on. Meaning to say, OtpRepository is not meant to be only consumed by OtpService.

The service layer which is performing the biz logic should use the underlying blocks to execute the biz logic. UserService can interact with other services or utils.

But if you feel it's weird still, we can do away with OtpService in favour of OtpUtils or something haha. So that we don't see it as a service layer

wdyt?

src/services/identity/UsersService.ts
harishv7: @seaerchin okay added destroy after checking isValidOtp

src/database/migrations/20230214055456-create-otps.js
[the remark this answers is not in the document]
seaerchin: #610

src/services/identity/UsersService.ts
harishv7: good point, added!

src/services/identity/UsersService.ts
harishv7: added

src/database/models/Otp.ts
```diff
@@ -0,0 +1,61 @@
...
+    validate: {
+      notEmpty: true,
+    },
+  })
+  hashedOtp!: string
+
+  // tracks number of times user attempts to submit the OTP code and log in
+  @Column({
+    type: DataType.INTEGER,
+    defaultValue: 0,
```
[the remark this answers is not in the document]
harishv7: This only counts the attempts. Since they have not yet "attempted" use the otp, default value is 0.

src/services/identity/UsersService.ts
harishv7: oops, removed

src/services/identity/OtpService.ts
harishv7: The return type of the function is a Promise<boolean> and is the same as bcrypt compare

```
export declare function compare(data: string | Buffer, encrypted: string): Promise<boolean>;
```

So for the upstream users of this method, either way it will resolve to an boolean

.env.test
harishv7: I mean, the CI command to run is `source .env.test && jest` right. Then how will the ci-test run if we remove the .env.test?

harishv7:
> Oh and could we add the `runInBand` within package.json as well? I remember this was fixed somewhere but not sure why it isnt present here :(

I think @alexanderleegs mentioned this is already done on a different branch so when we complete the merge, we should have it. 

@alexanderleegs Can you confirm this, else I can add it in this PR too

src/database/migrations/20230214055456-create-otps.js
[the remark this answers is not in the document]
harishv7: Thanks @seaerchin

src/services/identity/UsersService.ts
harishv7: done

src/services/identity/UsersService.ts
```diff
@@ -188,26 +210,120 @@ class UsersService {
...
-    const message = `Your OTP is ${otp}. It will expire in ${expiry} minutes. Please use this to verify your mobile number`
+    const message = `Your OTP is ${otp}. It will expire in ${milliSecondsToMinutes(
+      PARSED_EXPIRY
+    )} minutes. Please use this to verify your mobile number`
     await this.smsClient.sendSms(mobileNumber, message)
   }
 
-  verifyOtp(value: string, otp: string) {
-    return this.otp.verify(value, otp)
+  private async verifyOtp(otpEntry: Otp | null, otp: string) {
```
harishv7, line 256: @seaerchin @alexanderleegs @kishore03109 let me know if this sequence of verifyOtp logic makes sense

src/services/identity/UsersService.ts
seaerchin: file issue

.env.test
seaerchin: supposed to have `.env.test` cos it's used for testing and we **shouldn't** be using secrets for testing anyway

src/services/identity/UsersService.ts
```diff
@@ -188,26 +210,120 @@ class UsersService {
...
+      throw new BadRequestError("Max number of attempts reached")
+    }
+
+    if (!otpEntry?.hashedOtp) {
+      await otpEntry.destroy()
+      throw new BadRequestError("Hashed OTP not found")
+    }
+
+    // increment attempts
+    await otpEntry.update({ attempts: otpEntry.attempts + 1 })
```
seaerchin, line 276: not a blocker but just a site note that we don't increment attempts when the submitted otp is blank

seaerchin (approved):

src/services/identity/UsersService.ts
harishv7: decided together that this was acceptable for our use cases with rate limiter in place

PR #641, opened 2023-02-27
Feat/throttling

alexanderleegs:
Duplicate of https://github.com/isomerpages/isomercms-backend/pull/610 - i think we might have merged this when it was being tested on vapt branch


kishore03109 (changes requested):
Hey uhm I think this implementation might be wrong! 
I think `trust proxy` should be set to 2 as 
- our server uses borh the alb reverse proxy + cloudflare reverse proxy, so total of 2 requests. 
- We need to set a maximum number of hops, else we will have a global throttle rate, which isnt the intended design!

kishore03109 (approved):

PR #642, opened 2023-02-27
feat: add gitguardian hook and update readme

harishv7:
## Problem

While we do have checks post-commit to identify any secrets are committed, we should also attempt to prevent this even before a commit can occur.

After considering some solutions - https://github.com/awslabs/git-secrets and Git Guardian, we decided to go with GitGuardian since we are already using it.

## Solution

We add Git Guardian as a pre-commit hook which scans commits for any leaked secrets before actually performing the commit.

**Breaking Changes**

There is a breaking change for local development since this commit makes Git Guardian setup somewhat mandatory.

  - Details ...

**New dev dependencies**:

No new dependencies are added to our package. Git Guardian will be installed on the OS-level instead as a brew package.


README.md
```diff
@@ -1,3 +1,33 @@
...
+1. Install GitGuardian
+
+```
+brew install gitguardian/tap/ggshield
+```
+
+2. Add the API Key to your shell startup env (e.g. in `.zshrc`)
+
+```
+export GITGUARDIAN_API_KEY=abc123
```
QiluXie, line 21: I assume this API key is created by us manually on gitguardian portal, which is fine. But `abc123` is not exactly a good value for an API key, would it be easy to replace it with a more randomized value?

QiluXie, line 21: nit: is it better to put a comment explaining what value should be put here? 
Something like this:
`export GITGUARDIAN_API_KEY="abc123" //Service API key from gitguardian account`

.husky/pre-commit
```diff
@@ -2,3 +2,4 @@
 . "$(dirname "$0")/_/husky.sh"
 
 npx lint-staged
+ggshield secret scan pre-commit
```
QiluXie, line 5: nit: it's better to add a newline at the end of the file

harishv7: done

README.md
harishv7: sure, good idea. done!

kishore03109 (changes requested):
Running ggshield locally creates a`.cache__ggshield` file. By right, since this is a pre-commit hook, there isnt a real need to run ggshield directly. However, if there was a false positive and someone runs ggshield directly, I dont think we should be committing the `.cache__ggshield` to remote. I suggest to defensively add `.cache__ggshield` into our .gitignore, wdyt?

In addition to this, could we add this commit to develop BE + FE vapt + FE develop?

harishv7:
> Running ggshield locally creates a`.cache__ggshield` file. By right, since this is a pre-commit hook, there isnt a real need to run ggshield directly. However, if there was a false positive and someone runs ggshield directly, I dont think we should be committing the `.cache__ggshield` to remote. I suggest to defensively add `.cache__ggshield` into our .gitignore, wdyt?
> 
> In addition to this, could we add this commit to develop BE + FE vapt + FE develop?

@kishore03109 makes sense, added it to gitignore.

As for the FE, yes definitely we will be adding it. Just wanted to address comments on this PR, before I replicate the changes.

As for merging to develop as well, sure we can. But I thought eventually vapt is going into develop. Are we going to merge it soon after vapt?

kishore03109 (approved):
hmm not sure when we are going to merge vapt into develop, but sure, we can leave it here since it going to dev anyways!

.husky/pre-commit
kishore03109, line 5: Out of curiosity, could we source our env var `GITGUARDIAN_API_KEY` from here? that way we dont need to add anything to our global variable in our machines!

harishv7: Sure, I thought by installing globally, devs can use it across all repos and also the global CLI to scan. But I think .env also works

Will update the README and command

QiluXie (approved):
Thanks @harishv7, can be merged if no comment from @alexanderleegs

PR #644, opened 2023-03-02
fix(site creation): make rewrites to redirects

kishore03109:
## Problem

CSP headers are not preserved with rewrites. 

## Solution 

New sites created will have a redirect rather than a rewrite.


seaerchin (approved):
lgtm

PR #650, opened 2023-03-14
build(node): update to node 18

seaerchin:
## Problem
Node 14 is reaching EOL soon, so we should upgrade to Node 18.

## Solution
1. update `nvmrc` 
2. reinstall packages using new node version
3. pin engines >=16 (this does **not** mandate node version but triggers a warning)
    - using `engine-strict` would have the desired behaviour but cause packages that are incompat to fail to install (see [here](https://docs.npmjs.com/cli/v8/using-npm/config#engine-strict))


.nvmrc
```diff
@@ -1 +1 @@
-14
```
harishv7, line 1: what's this file? It shows empty changes

package.json
```diff
@@ -156,6 +156,6 @@
     ]
   },
   "engines": {
-    "node": ">=14.17.5"
-  }
+      "node" : ">=16.0.0"
```
harishv7, line 159: We are gonna use 16 or 18?

.nvmrc
seaerchin: it changed from 14-18, no? it's what node version manager (nvm) uses to determine the version of node to use (see **[here](https://github.com/nvm-sh/nvm#nvmrc))**

package.json
seaerchin: 18 - this is given by the `.nvmrc`; what `engines` specifies is the version of node that our app will work with. 

i specified 16 because it'll work with 16+ (technically 14 is acceptable but we want to outlaw it)

seaerchin:
concretely what this set-up means is:

1. use node 18 for development but 
2. use any node >16 for deployment

alexanderleegs (approved):
lgtm

.nvmrc
harishv7: oops sorry, somehow it showed up as empty file. i see the changes now after refreshing

package.json
harishv7: makes sense!

harishv7 (approved):

harishv7:
@seaerchin Looks good, let's deploy and test on staging EB once as well

kishore03109 (changes requested):
mostly lgtm, but CICD breaks since our github actions still use node 14 (in some cases even node 12 😱 ), we need to modify `.github/workflows/ci.yml` to rectify this

seaerchin:
@kishore03109 re-requested just in case i missed msht

kishore03109 (approved):

PR #655, opened 2023-03-16
feat(rr): construct cms link for users

seaerchin:
## Problem
Frontend requires a link to the cms for each page that was edited within the review request.. This PR adds that in. 

## Solution
After parsing the page, construct the cms link using the `kind` of the page and redirect teh users there

### DIFF
--- a/src/integration/Reviews.spec.ts
+++ b/src/integration/Reviews.spec.ts
@@ -749,7 +749,7 @@ describe("Review Requests Integration Tests", () => {
             type: ["page"],
             name: MOCK_GITHUB_FILENAME_ALPHA_ONE,
             path: [],
-            url: `${MOCK_DEPLOYMENT_DBENTRY_ONE.stagingUrl}${MOCK_PAGE_PERMALINK}`,
+            stagingUrl: `${MOCK_DEPLOYMENT_DBENTRY_ONE.stagingUrl}${MOCK_PAGE_PERMALINK}`,
             lastEditedBy: MOCK_USER_EMAIL_TWO, // TODO: This should be MOCK_USER_EMAIL_ONE
             lastEditedTime: new Date(MOCK_GITHUB_COMMIT_DATE_THREE).getTime(),
           },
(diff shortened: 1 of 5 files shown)


src/services/fileServices/MdPageServices/PageService.ts
```diff
@@ -317,6 +319,51 @@ export class PageService {
...
+      }
+      case "ResourceCategoryPage": {
+        return okAsync(
+          `${baseUrl}/resourceRoom/${pageName.resourceRoom}/resourceCategory/${pageName.resourceCategory}/editPage/${pageName.name}`
+        )
+      }
+      default: {
+        const unimplErr: never = pageName
+        return errAsync(
+          new BaseIsomerError(500, `Unimplemented pagetype: ${unimplErr}`)
```
kishore03109, line 361:
```suggestion
          new BaseIsomerError(500, `Unimplemented page type: ${unimplErr}`)
```

src/services/review/ReviewRequestService.ts
```diff
@@ -108,6 +111,11 @@ export default class ReviewRequestService {
           )
           // NOTE: We ignore the errors and use a placeholder
           .unwrapOr("")
+        const fileUrl = await pageNameRes
+          .andThen((pageName) =>
+            this.pageService.retrieveCmsPermalink(pageName, siteName)
+          )
+          .unwrapOr("")
```
kishore03109, line 118: Hmm I am confused here. Could you explain why we return an empty string here if it fails, rather than throwing/returning an error?

src/services/fileServices/MdPageServices/PageService.ts
```diff
@@ -35,6 +35,8 @@ import { ResourcePageService } from "./ResourcePageService"
 import { SubcollectionPageService } from "./SubcollectionPageService"
 import { UnlinkedPageService } from "./UnlinkedPageService"
 
+const FRONTEND_URL = process.env.FRONTEND_URL || "http://localhost:3000"
```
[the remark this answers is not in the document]
seaerchin: sian haven't rebase yet

src/services/fileServices/MdPageServices/PageService.ts
seaerchin: sure, looks good

src/services/review/ReviewRequestService.ts
seaerchin: going to raise a PR later on to resolve this and make the type parsing clearer

seaerchin: #662 - parsing logic clarified so that if we fail any preconditions to parse a file, we return a minimal object back to the frontend. 

this is because we can fail to parse any one file but still able to parse others successfully; this is abit of a cop-out because we choose to mix a file that failed to parse w/ successful parsing files but the FE is still able to display (some) relevant info

src/services/fileServices/MdPageServices/PageService.ts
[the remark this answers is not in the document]
seaerchin: going to rebase after approval to avoid dangling comments

src/integration/Reviews.spec.ts
alexanderleegs, line 1: mock data needs the new `fileUrl` param, i think this is the cause of the failing tests?

kishore03109 (commented):
Mostly lgtm pending CICD tests

kishore03109 (approved):

PR #659, opened 2023-03-22
Chore/change node env of local dev to expect dev

alexanderleegs:
For our `NODE_ENV` env var, we previously were using `LOCAL_DEV` for local dev and `DEV` for staging, which was confusing - this PR modifies it such that we use `dev` for local dev and `staging` for staging.

### DIFF
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -38,7 +38,7 @@ const config = convict({
   env: {
     doc: "The application environment.",
     env: "NODE_ENV",
-    format: ["dev", "test", "prod", "staging"],
+    format: ["dev", "test", "prod", "staging", "vapt"],
     default: "dev",
   },
   port: {
--- a/src/logger/logger.js
+++ b/src/logger/logger.js
@@ -11,7 +11,7 @@ const WinstonCloudwatch = require("winston-cloudwatch")
 const { config } = require("@config/config")
 
 // Env vars
-const NODE_ENV = config.get("env")
+const NODE_ENV = config.get("env").toLowerCase()
 
 // AWS
 const AWS_REGION_NAME = "ap-southeast-1"
@@ -25,7 +25,7 @@ const metadataRequest = Bluebird.promisify(
 // TODO: Check this env var as it is not in example
 const LOG_GROUP_NAME = `${process.env.AWS_BACKEND_EB_ENV_NAME}/nodejs.log`
 const IS_PROD_ENV =
-  NODE_ENV !== "LOCAL_DEV" && NODE_ENV !== "DEV" && NODE_ENV !== "test"
+  NODE_ENV !== "dev" && NODE_ENV !== "staging" && NODE_ENV !== "test"
 const IS_TEST_ENV = NODE_ENV === "test"
 
 function timestampGenerator() {
--- a/src/logger/logger.js
+++ b/src/logger/logger.js
@@ -11,7 +11,7 @@ const WinstonCloudwatch = require("winston-cloudwatch")
 const { config } = require("@config/config")
 
 // Env vars
-const NODE_ENV = config.get("env").toLowerCase()
+const NODE_ENV = config.get("env")
 
 // AWS
 const AWS_REGION_NAME = "ap-southeast-1"
--- a/src/routes/v1/auth.js
+++ b/src/routes/v1/auth.js
@@ -26,10 +26,9 @@ const router = express.Router()
 const CLIENT_ID = config.get("github.clientId")
 const CLIENT_SECRET = config.get("github.clientSecret")
 const REDIRECT_URI = config.get("github.redirectUri")
-const AUTH_TOKEN_EXPIRY_MS = config.get("auth.tokenExpiry")
 const CSRF_TOKEN_EXPIRY_MS = 600000
 const FRONTEND_URL = config.get("app.frontendUrl")
-const NODE_ENV = config.get("env")
+const NODE_ENV = config.get("env").toLowerCase()
 
 const CSRF_COOKIE_NAME = "isomer-csrf"
 const COOKIE_NAME = "isomercms"
@@ -53,8 +52,7 @@ async function authRedirect(req, res) {
   const cookieSettings = {
     expires: csrfTokenExpiry,
     httpOnly: true,
-    secure:
-      NODE_ENV !== "DEV" && NODE_ENV !== "LOCAL_DEV" && NODE_ENV !== "test",
+    secure: NODE_ENV !== "dev" && NODE_ENV !== "staging" && NODE_ENV !== "test",
   }
 
   const token = jwtUtils.signToken({ state })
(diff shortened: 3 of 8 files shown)


src/logger/logger.js
```diff
@@ -11,7 +11,7 @@ const WinstonCloudwatch = require("winston-cloudwatch")
 const { config } = require("@config/config")
 
 // Env vars
-const NODE_ENV = config.get("env")
+const NODE_ENV = config.get("env").toLowerCase()
```
kishore03109, line 14: Eh wait why are we not directly changing this ah in prod/staging ah?

src/config/config.ts
```diff
@@ -38,8 +38,8 @@ const config = convict({
   env: {
     doc: "The application environment.",
     env: "NODE_ENV",
-    format: ["LOCAL_DEV", "DEV", "test", "prod", "staging"],
-    default: "LOCAL_DEV",
+    format: ["dev", "test", "prod", "staging"],
```
kishore03109, line 41: could we use this opportunity to change "test" to "test-dev" / "local-test" instead?

src/logger/logger.js
seaerchin: not in favour of doing this here - should either do directly in the respective env as per ^ or we should normalise it in the convict function to ensure that it's case agnostic.

more in favour of the latter cos it'll save devs from having to remember if it's upper or lower

src/utils/auth-utils.js
```diff
@@ -1,9 +1,9 @@
 const { config } = require("@config/config")
 
-const NODE_ENV = config.get("env")
+const NODE_ENV = config.get("env").toLowerCase()
 
 function isSecure() {
-  return NODE_ENV !== "DEV" && NODE_ENV !== "LOCAL_DEV" && NODE_ENV !== "test"
+  return NODE_ENV !== "dev" && NODE_ENV !== "staging" && NODE_ENV !== "test"
 }
```
seaerchin, line 7: just realised this is the same as `IS_PROD_ENV`; can centralise but feel free to ignore (technically out of scope n small change)

src/config/config.ts
seaerchin: do you feel like we should add a `vapt` env here or should we do that separately in another PR that configures stuff for vapt

alexanderleegs: `test` is actually set by jest itself: see https://jestjs.io/docs/environment-variables

I think it makes sense to allow the vapt entry as well, have added it in https://github.com/isomerpages/isomercms-backend/pull/659/commits/0134f9647031bffc67dd09f61cb84bc738e9f811

src/utils/auth-utils.js
alexanderleegs: Hmm actually the only place this is being used is for our github auth redirect cookie settings - any idea why it's currently only set to secure on production and not staging? I was thinking it might be the same as the one we use in `server.js` instead to separate local dev/testing from staging/prod, we can probably refactor out the env checks if so

src/logger/logger.js
alexanderleegs: yeah that makes sense, have removed toLowerCase in https://github.com/isomerpages/isomercms-backend/pull/659/commits/0b7f6c13cdab0dd43a3ada79518a45bf7357d9bb

src/routes/v1/auth.js
```diff
@@ -53,8 +52,7 @@ async function authRedirect(req, res) {
   const cookieSettings = {
     expires: csrfTokenExpiry,
     httpOnly: true,
-    secure:
-      NODE_ENV !== "DEV" && NODE_ENV !== "LOCAL_DEV" && NODE_ENV !== "test",
+    secure: NODE_ENV !== "dev" && NODE_ENV !== "staging" && NODE_ENV !== "test",
```
seaerchin, line 55: use `isSecure`

src/server.js
```diff
@@ -92,8 +91,7 @@ const createError = require("http-errors")
 const NODE_ENV = config.get("env")
 const SESSION_SECRET = config.get("auth.sessionSecret")
 
-const isSecure =
-  NODE_ENV !== "DEV" && NODE_ENV !== "LOCAL_DEV" && NODE_ENV !== "test"
+const isSecure = NODE_ENV !== "dev" && NODE_ENV !== "test"
```
seaerchin, line 94: use `isSecure`

src/utils/auth-utils.js
seaerchin: i can't remember why it's not secure on `staging` tbh but it can't hurt HAHA

seaerchin (commented):
pls make convict env vars checking case insensitive

src/routes/v1/auth.js
alexanderleegs: https://github.com/isomerpages/isomercms-backend/pull/659/commits/aefef7b0b458be4c499e5f0d6a149bca068f44af

src/server.js
alexanderleegs: https://github.com/isomerpages/isomercms-backend/pull/659/commits/aefef7b0b458be4c499e5f0d6a149bca068f44af

alexanderleegs:
> pls make convict env vars checking case insensitive

As discussed offline, convict is case sensitive by default and doesn't seem to be able to convert env vars into lowercase - since this change only affects local dev, we'll leave it as case sensitive and using lowercase (same as the automatically set `test` value set by jest)

src/services/identity/index.ts
```diff
@@ -27,7 +27,7 @@ const NODE_ENV = config.get("env")
 const OTP_SECRET = config.get("auth.otpSecret")
 const OTP_EXPIRY = config.get("auth.otpExpiry")
 
-const IS_LOCAL_DEV = NODE_ENV === "LOCAL_DEV"
+const IS_LOCAL_DEV = NODE_ENV === "dev"
```
harishv7, line 30: should we just call it IS_DEV then?

src/services/utilServices/MailClient.ts
```diff
@@ -44,7 +44,7 @@ export default MailClient
 const NODE_ENV = config.get("env")
 const POSTMAN_API_KEY = config.get("postman.apiKey")
 
-const IS_LOCAL_DEV = NODE_ENV === "LOCAL_DEV"
+const IS_LOCAL_DEV = NODE_ENV === "dev"
```
harishv7, line 47: same here

harishv7:
> > pls make convict env vars checking case insensitive
> 
> As discussed offline, convict is case sensitive by default and doesn't seem to be able to convert env vars into lowercase - since this change only affects local dev, we'll leave it as case sensitive and using lowercase (same as the automatically set `test` value set by jest)

Adding on, it is intentional to be case sensitive

src/services/identity/index.ts
alexanderleegs: Sure, changed in https://github.com/isomerpages/isomercms-backend/pull/659/commits/7989ed4398f329f9d05a56a91cada7b54ee7004a

src/services/utilServices/MailClient.ts
alexanderleegs: https://github.com/isomerpages/isomercms-backend/pull/659/commits/7989ed4398f329f9d05a56a91cada7b54ee7004a

PR #662, opened 2023-03-22
feat(rr): config parsing

seaerchin:
## Problem
Our current model of parsing is simplified and only works for pages; this PR adds parsing for configuration files (nav + setting) 

Closes [IS-54]

## Solution
1. add a config service that handles whether a file is a configuration file or not
2. extend dto definitions to differentiate between the types 
3. in the event of failure, return the filename + folder path to the frontend to display

[IS-54]: https://isomeropengov.atlassian.net/browse/IS-54?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/routes/v2/authenticated/sites.ts
```diff
@@ -68,10 +69,10 @@ export class SitesRouter {
     )
 
     // Check for error and throw
-    if (possibleStagingUrl instanceof BaseIsomerError) {
-      return res.status(404).json({ message: possibleStagingUrl.message })
+    if (possibleStagingUrl.isErr()) {
+      return res.status(404).json({ message: possibleStagingUrl.error.message })
```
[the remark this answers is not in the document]
seaerchin: the type is `Result<T, E>` and the `err` here is a _variant_ of `Result` (either `ok` or `err`); the actual error itself is contained within `err`

src/services/fileServices/YmlFileServices/ConfigService.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  private extractYmlConfigFileName = ({
+    name: rawName,
+    path,
+  }: PathInfo): YmlConfigFileName => {
+    const name = Brand.fromString<"YmlConfig">(rawName)
+    return {
+      name,
+      path,
+      kind: "YmlConfig",
```
harishv7, line 40: use enums

src/errors/PageParseError.ts
```diff
@@ -0,0 +1,7 @@
+import { BaseIsomerError } from "@root/errors/BaseError"
+
+export default class PageParseError extends BaseIsomerError {
```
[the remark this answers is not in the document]
harishv7: So for example IsomerError(httpStatus, internalErrCode, errMsg)

src/services/review/ReviewRequestService.ts
```diff
@@ -71,70 +100,163 @@ export default class ReviewRequestService {
...
+        )
+      )
+
+  private compareDiffWithMappings = (
+    sessionData: UserWithSiteSessionData,
+    stagingLink: StagingPermalink,
+    files: RawFileChangeInfo[],
+    mappings: Record<
+      string,
+      {
```
harishv7, line 143: might want to abstract this as a type and use with a readable name

src/services/review/ReviewRequestService.ts
```diff
@@ -71,70 +100,163 @@ export default class ReviewRequestService {
...
+    >
+  ): ResultAsync<WithEditMeta<EditedItemDto>, BaseEditedItemDto>[] => {
     // Step 1: Get the site name
     const { siteName } = sessionData
 
-    // Step 2: Get the list of changed files using Github's API
-    // Refer here for details; https://docs.github.com/en/rest/commits/commits#compare-two-commits
-    // Note that we need a triple dot (...) between base and head refs
-    const { files, commits } = await this.apiService.getCommitDiff(siteName)
+    return files.map(({ filename, contents_url: contents }) => {
```
harishv7, line 152: why are we using snake_case here

src/services/review/ReviewRequestService.ts
```diff
@@ -1,34 +1,60 @@
...
-import { errAsync, okAsync, ResultAsync } from "neverthrow"
+import {
+  err,
+  errAsync,
+  ok,
+  okAsync,
+  Result,
+  ResultAsync,
+  combine,
+} from "neverthrow"
```
harishv7, line 11: have we decided to use neverthrow across our codebase?

harishv7: I think is an interesting way to handle errors, but it's quite a shift in the way conventional errors are handled in js

src/services/review/ReviewRequestService.ts
```diff
@@ -71,70 +100,163 @@ export default class ReviewRequestService {
...
+          this.pageService.retrieveCmsPermalink(pageName, siteName),
+          this.pageService.retrieveStagingPermalink(
+            sessionData,
+            stagingLink,
+            pageName
+          ),
+        ])
+      )
+      .map(([fileUrl, stagingUrl]) => ({
+        type: "page",
```
harishv7, line 219: what are the types there are? use enum?

src/services/review/ReviewRequestService.ts
```diff
@@ -71,70 +100,163 @@ export default class ReviewRequestService {
...
+  extractMediaInfo = ({
+    name,
+    path,
+  }: PathInfo): Result<EditedMediaDto, PageParseError> => {
+    const fileExt = getFileExt(name)
+    if (ALLOWED_FILE_EXTENSIONS.includes(fileExt)) {
+      return ok({
+        name,
+        path: path.unwrapOr([""]),
+        type: fileExt === "pdf" ? "file" : "image",
```
harishv7, line 238: use enums

src/services/review/ReviewRequestService.ts
```diff
@@ -71,70 +100,163 @@ export default class ReviewRequestService {
...
-          )
-          .unwrapOr("")
+        .map<WithEditMeta<EditedItemDto>>((item) => {
+          const items = contents.split("?ref=")
+          // NOTE: We have to compute sha this way rather than
+          // taking the file sha.
+          // This is because the sha present on the file is
+          // a checksum of the files contents.
+          // And the actual commit sha is given by the ref param
+          const hash = items[items.length - 1]
```
harishv7, line 177: shld we do a length check first?

is there a possibility ref= doesn't exist? if so, that would be an error state right (thought it technically wont break the code)

src/services/review/ReviewRequestService.ts
```diff
@@ -71,70 +100,163 @@ export default class ReviewRequestService {
...
-          stagingUrl,
-          fileUrl,
-          lastEditedBy: mappings[sha]?.author || "Unknown user",
-          lastEditedTime: mappings[sha]?.unixTime || 0,
-        }
+        .orElse<BaseEditedItemDto>(() => {
+          const { path, name } = pathInfo
+          return errAsync({
+            name,
+            path: path.unwrapOr([]),
```
harishv7, line 189: sorry can i check where does this unWrapOr come from? what does it do?

src/routes/v2/authenticated/review.ts
```diff
@@ -84,11 +85,9 @@ export class ReviewsRouter {
     return this.sitesService
       .getStagingUrl(userWithSiteSessionData)
       .andThen((stagingLink) =>
```
harishv7, line 87: i see a lot of these custom pipes - are these from neverthrow?

src/types/dto/review.ts
```diff
@@ -4,16 +4,33 @@ export type ReviewRequestStatus = "OPEN" | "APPROVED" | "MERGED" | "CLOSED"
...
 }
 
+export interface EditedPageDto extends BaseEditedItemDto {
+  type: "page"
+  stagingUrl: string
+  fileUrl: string
+}
+
+export interface EditedConfigDto extends BaseEditedItemDto {
+  type: "nav" | "setting"
```
harishv7, line 25: using enums will be cleaner

src/services/fileServices/YmlFileServices/ConfigService.ts
```diff
@@ -0,0 +1,75 @@
...
+import {
+  ConfigFileName,
+  HtmlConfigFileName,
+  YmlConfigFileName,
+} from "@root/types/configYml"
+import { PathInfo } from "@root/types/pages"
+import { Brand } from "@root/types/util"
+
+// eslint-disable-next-line import/prefer-default-export
+export class ConfigService {
```
harishv7, line 13: do we have a test file for this new service?

src/errors/ConfigParseError.ts
```diff
@@ -0,0 +1,7 @@
+import { BaseIsomerError } from "@root/errors/BaseError"
+
+export default class ConfigParseError extends BaseIsomerError {
+  constructor(fileName: string) {
+    super(500, `The given file: ${fileName} was not a config file!`)
```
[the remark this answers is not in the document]
seaerchin: u r right; we should also do this in 1 shot... i'll change the errors introduced here but we fundamentally need to rehaul our error hierarchy + structure sadge

could i trouble you to file an issue on this (error overhaul) either on github or jira?

src/errors/PageParseError.ts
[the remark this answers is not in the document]
seaerchin: re: 1 error file per error - no strong opinions, style when i was introduced to the codebase was 1/file, which is inline w/ other codebases so i'm ok with it too. our errors are classes so eslint might complain if there's more than a certain # per file

re: error hierarchy - agreed; should relook at this later on.

src/services/fileServices/YmlFileServices/ConfigService.ts
seaerchin: i like unions more tbh, i've ran into type errors w/ enums in the past due to how they are implemented in TS and hence, i've avoided them like the plague ever since.

see [here](https://fettblog.eu/tidy-typescript-avoid-enums/) or [here](https://www.niraj.life/blog/avoid-using-enums-in-typescript) for more details  but if you feel strongly about this, i can change it

src/services/review/ReviewRequestService.ts
seaerchin: caught me being lazy AHHAHHA wdyt about `shaMappings`

src/services/review/ReviewRequestService.ts
seaerchin: github returns as snake_case, renamed to `contents`

src/services/review/ReviewRequestService.ts
seaerchin: agreed, but throwing isn't good because then devs aren't inclined to handle errors properly + there's no oversight onto what errors are actually being returned. 

this forces errors to be explicit and handled properly rather than throwing and expecting a middleware to do so somewhere (as is done now)

src/routes/v2/authenticated/review.ts
seaerchin: yeap

src/services/review/ReviewRequestService.ts
seaerchin: `neverthrow` - if it's ok, returns the value in ok, otherwise uses a default

src/types/dto/review.ts
seaerchin: see above ;--;

src/services/fileServices/YmlFileServices/ConfigService.ts
seaerchin: not at present, will add in a future PR

src/services/review/ReviewRequestService.ts
seaerchin: see above ;--;

src/services/review/ReviewRequestService.ts
seaerchin: see above ;--;

src/services/review/ReviewRequestService.ts
seaerchin: re: this - the github schema [here](https://docs.github.com/en/rest/commits/commits?apiVersion=2022-11-28#compare-two-commits) has a `?ref=` at the end, which is why i made this assumption

i think it's possible that it doesn't exist but it does imply that github made a breaking change to their api

src/services/review/ReviewRequestService.ts
```diff
@@ -134,7 +137,7 @@ export default class ReviewRequestService {
   }
 
   // TODO
```
kishore03109: oof might be out of scope of this PR, but do you know what this todo is for?

src/utils/files.ts
```diff
@@ -1,3 +1,34 @@
...
+  const fullPath = pageName.split("/")
+  // NOTE: Name is guaranteed to exist
+  // as this method only accepts a string
+  // and we've validated that the string is not empty
+  const name = fullPath.pop()!
+
+  if (fullPath.length === 0) {
+    return ok({
+      name,
+      path: err([]),
```
kishore03109, line 26: I am confused here, why are we wrapping an err around an ok here ah?

src/services/fileServices/MdPageServices/PageService.ts
```diff
@@ -116,33 +115,40 @@ export class PageService {
...
       // If it is not, we assume that this is a collection page.
       // Because this method is invoked on existing file paths from the frontend,
       // this assumption will hold true.
-      .orElse(() => this.parseResourceRoomPage(pageName, sessionData))
-      .orElse(() => this.parseCollectionPage(pageName))
+      .orElse(() => this.parseResourceRoomPage(pathInfo, sessionData))
+      .orElse(() => this.parseCollectionPage(pathInfo))
+
+  isMarkdownPage = (pageName: string): Result<string, PageParseError> => {
+    if (pageName.endsWith(".md")) return ok(pageName)
```
kishore03109, line 136: Nit: can we just return a boolean value instead? Using neverthrow here feels too verbose for a function like this

src/services/fileServices/MdPageServices/PageService.ts
```diff
@@ -160,52 +166,46 @@ export class PageService {
...
+    path,
+  }: PathInfo): Result<ContactUsPageName, NotFoundError> => {
+    if (
+      path.isOk() &&
+      path.value.at(-1) === "pages" &&
+      name === "contact-us.md"
+    ) {
+      return ok({ name: Brand.fromString(name), kind: "ContactUsPage" })
+    }
+    return err(new NotFoundError())
```
kishore03109, line 196: Do we want to return the specific file that wasnt found here? (ie the path?)

src/services/fileServices/MdPageServices/PageService.ts
```diff
@@ -214,62 +214,56 @@ export class PageService {
...
+    path
+      .asyncAndThen<ResourceCategoryPageName, NotFoundError>((rawPath) => {
+        if (rawPath[0] !== resourceRoomName.name) {
+          return errAsync(new NotFoundError())
+        }
 
-            if (rawPath.length !== 3 && rawPath.at(-1) !== "_posts") {
-              return errAsync(new NotFoundError())
-            }
+        if (rawPath.length !== 3 && rawPath.at(-1) !== "_posts") {
```
kishore03109, line 238: May I get your help to add a comment of how a `rawPath` might look like? abit confused what the each element of `rawPath` represents

kishore03109:
CICD tests are failing, may I trouble you you to take a look at them?

src/services/review/ReviewRequestService.ts
seaerchin: i don't see this specific `TODO` actually, you might have reviewed an old commit

src/services/fileServices/MdPageServices/PageService.ts
seaerchin: actually the problem here is that we want to distinguish `MarkdownPages` from other things (eg: `YmlConfig` or `HtmlPages`) - the problem with returning `boolean` is that we lose context wrt what the `pageName` is beyond its immediate containing function.

i can change this to `string is MarkdownPage` and brand this string to prevent the verbosity - wdyt

src/services/fileServices/MdPageServices/PageService.ts
seaerchin: can but i'll add it onto the `PathInfo` type - this is to avoid repetition + set at the containing type. wdyt?

seaerchin: can but i'll add it onto the `PathInfo` type - this is to avoid repetition + set at the containing type. wdyt?

seaerchin:
> CICD tests are failing, may I trouble you you to take a look at them?

kek i think the source branch is `feat/buttons`, which had tests fixed in #587 and #588; i'll probably rebase + merge those first and then rebase this on top once approved to see if the tests are still failing

src/errors/ConfigParseError.ts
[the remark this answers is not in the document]
harishv7: Sure, here: https://isomeropengov.atlassian.net/browse/IS-37

src/utils/files.ts
kishore03109: uhm just checking in, we expect the caller to meaningfully use the `name` then? if yes, feel free to resolve this.

src/services/fileServices/MdPageServices/PageService.ts
kishore03109: hmm isnt that beyond the scope of this function, which is to just tell if this page is of type `MarkdownPages`? 
if not should we then rename this to `getPageType` right?

src/services/review/ReviewRequestService.ts
kishore03109: wait based on the schema, why are we not taking the `base_commit`.`sha` directly ah?

src/errors/PageParseError.ts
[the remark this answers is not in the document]
harishv7: Okay I think can keep like this, but we need a revamp of the error hierarchy soon

src/services/fileServices/YmlFileServices/ConfigService.ts
harishv7: @seaerchin That's a good point! I think I used to do it like this:

```
enum ESomeEnum {
    ValOne: "1",
    ValTwo: "2"
}

type EFinalType = ESomeEnum.ValOne | ESomeEnum.ValTwo
```

Then we use the EFinalType for the actual assertion

src/services/review/ReviewRequestService.ts
harishv7: Hahaha! 😂 

Shall we define one `shaMapping` and then `shaMappings` is a Record of string to `shaMapping`?

src/services/review/ReviewRequestService.ts
harishv7: Ooh okay, I think this syntax is going take some time to get used to. Do you think we need neverthrow though? I thought of discussing with you how best to standardise returns and responses across the codebase first

src/services/fileServices/YmlFileServices/ConfigService.ts
seaerchin: this actually results in the same thing HAHA

src/services/review/ReviewRequestService.ts
seaerchin: sorry, where did you see this? i saw mine under `Comparing two commits`, where the schema explicitly has a `?ref=` at the end (returned by github)

src/services/review/ReviewRequestService.ts
seaerchin: shure, can!

src/utils/files.ts
seaerchin: yea - if it's rooted at `/` (homepage is `/index.md` for eg), then the `name` will be `index.md` but path is empty (`err`).

hence, `name` is always semantically meaningful but path depends

src/errors/ConfigParseError.ts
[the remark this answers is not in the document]
seaerchin: marking as resolved cos we're tracking on jira

src/services/review/ReviewRequestService.ts
seaerchin: @kishore03109 i've a comment there (up above) and it's cos the sha present is a checksum of the files' contents but we want the sha referencing the actual file

src/services/fileServices/YmlFileServices/ConfigService.ts
seaerchin: oh ps, i haven't had time to look at this due to being caught up in the release but this has the same effect because the discriminated union still only allows the same set of values.

src/services/review/ReviewRequestService.ts
seaerchin: `Record` takes 2 type arguments, so i think sha mapping would be misleading (cos that's the actual user info that we'd be interested in).

wdyt about 
```ts
type GithubEditInfo = {
        author: string
        unixTime: number
}
type Sha = string
type ShaMappings = Record<Sha, GithubEditInfo>
```

src/services/fileServices/YmlFileServices/ConfigService.ts
harishv7: yes true, but felt it is cleaner with the use of discriminated unions + enums rather than strings

src/services/review/ReviewRequestService.ts
harishv7: okay makes sense to me

src/services/review/ReviewRequestService.ts
harishv7: back to the discussion the team had on using neverthrow, I guess we will be postponing the use of the library till we get clarity on some of the refactorings we need to do in our code right?

src/types/pages.ts
```diff
@@ -1,52 +1,28 @@
...
   collection: string
   subcollection: string
   kind: "SubcollectionPage"
 }
 
-export type CollectionPageName = {
-  name: string & { __kind: "CollectionPage" }
+export type CollectionPageName = FileNameBrand<"CollectionPage"> & {
   collection: string
-  kind: "CollectionPage"
```
kishore03109, line 27: sanity check: Is this removal of `kind` intended?

kishore03109 (approved):
[Tests] The full testing plan will be in the corresponding FE PR for buttons right? If not, please dismiss this review

src/types/pages.ts
seaerchin: ye shifted to util

src/routes/v2/authenticated/review.ts
```diff
@@ -125,14 +126,18 @@ export class ReviewsRouter {
...
+          userWithSiteSessionData,
+          stagingLink
         )
       )
       .map((items) => res.status(200).json({ items }))
+      .mapErr((err) => {
+        if (err instanceof MissingSiteError || err instanceof NotFoundError) {
+          return res.status(404).json({ message: err.message })
+        }
+        return res.status(500).json({ message: err.message })
```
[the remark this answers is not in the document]
seaerchin: the `NotFoundError` was constructed either with no arguments (defaults to generic error messages of `Something went wrong`) or a default string (no sensitive info) or a formatted string containing the requested `rawPath`, similarly for `MissingSiteError`, it's just constructed with a default message of `the requested site was not found in isomer`, which i think is ok.

lmk if `rawPath` is a concern

src/integration/Reviews.spec.ts
```diff
@@ -263,20 +266,20 @@ describe("Review Requests Integration Tests", () => {
...
           {
-            type: ["page"],
+            type: "page",
             name: MOCK_GITHUB_FILENAME_ALPHA_ONE,
             path: [],
             stagingUrl: `${MOCK_DEPLOYMENT_DBENTRY_ONE.stagingUrl}${MOCK_PAGE_PERMALINK}`,
-            fileUrl: `${FRONTEND_URL}/sites/${MOCK_REPO_NAME_ONE}/homepage`,
-            lastEditedBy: MOCK_USER_EMAIL_TWO, // TODO: This should be MOCK_USER_EMAIL_ONE
+            cmsFileUrl: `${FRONTEND_URL}/sites/${MOCK_REPO_NAME_ONE}/homepage`,
+            lastEditedBy: MOCK_USER_EMAIL_TWO,
```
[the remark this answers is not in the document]
seaerchin: can i clarify this - isn't the removed property added back below?

src/integration/Reviews.spec.ts
```diff
@@ -750,22 +753,22 @@ describe("Review Requests Integration Tests", () => {
         reviewRequestedTime: new Date(MOCK_GITHUB_DATE_ONE).getTime(),
         changedItems: [
           {
-            type: ["page"],
+            type: "page",
             name: MOCK_GITHUB_FILENAME_ALPHA_ONE,
             path: [],
-            stagingUrl: `${MOCK_DEPLOYMENT_DBENTRY_ONE.stagingUrl}${MOCK_PAGE_PERMALINK}`,
-            fileUrl: `${FRONTEND_URL}/sites/${MOCK_REPO_NAME_ONE}/homepage`,
-            lastEditedBy: MOCK_USER_EMAIL_TWO, // TODO: This should be MOCK_USER_EMAIL_ONE
```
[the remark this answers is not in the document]
seaerchin: can i clarify this - isn't the removed property added back below?

PR #663, opened 2023-03-24
Fix/convict fixes

harishv7:
## Problem

- Since convict runs at service startup, EB does not do a rollback on prod since it thinks build is successful. Resulting effect is an infinite loop of retrying to start the service
- Validation for required-positive-number misses corner case of 0
- Also, when running db:migrate, sequelize CLI doesn't support importing TS config file into JS which breaks the migration command

## Solution

- As discussed, I have added a try-catch for convict errors to prevent hard failure on production for safety.
- Added check for 0 in validation of positive number
- Changed to use process.env in database/config.js

### DIFF
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -9,7 +9,7 @@ convict.addFormat({
 })
 
 convict.addFormat({
-  name: "required-whole-number",
+  name: "required-natural-number",
   validate: (val: any) => {
     if (val === null || val === undefined || val === "")
       throw new Error("value cannot be empty, null or undefined")
@@ -19,11 +19,11 @@ convict.addFormat({
     const coercedVal = Number(val)
     if (isNaN(coercedVal)) {
       throw new Error(
-        "value provided is not a whole number. please provide a valid whole number"
+        "value provided is not a natural number. please provide a valid natural number"
       )
     }
-    if (coercedVal < 0) {
-      throw new Error("value must be more than or equal to zero")
+    if (coercedVal <= 0) {
+      throw new Error("value must be more than zero")
     }
     return coercedVal
   },
@@ -50,7 +50,7 @@ const config = convict({
   port: {
     doc: "The port to bind.",
     env: "PORT",
-    format: "required-whole-number",
+    format: "required-natural-number",
     default: 8081,
   },
   gitGuardian: {
@@ -85,7 +85,7 @@ const config = convict({
     pageCount: {
       doc: "Number of pages of repos to retrieve from GitHub API",
       env: "ISOMERPAGES_REPO_PAGE_COUNT",
-      format: "required-whole-number",
+      format: "required-natural-number",
       default: 10,
     },
   },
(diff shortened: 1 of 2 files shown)


src/config/config.ts
```diff
@@ -331,7 +337,12 @@ const config = convict({
 })
 
 // Perform validation
-config.validate({ allowed: "strict" })
+// TODO: remove try-catch after prod deployment is successful to avoid blocking
+try {
+  config.validate({ allowed: "strict" })
+} catch (e) {
+  console.log(e)
+}
```
alexanderleegs, line 345: Do we intend to leave this in to catch any env vars we fail to add? If so, we might want to make the log statement easier to search e.g. `Convict error, missing env vars: ${e}`

src/config/config.ts
```diff
@@ -9,15 +9,21 @@ convict.addFormat({
 })
 
 convict.addFormat({
-  name: "required-positive-number",
+  name: "required-whole-number",
```
kishore03109, line 12: Hmm I am abit confused about this change :( 
there is no case where any of these env vars (ISOMERPAGES_REPO_PAGE_COUNT, AUTH_TOKEN_EXPIRY_DURATION_IN_MILLISECONDS, MAX_NUM_OTP_ATTEMPTS, OTP_EXPIRY,DB_MIN_POOL and DB_MAX_POOL) to have the value of 0 right? 

if any of the above variables can have a value of 0, does it makes sense to have two schemas instead? one that allows 0 and one that doesnt?

harishv7: actually yea, i wanted to check with you guys if there is any env vars that you think 0 will be a valid option for in the current list
Else i can just change it to required-natural-number

harishv7: after looking through I think 0 does not seem to be a valid option for any of the env vars. will proceed to rename the check and asset for > 0 instead
let me know if any concerns @kishore03109 @seaerchin @alexanderleegs

src/config/config.ts
harishv7: Ah that makes sense, let me update it

harishv7: done

src/database/config.js
```diff
@@ -1,14 +1,15 @@
...
 
-const DB_URI = config.get("database.dbUri")
-const DB_MIN_POOL = config.get("database.dbMinPool")
-const DB_MAX_POOL = config.get("database.dbMaxPool")
+// Note: We are using process.env here instead of convict's config.get() as sequelize-cli is unable
+// to support import of TS files inside JS. Note that validation of these envs will still be
+// performed by convict in src/config/config.ts.
+const { DB_URI } = process.env
+const DB_MIN_POOL = parseInt(process.env.DB_MIN_POOL)
+const DB_MAX_POOL = parseInt(process.env.DB_MAX_POOL)
```
kishore03109, line 12: Style: ESlint error: Missing radix parameter. Just need to change
```
const DB_MIN_POOL = parseInt(process.env.DB_MIN_POOL)
const DB_MAX_POOL = parseInt(process.env.DB_MAX_POOL)
```
to 
```
const DB_MIN_POOL = parseInt(process.env.DB_MIN_POOL, 10)
const DB_MAX_POOL = parseInt(process.env.DB_MAX_POOL, 10)
```

kishore03109 (commented):
LTGM, just some eslint stuff

alexanderleegs (approved):
lgtm once kishore's points have been addressed!

src/database/config.js
harishv7: thanks for noticing this, fixed

PR #664, opened 2023-03-27
Fix/issecure

alexanderleegs:
This PR fixes a bug introduced in https://github.com/isomerpages/isomercms-backend/pull/659 - we were using `isSecure` as a boolean value instead of a function in some places. This PR changes `isSecure` into a constant variable instead of a function and makes the appropriate changes to fix this behaviour.

### DIFF
--- a/src/routes/v2/auth.js
+++ b/src/routes/v2/auth.js
@@ -17,7 +17,12 @@ const CSRF_COOKIE_NAME = "isomer-csrf"
 const COOKIE_NAME = "isomercms"
 
 class AuthRouter {
-  constructor({ authService, authenticationMiddleware, apiLogger, rateLimiter }) {
+  constructor({
+    authService,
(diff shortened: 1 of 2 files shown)


src/routes/v2/auth.js
```diff
@@ -46,7 +51,7 @@ class AuthRouter {
     const cookieSettings = {
       expires: csrfTokenExpiry,
       httpOnly: true,
-      secure: isSecure(),
+      secure: isSecure,
```
kishore03109, line 54: Oof, this seems error prone, could we define CookieSettings interface for stricter type setting? 
(I am assuming this is low effort, not sure in how many other places this object is being used in)

kishore03109: HAHA just realised this JS file psps

kishore03109 (approved):

PR #674, opened 2023-03-31
hotfix(0.18.1): merge into develop

seaerchin:
## Problem
1. old sites that are on github but not on db are mistakenly returned as 404 as the db query is empty
2. `DOMPurify` too restrictive on BE
## Solution
1. check data on github in addition to checking sites. 
2. use same setup as FE


kishore03109 (approved):
Nits: 
we shouldnt have an additional call to github if the site already exists in our db. 
But since this is non-functional and more of an optimisation, we can tackle this after our hotfix!

src/routes/v2/authenticated/review.ts
```diff
@@ -70,6 +96,16 @@ export class ReviewsRouter {
     const { userWithSiteSessionData } = res.locals
     const { siteName } = req.params
 
+    const site = await this.sitesService.getBySiteName(siteName)
+    const doesUnmigratedSiteExist = await this.doesUnmigratedSiteExist(
+      userWithSiteSessionData
+    )
+
+    console.log("DOES IT", doesUnmigratedSiteExist)
```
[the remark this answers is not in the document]
seaerchin: removed

seaerchin:
> Nits: we shouldnt have an additional call to github if the site already exists in our db. But since this is non-functional and more of an optimisation, we can tackle this after our hotfix!

yea but then it becomes q hard to read cos you have 2 nested `ifs`

src/routes/v2/authenticated/review.ts
```diff
@@ -41,25 +42,50 @@ export class ReviewsRouter {
...
+    sessionData: UserWithSiteSessionData
+  ): Promise<boolean> => {
+    try {
+      // Site exists on github but isn't migrated
+      await this.githubService.getRepoInfo(sessionData)
+      return true
+    } catch {
+      // If there are any errors, we failed to fetch the site
+      // and the site is assumed to not exist.
+      return false
```
harishv7, line 82: in the event site does not exist, what response does the github API return?

Also, when we say errors, do we mean any response apart from 2xx goes into the catch block?

src/routes/v2/authenticated/review.ts
```diff
@@ -70,6 +96,14 @@ export class ReviewsRouter {
     const { userWithSiteSessionData } = res.locals
     const { siteName } = req.params
 
+    const site = await this.sitesService.getBySiteName(siteName)
+    const doesUnmigratedSiteExist = await this.doesUnmigratedSiteExist(
```
harishv7, line 100: Both var and function is same name, can we rename one of it?

src/routes/v2/authenticated/review.ts
seaerchin: returns 404 on our side - this means that it falls through to the default case where the site really doesn't exist

re: second point, yes. broadly, if it's a 4xx, we can omit the response from github because it's user based so our subsequent flow should cover it. if it's truly github's fault (5xx), we can't determine if the site exists or not, so returning that it **doesn't** is a safer choice than saying it does (and possibly being wrong)

kishore03109 (approved):

src/routes/v2/authenticated/review.ts
harishv7: okay makes sense

PR #679, opened 2023-04-05
Feat/site creation form email

alexanderleegs:
This PR modifies our existing site creation form to utilise the new email login flow instead. Currently, our site creation form creates a new website, but does not assign site members to the new site - thus our creation flow currently creates an Amplify site that is accessed via Github login, with access manually granted to users via Github. This PR changes the creation flow to associate an existing isomer user to the specified site, or create a new isomer user using the specified email if no such user exists. A new form has also been created for both staging and production ([staging form link](https://form.gov.sg/642be6e0682ef3001171d860)). Note that the person filling in the form is still specified as the owner of the form.

Env var changes before merging:
- Modify `SITE_CREATE_FORM_KEY` on production - we're using a new site create form (v3) just to be safe, so the form key needs to be swapped out to the new secret key

### DIFF
--- a/src/routes/formsgSiteCreation.ts
+++ b/src/routes/formsgSiteCreation.ts
@@ -20,6 +20,7 @@ const SITE_CREATE_FORM_KEY = config.get("formSg.siteCreateFormKey")
 const REQUESTER_EMAIL_FIELD = "Government E-mail"
 const SITE_NAME_FIELD = "Site Name"
 const REPO_NAME_FIELD = "Repository Name"
+const OWNER_NAME_FIELD = "Site Owner E-mail"
 
 export interface FormsgRouterProps {
   usersService: UsersService
@@ -51,6 +52,7 @@ export class FormsgRouter {
     const requesterEmail = getField(responses, REQUESTER_EMAIL_FIELD)
     const siteName = getField(responses, SITE_NAME_FIELD)
     const repoName = getField(responses, REPO_NAME_FIELD)
+    const ownerEmail = getField(responses, OWNER_NAME_FIELD)?.toLowerCase()
 
     logger.info(
       `Create site form submission [${submissionId}] (repoName '${repoName}', siteName '${siteName}') requested by <${requesterEmail}>`
@@ -75,16 +77,18 @@ export class FormsgRouter {
         await this.sendCreateError(requesterEmail, repoName, submissionId, err)
         return res.sendStatus(200)
       }
-      const foundUser = await this.usersService.findByEmail(requesterEmail)
-      if (!foundUser) {
+      const foundRequester = await this.usersService.findByEmail(requesterEmail)
+      if (!foundRequester) {
         const err = `Form submitter ${requesterEmail} is not an Isomer user. Register an account for this user and try again.`
         await this.sendCreateError(requesterEmail, repoName, submissionId, err)
         return res.sendStatus(200)
       }
+      const foundOwner = await this.usersService.findOrCreateByEmail(ownerEmail)
 
       // 3. Use service to create site
       const { deployment } = await this.infraService.createSite(
-        foundUser,
+        foundRequester,
+        foundOwner,
         siteName,
         repoName
       )
--- a/src/routes/formsgSiteCreation.ts
+++ b/src/routes/formsgSiteCreation.ts
@@ -77,8 +77,10 @@ export class FormsgRouter {
         await this.sendCreateError(requesterEmail, repoName, submissionId, err)
         return res.sendStatus(200)
       }
-      const foundRequester = await this.usersService.findByEmail(requesterEmail)
-      if (!foundRequester) {
+      const foundIsomerRequester = await this.usersService.findByEmail(
+        requesterEmail
+      )
+      if (!foundIsomerRequester) {
         const err = `Form submitter ${requesterEmail} is not an Isomer user. Register an account for this user and try again.`
         await this.sendCreateError(requesterEmail, repoName, submissionId, err)
         return res.sendStatus(200)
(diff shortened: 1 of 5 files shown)


src/routes/formsgSiteCreation.ts
```diff
@@ -75,16 +77,18 @@ export class FormsgRouter {
         await this.sendCreateError(requesterEmail, repoName, submissionId, err)
         return res.sendStatus(200)
       }
-      const foundUser = await this.usersService.findByEmail(requesterEmail)
-      if (!foundUser) {
+      const foundRequester = await this.usersService.findByEmail(requesterEmail)
```
kishore03109, line 80: Nit: Could we rename this to `foundIsomerRequester` so we don't confuse this person with a `agencyFoundRequester`

src/routes/formsgSiteCreation.ts
```diff
@@ -75,16 +77,18 @@ export class FormsgRouter {
...
+      if (!foundRequester) {
         const err = `Form submitter ${requesterEmail} is not an Isomer user. Register an account for this user and try again.`
         await this.sendCreateError(requesterEmail, repoName, submissionId, err)
         return res.sendStatus(200)
       }
+      const foundOwner = await this.usersService.findOrCreateByEmail(ownerEmail)
```
kishore03109, line 86: Do we need to handle the case where `foundOwner` is not defined here?

src/services/infra/InfraService.ts
```diff
@@ -49,27 +51,45 @@ export default class InfraService {
...
+      throw new Error(
+        `createSite: initial member for ${siteName} does not have associated email`
+      )
+    }
     try {
       // 1. Create a new site record in the Sites table
       const newSiteParams = {
         name: siteName,
-        apiTokenName: "", // TODO: figure this out
+        apiTokenName: "", // TODO: remove once DB has removed this param
```
kishore03109, line 92: do we have an issue/jira for this?

src/services/infra/InfraService.ts
```diff
@@ -96,6 +116,7 @@ export default class InfraService {
 
       // 5. Set up permissions
       await this.reposService.setRepoAndTeamPermissions(repoName)
+      await this.collaboratorsService.create(repoName, memberEmail, true)
```
kishore03109, line 119: Hmm just to clarify, why is acknowledged = true over here ah? In what case will acknowledged = false ah?

kishore03109 (commented):
Do we have an example of a repo that has been created via this new form as a sanity check that this works?

src/routes/formsgSiteCreation.ts
```diff
@@ -75,16 +77,18 @@ export class FormsgRouter {
         await this.sendCreateError(requesterEmail, repoName, submissionId, err)
         return res.sendStatus(200)
       }
-      const foundUser = await this.usersService.findByEmail(requesterEmail)
-      if (!foundUser) {
+      const foundRequester = await this.usersService.findByEmail(requesterEmail)
+      if (!foundRequester) {
         const err = `Form submitter ${requesterEmail} is not an Isomer user. Register an account for this user and try again.`
         await this.sendCreateError(requesterEmail, repoName, submissionId, err)
```
harishv7, line 83: sorry a bit confused, this is an error flow right? Why are we returning 200 res after sendCreateError?

src/routes/formsgSiteCreation.ts
alexanderleegs: Fixed in https://github.com/isomerpages/isomercms-backend/pull/679/commits/f6cdb2b61f8f92ed453bf8782a2acae660fc95ad!

src/routes/formsgSiteCreation.ts
alexanderleegs: Nope - foundOwner retrieves an existing user if one with that email exists in our db, and creates one otherwise, so a user is always returned

src/services/infra/InfraService.ts
alexanderleegs: Added in https://github.com/isomerpages/isomercms-backend/pull/679/commits/926b0023857832cb6191e15e7932b39024ab3612!

src/services/infra/InfraService.ts
alexanderleegs: We're reusing the collaborators method here - this was a verification that when adding collaborators (non govt email users), that the user had acknowledged on the frontend. It's not applicable here, so it's always true

src/routes/formsgSiteCreation.ts
alexanderleegs: This endpoint is only triggered by forms, which expects a 200 response! The error handling is done separately via email

src/routes/formsgSiteCreation.ts
```diff
@@ -51,6 +52,7 @@ export class FormsgRouter {
     const requesterEmail = getField(responses, REQUESTER_EMAIL_FIELD)
     const siteName = getField(responses, SITE_NAME_FIELD)
     const repoName = getField(responses, REPO_NAME_FIELD)
+    const ownerEmail = getField(responses, OWNER_NAME_FIELD)?.toLowerCase()
```
kishore03109, line 55: Nit: could we do a strip on the string? 
worried if ops put space then the string equality operation fails and it registers as new email 
(maybe this error case is captured downstream)

kishore03109 (approved):

src/routes/formsgSiteCreation.ts
alexanderleegs: good catch, added in https://github.com/isomerpages/isomercms-backend/pull/679/commits/80cfbbd63ade3c5d9dc78a2992b19c8edbf1a89a

PR #680, opened 2023-04-05
fix(utils): change order of ops and rec sanitization

seaerchin:
## Problem
the problem is that sanitizing in special pages (homepage, contact us, nav) leads to html tags being removed. though this is expected behaviour, it destructively interferes with user edits.

an example of this is when the user uses <a href = "something" /> to direct users to their internal link; this gets removed sadly and causes errors when we do yaml.parse post sanitization.

## Solution
sanitize the values post parsing rather than treating the yaml as a raw blob

## Dependencies
Upgraded `yaml` from 1.x to 2.x in order to use the `reviver/replacer` functionality. This operates on the value post parsing, which is what we desire. See [here](https://eemeli.org/yaml/#parse-amp-stringify) for more info on this.

This is a [breaking change](https://github.com/eemeli/yaml/releases/tag/v2.0.0) but we only use `.parse` (verified by global search on codebase) so this a safe upgrade for us

### DIFF
--- a/src/utils/yaml-utils.ts
+++ b/src/utils/yaml-utils.ts
@@ -1,12 +1,29 @@
+import _ from "lodash"
 import yaml from "yaml"
 
 import { sanitizer } from "@services/utilServices/Sanitizer"
+
+type YamlRecord = {
+  [key: string]: YamlRecord | string
+}
+
+const sanitizeYamlRecord = (yamlRec: YamlRecord): YamlRecord =>
+  _(yamlRec)
+    .mapValues((val) => {
+      if (typeof val === "string") return sanitizer.sanitize(val)
+      return sanitizeYamlRecord(val)
+    })
+    .value()
 
 // Note: `yaml.parse()` and `yaml.stringify()` should not be used anywhere
 // else in the codebase.
 export const sanitizedYamlParse = (
   unparsedContent: string
-): Record<string, unknown> => yaml.parse(sanitizer.sanitize(unparsedContent))
+): Record<string, unknown> => {
+  const parsedContent = yaml.parse(unparsedContent) as YamlRecord
+
+  return sanitizeYamlRecord(parsedContent)
+}
 
 export const sanitizedYamlStringify = (prestringifiedContent: object): string =>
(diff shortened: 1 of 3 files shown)


alexanderleegs (commented):
converting to draft - debugging issue where nav bar is not loading the sanitised content properly

src/utils/yaml-utils.ts
```diff
@@ -6,7 +6,12 @@ import { sanitizer } from "@services/utilServices/Sanitizer"
 // else in the codebase.
 export const sanitizedYamlParse = (
   unparsedContent: string
-): Record<string, unknown> => yaml.parse(sanitizer.sanitize(unparsedContent))
+): Record<string, unknown> =>
+  yaml.parse(unparsedContent, (key, value) =>
+    typeof value === "string" ? sanitizer.sanitize(value) : value
+  )
```
alexanderleegs, line 12: Just checking, wouldn't this mean that any nested fields (e.g. 
```
operating_hours:
      - days: aa
        time: "12"
        description: "123"
      - days: bb
        time: "13"
        description: <script>alert('hi')</script>
```

would end up unsanitised? It gets blocked by our csp right now, but the page itself is able to save this without sanitisation

seaerchin: since the doc for yaml itself points you to the `json.parse` [page](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse#using_the_reviver_parameter), i'm just gonna paste the ss here to save you the effort HAHA

Screenshot 2023-04-06 at 1 25 10 PM

the effect of this means that only **leaf nodes** of type `string` gets sanitized and we leave safe terminal values of other types (yaml has `null | boolean | number | string` for scalar values) untouched

seaerchin: turns out this affects `contact-us` and more due to it calling the v1 endpoint to update pages. this isn't caused by this PR and has been found to work on prod (albeit no impact due to CSP blocking). 

filed as IS-75 due i'll be resolving this as it's out of scope

alexanderleegs (approved):

PR #681, opened 2023-04-05
Chore/update node

alexanderleegs:
Duplicate of https://github.com/isomerpages/isomercms-backend/pull/650 - we dropped this commit in our previous release because we wanted to separate the release of identity features and the node upgrade, which involved a change in our eb environment.

This PR also additionally modifies our ci, to use our new staging and production environments.


.nvmrc
```diff
@@ -1 +1 @@
-14
\ No newline at end of file
+18
```
harishv7, line 1: 16 or 18?

alexanderleegs: 18 - this is for our local dev, and what engines specifies is the version of node that our app will use

so overall we'll be using node 18 for local dev, and node 16 for deployment - this is fine since we're not using any node 18 specific features

harishv7: makes sense, thanks!

harishv7 (approved):

kishore03109 (approved):

PR #683, opened 2023-04-06
chore(review): fix tests for review router

seaerchin:
## Problem
Last release added a bunch of checks for sites that are undergoing a migration from github into our db (but not yet completed); this clashes with our code, which expects a **fully** migrated site (sites table in db populated, site members retrieved from db etc). 

This leads to a couple of tests failing in `review.spec.ts` as the calls to calculate those pre-conditions are not mocked. This PR fixes this by mocking said calls.

## Solution
for methods that have the new checks, there is a new `beforeEach` hook added to mock the call to the precondition. This causes the check to be false (ie, that the site has underwent a full migration), and continues on with the business logic rather than short-circuiting

### DIFF
--- a/src/routes/v2/authenticated/__tests__/review.spec.ts
+++ b/src/routes/v2/authenticated/__tests__/review.spec.ts
@@ -137,7 +137,8 @@ describe("Review Requests Router", () => {
 
   describe("compareDiff", () => {
     beforeEach(() => {
-      // NOTE: Skip preconditions check for unmigrated site
+      // TODO (IS-58): Skip preconditions check for unmigrated site
+      // Remove this when sites are fully migrated over to email login
       mockCollaboratorsService.list.mockResolvedValueOnce([{}])
     })
     it("should return 200 with the list of changed files", async () => {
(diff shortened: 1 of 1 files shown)


alexanderleegs (approved):
I think the cleaner way of doing this would be to also insert the list of collaborators by adding user/site db entries before each of these tests. I'm fine with releasing this as is for now, but we should change it later on

src/routes/v2/authenticated/__tests__/review.spec.ts
```diff
@@ -136,6 +136,10 @@ describe("Review Requests Router", () => {
   })
 
   describe("compareDiff", () => {
+    beforeEach(() => {
+      // NOTE: Skip preconditions check for unmigrated site
```
harishv7, line 140: Can we put as a TODO to remove this after email migration is complete?

seaerchin:
> I think the cleaner way of doing this would be to also insert the list of collaborators by adding user/site db entries before each of these tests. I'm fine with releasing this as is for now, but we should change it later on

this isn't an integration test so the service itself is mocked and there's no actual db calls being made. the integration test itself is fine as those actually exist

src/routes/v2/authenticated/__tests__/review.spec.ts
seaerchin: sure, changing to `TODO (IS-58)` inline with the actual code itself

PR #689, opened 2023-04-10
chore: add ga4 to request schema

alexanderleegs:
This PR adds the `google_analytics_ga4` field to our request schema to support the addition of the GA4 field. To be reviewed in conjunction with PR #[1229](https://github.com/isomerpages/isomercms-frontend/pull/1229) on the isomercms-frontend repo.


src/fixtures/config.js
```diff
@@ -56,6 +56,7 @@ is_government: false
 shareicon: /images/isomer-logo.svg
 facebook-pixel: "true"
 google_analytics: UA-39345131-33
+google_analytics_ga4: GA-test
```
harishv7, line 59: where will the test value be replaced with an actual value?

src/classes/Settings.js
```diff
@@ -109,6 +109,7 @@ class Settings {
       is_government: configContent.is_government,
       facebook_pixel: configContent["facebook-pixel"],
       google_analytics: configContent.google_analytics,
```
harishv7, line 111: guessing this is the old GA tag that we will deprecate after a while?

alexanderleegs: Yup, we'll be deprecating this after July when google removes support for the old google analytics (UA)

src/fixtures/config.js
alexanderleegs: We never actually test this with actual values - since this is only retrieved from github, and we mock github calls in all our tests, none of our backend tests would use an actual value of this, and the other values here are dummy values taken from our default template params iirc

We only check for actual values along with the settings e2e tests on the frontend, but those tests are currently not working properly

src/classes/Settings.js
harishv7: got it

src/validators/RequestSchema.js
```diff
@@ -251,6 +251,7 @@ const UpdateSettingsRequestSchema = Joi.object().keys({
   favicon: Joi.string(),
   "facebook-pixel": Joi.string().allow(""),
   google_analytics: Joi.string().allow(""),
```
kishore03109, line 253: Style: could we rename `google_analytics` to `google_analytics_ua` for clarity?

harishv7 (approved):

kishore03109 (approved):
Mostly lgtm, would prefer to rename all instances of `google_analytics` to `google_analytics_ua` for clarity

src/validators/RequestSchema.js
alexanderleegs: I'm reluctant to do this because the `google_analytics` param already exists on repos currently - we'd need to run a migration on all existing sites to change `google_analytics` to `google_analytics_ua` in their `config.yml`

PR #692, opened 2023-04-11
Fix/rename subcollection and rename/delete resource room

alexanderleegs:
## Problem

This PR fixes an issue with our subcollection rename and resource room rename/delete methods. For subcollection rename, when renaming subcollections, we no longer use the `baseDirectoryService.rename` but instead delete and recreate each individual file, meaning it was no longer necessary to pass `githubSessionData`, but our call site still passed this as an argument, which resulted in subcollection renaming not working as expected. The resource room delete/rename methods were missing the `githubSessionData` argument. This PR fjxes this behaviour.

In addition, the existing use of `reqDetails` was changed to `sessionData` to bring it in line with the rest of our methods to reduce confusion.

### DIFF
--- a/src/routes/v2/authenticatedSites/__tests__/Collections.spec.js
+++ b/src/routes/v2/authenticatedSites/__tests__/Collections.spec.js
@@ -330,15 +330,11 @@ describe("Collections Router", () => {
         .expect(200)
       expect(
         mockSubcollectionDirectoryService.renameDirectory
-      ).toHaveBeenCalledWith(
-        mockUserWithSiteSessionData,
-        mockGithubSessionData,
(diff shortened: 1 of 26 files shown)


seaerchin (approved):
lgtm, there wasn't any logic changes beyond `src/routes/v2/authenticatedSites/collections.js` right? (where we deleted the extra param)

alexanderleegs:
> lgtm, there wasn't any logic changes beyond `src/routes/v2/authenticatedSites/collections.js` right? (where we deleted the extra param)

same logic changes for resource room renaming and deletion (remove extra param) - other than that it was just var renaming

PR #694, opened 2023-04-11
fix(constants): fixed import

seaerchin:
## Problem
Old constant was being imported as `HOMEPAGE_NAME` and not `HOMEPAGE_FILENAME`

## Solution
Fixed import


alexanderleegs (approved):
lgtm, remember to rebase

PR #698, opened 2023-04-12
chore(site launch form): trim site launch form input

kishore03109:
## Problem

The form expects string values. When typing, it is possible for someone to enter a space in the string

## Solution

Defensively trim all inputs

### DIFF
--- a/src/utils/formsg-utils.ts
+++ b/src/utils/formsg-utils.ts
@@ -7,7 +7,25 @@ export function getField(
 ): string | undefined {
   const response = responses.find(({ question }) => question === name)
 
-  return response?.answer
+  return response?.answer?.trim()
+}
+
+function trimAllStrings(
+  responseArray: string[] | string[][]
+): string[] | string[][] {
+  if (Array.isArray(responseArray)) {
+    responseArray.map((item) => {
+      if (Array.isArray(item)) {
+        return trimAllStrings(item)
+      }
+      if (typeof item === "string") {
+        return item.trim()
+      }
+      return item
+    })
+    return responseArray
+  }
+  return responseArray
 }
 
 export function getFieldsFromTable(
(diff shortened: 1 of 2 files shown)


alexanderleegs (commented):
Actually, is defensive trimming something we should have as a default on all our methods which use formsg input? I'm thinking it might be better to trim from `getField` in `formsg-utils` instead

kishore03109:
@alexanderleegs oh hey this is fair!

src/utils/formsg-utils.ts
```diff
@@ -7,13 +7,35 @@ export function getField(
...
+        return trimAllStrings(item)
+      }
+      if (typeof item === "string") {
+        return item.trim()
+      }
+      return item
+    })
+    return responseArray
+  }
+  return responseArray
```
alexanderleegs, line 28: Thoughts on reducing the checks here, given that the method signature is for string[] | string[][]?

src/utils/formsg-utils.ts
```diff
@@ -7,13 +7,35 @@ export function getField(
...
 export function getFieldsFromTable(
   responses: FormField[],
   name: string
 ): string[] | string[][] | undefined {
   const response = responses.find(({ question }) => question === name)
+  let answers = response?.answerArray
+  if (answers) {
+    answers = trimAllStrings(answers)
+  }
   return response?.answerArray
```
alexanderleegs, line 40: Just checking - does this modify `response`? I think it might be better to rewrite it so it's clearly regardless

kishore03109: nope, this is a bug, fixed here [08433bb](https://github.com/isomerpages/isomercms-backend/pull/698/commits/08433bb83269ceb74044f3e9e7a6dc6e1c4fd4e8)

src/utils/formsg-utils.ts
kishore03109: TS complains if we dont have this because it is not able to infer that 
if (Array.isArray(responseArray)), then responseArray is of type []. I have added in a comment for clarity here: 
[08433bb](https://github.com/isomerpages/isomercms-backend/pull/698/commits/08433bb83269ceb74044f3e9e7a6dc6e1c4fd4e8)

alexanderleegs (approved):
lgtm

PR #699, opened 2023-04-12
Is 59/fix/disallow special character as first character

alexanderleegs:
This PR replicates the frontend changes made to restrict the use of `#` as the first character in files and subcollection titles. To be reviewed in conjunction with PR [#1231](https://github.com/isomerpages/isomercms-frontend/pull/1231) on the isomercms frontend repo.

This PR also updates our logging for failure cases on these create/rename endpoints, as the error message was previously not very informative.

### DIFF
--- a/src/validators/validators.js
+++ b/src/validators/validators.js
@@ -1,4 +1,5 @@
 const specialCharactersRegexTest = /[~%^*_+\-./\\`;~{}[\]"<>]/
+const jekyllFirstCharacterRegexTest = /^[._#~]/
 const dateRegexTest = /^[0-9]{4}-[0-9]{2}-[0-9]{2}$/
 
 const mediaSpecialCharactersRegexTest = /[~%^#*+\./\\`;~{}[\]"<>]/ // Allows dashes
@@ -11,7 +12,10 @@ const titleSpecialCharCheck = ({ title, isFile = false }) => {
     // Remove .md
     testTitle = title.replace(/.md$/, "")
   }
-  return specialCharactersRegexTest.test(testTitle)
+  return (
+    specialCharactersRegexTest.test(testTitle) ||
+    jekyllFirstCharacterRegexTest.test(testTitle)
+  )
 }
 
 const isDateValid = (date) => dateRegexTest.test(date)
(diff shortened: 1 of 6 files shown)


seaerchin (commented):
changes made look fine but just asking if we need to make the corresponding changes also onto `v1` endpoints.

from the jekyll docs [here](https://jekyllrb.com/docs/structure/), it seems like the special characters also shouldn't be part of directory names. at present, this is also only done for `SubcollectionDirectoryService` -  should we do for collections + resources as a whole?

seaerchin (approved):
approving first because this fixes a known issue but it might be a partial fix

src/validators/validators.js
```diff
@@ -11,7 +12,10 @@ const titleSpecialCharCheck = ({ title, isFile = false }) => {
     // Remove .md
     testTitle = title.replace(/.md$/, "")
   }
-  return specialCharactersRegexTest.test(testTitle)
+  return (
+    specialCharactersRegexTest.test(testTitle) ||
+    jekyllFirstCharacterRegexTest.test(testTitle)
```
kishore03109, line 17: Hi shouldnt this be a '&&'  rather than a "||"?

alexanderleegs:
> changes made look fine but just asking if we need to make the corresponding changes also onto `v1` endpoints.
> 
> from the jekyll docs [here](https://jekyllrb.com/docs/structure/), it seems like the special characters also shouldn't be part of directory names. at present, this is also only done for `SubcollectionDirectoryService` - should we do for collections + resources as a whole?



> changes made look fine but just asking if we need to make the corresponding changes also onto `v1` endpoints.
> 
> from the jekyll docs [here](https://jekyllrb.com/docs/structure/), it seems like the special characters also shouldn't be part of directory names. at present, this is also only done for `SubcollectionDirectoryService` - should we do for collections + resources as a whole?

V1 endpoints for file/resource/directories don't have any type checking - we've thankfully managed to move completely off v1 for these endpoints as far as i know.

For the collections/resources, those already have even stricter restrictions (no symbols at all), so no changes to them required

src/validators/validators.js
alexanderleegs: `||` is correct - this method is checking for the existence of any invalid characters, but I think the method name is confusing about this. I've renamed the method to `hasSpecialCharInTitle` in https://github.com/isomerpages/isomercms-backend/pull/699/commits/b8a41a3b1937f5aa2852f58780fed2d4e660d99f

kishore03109 (approved):

PR #700, opened 2023-04-12
fix(authservice): return `userId` when e2e

seaerchin:
Related to IS-87

## Solution
add `userId` property when e2e

### DIFF
--- a/src/services/utilServices/AuthService.js
+++ b/src/services/utilServices/AuthService.js
@@ -141,7 +141,11 @@ class AuthService {
   async getUserInfo(sessionData) {
     try {
       if (sessionData.isomerUserId === E2E_ISOMER_ID) {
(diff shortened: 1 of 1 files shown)


harishv7 (approved):

harishv7:
> we should probably maintain 2 types of users for e2e (github + email) or, if we decide that we want to deprecate github in the future, make our existing tests compat w/ email flow

Have we created an epic/tickets for this?

seaerchin:
> > we should probably maintain 2 types of users for e2e (github + email) or, if we decide that we want to deprecate github in the future, make our existing tests compat w/ email flow
> 
> Have we created an epic/tickets for this?

not at the moment cos i wanted to see which flow y'all would be more comfortable with. i'm just gonna file the comment as an issue as-is HAHA but just putting it here that i think maintaining 2 types of E2E (at first) is more robust

seaerchin:
file as IS-99

PR #704, opened 2023-04-13
Feat(Site Launch): Mock calls to Amplify 

kishore03109:
## Problem

During development, the CreateDomainAssociation calls are rate-limited to 10 per hour. This low rate limit can cause issues during actual site launches that may affect operations.

## Solution

To avoid impeding with the actual infrastructure rate limits, this PR mocks all calls to CreateDomainAssociation and GetDomainAssociation in LaunchesService. Only staging, vapt, and prod environments should be making actual calls to Amplify. Additionally, since we are mocking Amplify calls now, we no longer need to wait for 90 seconds after creating domain associations for Amplify to generate the required values.

The mocked objects may seem verbose, but they closely resemble the actual values received when calling Amplify.

## Breaking Changes

There are no breaking changes.

## Reviewer Notes
I am concerned that future devs working around site launch feature might be confused why they are not able to create domain associations during development. Not too sure if there is a way to highlight clearly in code that the calls are mocks apart from the proposed solution of already naming them  `mockCreateDomainAssociationOutput` and `mockCreateDomainAssociationOutput` in `LaunchesClient`. Open to feedback on this! 

## Tests

To mock form responses, you can call the following code directly from server.js:
```
const formResponses = [
  {
    submissionId: "",
    requesterEmail: "kishore@open.gov.sg",
    repoName: "kishore-test",
    primaryDomain: "kishorewithwww.isomer.gov.sg",
    redirectionDomain: "www.kishorewithwww.isomer.gov.sg",
    agencyEmail: "alexander@open.gov.sg",
  },
  {
    submissionId: "",
    requesterEmail: "kishore@open.gov.sg",
    repoName: "test-vincent",
    primaryDomain: "kishorewithoutwww.isomer.gov.sg",
    redirectionDomain: "",
    agencyEmail: "alexander@open.gov.sg",
  },
]

formsgSiteLaunchRouter.handleSiteLaunchResults(formResponses, "test")
```
After mocking a site launch form submission, the launches and redirections in the database are populated, and emails are sent as expected.
Screenshot 2023-04-13 at 2 12 43 PM
Screenshot 2023-04-13 at 2 12 57 PM

## Deploy Notes 

- `MOCK_AMPLIFY_CREATE_DOMAIN_CALLS` : mocking amplify calls so as to not impede with operations 

Values in 1pw prod and staging env vars are also updated. TLDR; MOCK_AMPLIFY_CREATE_DOMAIN_CALLS is `true` in prod and staging env, `false` in local dev.

### DIFF
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -172,6 +172,12 @@ const config = convict({
         format: String,
         default: "",
       },
+      mockAmplifyCreateDomainAssociationCalls: {
+        doc: "Mock createDomainAssociation calls to Amplify ",
+        env: "MOCK_AMPLIFY_CREATE_DOMAIN_ASSOCIATION_CALLS",
+        format: "required-boolean",
+        default: true,
+      },
     },
     sqs: {
       incomingQueueUrl: {
--- a/src/routes/formsgSiteLaunch.ts
+++ b/src/routes/formsgSiteLaunch.ts
@@ -208,7 +208,7 @@ export class FormsgSiteLaunchRouter {
       ) || []
 
     res.sendStatus(200) // we have received the form and obtained relevant field
-    await this.handleSiteLaunchResults(formResponses, submissionId)
+    this.handleSiteLaunchResults(formResponses, submissionId)
   }
 
   sendLaunchError = async (
@@ -301,33 +301,39 @@ export class FormsgSiteLaunchRouter {
     formResponses: FormResponsesProps[],
     submissionId: string
   ) {
-    const launchResults = await Promise.all(
-      formResponses.map(this.launchSiteFromForm)
-    )
-    const successResults: DnsRecordsEmailProps[] = []
-    launchResults.forEach((launchResult) => {
-      if (launchResult.isOk()) {
-        successResults.push(launchResult.value)
-      }
-    })
+    try {
+      const launchResults = await Promise.all(
+        formResponses.map(this.launchSiteFromForm)
+      )
+      const successResults: DnsRecordsEmailProps[] = []
+      launchResults.forEach((launchResult) => {
+        if (launchResult.isOk()) {
+          successResults.push(launchResult.value)
+        }
+      })
 
-    await this.sendDNSRecords(submissionId, successResults)
+      await this.sendDNSRecords(submissionId, successResults)
 
-    const failureResults: LaunchFailureEmailProps[] = []
+      const failureResults: LaunchFailureEmailProps[] = []
 
-    launchResults.forEach((launchResult) => {
-      if (launchResult.isErr() && launchResult.error) {
-        failureResults.push(launchResult.error)
-        return
-      }
-      if (launchResult.isErr()) {
-        failureResults.push({
-          error: "Unknown error",
-        })
-      }
-    })
+      launchResults.forEach((launchResult) => {
+        if (launchResult.isErr() && launchResult.error) {
+          failureResults.push(launchResult.error)
+          return
+        }
+        if (launchResult.isErr()) {
+          failureResults.push({
+            error: "Unknown error",
+          })
+        }
+      })
 
-    await this.sendLaunchError(submissionId, failureResults)
+      await this.sendLaunchError(submissionId, failureResults)
+    } catch (e) {
+      logger.error(
+        `Something unexpected went wrong when launching sites from form submission ${submissionId}. Error: ${e}`
+      )
+    }
   }
 
   getRouter() {
--- a/src/services/identity/LaunchClient.ts
+++ b/src/services/identity/LaunchClient.ts
@@ -155,8 +155,8 @@ class LaunchClient {
         statusReason: undefined,
       },
     }
-    const hasCreated = true // this is a get call, assume domain has already been created
-    const subDomains = this.getSubDomains(input.subDomainSettings, hasCreated)
+    const isCreated = true // this is a get call, assume domain has already been created
+    const subDomains = this.getSubDomains(input.subDomainSettings, isCreated)
 
     // We know that domainAssociation is not undefined, so we can use the non-null assertion operator
     // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
--- a/src/services/identity/LaunchClient.ts
+++ b/src/services/identity/LaunchClient.ts
@@ -10,7 +10,7 @@ import {
 } from "@aws-sdk/client-amplify"
 import { SubDomain } from "aws-sdk/clients/amplify"
 
-import { config } from "@config/config"
+import { isSecure } from "@root/utils/auth-utils"
 
 // create a new interface that extends GetDomainAssociationCommandInput
 interface MockGetDomainAssociationCommandInput
@@ -114,7 +114,7 @@ class LaunchClient {
   }
 
   private isTestEnv() {
-    return config.get("env") === "test" || config.get("env") === "dev"
+    return isSecure
   }
 
   private getSubDomains(subDomainList: SubDomainSetting[], hasCreated = false) {
--- a/src/services/identity/LaunchClient.ts
+++ b/src/services/identity/LaunchClient.ts
@@ -8,10 +8,15 @@ import {
   GetDomainAssociationCommandOutput,
   SubDomainSetting,
 } from "@aws-sdk/client-amplify"
-import { options } from "joi"
+import { SubDomain } from "aws-sdk/clients/amplify"
 
-import logger from "@root/logger/logger"
+import config from "@config/config"
 
+// create a new interface that extends GetDomainAssociationCommandInput
+interface MockGetDomainAssociationCommandInput
+  extends GetDomainAssociationCommandInput {
+  subDomainSettings: SubDomainSetting[]
+}
 
 class LaunchClient {
   private readonly amplifyClient: InstanceType<typeof AmplifyClient>
@@ -35,21 +40,130 @@ class LaunchClient {
 
   sendCreateDomainAssociation = (
     input: CreateDomainAssociationCommandInput
-  ): Promise<CreateDomainAssociationCommandOutput> =>
-    this.amplifyClient.send(new CreateDomainAssociationCommand(input))
+  ): Promise<CreateDomainAssociationCommandOutput> => {
+    if (this.isTestEnv()) {
+      return this.mockCreateDomainAssociationOutput(input)
+    }
+    const output = this.amplifyClient.send(
+      new CreateDomainAssociationCommand(input)
+    )
+    return output
+  }
 
   createGetDomainAssociationCommandInput = (
     appId: string,
-    domainName: string
-  ): GetDomainAssociationCommandInput => ({
+    domainName: string,
+    subDomainSettings: SubDomainSetting[]
+  ):
+    | GetDomainAssociationCommandInput
+    | MockGetDomainAssociationCommandInput => ({
     appId,
     domainName,
+    ...(this.isTestEnv() && { subDomainSettings }),
   })
 
   sendGetDomainAssociationCommand = (
-    input: GetDomainAssociationCommandInput
-  ): Promise<GetDomainAssociationCommandOutput> =>
-    this.amplifyClient.send(new GetDomainAssociationCommand(input))
+    input:
+      | GetDomainAssociationCommandInput
+      | MockGetDomainAssociationCommandInput
+  ): Promise<GetDomainAssociationCommandOutput> => {
+    const isMockInput = "subDomainSettings" in input
+    if (isMockInput) {
+      // handle mock input
+      return this.mockGetDomainAssociationOutput(input)
+    }
+    return this.amplifyClient.send(new GetDomainAssociationCommand(input))
+  }
+
+  /**
+   * The rate limit for Create Domain Association is 10 per hour.
+   * We want to limit interference with operations, as such we mock this call during development.
+   * @returns Mocked output for CreateDomainAssociationCommand
+   */
+  mockCreateDomainAssociationOutput = (
+    input: CreateDomainAssociationCommandInput
+  ): Promise<CreateDomainAssociationCommandOutput> => {
+    const mockResponse: CreateDomainAssociationCommandOutput = {
+      $metadata: {
+        httpStatusCode: 200,
+      },
+      domainAssociation: {
+        autoSubDomainCreationPatterns: [],
+        autoSubDomainIAMRole: undefined,
+        certificateVerificationDNSRecord: undefined,
+        domainAssociationArn: `arn:aws:amplify:ap-southeast-1:11111:apps/${input.appId}/domains/${input.domainName}`,
+        domainName: input.domainName,
+        domainStatus: "CREATING",
+        enableAutoSubDomain: false,
+        statusReason: undefined,
+        subDomains: undefined,
+      },
+    }
+
+    const subDomainSettingsList = input.subDomainSettings
+    if (!subDomainSettingsList || subDomainSettingsList.length === 0) {
+      return Promise.resolve(mockResponse)
+    }
+
+    const subDomains: SubDomain[] = this.getSubDomains(subDomainSettingsList)
+
+    // We know that domainAssociation is not undefined, so we can use the non-null assertion operator
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    mockResponse.domainAssociation!.subDomains = subDomains
+    return Promise.resolve(mockResponse)
+  }
+
+  private isTestEnv() {
+    return config.get("env") === "test" || config.get("env") === "dev"
+  }
+
+  private getSubDomains(subDomainList: SubDomainSetting[], hasCreated = false) {
+    const subDomainPrefixList = subDomainList
+      .map((subDomain) => subDomain.prefix)
+      .filter((prefix) => prefix !== undefined) as string[]
+
+    const subDomains: SubDomain[] = []
+    subDomainPrefixList.forEach((subDomainPrefix) => {
+      subDomains.push({
+        dnsRecord: `${subDomainPrefix} CNAME ${
+          hasCreated ? "test.cloudfront.net" : "<pending>"
+        }`,
+        subDomainSetting: {
+          branchName: "master",
+          prefix: subDomainPrefix,
+        },
+        verified: false,
+      })
+    })
+    return subDomains
+  }
+
+  mockGetDomainAssociationOutput(
+    input: MockGetDomainAssociationCommandInput
+  ): Promise<GetDomainAssociationCommandOutput> {
+    const mockResponse: GetDomainAssociationCommandOutput = {
+      $metadata: {
+        httpStatusCode: 200,
+      },
+      domainAssociation: {
+        certificateVerificationDNSRecord: `testcert.${input.domainName}. CNAME testcert.acm-validations.aws.`,
+        domainAssociationArn: `arn:aws:amplify:ap-southeast-1:11111:apps/${input.appId}/domains/${input.domainName}`,
+        domainName: input.domainName,
+        domainStatus: "PENDING_VERIFICATION",
+        enableAutoSubDomain: false,
+        subDomains: undefined,
+        statusReason: undefined,
+      },
+    }
+    const hasCreated = true // this is a get call, assume domain has already been created
+    const subDomains = this.getSubDomains(input.subDomainSettings, hasCreated)
+
+    // We know that domainAssociation is not undefined, so we can use the non-null assertion operator
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    mockResponse.domainAssociation!.subDomains = subDomains
+
+    return Promise.resolve(mockResponse)
+  }
 }
 
 export default LaunchClient
--- a/src/services/identity/LaunchClient.ts
+++ b/src/services/identity/LaunchClient.ts
@@ -119,31 +119,33 @@ class LaunchClient {
 
   private getSubDomains(
     subDomainList: SubDomainSetting[],
-    hasCreated = false
+    isCreated = false
   ): SubDomain[] {
-    const subDomainPrefixList = subDomainList
+    const subDomains = subDomainList
       .map((subDomain) => subDomain.prefix)
-      .filter((prefix) => prefix !== undefined) as string[]
-
-    const subDomains: SubDomain[] = []
-    subDomainPrefixList.forEach((subDomainPrefix) => {
-      subDomains.push({
+      .filter((prefix) => prefix !== undefined && prefix !== null)
+      .map((subDomainPrefix) => ({
         dnsRecord: `${subDomainPrefix} CNAME ${
-          hasCreated ? "test.cloudfront.net" : "<pending>"
+          isCreated ? "test.cloudfront.net" : "<pending>"
         }`,
         subDomainSetting: {
           branchName: "master",
           prefix: subDomainPrefix,
         },
         verified: false,
-      })
-    })
+      })) as SubDomain[]
     return subDomains
   }
 
   mockGetDomainAssociationOutput(
     input: MockGetDomainAssociationCommandInput
   ): Promise<GetDomainAssociationCommandOutput> {
+    const isSubDomainCreated = true // this is a `get` call, assume domain has already been created
+    const subDomains = this.getSubDomains(
+      input.subDomainSettings,
+      isSubDomainCreated
+    )
+
     const mockResponse: GetDomainAssociationCommandOutput = {
       $metadata: {
         httpStatusCode: 200,
@@ -154,16 +156,10 @@ class LaunchClient {
         domainName: input.domainName,
         domainStatus: "PENDING_VERIFICATION",
         enableAutoSubDomain: false,
-        subDomains: undefined,
+        subDomains,
         statusReason: undefined,
       },
     }
-    const isCreated = true // this is a get call, assume domain has already been created
-    const subDomains = this.getSubDomains(input.subDomainSettings, isCreated)
-
-    // We know that domainAssociation is not undefined, so we can use the non-null assertion operator
-    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-    mockResponse.domainAssociation!.subDomains = subDomains
 
     return Promise.resolve(mockResponse)
   }
--- a/src/services/identity/LaunchClient.ts
+++ b/src/services/identity/LaunchClient.ts
@@ -83,7 +83,7 @@ class LaunchClient {
    * We want to limit interference with operations, as such we mock this call during development.
    * @returns Mocked output for CreateDomainAssociationCommand
    */
-  mockCreateDomainAssociationOutput = (
+  private mockCreateDomainAssociationOutput = (
     input: CreateDomainAssociationCommandInput
   ): Promise<CreateDomainAssociationCommandOutput> => {
     // We are mocking the response from the Amplify API, so we need to store the input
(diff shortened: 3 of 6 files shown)


src/services/identity/LaunchClient.ts
```diff
@@ -35,21 +40,130 @@ class LaunchClient {
...
+
+    const subDomains: SubDomain[] = this.getSubDomains(subDomainSettingsList)
+
+    // We know that domainAssociation is not undefined, so we can use the non-null assertion operator
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    mockResponse.domainAssociation!.subDomains = subDomains
+    return Promise.resolve(mockResponse)
+  }
+
+  private isTestEnv() {
```
harishv7, line 116: should we make this a generic util function?

src/services/identity/LaunchClient.ts
```diff
@@ -35,21 +40,130 @@ class LaunchClient {
...
+    return config.get("env") === "test" || config.get("env") === "dev"
+  }
+
+  private getSubDomains(subDomainList: SubDomainSetting[], hasCreated = false) {
```
harishv7, line 120: nit: is `isCreated` better?

src/services/identity/LaunchClient.ts
```diff
@@ -35,21 +40,130 @@ class LaunchClient {
...
+    const subDomainPrefixList = subDomainList
+      .map((subDomain) => subDomain.prefix)
+      .filter((prefix) => prefix !== undefined) as string[]
+
+    const subDomains: SubDomain[] = []
```
harishv7, line 125: nit: `subdomain` is a single word so don't need camelcase technically

src/services/identity/LaunchClient.ts
```diff
@@ -35,21 +40,130 @@ class LaunchClient {
...
+  ): Promise<GetDomainAssociationCommandOutput> {
+    const mockResponse: GetDomainAssociationCommandOutput = {
+      $metadata: {
+        httpStatusCode: 200,
+      },
+      domainAssociation: {
+        certificateVerificationDNSRecord: `testcert.${input.domainName}. CNAME testcert.acm-validations.aws.`,
+        domainAssociationArn: `arn:aws:amplify:ap-southeast-1:11111:apps/${input.appId}/domains/${input.domainName}`,
+        domainName: input.domainName,
+        domainStatus: "PENDING_VERIFICATION",
```
harishv7, line 152: Is this the expected status? What's the terminal status? "Verified"?

src/services/identity/LaunchClient.ts
```diff
@@ -35,21 +40,130 @@ class LaunchClient {
...
+    const subDomainSettingsList = input.subDomainSettings
+    if (!subDomainSettingsList || subDomainSettingsList.length === 0) {
+      return Promise.resolve(mockResponse)
+    }
+
+    const subDomains: SubDomain[] = this.getSubDomains(subDomainSettingsList)
+
+    // We know that domainAssociation is not undefined, so we can use the non-null assertion operator
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    mockResponse.domainAssociation!.subDomains = subDomains
```
harishv7, line 112: We need `!` because `domainAssociation` is optional in `CreateDomainAssociationCommandOutput`?

src/services/identity/LaunchClient.ts
```diff
@@ -35,21 +40,130 @@ class LaunchClient {
...
+    return Promise.resolve(mockResponse)
+  }
+
+  private isTestEnv() {
+    return config.get("env") === "test" || config.get("env") === "dev"
```
harishv7, line 117: ideally we should be using enums for comparison like ` === NODE_ENVS.TEST`

src/services/identity/LaunchClient.ts
harishv7, line 120: what's the return type of this function?

src/services/identity/LaunchClient.ts
```diff
@@ -35,21 +40,130 @@ class LaunchClient {
...
+  }
+
+  private getSubDomains(subDomainList: SubDomainSetting[], hasCreated = false) {
+    const subDomainPrefixList = subDomainList
+      .map((subDomain) => subDomain.prefix)
```
harishv7, line 122: what does a prefix of a subdomain mean?

src/services/identity/LaunchClient.ts
```diff
@@ -35,21 +40,130 @@ class LaunchClient {
...
+      },
+    }
+    const hasCreated = true // this is a get call, assume domain has already been created
+    const subDomains = this.getSubDomains(input.subDomainSettings, hasCreated)
+
+    // We know that domainAssociation is not undefined, so we can use the non-null assertion operator
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    mockResponse.domainAssociation!.subDomains = subDomains
+
+    return Promise.resolve(mockResponse)
```
harishv7, line 165: thought this is not async, we use Promise here to simulate an actual `GetDomainAssociationCommandOutput` call right?

src/services/infra/InfraService.ts
```diff
@@ -226,23 +228,28 @@ export default class InfraService {
...
+      // since we mock values during development, we don't have to await for the dns records
+      if (!isTestEnv) {
+        /**
+         * note: we wait for ard 90 sec as there is a time taken
+         * for amplify to generate the certification manager in the first place
+         * This is a dirty workaround for now, and will cause issues when we integrate
+         * this directly within the Isomer CMS.
+         * todo: push this check into a queue-like system when integrating this with cms
+         */
+        await new Promise((resolve) => setTimeout(resolve, 90000))
```
harishv7, line 242: This is going to block the main thread for 90s right?

Seems like "todo: push this check into a queue-like system when integrating this with cms" is crucial for integration with CMS

harishv7:
> I am concerned that future devs working around site launch feature might be confused why they are not able to create domain associations during development. Not too sure if there is a way to highlight clearly in code that the calls are mocks apart from the proposed solution of already naming them mockCreateDomainAssociationOutput and mockCreateDomainAssociationOutput in LaunchesClient. Open to feedback on this!

@kishore03109 maybe do you want to have an actual dev config like "MOCK_AMPLIFY_CALLS = true or false" which gives devs more "conscious" control over this? If you feel this could extend beyond in future as the launch service develops, you might want to have a great detail of specificity apart from `isTestEnv()`.

src/services/identity/LaunchClient.ts
kishore03109: 4bc54963

src/services/identity/LaunchClient.ts
kishore03109: bc6aa75c already have util function 
 keeping the `isTestEnv()` function because I feel using that over `isSecure` represents the intent of the code better

src/services/identity/LaunchClient.ts
kishore03109: hm Amplify client uses this spelling, you ok if I keep it consistent throughout?

src/services/identity/LaunchClient.ts
kishore03109: Yup, the moment we create the domain and get the required DNS values, we dont expect a success state at the moment. the full list of status can be found [here](https://repost.aws/questions/QUxsRQ8_2KSJ68j-Dk5M7Xvg/amplify-domain-creation-statuses-what-do-they-mean) are  
PENDING_VERIFICATION | IN_PROGRESS | AVAILABLE | PENDING_DEPLOYMENT | FAILED | CREATING | REQUESTING_CERTIFICATE | UPDATING

src/services/identity/LaunchClient.ts
kishore03109: yup

src/services/identity/LaunchClient.ts
kishore03109: I am in favour of not using enums tho, in line with Chin's reasoning in the previous PRs. Shall we discuss offline about this?

src/services/identity/LaunchClient.ts
kishore03109: Updated here 117ef09c

src/services/identity/LaunchClient.ts
kishore03109: eg. when my subdomain is `www.blah.sg`
my prefix is `www`

when my subdomain is `blah.sg`
my prefix is ``

src/services/identity/LaunchClient.ts
kishore03109: yup, agreed

src/services/infra/InfraService.ts
kishore03109: hmm correct me if I am wrong, but set setTimeout does [not block the main thread right](https://developer.mozilla.org/en-US/docs/Web/API/setTimeout)?

kishore03109:
@harishv7 
> @kishore03109 maybe do you want to have an actual dev config like "MOCK_AMPLIFY_CALLS = true or false" which gives devs more "conscious" control over this? If you feel this could extend beyond in future as the launch service develops, you might want to have a great detail of specificity apart from isTestEnv().

Hey thanks for the suggestion, think this is a cleaner solution, [updated](fa70525b) to use an env var. Regarding greater detail of specificity, I think we can revisit this as the service develops, I think this is ok for now!

src/services/identity/LaunchClient.ts
harishv7: okay can

src/services/identity/LaunchClient.ts
harishv7: Then do we return `AVAILABLE` or `VERIFIED`? I don't see those statuses in the mock response

src/services/infra/InfraService.ts
harishv7: @kishore03109 hmm, you can do a simple test on console -
```
async function test() { if(true) { await new Promise((resolve) => setTimeout(resolve, 2000)) } console.log("DONE") }
```

"DONE" wont be printed until the promise is done. you might want to double check the actual behaviour on this

harishv7:
> @harishv7
> 
> > @kishore03109 maybe do you want to have an actual dev config like "MOCK_AMPLIFY_CALLS = true or false" which gives devs more "conscious" control over this? If you feel this could extend beyond in future as the launch service develops, you might want to have a great detail of specificity apart from isTestEnv().
> 
> Hey thanks for the suggestion, think this is a cleaner solution, [updated](fa70525b) to use an env var. Regarding greater detail of specificity, I think we can revisit this as the service develops, I think this is ok for now!

Sounds good, actually `MOCK_AMPLIFY_CREATE_DOMAIN_CALLS` is very specific already, as it specifies exactly that only "Create Domain" calls are mocked.

If there are a lot of mocking calls required, then a more generic env like `MOCK_ALL_AMPLIFY_CALLS` may be required. You can revise this in future if required

src/services/identity/LaunchClient.ts
kishore03109: The sequence of events is 
1. Send CreateDomainAssociation Call <-- at this point we don't know the DNS records to set, Amplify still generating 
2. Send GetDomainAssociationCall <-- at this point we call this just to get the new DNS records to set 

At the later point, the call is not meant to be `Available` or `Verified`. This is not the part of the process where the user has already put in the DNS records. As such, to mimic the actual behaviour from Amplify, we use `PENDING_VERIFICATION`

src/services/infra/InfraService.ts
kishore03109: Hmm wait isnt that the point? I want to only resolve this after 90 seconds to give Amplify enough time to generate the DNS results

eg for the code
"""
async function test() {
  await new Promise((resolve) => setTimeout(resolve, 10000))
  console.log("PROMISE DONE")
}
console.log("STARTING")
test()
console.log("ENDED")
"""

in the console log, we would see:
STARTING
ENDED
PROMISE DONE

thus the other functions in the thread will not be blocked?

src/services/identity/LaunchClient.ts
```diff
@@ -35,21 +40,133 @@ class LaunchClient {
...
+  private getSubDomains(
+    subDomainList: SubDomainSetting[],
+    hasCreated = false
+  ): SubDomain[] {
+    const subDomainPrefixList = subDomainList
+      .map((subDomain) => subDomain.prefix)
+      .filter((prefix) => prefix !== undefined) as string[]
+
+    const subDomains: SubDomain[] = []
+    subDomainPrefixList.forEach((subDomainPrefix) => {
```
seaerchin, line 129: nit: can chain this off the previous `filter` as a `map`

src/services/identity/LaunchClient.ts
```diff
@@ -35,21 +40,133 @@ class LaunchClient {
...
+        subDomains: undefined,
+        statusReason: undefined,
+      },
+    }
+    const isCreated = true // this is a get call, assume domain has already been created
+    const subDomains = this.getSubDomains(input.subDomainSettings, isCreated)
+
+    // We know that domainAssociation is not undefined, so we can use the non-null assertion operator
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    mockResponse.domainAssociation!.subDomains = subDomains
```
seaerchin, line 166: we can structure this whole chunk better by placing this up top and using `subDomains` to create `mockResponse`

src/services/infra/InfraService.ts
seaerchin: think the confusion stems from the `await` - if you `await` in an async function: 

```ts
const x = await work()
doOtherWork()
const y = x
```

the main thread is free to continue on until the result of the `await`-ed promise is used (ie, `const y = x`). however, because we `await` here, i think we'll pause execution for 90s

src/services/infra/InfraService.ts
```diff
@@ -368,7 +375,7 @@ export default class InfraService {
     await Promise.all(
       messages.map(async (message) => {
         const site = await this.sitesService.getBySiteName(message.repoName)
-        if (!site) {
+        if (!site || site.isErr()) {
```
seaerchin, line 378: nit: delete `!site`

seaerchin (commented):
generally when we mock, we want to preserve the same function contract to ensure that calling code behaves as intended. moreover, because of how we use DI, we are able to substitute out the inner client (`amplifyClient`) for a mocked one, which ensures that our calling code is never aware of the fact that it's calling a fake client.

in here, the API contract is changed, which might lead to callers not upholding the contract and test logic is intermingled w/ actual biz logic, which seems a little messy.

wdyt about creating a `MOCK_AMPLIFY_CLIENT` that returns const data and passing in the appropriate client based on teh result of `isTest`?

kishore03109:
@seaerchin 
> generally when we mock, we want to preserve the same function contract to ensure that calling code behaves as intended. moreover, because of how we use DI, we are able to substitute out the inner client (amplifyClient) for a mocked one, which ensures that our calling code is never aware of the fact that it's calling a fake client.
in here, the API contract is changed, which might lead to callers not upholding the contract and test logic is intermingled w/ actual biz logic, which seems a little messy.
wdyt about creating a MOCK_AMPLIFY_CLIENT that returns const data and passing in the appropriate client based on teh result of isTest?

Yup, you are right to say that this is not ideal where test logic is intermingled with actual biz logic, and seems messy. 

We primarily have two main calls to Amplify here: 
1) `CreateDomainAssociation` 
2) `GetDomainAssociation`

Originally, when actually calling Amplify, the `getDomainAssociation` only takes in the arguments `domainName` and `appId`, since the state of the application (specifically the SubDomainSettings[]) are stored in Amplify itself. To fully mock this behaviour, one must store this state, either in the database (which is def an over-engineered solution) or as a static variable in `MockAmplifyClient`. 

Say we store this state in some static variable `AmplifyDomains` in a new class `MockAmplifyClient` . To fully mock the `GetDomainAssociation` during development, one must continue to manipulate `AmplifyDomains` to fit the expected mocking values. I am foreseeing that this would make writing tests in the future harder, because it might not be immediately obvious to the developer that some variable in `MockAmplifyClient` needs to be modified in order for an accurate representation of what the Amplify Client would have returned if this was not mocked. This leads to an implicit dependency here, since the precondition for receiving an accurate mock from `GetDomainAssociation` is to set the static variable `AmplifyDomains` properly.

To be clear, changing the API contract here is a _**concession**_ to give more power to the consumer to give a less ambiguous, clearer response of what the Amplify client will return if the values are not mocked. Does this reasoning make sense?

src/services/identity/LaunchClient.ts
kishore03109: 6f7872fa

src/services/infra/InfraService.ts
kishore03109: good catch 6f7872fa

src/services/identity/LaunchClient.ts
harishv7: ok got it

src/services/identity/LaunchClient.ts
harishv7: for this, i recommended moving it to a `utils/` directory as it seems common enough for other services to use apart from LaunchClient

Regarding enums, we can always use a simple object instead right?

```
const Roles = {
  Admin: "admin",
  Writer: "writer",
  Reader: "reader"
} as const;

// Convert object key in a type
type RoleKeys = typeof Roles[keyof typeof Roles]

declare function hasAccess(role: RoleKeys): void;
```
Ref: https://dev.to/ivanzm123/dont-use-enums-in-typescript-they-are-very-dangerous-57bh

src/services/identity/LaunchClient.ts
harishv7: got it thanks

src/services/identity/LaunchClient.ts
kishore03109: oh yes, agreed! Since this file no longer checks the env var (it checks to a new var `mockAmplifyCreateDomainAssociationCalls`), this is beyond the scope of the pr, happy to tackle this in a separate issue instead! Feel free to resolve this if you have no other concerns!

src/services/infra/InfraService.ts
harishv7: @kishore03109 if you added `await` on test:

```
console.log("STARTING")
await test()
console.log("ENDED")
```

it will wait for promise to be done first

seaerchin:
> @seaerchin
> 
> > generally when we mock, we want to preserve the same function contract to ensure that calling code behaves as intended. moreover, because of how we use DI, we are able to substitute out the inner client (amplifyClient) for a mocked one, which ensures that our calling code is never aware of the fact that it's calling a fake client.
> > in here, the API contract is changed, which might lead to callers not upholding the contract and test logic is intermingled w/ actual biz logic, which seems a little messy.
> > wdyt about creating a MOCK_AMPLIFY_CLIENT that returns const data and passing in the appropriate client based on teh result of isTest?
> 
> Yup, you are right to say that this is not ideal where test logic is intermingled with actual biz logic, and seems messy.
> 
> We primarily have two main calls to Amplify here:
> 
> 1. `CreateDomainAssociation`
> 2. `GetDomainAssociation`
> 
> Originally, when actually calling Amplify, the `getDomainAssociation` only takes in the arguments `domainName` and `appId`, since the state of the application (specifically the SubDomainSettings[]) are stored in Amplify itself. To fully mock this behaviour, one must store this state, either in the database (which is def an over-engineered solution) or as a static variable in `MockAmplifyClient`. (Please let me know if there is another way that I missed :/ )
> 
> Say we store this state in some static variable `AmplifyDomains` in a new class `MockAmplifyClient` . To fully mock the `GetDomainAssociation` during development, one must continue to manipulate `AmplifyDomains` to fit the expected mocking values. I am foreseeing that this would make writing tests in the future harder, because it might not be immediately obvious to the developer that some variable in `MockAmplifyClient` needs to be modified in order for an accurate representation of what the Amplify Client would have returned if this was not mocked. This leads to an implicit dependency here, since the precondition for receiving an accurate mock from `GetDomainAssociation` is to set the static variable `AmplifyDomains` properly.
> 
> To be clear, changing the API contract here is a _**concession**_ to give more power to the consumer to give a less ambiguous, clearer response of what the Amplify client will return if the values are not mocked. Does this reasoning make sense?

hmm i don't understand what you're trying to say here - this only applies if we ever hit the amplify endpoint, which we're not doing (by design). so why can't we just give a constant result of the expected shape back to the caller? we don't have to store anything/compute anything since the code in `localdev` won't ever hit amplify anyway, no?

kishore03109:
@seaerchin 
>hmm i don't understand what you're trying to say here - this only applies if we ever hit the amplify endpoint, which we're not doing (by design). so why can't we just give a constant result of the expected shape back to the caller? we don't have to store anything/compute anything since the code in localdev won't ever hit amplify anyway, no?

the issue stems from the fact that the result of `GetDomainAssociationCall` semantically differs depending on arguments given to `CreateDomainAssociation` to trigger the domain association process. 

super high level code to demonstrate this:

when actually calling Amplify: 
```
createDomainAssociation({appId: blah, domainName: 'test.gov.sg', subDomainSettings: 'www'})
// some code to wait 90 secs
const res = getDomainAssociation({appId: blah, domainName:'test.gov.sg'})
if (res.subDomains) {
  // lets call this node in the CFG as node A 
  handleSubdomains(res) 
}  else {
  // lets call this node in the CFG as node B 
  handleDomainsWithoutSubdomains(res)
}
```

When mocking in local dev, if we returned an object of 
const res = {
  subDomains: "www"
...otherProps
}, node A will never be run. 

When mocking, if we returned an object of 
const res = {
  ...otherProps
}, node B will never be run. 

To give more flexibility to the caller, the suggested solution proposes adding the prop `subDomainSettings` to fine-tune the mocked result returned!
Concretely I can call 
getDomainAssociation({appId: blah, domainName: 'test.gov.sg', subDomainSettings: 'www'}) or 
getDomainAssociation({appId: blah, domainName: 'test.gov.sg'}) depending on whether or not I wish to receive a mocked Response from amplify with subdomains or no subdomains.

src/services/infra/InfraService.ts
kishore03109: @harishv7, for ease of mind, I have made the main caller (`this.handleSiteLaunchResults`) function non-blocking (ie, removing the `await`), since we technically can run this async after returning the 200 response. However, this means that any error thrown by this will not be caught by `handleSiteLaunchFormRequest`. Therefore, I am wrapping a try catch around and logging the error to any unintended effects of some unexpected error. Ideally, during an error, this would return something to the user, but I think this can be revisited when we are integrating the features with the frontend!

kishore03109:
@seaerchin as discussed offline, have [modified](https://github.com/isomerpages/isomercms-backend/pull/704/commits/a1fa956f10772bd191626759710a01343e2597da) to retain the original API contract. The state is now stored in `LaunchesClient`, who is the only one who is aware of any mocked values.

src/routes/formsgSiteLaunch.ts
```diff
@@ -301,33 +301,39 @@ export class FormsgSiteLaunchRouter {
...
-    const successResults: DnsRecordsEmailProps[] = []
-    launchResults.forEach((launchResult) => {
-      if (launchResult.isOk()) {
-        successResults.push(launchResult.value)
-      }
-    })
+    try {
+      const launchResults = await Promise.all(
+        formResponses.map(this.launchSiteFromForm)
+      )
```
seaerchin, line 307: hmm just beware that `Promise.all` means the resulting `Promise` succeeds only when **all** the promises succeed. 

if what you want is just for the promise to _complete_, the behaviour you want is `Promise.allSettled`

see [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled) for the `allSettled` behaviour

src/config/config.ts
```diff
@@ -172,6 +172,12 @@ const config = convict({
         format: String,
         default: "",
       },
+      mockAmplifyCreateDomainAssociationCalls: {
+        doc: "Mock createDomainAssociation calls to Amplify ",
+        env: "MOCK_AMPLIFY_CREATE_DOMAIN_ASSOCIATION_CALLS",
+        format: "required-boolean",
+        default: true,
+      },
```
seaerchin, line 180: pls update staging/prod 1PW so we don't default to `true` :monkas:

src/services/identity/LaunchClient.ts
```diff
@@ -1,26 +1,41 @@
...
   SubDomainSetting,
 } from "@aws-sdk/client-amplify"
-import { options } from "joi"
+import { SubDomain } from "aws-sdk/clients/amplify"
 
-import logger from "@root/logger/logger"
+import { config } from "@config/config"
 
+// stricter typing to interact with Amplify SDK
+type CreateDomainAssociationCommandInput = {
```
seaerchin, line 16: fyi, if the intention here is to also make all the keys required, then you'd need to add `Required` around the curly braces.

otherwise, this would still allow `undefined` for optional keys. (can be seen by mousing over)

src/services/identity/LaunchClient.ts
```diff
@@ -48,8 +70,107 @@ class LaunchClient {
...
+      .map((subDomainPrefix) => ({
+        dnsRecord: `${subDomainPrefix} CNAME ${
+          isCreated ? "test.cloudfront.net" : "<pending>"
+        }`,
+        subDomainSetting: {
+          branchName: "master",
+          prefix: subDomainPrefix,
+        },
+        verified: false,
+      })) as SubDomain[]
```
seaerchin, line 141: why do we need the cast here?

src/services/identity/LaunchClient.ts
```diff
@@ -48,8 +70,107 @@ class LaunchClient {
...
+    return subDomains
+  }
+
+  mockGetDomainAssociationOutput(
```
seaerchin, line 145: can make this `private`? internals shouldn't be exposed

src/services/infra/InfraService.ts
```diff
@@ -226,23 +228,28 @@ export default class InfraService {
...
+         * todo: add some level of retry logic if get domain association command
+         * does not contain the DNS redirections info.
+         */
+      }
 
       const dnsInfo = await this.launchesService.getDomainAssociationRecord(
         primaryDomain,
-        appId
+        appId,
+        subDomainSettings
```
seaerchin, line 252: seems to be wrong? this method only takes 2 arguments

src/services/infra/InfraService.ts
seaerchin: i think this is non-blocking - top-most caller is `handleSiteLaunchFormRequest` in `formsgSiteLaunch` and it calls `this.handleSiteLaunchResults(formResponses, submissionId)` without `await` (which calls this method later on)

seaerchin (commented):
wrong method signature

src/routes/formsgSiteLaunch.ts
kishore03109: good point, since this the 'launchSiteFromForm' never throws an error but returns it, i shall to keep this behaviour!

src/config/config.ts
kishore03109: Yes, 1pw values are already updated!

src/services/identity/LaunchClient.ts
kishore03109: nope, got some keys that we dont need to configure! I was thinking we can be stricter with our typing if we need to modify those values in the future!

src/services/identity/LaunchClient.ts
kishore03109: Filter doesn't narrow types :( 
Primarily this stems from `subDomainPrefix` not being narrowed to `string` after the filer, I have [changed](602b6f8bc03138ee091d291b2d54864d262f466c) the code to reflect this. 

Am opting against using a type guard here since this only checks if a string is undefined, opting for the more simplistic solution here

src/services/identity/LaunchClient.ts
kishore03109: 10e26d10  good point, mocking both create and get calls!

src/services/infra/InfraService.ts
kishore03109: 9e661017 sorry, must have missed this :(

seaerchin (approved):
lgtm

PR #718, opened 2023-04-19
[LS-81] fix(markdown-utils): change sanitization process + add unescape

seaerchin:
## Problem
Previously, there was inconsistent behaviour caused by sanitization on the backend. This is because of dompurify's sanitization config, where it will automatically html encode certain special characters if it detects that there is a html tag. 

This process affects not just content within the tag, but the string as a whole. For example, 
```ts
const x = "& <b>&something</b>"
```
will have both ampersands encoded even though the first one is outside of the b-tag. 

Closes LS-81

## Solution
In order to make sure that sanitization takes place properly, this PR establishes an invariant, as follows: 
**frontmatter content is never html encoded**. 

This is chosen over html encoding all content in our frontmatter due to the existence of the 3 special pages (homepage/nav/contact-us), where users are able to input html (:sadge:)

In order to preserve this property, a few rules have to be followed (done alr at present)
1. devs never call `sanitize` directly for markdown files but through a given interface (`<convert|retrieve>DataFromMarkdown`). 
2. the respective conversion/retrieval functions separately sanitize frontmatter/body 
3. we run `unescape` after sanitizing frontmatter. 

This allows us

## Testing 
- [ ] navigate to the CMS and choose a testing site
- [ ] make the page have special character in frontmatter (can be done through editing the page directly on github if lazy)
- [ ] rename a page to have a special character in the title
- [ ] check that clicking settings on the page does **not** show a html encoded title

**Notes**
This change **might be destructive** - existing pages w html encoded properties in their frontmatter will have them unescaped. I'm not entirely sure how this impacts things like `permalink` etc but it's also possible to guarantee only for the `title` property

### DIFF
--- a/src/utils/markdown-utils.js
+++ b/src/utils/markdown-utils.js
@@ -12,11 +12,17 @@ const getTrailingSlashWithPermalink = (permalink) =>
 
 const retrieveDataFromMarkdown = (fileContent) => {
   // eslint-disable-next-line no-unused-vars
-  const [unused, encodedFrontMatter, ...pageContent] = sanitizer
-    .sanitize(fileContent)
-    .split("---")
+  const [unused, encodedFrontMatter, ...pageContent] = fileContent.split("---")
+  // NOTE: We separate the sanitization into 2 steps.
+  // This is because DOMPurify does URL encoding when it detects html in the string.
+  // For example, `<b>&something</b>` will get escaped to `<b>&amp;something</b>`.
+  // To prevent this behaviour from affecting our frontmatter, we do the sanitization separately
+  // on the frontmatter and the content
   const frontMatter = sanitizedYamlParse(encodedFrontMatter)
-  return { frontMatter, pageContent: pageContent.join("---").trim() }
+  return {
+    frontMatter,
+    pageContent: sanitizer.sanitize(pageContent.join("---")).trim(),
+  }
 }
 
 const isResourceFileOrLink = (frontMatter) => {
--- a/src/utils/markdown-utils.js
+++ b/src/utils/markdown-utils.js
@@ -15,10 +15,20 @@ const retrieveDataFromMarkdown = (fileContent) => {
   const [unused, encodedFrontMatter, ...pageContent] = fileContent.split("---")
   // NOTE: We separate the sanitization into 2 steps.
   // This is because DOMPurify does URL encoding when it detects html in the string.
-  // For example, `<b>&something</b>` will get escaped to `<b>&amp;something</b>`.
+  // For example, `<b>&something</b>` will get HTML encoded to `<b>&amp;something</b>`.
   // To prevent this behaviour from affecting our frontmatter, we do the sanitization separately
   // on the frontmatter and the content
-  const frontMatter = sanitizedYamlParse(encodedFrontMatter)
+  const frontMatter = _.mapValues(
+    sanitizedYamlParse(encodedFrontMatter),
+    // NOTE: We call `unescape` here to transform `&amp` into `&`.
+    // Because of the above property, where DOMPurify does html encoding on detection of a html tag,
+    // there might be special characters that are encoded into their html form.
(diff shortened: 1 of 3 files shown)


src/utils/markdown-utils.js
```diff
@@ -12,11 +12,27 @@ const getTrailingSlashWithPermalink = (permalink) =>
...
+  const frontMatter = _.mapValues(
+    sanitizedYamlParse(encodedFrontMatter),
+    // NOTE: We call `unescape` here to transform `&amp` into `&`.
+    // Because of the above property, where DOMPurify does html encoding on detection of a html tag,
+    // there might be special characters that are encoded into their html form.
+    // This is a safe transformation to run because the original value was already a special character
+    // so this does not do anything destructive.
+    // Do note that frontmatter containing pre-existing html encoded characters (&amp;)
+    // will get transformed regardless.
+    (val) => _.unescape(val)
```
kishore03109, line 30: Nit: Hmm this solution escapes all html characters, wdyt about only escaping `&`, since that is the most common case that we are encountering at the moment?

kishore03109 (commented):
> This change might be destructive - existing pages w html encoded properties in their frontmatter will have them unescaped. I'm not entirely sure how this impacts things like permalink etc but it's also possible to guarantee only for the title property

hmm, if we are not sure the possible implications of this, is there a reason why we don't make this change only for the title property then to reduce surface area of bugs?

alexanderleegs (commented):
Screenshot 2023-04-20 at 11 29 45 AM

creating new pages creates this commented out thing at the moment!

alexanderleegs:
> @alexanderleegs tagging you separately in case there is anything that isn't back-compat

I think this is fine, we previously didn't do any html encoding for the special pages either as far as i know?

src/utils/markdown-utils.js
seaerchin: the common case isn't the _only_ case - this means that the same bug can appear, just with a different character that's escaped. in the event that it happens, we'd have to expend eng resources to dig through + fix so i'd rather just escape all

kishore03109: Do you think there are going to be additional security concerns that we might have with allowing *all* html for all frontmatter?
Considering we have CSP headers + this is already the case for some pages, it seems ok, but just checking in

seaerchin:
> Screenshot 2023-04-20 at 11 29 45 AM
> 
> creating new pages creates this commented out thing at the moment!

does this actually impact editing experience? this is injected due to `sanitize` encountering an empty body and [injecting it into a document](https://github.com/cure53/DOMPurify/blob/main/src/purify.js#L1395). if it's concerning, we could always avoid sanitization if it's an empty string

src/utils/markdown-utils.js
seaerchin: this PR doesn't allow/disallow html, it just escapes encoded html present in frontmatter. the sanitization invariant is still preserved (front matter still sanitised)

kishore03109 (approved):
nit: 

> does this actually impact editing experience? this is injected due to sanitize encountering an empty body and [injecting it into a document](https://github.com/cure53/DOMPurify/blob/main/src/purify.js#L1395). if it's concerning, we could always avoid sanitization if it's an empty string

this does seem confusing at the first glance... could we add a test case to show this behaviour is expected?

alexanderleegs:
> > Screenshot 2023-04-20 at 11 29 45 AM
> > creating new pages creates this commented out thing at the moment!
> 
> does this actually impact editing experience? this is injected due to `sanitize` encountering an empty body and [injecting it into a document](https://github.com/cure53/DOMPurify/blob/main/src/purify.js#L1395). if it's concerning, we could always avoid sanitization if it's an empty string

Could we put in the check for empty string then? As a user creating a new page, having something you didn't input immediately show up in your editing view probably isn't ideal

alexanderleegs (approved):

PR #719, opened 2023-04-19
style(env var): change naming of var

kishore03109:
## Problem
Technically we mock both `createDomainAssociation` and `getDomainAssociation` calls, therefore var name is not really accurate
## Solution
changing the var name before release today to reflect this. 

1pw already updated with the new values. 

## Deploy notes

new env var: 
export MOCK_AMPLIFY_DOMAIN_ASSOCIATION_CALLS="false"


seaerchin (approved):
Lgtm, remember to update 1pw if needed

PR #720, opened 2023-04-20
feat(site launch): Allow for ops to remove timed out domain associations

kishore03109:
## Problem

If we try to associate the same domain twice, Amplify will generally throw a `BadRequestException: One or more domains requested are already associated with another Amplify app`, which can inform the user to delete + retry. 

However, this did not hold true if the domain association process timed out, in which case Amplify did _**not**_ throw an error, and neither did it create a new Domain Association during the `CreateDomainAssociation` call. 

To replicate the bug, this sequence of errors would need to happen: 
1) Submit the form for blah.gov.sg for blah-test app.
2) Receive an email for the DNS Results
3) Wait for 24 hours for domain association to time out
4) Retry step 1 
5) Receive an email with the exact same DNS results as step 2. 
One would think that the email received in step 5 is the most recent values, but it is in fact a lingering DomainAssociation that has yet to be deleted. 

Closes IS93

## Solution

Before calling a `CreateDomainAssociation`, delete any failed domain associations for the application. 

**Breaking Changes**

  - Details ...

## Testing 
While I have tested it manually on the `kishore-test` app, it is a bit hard to easily replicate this test this as it requires the existence of an App timeout over 24 hours. :/

### DIFF
--- a/src/services/identity/LaunchClient.ts
+++ b/src/services/identity/LaunchClient.ts
@@ -7,10 +7,15 @@ import {
   GetDomainAssociationCommandInput as AmplifySDKGetDomainAssociationCommandInput,
   GetDomainAssociationCommandOutput,
   SubDomainSetting,
+  DeleteDomainAssociationCommandInput as AmplifySDKDeleteDomainAssociationCommandInput,
+  DeleteDomainAssociationCommand,
+  NotFoundException,
 } from "@aws-sdk/client-amplify"
 import { SubDomain } from "aws-sdk/clients/amplify"
 
 import { config } from "@config/config"
+
+import { AmplifyError } from "@root/types"
 
 // stricter typing to interact with Amplify SDK
 type CreateDomainAssociationCommandInput = {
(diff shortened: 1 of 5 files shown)


src/services/identity/LaunchClient.ts
```diff
@@ -142,13 +168,33 @@ class LaunchClient {
...
+    }
+    await this.amplifyClient.send(new DeleteDomainAssociationCommand(input))
+  }
+
+  mockDeleteDomainAssociationOutput(
+    input: DeleteDomainAssociationCommandInput
+  ) {
+    if (!this.mockDomainAssociations.has(input.domainName)) {
+      throw new AmplifyError(
+        `NotFoundException: Domain association ${input.domainName} not found.`
```
kishore03109, line 185: This mocks the behaviour of the real amplify client and throws an exception if domain hasnt been created yet

src/services/identity/LaunchesService.ts
```diff
@@ -159,6 +161,41 @@ export class LaunchesService {
...
+      const getDomainAssociationCommandInput = this.launchClient.createGetDomainAssociationCommandInput(
+        appIdResult.value,
+        domainName
+      )
+
+      const getDomainAssociationResult = await this.launchClient.sendGetDomainAssociationCommand(
+        getDomainAssociationCommandInput
+      )
+      const hasDomainAssociationFailed =
+        getDomainAssociationResult?.domainAssociation?.domainStatus === "FAILED"
```
kishore03109, line 175: Refer [here](https://repost.aws/questions/QUxsRQ8_2KSJ68j-Dk5M7Xvg/amplify-domain-creation-statuses-what-do-they-mean) for types of domain association result

src/services/identity/LaunchesService.ts
```diff
@@ -159,6 +161,41 @@ export class LaunchesService {
       throw siteIdResult.error
     }
 
+    try {
```
kishore03109, line 164: Wrapped in a try catch since AmplifySDK will throw an (expected) not found error, which is ok

alexanderleegs (approved):

PR #723, opened 2023-04-20
[LS-97] chore(stats): add logout tracking

seaerchin:
## Problem
At present, there is no way of knowing how many users are logged into our cms. This is a useful metric to have to give us some insight into how heavily our cms is used.

Closes [LS-105]

## Solution
This could be achieved in 2 ways - either through querying our db or to track all logins - all logouts. The second way was chosen because it tracks ephemeral metrics as well as allowing us to derive the desired metric. Moreover, querying the database requires some contortion on our end (need to hit the `SequelizeStore` + pass in its deps).

This will allow us to graph the # of active users in dd using a formula

**Notes**
Do note that this implementation **might** overcount the # of currently logged in users because the tracking call is done **when the endpoint is hit**, not when the login is successful. This means that a high # of 401s could potentially cause the metric to be wrongly over-represented. a way to fix this is to call the stat tracking only at point of success (ideally). 

However, if this way (to call at point of success) is chosen, it will be done in a future PR (out of scope for now). Also, we might want to consider differentiating between _successful_ logins and _all_ logins.

### DIFF
--- a/src/middleware/stats.ts
+++ b/src/middleware/stats.ts
@@ -53,6 +53,14 @@ export class StatsMiddleware {
     this.statsService.trackGithubLogins(Versions.V2)
   )
 
+  trackV1Logout = wrapAsRequestHandler(async () =>
+    this.statsService.trackLogout(Versions.V1)
+  )
+
+  trackV2Logout = wrapAsRequestHandler(async () =>
+    this.statsService.trackLogout(Versions.V2)
+  )
+
   trackEmailLogins = wrapAsRequestHandler(async () =>
     this.statsService.trackEmailLogins()
   )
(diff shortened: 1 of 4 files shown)


kishore03109 (approved):

QiluXie:
Super Nit: typo in the title and description of the PR, the Jira key should be IS-105 instead of LS-105. It doesn't affect anything because the branch name has the right jira key which automatically links this PR.

seaerchin:
Closing this PR for two reasons: 

1. we can get this by using inbuilt trace metrics and searching for `delete_/v2/.../logout` 
2. the actual metric generated here from logins - logouts is negative oddly enough. 

i'm probably going to do up a custom endpoint for stats `/v2/stats/active-users` and hit it using an event bridge scheduler every 10 mins. just fyi

PR #726, opened 2023-04-21
chore/remove api token name from sites table

alexanderleegs:
This PR removes the unused `apiTokenName` param from our Sites table. Resolves IS-76.

## Deploy Notes
Before merging, run db:migrate on staging/production
Between running migration on prod and release, ensure that no sites are created using the site creation form - that endpoint is the only one which still uses `apiTokenName`

### DIFF
--- a/src/database/migrations/20230421020459-remove-sites-api-token-name.js
+++ b/src/database/migrations/20230421020459-remove-sites-api-token-name.js
@@ -1,0 +1,21 @@
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    await queryInterface.removeColumn(
+      "sites", // name of Source Model
+      "api_token_name" // name of column we want to remove
+    )
+  },
+
+  async down(queryInterface, Sequelize) {
+    await queryInterface.addColumn(
+      "sites", // name of Source model
+      "api_token_name", // name of column we're adding
+      {
+        allowNull: true,
+        defaultValue: "",
+        type: Sequelize.TEXT,
+      }
+    )
+  },
+}
--- a/src/database/models/Site.ts
+++ b/src/database/models/Site.ts
@@ -39,12 +39,6 @@ export class Site extends Model {
 
   @Column({
     allowNull: false,
-    type: DataType.TEXT,
-  })
-  apiTokenName!: string
-
-  @Column({
-    allowNull: false,
     type: DataType.ENUM(...Object.values(SiteStatus)),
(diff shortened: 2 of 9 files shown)


harishv7:
@alexanderleegs How is this endpoint using it? Have we removed those references? What will happen if a new user uses the SiteCreation form after migration?

> that endpoint is the only one which still uses apiTokenName

alexanderleegs:
> Hey we are missing changes for `SitesService.ts`! (which references `apiTokenName`) Apart from that, code wise looks ok to me, but I am also quite confused by the context of this :( Why did we use to have this, and is the deletion of this a safe operation? Have we decommissioned the form to prevent wrongful submission from a depreciated form?

Good catch, fixed in https://github.com/isomerpages/isomercms-backend/pull/726/commits/d507c82975b7d16435e645e8f2ce21b1f5b41ec3

alexanderleegs:
> @alexanderleegs How is this endpoint using it? Have we removed those references? What will happen if a new user uses the SiteCreation form after migration?
> 
> > that endpoint is the only one which still uses apiTokenName

It's currently just being used as a param added to new `Sites` created - we'll just need to hold off on using the site creation form between running migrations on prod and merging the change, otherwise the site creation form will fail. I think this is fine as it's only being used by ops, so we can just let them know not to use it for a couple of hours

QiluXie:
@alexanderleegs Code looks good to me and it's good that we have the revert logic to undo the migration if anything is wrong with the release. Two questions here:

1. Why did we have this `api_token_name` param before? Is it because we used to use individual github token to call github API? 
2. Would it help to avoid the down time if we do the release first and then do the db migration, since `api_token_name` will default to empty string if not submitted by the form?

alexanderleegs:
> @alexanderleegs Code looks good to me and it's good that we have the revert logic to undo the migration if anything is wrong with the release. Two questions here:
> 
> 1. Why did we have this `api_token_name` param before? Is it because we used to use individual github token to call github API?
> 2. Would it help to avoid the down time if we do the release first and then do the db migration, since `api_token_name` will default to empty string if not submitted by the form?

In the past we were considering each site having their own api token to use! We discarded this idea relatively early on because github accounts can't be created automatically, and it doesn't make sense to assign one of our machine users to each site because different sites have different loads, so it's harder to split evenly

I think at the moment not having `api_token_name` will result in an error, so I think we should just inform ops and go ahead with it

harishv7 (approved):

kishore03109 (approved):

PR #727, opened 2023-04-24
Feat(dynamo db): Crud operations for Dynamo DB 

kishore03109:
## Problem

Currently, we are migrating from serverless to pulumi and in the process, removing any usage of queues and replacing them with the usage of dynamo db. This is the first PR of a 3 part PR that that focuses on CRUD operations from the dynamoDB. The creation of the DB is done in the Infra PR [here](https://github.com/isomerpages/isomer-infra/pull/1). 

Closes IS-88

## Solution

Ability to perform CRUD operations on the DB. 

## Reviewer notes 
This is not the full implementation, I am keeping the integration work separate from this implementation. This PR is meant to only provide CRUD operations. To aid in the initial testing, I have mocked the values that are to be passed into the `LaunchesService` class. This is the reason why you will notice that the arguments to the methods are **not** used at the moment. 

**Breaking Changes**

  - Details ...

## Tests

1. Update the env var for `SITE_LAUNCH_DYNAMO_DB_TABLE_NAME` from 1pw 
2. Write this script in `server.js`
```
import DynamoDBService from "@root/services/infra/DynamoDBService"
const dynamoDbService = new DynamoDBService()
const dynamoDBTests = async () => {
  // put item in dynamoDb
  const createResult = await dynamoDbService.createItem()
  // get item from dynamoDB
  const getResult = await dynamoDbService.getItem()
  console.log("result: ", getResult)
  // update item in dynamoDB

  const updateResult = await dynamoDbService.updateItem()
  console.log("result: ", updateResult.$metadata.httpStatusCode)
  // delete item from dynamoDB
  const deleteResult = await dynamoDbService.deleteItem()
  console.log("result: ", deleteResult.$metadata.httpStatusCode)
}
dynamoDBTests()
  .then(() => {
    logger.info("Successfully connected to DynamoDB")
  })
  .catch((err) => {
    logger.error(`Unable to connect to DynamoDB: ${err}`)
  })

```
3. Note the lack of errors
4. Heading over to the aws console to look into the dynamo db there also suggests the clean removal of the entry from the dynamo db

## Deploy Notes

nil, this code has no impact to production

- `SITE_LAUNCH_DYNAMO_DB_TABLE_NAME` : copy from 1pw for dev, staging and prod 

**New dependencies**:

- `DynamoDBService` : dynamodb sdk

### DIFF
--- a/.env-example
+++ b/.env-example
@@ -65,3 +65,6 @@ export GITGUARDIAN_API_KEY=""
 export INCOMING_QUEUE_URL="incoming"
 export OUTGOING_QUEUE_URL="outgoing"
 export SITE_LAUNCH_QUEUE_URL="site_launch"
+
+# Dynamo DB 
+export SITE_LAUNCH_DYNAMO_DB_TABLE_NAME="site_launch"
--- a/src/services/infra/DynamoDBClient.ts
+++ b/src/services/infra/DynamoDBClient.ts
@@ -55,7 +55,7 @@ export default class DynamoDBDocClient {
     type: "create" | "get" | "delete" | "update"
   ): Promise<T> => {
     try {
-      return promise
+      return await promise
     } catch (e) {
       logger.error(
         `Error in operation ${type} item for the dynamoDB table: ${e}`
--- a/src/services/infra/DynamoDBClient.ts
+++ b/src/services/infra/DynamoDBClient.ts
@@ -7,7 +7,6 @@ import {
   UpdateCommand,
   DeleteCommand,
   UpdateCommandInput,
-  PutCommandOutput,
 } from "@aws-sdk/lib-dynamodb"
 import autoBind from "auto-bind"
 
--- a/src/services/infra/DynamoDBClient.ts
+++ b/src/services/infra/DynamoDBClient.ts
@@ -1,4 +1,3 @@
-/* eslint-disable consistent-return */ // todo remove this
 import { DynamoDBClient } from "@aws-sdk/client-dynamodb"
 import {
   DynamoDBDocumentClient,
@@ -7,6 +6,9 @@ import {
   UpdateCommand,
   DeleteCommand,
   UpdateCommandInput,
+  PutCommandOutput,
+  UpdateCommandOutput,
+  DeleteCommandOutput,
 } from "@aws-sdk/lib-dynamodb"
 import autoBind from "auto-bind"
 
@@ -46,7 +48,10 @@ export default class DynamoDBDocClient {
     autoBind(this)
   }
 
-  createItem = async (tableName: string, item: MessageBody) => {
+  createItem = async (
+    tableName: string,
+    item: MessageBody
+  ): Promise<PutCommandOutput> => {
     const params = {
       TableName: tableName,
       Item: {
@@ -55,13 +60,18 @@ export default class DynamoDBDocClient {
       },
     }
     try {
-      return await this.dynamoDBDocClient.send(new PutCommand(params))
+      const response = await this.dynamoDBDocClient.send(new PutCommand(params))
+      return response
     } catch (error) {
       logger.error(`Error in createItem for the table ${tableName}: ${error}`)
+      throw error
     }
   }
 
-  getItem = async (tableName: string, key: string) => {
+  getItem = async (
+    tableName: string,
+    key: string
+  ): Promise<Record<string, unknown> | undefined> => {
     const params = {
       TableName: tableName,
       Key: { appId: key },
@@ -70,8 +80,9 @@ export default class DynamoDBDocClient {
     try {
       const result = await this.dynamoDBDocClient.send(new GetCommand(params))
       return result.Item
-    } catch (err) {
-      logger.error(`Error in getItem for the table ${tableName}: ${err}`)
+    } catch (error) {
+      logger.error(`Error in getItem for the table ${tableName}: ${error}`)
+      throw error
     }
   }
 
@@ -81,7 +92,7 @@ export default class DynamoDBDocClient {
     UpdateExpression,
     ExpressionAttributeNames,
     ExpressionAttributeValues,
-  }: UpdateParams) => {
+  }: UpdateParams): Promise<UpdateCommandOutput> => {
     try {
       const params: UpdateCommandInput = {
         TableName,
@@ -96,10 +107,14 @@ export default class DynamoDBDocClient {
       return result
     } catch (error) {
       logger.error(`Error in updateItem for the table ${TableName}: ${error}`)
+      throw error
     }
   }
 
-  deleteItem = async (tableName: string, key: Record<string, string>) => {
+  deleteItem = async (
+    tableName: string,
+    key: Record<string, string>
+  ): Promise<DeleteCommandOutput> => {
     const params = {
       TableName: tableName,
       Key: key,
@@ -108,6 +123,7 @@ export default class DynamoDBDocClient {
       return await this.dynamoDBDocClient.send(new DeleteCommand(params))
     } catch (error) {
       logger.error(`Error in deleteItem for the table ${tableName}: ${error}`)
+      throw error
     }
   }
 }
--- a/src/services/infra/DynamoDBClient.ts
+++ b/src/services/infra/DynamoDBClient.ts
@@ -1,0 +1,114 @@
+/* eslint-disable consistent-return */ // todo remove this
+import { DynamoDBClient } from "@aws-sdk/client-dynamodb"
+import {
+  DynamoDBDocumentClient,
+  PutCommand,
+  GetCommand,
+  UpdateCommand,
+  DeleteCommand,
+  UpdateCommandInput,
+  PutCommandOutput,
+} from "@aws-sdk/lib-dynamodb"
+import autoBind from "auto-bind"
+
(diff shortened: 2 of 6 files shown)


QiluXie:
@kishore03109 Nit: I think your forget to add the link to the infra PR in the description on this line `The creation of the DB is done in the Infra PR here.`

src/services/infra/DynamoDBClient.ts
```diff
@@ -0,0 +1,113 @@
...
+      convertClassInstanceToMap: false,
+    }
+
+    const unmarshallOptions = {
+      // Whether to return numbers as a string instead of converting them to native JavaScript numbers.
+      wrapNumbers: false,
+    }
+
+    this.dynamoDBDocClient = DynamoDBDocumentClient.from(
+      new DynamoDBClient({ region: "ap-southeast-1" }),
```
QiluXie, line 42: It's better to use env variable for `region` than hardcoding it in the code here.

QiluXie:
@kishore03109 found the jira ticket number is not correct, updated it to IS-88 so this PR is linked to the correct Jira ticket.

src/services/infra/DynamoDBClient.ts
```diff
@@ -0,0 +1,113 @@
+/* eslint-disable consistent-return */ // todo remove this
```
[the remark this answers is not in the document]
kishore03109: af6cd203

src/services/infra/DynamoDBClient.ts
kishore03109: 4d2bcc09

src/services/infra/DynamoDBClient.ts
```diff
@@ -0,0 +1,130 @@
...
+      },
+    }
+    try {
+      const response = await this.dynamoDBDocClient.send(new PutCommand(params))
+      return response
+    } catch (error) {
+      logger.error(`Error in createItem for the table ${tableName}: ${error}`)
+      throw error
+    }
+  }
```
seaerchin, line 70: feels like the `try/catch` is placed at the wrong site. because we want to catch errors at their source (`client.send` here), maybe we could consider doing something similar to the following

 ```ts
  private withLogger = async <T>(
    p: Promise<T>,
    tableName: string
    type: "create" | "get" | "delete" | "update"
  ): Promise<T> => {
    try {
      return p
    } catch (e) {
      logger.error(`Error in ${type}Item for the table ${tableName}: ${e}`)
      throw e
    }
  }
```

seems to work fine and we can simplify our method to 
```ts
  createItem = async (
    tableName: string,
    item: MessageBody
  ): Promise<PutCommandOutput> => {
    const params = {...}

    return this.withLogger(
      this.dynamoDBDocClient.send(new PutCommand(params)),
      tableName
    )
  }
```

src/services/infra/DynamoDBService.ts
```diff
@@ -0,0 +1,72 @@
...
+} from "@root/../microservices/site-launch/shared/types"
+
+import DynamoDBClient, { UpdateParams } from "./DynamoDBClient"
+
+export default class DynamoDBService {
+  private readonly dynamoDBClient: DynamoDBClient
+
+  private readonly TABLE_NAME: string
+
+  // todo: delete these mock after integration
```
seaerchin, line 17: nit: prefer to have these tracked in an issue

src/services/infra/DynamoDBService.ts
```diff
@@ -0,0 +1,72 @@
...
+    // todo: delete mocking after integration
+    this.mockLaunch.status = SiteLaunchLambdaStatus.SUCCESS
+    const updateParams: UpdateParams = {
+      TableName: this.TABLE_NAME,
+      Key: { appId: this.mockLaunch.appId },
+      UpdateExpression: "set #s = :s",
+      ExpressionAttributeNames: { "#s": "status" },
+      ExpressionAttributeValues: {
+        ":s": this.mockLaunch.status,
+      },
```
seaerchin, line 62: could we have a short comment for what these do? this is cos the `<#|:>s` isn't very understandable

src/services/infra/DynamoDBService.ts
```diff
@@ -0,0 +1,72 @@
...
+    this.dynamoDBClient = new DynamoDBClient()
+    this.TABLE_NAME = config.get("aws.dynamodb.siteLaunchTableName")
+    autoBind(this)
+  }
+
+  async createItem(message: MessageBody): Promise<void> {
+    await this.dynamoDBClient.createItem(this.TABLE_NAME, this.mockLaunch)
+  }
+
+  async getItem(message: MessageBody) {
```
seaerchin, line 48: nit: could we add the return types for this and the other methods in the file?

src/services/infra/DynamoDBService.ts
```diff
@@ -0,0 +1,72 @@
...
+
+import DynamoDBClient, { UpdateParams } from "./DynamoDBClient"
+
+export default class DynamoDBService {
+  private readonly dynamoDBClient: DynamoDBClient
+
+  private readonly TABLE_NAME: string
+
+  // todo: delete these mock after integration
+  private mockLaunch: MessageBody = {
```
seaerchin, line 18: nit: prefer shifting this outside and not exporting + rename to `SCREAMING_SNAKE_CASE`

seaerchin (commented):
minor comments

seaerchin (commented):
minor comments

seaerchin (commented):
minor comments

.env-example
```diff
@@ -64,4 +64,7 @@ export GITGUARDIAN_API_KEY=""
 # SQS Queue
 export INCOMING_QUEUE_URL="incoming"
 export OUTGOING_QUEUE_URL="outgoing"
-export SITE_LAUNCH_QUEUE_URL="site_launch"
\ No newline at end of file
+export SITE_LAUNCH_QUEUE_URL="site_launch"
+
+# Dynamo DB 
+export SITE_LAUNCH_DYNAMO_DB_TABLE_NAME="site_launch"
```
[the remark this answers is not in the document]
kishore03109: hmm the full url includes our AWS account number! while it is not in theory sensitive information, it is best practice to not put them publicly!

src/services/infra/DynamoDBClient.ts
kishore03109: Thanks for the clear suggestion! implemented [here](e76d7665)

.env-example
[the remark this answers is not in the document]
seaerchin: just so i understand this correctly, the table name itself contains our account number?

src/services/infra/DynamoDBService.ts
kishore03109: added issue here: d9fffcfc

src/services/infra/DynamoDBService.ts
harishv7: super small nit: can we put as `TODO:`
Vscode tools like https://marketplace.visualstudio.com/items?itemName=wayou.vscode-todo-highlight help highlight these annotations like TODO and FIXME

src/services/infra/DynamoDBService.ts
```diff
@@ -0,0 +1,72 @@
...
+      ExpressionAttributeValues: {
+        ":s": this.mockLaunch.status,
+      },
+    }
+    return this.dynamoDBClient.updateItem(updateParams)
+  }
+
+  async deleteItem(message: MessageBody) {
+    return this.dynamoDBClient.deleteItem(this.TABLE_NAME, {
+      appId: this.mockLaunch.appId || "",
```
harishv7, line 75: why does it need the || ""? if app id doesn't exist should throw error

src/services/infra/DynamoDBService.ts
kishore03109: ok, added comments for clarity 763b411d

src/services/infra/DynamoDBService.ts
kishore03109: oh hey, this should be clearer! `message` is guaranteed to have `appId` is defined, so this is redundant, have changed it here 77197599

src/services/infra/DynamoDBService.ts
kishore03109: @harishv7 823d1059

.env-example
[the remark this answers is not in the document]
seaerchin: ok, if it changes by environment then please keep it as an env var! (but why does it vary by environment???)

src/services/infra/DynamoDBService.ts
kishore03109: @seaerchin done! 9c38a9

.env-example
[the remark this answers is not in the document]
kishore03109: we dont want developers in local dev to poll from the db and mutate values of actual site launches!

src/services/infra/DynamoDBClient.ts
kishore03109: oh right opps good catch, done here : 761fc5a9b

kishore03109: 761fc5a9b26a5b15fd77f5608ae4a4949221f081

seaerchin (approved):

.env-example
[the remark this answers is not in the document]
kishore03109: yup!

PR #738, opened 2023-04-28
Fix: add svg to allowed image types

alexanderleegs:
This PR fixes an issue where svg images were not allowed to be uploaded.

### DIFF
--- a/src/utils/file-upload-utils.js
+++ b/src/utils/file-upload-utils.js
@@ -18,6 +18,7 @@ const ALLOWED_FILE_EXTENSIONS = [
   "tif",
   "bmp",
   "ico",
+  "svg",
 ]
 const defaultCloudmersiveClient = CloudmersiveVirusApiClient.ApiClient.instance
 


harishv7 (dismissed):

harishv7:
@alexanderleegs Just to check, is this the first time we are allowing SVGs? If so, have we tried upload and embedding SVGs and if they work fine?

alexanderleegs:
> @alexanderleegs Just to check, is this the first time we are allowing SVGs? If so, have we tried upload and embedding SVGs and if they work fine?

It's really strange - we've allowed svgs for quite a while now, but it seems like editing svgs specifically is buggy - actually let me take a look at the root cause, i think we might not be doing something correctly here

kishore03109:
@alexanderleegs 
> we've allowed svgs for quite a while now, but it seems like editing svgs specifically is buggy - actually let me take a look at the root cause, i think we might not be doing something correctly here

sorry so does this PR fix the issue or is it something like there is a root cause and this is a bandage fix? (if yes, why is this a bandage fix ah?)

kishore03109 (approved):
am ok with this as a bandage fix, but I am confused about the context around the root cause + what would be the permanent fix. 
Assuming that that will be addressed in a separate pr (correct me if I am wrong here), will just gather the context from there ya

alexanderleegs:
> am ok with this as a bandage fix, but I am confused about the context around the root cause + what would be the permanent fix. Assuming that that will be addressed in a separate pr (correct me if I am wrong here), will just gather the context from there ya

I verified what's the reason behind the different behaviour of create and rename - the reason why creating svgs works currently is that we call `validateAndSanitizeFileUpload`, which has a custom check and handling for svgs - the rename instead only checks against the list of valid file extensions (this is fine because the file content itself does not change), so this fix should work fine

PR #739, opened 2023-05-02
feat(site creation): add support for email + github logins

kishore03109:
## Problem

Currently, the new site creation only allows for email based logins. This is disruptive for ops as there are still some agencies that use the github based logins, and it might be confusing for them to have some websites in the email flow, and some in the github flow. 

Closes 154

## Solution

Rather than only allowing for email based logins, ops will now have the choice of either email or github logins. 
This PR effectively brings back the code [here](https://github.com/isomerpages/isomercms-backend/pull/713/files) and [here](https://github.com/isomerpages/isomercms-backend/pull/679), but only executes the relevant code iff ops wants to use the email login flow.

## Other notes 
The member email is disregarded if ops select the github login flow. 

**Breaking Changes**

  - Details ...

## Testing 
Staging is now live with this form, feel free to test it out directly. I have tested out for both the email and github flow, both seems to work fine.

### DIFF
--- a/src/routes/formsgSiteCreation.ts
+++ b/src/routes/formsgSiteCreation.ts
@@ -87,8 +87,10 @@ export class FormsgRouter {
         await this.sendCreateError(requesterEmail, repoName, submissionId, err)
         return res.sendStatus(200)
       }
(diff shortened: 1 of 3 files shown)


alexanderleegs (approved):
lgtm

PR #740, opened 2023-05-02
Feat/modify private image retrieval

alexanderleegs:
This PR modifies our image retrieval method for private repos. 

### Context
The image retrieval operation on private repos is very expensive currently, as we have to retrieve each image via a github call separately - this uses up a lot of our token capacity and is a major blocker for repo privatisation. This is because the image view in the main album overview needs to show a preview of all our images - on public repos, we can retrieve the image to be used as preview by querying raw.githubusercontent.com/, but this same link doesn't work for private repos (it returns a 404 instead as read access to the repo is restricted). Previously, we were making a query via the github API to retrieve the image preview instead, which is very expensive, especially for sites which have a large number of images.

### Solution
The solution to this requires us to retrieve image information with relatively low token cost as far as possible. It turns out that images on private repos can still be retrieved from `<token>@raw.githubusercontent.com/` - this has been verified to not use up our token capacity as well. However, this url cannot be linked directly on our frontend, as this would expose our access tokens - instead, we pass a dataUrl to obfuscate the image origin source.

Tests are updated in #741

### DIFF
--- a/src/services/directoryServices/MediaDirectoryService.js
+++ b/src/services/directoryServices/MediaDirectoryService.js
@@ -47,15 +47,21 @@ class MediaDirectoryService {
     )
 
     const resp = await Promise.all(
-      files.map((curr) =>
-        getMediaFileInfo({
+      files.map((curr) => {
+        if (curr.type === "dir") {
+          return {
+            name: curr.name,
+            type: "dir",
+          }
+        }
+        return getMediaFileInfo({
           file: curr,
           siteName,
           directoryName,
           mediaType,
           isPrivate,
         })
-      )
+      })
     )
 
     return resp
--- a/src/services/directoryServices/MediaDirectoryService.js
+++ b/src/services/directoryServices/MediaDirectoryService.js
@@ -48,7 +48,13 @@ class MediaDirectoryService {
 
     const resp = await Promise.all(
       files.map((curr) =>
-        getMediaFileInfo(curr, siteName, directoryName, mediaType, isPrivate)
+        getMediaFileInfo({
+          file: curr,
+          siteName,
+          directoryName,
+          mediaType,
+          isPrivate,
+        })
       )
     )
 
--- a/src/services/directoryServices/MediaDirectoryService.js
+++ b/src/services/directoryServices/MediaDirectoryService.js
@@ -1,12 +1,10 @@
-const { config } = require("@config/config")
-
 const { BadRequestError } = require("@errors/BadRequestError")
 
-const GITHUB_ORG_NAME = config.get("github.orgName")
+const { isMediaPathValid } = require("@validators/validators")
+
+const { getMediaFileInfo } = require("@root/utils/media-utils")
 
 const PLACEHOLDER_FILE_NAME = ".keep"
-
-const { isMediaPathValid } = require("@validators/validators")
 
 class MediaDirectoryService {
   constructor({ baseDirectoryService, gitHubService }) {
@@ -40,39 +38,20 @@ class MediaDirectoryService {
     const { private: isPrivate } = await this.gitHubService.getRepoInfo(
       sessionData
     )
-    const files = await this.listWithDefault(sessionData, { directoryName })
+    const files = (
+      await this.listWithDefault(sessionData, { directoryName })
+    ).filter(
+      (file) =>
+        (file.type === "file" || file.type === "dir") &&
+        file.name !== PLACEHOLDER_FILE_NAME
+    )
 
-    const resp = []
-    for (const curr of files) {
-      if (curr.type === "dir") {
-        resp.push({
-          name: curr.name,
-          type: "dir",
-        })
-      }
-      if (curr.type !== "file" || curr.name === PLACEHOLDER_FILE_NAME) continue
-      const fileData = {
-        mediaUrl: `https://raw.githubusercontent.com/${GITHUB_ORG_NAME}/${siteName}/staging/${curr.path
-          .split("/")
-          .map((v) => encodeURIComponent(v))
-          .join("/")}${curr.path.endsWith(".svg") ? "?sanitize=true" : ""}`,
-        name: curr.name,
-        sha: curr.sha,
-        mediaPath: `${directoryName}/${curr.name}`,
-        type: curr.type,
-      }
-      if (mediaType === "images" && isPrivate) {
-        // Generate blob url
-        const imageExt = curr.name.slice(curr.name.lastIndexOf(".") + 1)
-        const contentType = `image/${imageExt === "svg" ? "svg+xml" : imageExt}`
-        const { content } = await this.gitHubService.readMedia(sessionData, {
-          fileSha: curr.sha,
-        })
-        const blobURL = `data:${contentType};base64,${content}`
-        fileData.mediaUrl = blobURL
-      }
-      resp.push(fileData)
-    }
+    const resp = await Promise.all(
+      files.map((curr) =>
+        getMediaFileInfo(curr, siteName, directoryName, mediaType, isPrivate)
+      )
+    )
+
     return resp
   }
 
--- a/src/types/media.ts
+++ b/src/types/media.ts
@@ -1,4 +1,5 @@
 export type ItemType = "dir" | "file"
+export type MediaType = "images" | "files"
 
 export interface MediaFile {
   name: string
@@ -11,8 +12,8 @@ export interface MediaFileInput {
   file: MediaFile
   siteName: string
   directoryName: string
-  mediaType: string
-  isPrivate: boolean
+  mediaType: MediaType
+  isPrivate?: boolean
 }
 
 export interface MediaFileOutput {
--- a/src/types/media.ts
+++ b/src/types/media.ts
@@ -1,0 +1,24 @@
+export type ItemType = "dir" | "file"
+
+export interface MediaFile {
+  name: string
+  type: ItemType
+  sha: string
+  path: string
+}
+
+export interface MediaFileInput {
+  file: MediaFile
+  siteName: string
+  directoryName: string
+  mediaType: string
+  isPrivate: boolean
+}
+
+export interface MediaFileOutput {
+  name: string
+  sha: string
+  mediaUrl: string
+  mediaPath: string
+  type: ItemType
+}
--- a/src/utils/media-utils.ts
+++ b/src/utils/media-utils.ts
@@ -8,7 +8,7 @@ const GITHUB_ORG_NAME = config.get("github.orgName")
 
 type ItemType = "dir" | "file"
 
-export interface MediaFile {
+interface MediaFile {
   name: string
   type: ItemType
   sha: string
--- a/src/utils/media-utils.ts
+++ b/src/utils/media-utils.ts
@@ -1,0 +1,53 @@
+import axios from "axios"
+
+import { config } from "@config/config"
+
+import { getAccessToken } from "@root/utils/token-retrieval-utils"
+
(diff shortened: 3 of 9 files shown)


src/utils/media-utils.ts
```diff
@@ -0,0 +1,80 @@
...
+
+import { config } from "@config/config"
+
+import { getAccessToken } from "@root/utils/token-retrieval-utils"
+
+const GITHUB_ORG_NAME = config.get("github.orgName")
+
+type ItemType = "dir" | "file"
+
+export interface MediaFile {
```
seaerchin, line 11: i think we should separate our utils from our types - it's not intuitive for devs to find the type defs for inside a util file and leads to duplication or worse, inaccurate type defs. 

we should shift this out into the `types` folder if they are going to be exported or remove the export otherwise

src/utils/media-utils.ts
```diff
@@ -0,0 +1,80 @@
...
+  directoryName,
+  mediaType,
+  isPrivate,
+}: MediaFileInput): Promise<MediaFileOutput | MediaDirOutput> => {
+  if (file.type === "dir") {
+    return {
+      name: file.name,
+      type: "dir",
+    }
+  }
```
seaerchin, line 55: this method feels too bloated alr - we are trying to smush file/dir retrieval into one method, which doesn't lead itself well to future extension.

as the caller already knows the file type here, i don't see why we can't just separate this into `getMediaFileInfo` and `getMediaDirectoryInfo`. this also makes it more apparent that the media directory info relies only on the type and the name.

src/utils/media-utils.ts
```diff
@@ -0,0 +1,80 @@
...
+    const accessToken = await getAccessToken()
+    // Accessing images in this way avoids token usage
+    const endpoint = `https://${accessToken}@raw.githubusercontent.com/${GITHUB_ORG_NAME}/${siteName}/staging/${file.path
+      .split("/")
+      .map((v) => encodeURIComponent(v))
+      .join("/")}${file.path.endsWith(".svg") ? "?sanitize=true" : ""}`
+    const resp = await axios.get(endpoint, { responseType: "arraybuffer" })
+    const data = Buffer.from(resp.data, "binary").toString("base64")
+    const dataUrl = `data:${resp.headers["content-type"]};base64,${data}`
+    fileData.mediaUrl = dataUrl
```
seaerchin, line 77: actually we want to be deliberate about this because data URIs are very different from resource URIs. 

the data uri itself implies that the data is already fetched and the caller doesn't need to do anything else. conversely, `mediaUrl` implies that the caller needs to do additional work to render the image inline. 

yes, if this is just given as href then the treatment is the same but could we recognise this distinction in the types? we could do this by giving a `dataUrl` when we fetch!

src/utils/media-utils.ts
alexanderleegs: Yeah that makes sense, removed in https://github.com/isomerpages/isomercms-backend/pull/740/commits/6fb5f790b5ed342ad0411fb89e4bfaf590ff4eb1

src/utils/media-utils.ts
alexanderleegs: Yeah that makes sense, I've removed the directory info portion and moved that directly into the retrieval method, since the check alone basically already gives the return output - https://github.com/isomerpages/isomercms-backend/pull/740/commits/2d69f03b73cb67e275eeceaf394674ea9031def9

src/utils/media-utils.ts
alexanderleegs: As discussed offline - as this is relatively time consuming to change on the frontend, and our current usage is solely for being placed in the `href`, I've filed an issue for this [here](https://isomeropengov.atlassian.net/browse/IS-196?atlOrigin=eyJpIjoiZGViZmU0ZGFiMWYzNGRlYzk0Njk4MGMzODkzNjM2NzQiLCJwIjoiaiJ9) to be tackled in a future PR

src/services/directoryServices/MediaDirectoryService.js
```diff
@@ -40,39 +38,26 @@ class MediaDirectoryService {
...
-          fileSha: curr.sha,
+    const files = (
+      await this.listWithDefault(sessionData, { directoryName })
+    ).filter(
+      (file) =>
+        (file.type === "file" || file.type === "dir") &&
+        file.name !== PLACEHOLDER_FILE_NAME
+    )
+
+    const resp = await Promise.all(
```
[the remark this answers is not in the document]
alexanderleegs: That makes a lot of sense, but I think this might be better to handle within `getMediaFileInfo` instead? I've inserted a try/catch block in https://github.com/isomerpages/isomercms-backend/pull/740/commits/bb93aa2e83e8c35c193924db891c772eb5308da3 instead, as we should still need some file details which are extracted by the method, and we just need to convert the `mediaUrl` to a generic value to be able to handle individual file failures

As discussed offline, the frontend currently has a generic image when it is unable to retrieve an image - the user can still attempt to fetch the actual file data again by clicking on the image, which loads up a modal where the image details are refetched.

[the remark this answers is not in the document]
harishv7: Cool sounds good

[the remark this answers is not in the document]
harishv7: Are we still keeping Promise.all though? So we want the request to fail if any of the loading fails?

src/utils/media-utils.ts
```diff
@@ -0,0 +1,51 @@
...
+
+import { config } from "@config/config"
+
+import logger from "@root/logger/logger"
+import { MediaFileInput, MediaFileOutput } from "@root/types"
+import { getAccessToken } from "@root/utils/token-retrieval-utils"
+
+const GITHUB_ORG_NAME = config.get("github.orgName")
+
+export const isMediaFileOutput = (t: MediaFileOutput): t is MediaFileOutput =>
```
seaerchin, line 11: suggest changing the input type to `unknown`. because this function's precondition and postcondition is identical, we are asserting something **we already know** to use it. also realised this has 0 refs, so we can also delete it ._.

src/utils/media-utils.ts
```diff
@@ -0,0 +1,51 @@
...
+        .map((v) => encodeURIComponent(v))
+        .join("/")}${file.path.endsWith(".svg") ? "?sanitize=true" : ""}`
+      const resp = await axios.get(endpoint, { responseType: "arraybuffer" })
+      const data = Buffer.from(resp.data, "binary").toString("base64")
+      const dataUrl = `data:${resp.headers["content-type"]};base64,${data}`
+      fileData.mediaUrl = dataUrl
+    } catch (err) {
+      // If an error occurs while retrieving the image, we log but continue to return generic values
+      logger.error(err)
+      fileData.mediaUrl = ""
```
seaerchin, line 47: this means the frontend will have an invalid image right? but we're ok with this because the rest of the data would still be correct

src/utils/media-utils.ts
```diff
@@ -0,0 +1,51 @@
...
+  (t as MediaFileOutput).sha !== undefined
+
+export const getMediaFileInfo = async ({
+  file,
+  siteName,
+  directoryName,
+  mediaType,
+  isPrivate,
+}: MediaFileInput): Promise<MediaFileOutput> => {
+  const fileData = {
```
seaerchin, line 21: there's some repeated code here, which makes it abit confusing to follow. wdyt about: 

1. shifting the encoding of `file.path` out into a util function (maybe `encodeFilePathAsUriComponent`)
2. either making another method that computes the github URL or making a util method that decides if sanitization is needed
3. shifting the shared properties out (only `mediaUrl` is diff b/w public/private) then using object spread rather than setting `.mediaUrl`

src/utils/media-utils.ts
```diff
@@ -0,0 +1,51 @@
...
+import logger from "@root/logger/logger"
+import { MediaFileInput, MediaFileOutput } from "@root/types"
+import { getAccessToken } from "@root/utils/token-retrieval-utils"
+
+const GITHUB_ORG_NAME = config.get("github.orgName")
+
+export const isMediaFileOutput = (t: MediaFileOutput): t is MediaFileOutput =>
+  (t as MediaFileOutput).sha !== undefined
+
+export const getMediaFileInfo = async ({
```
seaerchin, line 14: nit: could we change the `isPrivate` to be either an optional parameter or make it optional on the type?

src/utils/media-utils.ts
```diff
@@ -0,0 +1,51 @@
...
+      // Generate data url
+      const accessToken = await getAccessToken()
+      // Accessing images in this way avoids token usage
+      const endpoint = `https://${accessToken}@raw.githubusercontent.com/${GITHUB_ORG_NAME}/${siteName}/staging/${file.path
+        .split("/")
+        .map((v) => encodeURIComponent(v))
+        .join("/")}${file.path.endsWith(".svg") ? "?sanitize=true" : ""}`
+      const resp = await axios.get(endpoint, { responseType: "arraybuffer" })
+      const data = Buffer.from(resp.data, "binary").toString("base64")
+      const dataUrl = `data:${resp.headers["content-type"]};base64,${data}`
```
kishore03109, line 42: Nit: shall we rename this to `dataUri` for clarity?

kishore03109:
Hey I am confused about some the aspects of this pr. Correct me if I am wrong, but does this PR need to have a corresponding FE PR to display the data URI if exists, else display the URL? I understand that there is a corresponding PR for updating our unit tests, but are we confident that this wouldnt break pre-existing private repos?

src/utils/media-utils.ts
harishv7: yea we wanted to ensure like a few broken images can be reloaded from FE by user but the overall API call will try to return as many imageURLs as it can successfully fetch

src/utils/media-utils.ts
alexanderleegs: This was originally used for the tests when the output type was variable - since we've moved out the directory functionality of this method, this is no longer necessary! removed in https://github.com/isomerpages/isomercms-backend/pull/740/commits/7e317a77d5238c6f9f1ce57bcf4055f4cd531275

src/types/media.ts
```diff
@@ -0,0 +1,24 @@
...
+  type: ItemType
+  sha: string
+  path: string
+}
+
+export interface MediaFileInput {
+  file: MediaFile
+  siteName: string
+  directoryName: string
+  mediaType: string
```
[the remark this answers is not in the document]
alexanderleegs: Makes sense, changed in https://github.com/isomerpages/isomercms-backend/pull/740/commits/65a108644ca42d6daa3676ed3c597be2ba13e0d3

src/utils/media-utils.ts
alexanderleegs: Yeah refactoring makes sense, I've moved the main url generation into a separate method in https://github.com/isomerpages/isomercms-backend/pull/740/commits/7e317a77d5238c6f9f1ce57bcf4055f4cd531275

src/utils/media-utils.ts
alexanderleegs: Made it optional in https://github.com/isomerpages/isomercms-backend/pull/740/commits/65a108644ca42d6daa3676ed3c597be2ba13e0d3

src/utils/media-utils.ts
alexanderleegs: Renamed in https://github.com/isomerpages/isomercms-backend/pull/740/commits/7e317a77d5238c6f9f1ce57bcf4055f4cd531275

seaerchin (approved):

PR #741, opened 2023-05-02
Test/image retrieval tests

alexanderleegs:
This PR introduces tests for PR #740. It removes existing tests for private repos in `MediaDirectoryService` and `MediaFileService` and moves them to the newly created `media-utils` test instead, to be able to better mock and handle axios calls.

### DIFF
--- a/src/fixtures/media.ts
+++ b/src/fixtures/media.ts
@@ -1,8 +1,9 @@
-import { MediaFile } from "@root/types"
+import { MediaType, MediaFile } from "@root/types"
 
 export const MEDIA_FILE_NAME = "test file"
 export const MEDIA_SITE_NAME = "site"
 export const MEDIA_DIRECTORY_NAME = "dir"
+export const MEDIA_SUBDIRECTORY_NAME = "sub dir"
 export const MEDIA_FILE_SHA = "sha"
 
 export const MEDIA_DIR: MediaFile = {
@@ -19,6 +20,11 @@ const BASE_MEDIA_FILE: MediaFile = {
   path: `${MEDIA_DIRECTORY_NAME}/${MEDIA_FILE_NAME}`,
 }
 
+export const NESTED_MEDIA_FILE: MediaFile = {
+  ...BASE_MEDIA_FILE,
+  path: `${MEDIA_DIRECTORY_NAME}/${MEDIA_SUBDIRECTORY_NAME}/${MEDIA_FILE_NAME}`,
+}
+
 export const SVG_FILE = {
   ...BASE_MEDIA_FILE,
   name: `${MEDIA_FILE_NAME}.svg`,
@@ -33,15 +39,21 @@ const BASE_INPUT = {
 export const DIR_INPUT = {
   ...BASE_INPUT,
   file: MEDIA_DIR,
-  mediaType: "images",
+  mediaType: "images" as MediaType,
   isPrivate: false,
 }
 
 export const IMAGE_FILE_PUBLIC_INPUT = {
   ...BASE_INPUT,
   file: BASE_MEDIA_FILE,
-  mediaType: "images",
+  mediaType: "images" as MediaType,
   isPrivate: false,
+}
+
+export const NESTED_IMAGE_FILE_PUBLIC_INPUT = {
+  ...IMAGE_FILE_PUBLIC_INPUT,
+  file: NESTED_MEDIA_FILE,
+  directoryName: `${MEDIA_DIRECTORY_NAME}/${MEDIA_SUBDIRECTORY_NAME}`,
 }
 
 export const SVG_FILE_PUBLIC_INPUT = {
@@ -61,7 +73,7 @@ export const SVG_FILE_PRIVATE_INPUT = {
 
 export const PDF_FILE_PUBLIC_INPUT = {
   ...IMAGE_FILE_PUBLIC_INPUT,
-  mediaType: "files",
+  mediaType: "files" as MediaType,
 }
 
 export const PDF_FILE_PRIVATE_INPUT = {
--- a/src/fixtures/media.ts
+++ b/src/fixtures/media.ts
@@ -1,0 +1,70 @@
+import { MediaFile } from "@root/utils/media-utils"
+
+export const MEDIA_FILE_NAME = "test file"
+export const MEDIA_SITE_NAME = "site"
+export const MEDIA_DIRECTORY_NAME = "dir"
+export const MEDIA_FILE_SHA = "sha"
+
+export const mediaDir: MediaFile = {
+  name: "directory",
+  type: "dir",
+  sha: MEDIA_FILE_SHA,
+  path: `${MEDIA_DIRECTORY_NAME}/directory`,
+}
+
+const baseMediaFile: MediaFile = {
+  name: MEDIA_FILE_NAME,
+  type: "file",
+  sha: MEDIA_FILE_SHA,
+  path: `${MEDIA_DIRECTORY_NAME}/${MEDIA_FILE_NAME}`,
+}
+
+export const svgFile = {
+  ...baseMediaFile,
+  name: `${MEDIA_FILE_NAME}.svg`,
+  path: `${MEDIA_DIRECTORY_NAME}/${MEDIA_FILE_NAME}.svg`,
+}
+
+const baseInput = {
+  siteName: MEDIA_SITE_NAME,
+  directoryName: MEDIA_DIRECTORY_NAME,
+}
+
+export const dirInput = {
+  ...baseInput,
+  file: mediaDir,
+  mediaType: "images",
+  isPrivate: false,
+}
+
+export const imageFilePublicInput = {
+  ...baseInput,
+  file: baseMediaFile,
+  mediaType: "images",
+  isPrivate: false,
+}
+
+export const svgFilePublicInput = {
+  ...imageFilePublicInput,
+  file: svgFile,
+}
+
+export const imageFilePrivateInput = {
+  ...imageFilePublicInput,
+  isPrivate: true,
+}
+
+export const svgFilePrivateInput = {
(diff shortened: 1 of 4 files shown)


src/fixtures/media.ts
```diff
@@ -0,0 +1,70 @@
+import { MediaFile } from "@root/utils/media-utils"
+
+export const MEDIA_FILE_NAME = "test file"
+export const MEDIA_SITE_NAME = "site"
+export const MEDIA_DIRECTORY_NAME = "dir"
+export const MEDIA_FILE_SHA = "sha"
+
+export const mediaDir: MediaFile = {
```
seaerchin, line 8: curious: why not SCREAMING_SNAKE_CASE? see the google style guide [here](https://google.github.io/styleguide/tsguide.html#identifiers-constants)

src/services/directoryServices/__tests__/MediaDirectoryService.spec.js
```diff
@@ -129,46 +129,6 @@ describe("Media Directory Service", () => {
         directoryName: imageDirectoryName,
       })
     })
-    mockGitHubService.getRepoInfo.mockResolvedValueOnce({
```
seaerchin, line 132: could i get some context behind the mass deletion fo this chunk? is this for media-utils test?

src/utils/__tests__/media-utils.spec.ts
```diff
@@ -0,0 +1,149 @@
...
+  getAccessToken: jest.fn().mockResolvedValue("token"),
+}))
+jest.mock("axios", () => ({
+  get: jest.fn().mockResolvedValue({
+    data: "blahblah",
+    headers: {
+      "content-type": "blahblah",
+    },
+  }),
+}))
```
seaerchin, line 34: style: we alr have `mock-axios` btw so this is done automatically and we should prefer using that instead

src/utils/__tests__/media-utils.spec.ts
```diff
@@ -0,0 +1,149 @@
...
+      sha: MEDIA_FILE_SHA,
+      mediaPath: `${MEDIA_DIRECTORY_NAME}/${MEDIA_FILE_NAME}`,
+      type: "file",
+    }
+    expect(await getMediaFileInfo(imageFilePublicInput)).toStrictEqual(
+      expectedResp
+    )
+  })
+
+  it("should return normal information for svgs in public repos", async () => {
```
seaerchin, line 59: what is "normal" information? and when will we get "not normal" information?

seaerchin (commented):
actually i peeked at `getMediaFileInfo` and i think there are some invariants that we want to ensure that isn't immediately apparent in the test

1. if the file type is `dir`, we don't do anything else. 
2. we only call `getAccessToken` for private images (why not private files?)
3. we always sanitize `svg` if 2 is true (ie, if `getAccessToken` is called and it is a `svg`, we should sanitize); separately, do we not need to sanitize anything else?
4. if 2 is true, we will always replace the `mediaUrl` with a data URI

should we be more comprehensive wrt ^

src/fixtures/media.ts
alexanderleegs: Changed in https://github.com/isomerpages/isomercms-backend/pull/741/commits/625680525dcc9f0aa9bea12c5f8fa3a7cd172070!

src/services/directoryServices/__tests__/MediaDirectoryService.spec.js
alexanderleegs: Yup this was no longer easy to test in the directory service itself - the equivalent tests (for private repos) has been moved into media-utils!

alexanderleegs:
> actually i peeked at `getMediaFileInfo` and i think there are some invariants that we want to ensure that isn't immediately apparent in the test
> 
> 1. if the file type is `dir`, we don't do anything else.
> 2. we only call `getAccessToken` for private images (why not private files?)
> 3. we always sanitize `svg` if 2 is true (ie, if `getAccessToken` is called and it is a `svg`, we should sanitize); separately, do we not need to sanitize anything else?
> 4. if 2 is true, we will always replace the `mediaUrl` with a data URI
> 
> should we be more comprehensive wrt ^

This is what I've tried to do in the media utils tests, which have 7 (now 6) main flows, private/public x regular image/svg/files

(The dirs test has been removed because it's now being done in `MediaDirectoryService` directly)
The private image and svg tests look for the mock axios call to be called and the mediaUrl to be modified to use the generated dataUri instead, because the axios call is used only to retrieve the actual image which needs to be encoded - as we can't really determine the final encoded value, we look for the an entry starting with `data:` instead

The svg tests look for sanitization being added to the raw github url - github itself handles the rest. This was a point raised by our VAPT, that svgs specifically were vulnerable

Files have the same behaviour in both public and private repos, and we test that it doesn't use the private image flow - this is because files have no need for a preview image, and we don't need the mediaUrl at all

src/utils/__tests__/media-utils.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+describe("Media utils test", () => {
+  it("should return mediaUrl as raw github information for images in public repos", async () => {
+    const expectedResp = {
+      mediaUrl: `https://raw.githubusercontent.com/${GITHUB_ORG_NAME}/${MEDIA_SITE_NAME}/staging/${MEDIA_DIRECTORY_NAME}/${encodeURIComponent(
+        MEDIA_FILE_NAME
+      )}`,
+      name: MEDIA_FILE_NAME,
+      sha: MEDIA_FILE_SHA,
+      mediaPath: `${MEDIA_DIRECTORY_NAME}/${MEDIA_FILE_NAME}`,
+      type: "file",
```
seaerchin, line 43: should this be `images`? also this seems to pass so...

src/utils/__tests__/media-utils.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+      expectedResp
+    )
+  })
+
+  it("should handle mediaUrl for images in private repos", async () => {
+    const expectedPartialResp = {
+      name: MEDIA_FILE_NAME,
+      sha: MEDIA_FILE_SHA,
+      mediaPath: `${MEDIA_DIRECTORY_NAME}/${MEDIA_FILE_NAME}`,
+      type: "file",
```
seaerchin, line 85: :sus:

src/utils/__tests__/media-utils.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+      { responseType: "arraybuffer" }
+    )
+  })
+
+  it("should handle mediaUrl for svgs and sanitise image in private repos", async () => {
+    const expectedPartialResp = {
+      name: `${MEDIA_FILE_NAME}.svg`,
+      sha: MEDIA_FILE_SHA,
+      mediaPath: `${MEDIA_DIRECTORY_NAME}/${MEDIA_FILE_NAME}.svg`,
+      type: "file",
```
seaerchin, line 108: :sus:

src/utils/__tests__/media-utils.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+      mockGenericAxios.get
+    ).toHaveBeenCalledWith(
+      `https://token@raw.githubusercontent.com/${GITHUB_ORG_NAME}/${MEDIA_SITE_NAME}/staging/${MEDIA_DIRECTORY_NAME}/${encodeURIComponent(
+        MEDIA_FILE_NAME
+      )}`,
+      { responseType: "arraybuffer" }
+    )
+  })
+
+  it("should handle mediaUrl for svgs and sanitise image in private repos", async () => {
```
seaerchin, line 103:
```suggestion
  it("should return the mediaUrl as a data url for svgs and sanitise the svgs for svgs in private repos", async () => {
```

src/utils/__tests__/media-utils.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+      sha: MEDIA_FILE_SHA,
+      mediaPath: `${MEDIA_DIRECTORY_NAME}/${MEDIA_FILE_NAME}`,
+      type: "file",
+    }
+    expect(await getMediaFileInfo(PDF_FILE_PUBLIC_INPUT)).toStrictEqual(
+      expectedResp
+    )
+  })
+
+  it("should handle mediaUrl for images in private repos", async () => {
```
seaerchin, line 80:
```suggestion
  it("should return the mediaUrl as a data url for images in private repos", async () => {
```

seaerchin (commented):
could we have a test case for complex file names (`a/b/c`)? this is because our function internally does a spilt then a join by `/` and it's important to ensure this works as expected. 

side note: maybe can structure by like public/private repos using a nested `describe` but this isn't v impt

src/utils/__tests__/media-utils.spec.ts
alexanderleegs: This is actually the github response! We're only expecting `file` or `dir`

kishore03109 (approved):

src/utils/__tests__/media-utils.spec.ts
seaerchin: for clarification, github only returns either `file | dir` even for `images` right? i'm asking this because the test description said that this was for `... images in public repos`, so this confused me

alexanderleegs:
> 6ed1774

Yeah that makes sense, I've added a new test case in https://github.com/isomerpages/isomercms-backend/pull/741/commits/6ed1774e78bb0987c33d47d56f81333c372d38f1

src/utils/__tests__/media-utils.spec.ts
alexanderleegs: yup, github returns directory item information - any item which is a single item (even .png or .svg files) returns `file`, any subdirectory returns `dir`

PR #752, opened 2023-05-04
Fix: publish button delay

alexanderleegs:
This PR adds an additional guard when publishing changes to check if the review request has been approved. To be reviewed in conjunction with PR [#1263](https://github.com/isomerpages/isomercms-frontend/pull/1263) on the isomer cms frontend repo.

### DIFF
--- a/src/integration/Reviews.spec.ts
+++ b/src/integration/Reviews.spec.ts
@@ -1265,6 +1265,23 @@ describe("Review Requests Integration Tests", () => {
       })
     })
 
+    it("should return 404 if review request is not approved", async () => {
+      // Arrange
+      const app = generateRouterForUserWithSite(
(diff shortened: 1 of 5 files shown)


kishore03109 (commented):
hey, would it be ok if we shifted this check _**before**_ calling `mergeReviewRequest`? 

Idea is that we can directly return this as a 404 rather than throwing it and letting the error bubble up, wdyt? 
Screenshot 2023-05-05 at 9 02 45 AM

alexanderleegs:
> 63efe6e

Sure, changed in https://github.com/isomerpages/isomercms-backend/pull/752/commits/63efe6e28972a2fec08fb4d80b77ed50e679afcb

kishore03109 (approved):

PR #753, opened 2023-05-05
feat(site launch enums): more meaningful naming

kishore03109:
Modify enums to be more meaningful

### DIFF
--- a/microservices/site-launch/shared/types.ts
+++ b/microservices/site-launch/shared/types.ts
@@ -5,9 +5,12 @@ export enum SiteLaunchLambdaType {
 }
 
 export enum SiteLaunchLambdaStatus {
-  SUCCESS = "success",
-  FAILURE = "failure",
-  PENDING = "pending",
+  SUCCESS_SITE_LIVE = "success - site live",
+  SUCCESS_PROPAGATING = "success - propagating",
+  FAILURE_WRONG_CLOUDFRONT_DISTRIBUTION = "failure - wrong cloudfront distribution",
+  FAILURE_CLOUDFRONT_ALIAS_CLASH = "failure - cloudfront alias clash",
+  PENDING_DURING_SITE_LAUNCH = "pending - during site launch",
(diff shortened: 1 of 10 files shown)


microservices/site-launch/shared/types.ts
```diff
@@ -5,12 +5,15 @@ export enum SiteLaunchLambdaType {
...
 export enum SiteLaunchLambdaStatus {
-  SUCCESS = "success",
-  FAILURE = "failure",
-  PENDING = "pending",
+  SUCCESS_SITE_LIVE = "success - site live",
+  SUCCESS_PROPAGATING = "success - propagating",
+  FAILURE_WRONG_CLOUDFRONT_DISTRIBUTION = "failure - wrong cloudfront distribution",
+  FAILURE_CLOUDFRONT_ALIAS_CLASH = "failure - cloudfront alias clash",
+  PENDING_DURING_SITE_LAUNCH = "pending - during site launch",
+  PENDING_PRE_SITE_LAUNCH = "pending - pre site launch",
```
seaerchin, line 13: i suddenly realised we don't tag the state here but have to parse the string manually... 
eg: 
```ts 
export type SiteLaunchStatus = { 
  state: "success" | "failure" | "pending"
  message: keyof typeof SiteLaunchLambdaStatus
}  
```

makes the enum more complicated to work with because the state is now implicitly tied to the portion before `-`, which might not be upheld/noticed

seaerchin (commented):
comment regarding the type itself, rest is ok

microservices/site-launch/shared/types.ts
kishore03109: oo this is a good suggestion, thanks!

seaerchin (approved):
lgtm, i skimmed over the remaining changes as they appeared to be largely changes related to renaming + using the new type.

pls dismiss this review if that's not the case

seaerchin (approved):
lgtm, i skimmed over the remaining changes as they appeared to be largely changes related to renaming + using the new type.

pls dismiss this review if that's not the case

PR #754, opened 2023-05-05
feat: return sites from db for email login

harishv7:
## Problem

We are making unnecessary Github calls on /sites page especially on email login flow when this could be a call to our DB. In light of us reaching the limits of Github api, we should try to prevent excessive calls.

Closes IS-165

## Solution

Added a check on getSites to return the list of site names for the user from our DB (only if it is a non-admin and non-github flow user).

**Breaking Changes**

  - Details ...

## Deploy Notes

Note minor dependency on https://github.com/isomerpages/isomercms-frontend/pull/1265
Needs to be deployed together

### DIFF
--- a/src/routes/v2/authenticated/sites.ts
+++ b/src/routes/v2/authenticated/sites.ts
@@ -128,7 +128,6 @@ export class SitesRouter {
 
     router.get(
       "/",
-      this.statsMiddleware.countGithubSites,
       this.statsMiddleware.countMigratedSites,
       attachReadRouteHandlerWrapper(this.getSites)
     )
--- a/src/services/identity/SitesService.ts
+++ b/src/services/identity/SitesService.ts
@@ -331,14 +331,10 @@ class SitesService {
       })
     )
 
-    // helper type guard
-    // note: empty strings will return false as well
-    const isString = (val: string | undefined): val is string => !!val
-
     // get sites from DB for email login users
     if (!isAdminUser && isEmailUser) {
       const retrievedSitesByEmail = await this.getSitesForEmailUser(userId)
-      const filteredValidSites = retrievedSitesByEmail.filter(isString)
+      const filteredValidSites = retrievedSitesByEmail.filter(_.isString)
 
       const repoData: RepositoryData[] = filteredValidSites.map((site) => ({
         repoName: site,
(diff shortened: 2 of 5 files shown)


src/routes/v2/authenticated/sites.ts
```diff
@@ -128,7 +128,6 @@ export class SitesRouter {
 
     router.get(
       "/",
-      this.statsMiddleware.countGithubSites,
```
harishv7, line 131: Removing the countGithubSites middleware to prevent excessive calls to github

cc @isomerpages/iso-engineers

src/types/repoInfo.ts
```diff
@@ -15,10 +15,10 @@ export type GitHubRepositoryData = {
 }
 
 export type RepositoryData = {
-  lastUpdated: GitHubRepositoryData["pushed_at"]
-  permissions: GitHubRepositoryData["permissions"]
+  lastUpdated?: GitHubRepositoryData["pushed_at"]
+  permissions?: GitHubRepositoryData["permissions"]
   repoName: GitHubRepositoryData["name"]
-  isPrivate: GitHubRepositoryData["private"]
+  isPrivate?: GitHubRepositoryData["private"]
```
alexanderleegs, line 21: hmm actually removing isPrivate has implications specifically for loading homepages on private repos - without this param available the images on editHomepage don't load properly. I feel like this can be a future fix though since the affected repos should be small, but let's check with prod ops first if any existing email login sites are private

src/services/identity/SitesService.ts
```diff
@@ -331,6 +331,21 @@ class SitesService {
       })
     )
 
+    // helper type guard
+    // note: empty strings will return false as well
+    const isString = (val: string | undefined): val is string => !!val
```
kishore03109, line 336: Style: could we just use the lodash `isString` for this?

src/types/repoInfo.ts
harishv7: We should then be tracking this info in our DB right?

kishore03109 (approved):
LGTM for the scope of this pr, but I do share the concens that Alex has, we would need to tell ops about this + plan to rectify this. I am sure that we will have private repos for email login in the future.

src/services/identity/SitesService.ts
harishv7: oh yea! great suggestion haha

PR #755, opened 2023-05-06
Is 166 race condition when editing a page inside a deleted folder

kishore03109:
## Problem

User A creates a folder A

User A creates a page `testPage` inside folder A

User B starts editing page testPage

User A deletes folder A

User B saves page testPage

Observed behaviour: 

Folder A exists, but is no longer accessible in the CMS. (404 error thrown) 

## Solution

Do a read call to ensure that the page still exists before making and mutation to the repo in github. 
Modifying the file just makes the user see the 404 page, and they have to navigate back to their workspace, which, while annoying, preserves the valid state of the repo.

## Limitations 
Since we are working with stale data, one would need to do a read call to make sure that the file exists in the particular folder. This would incur an additional call, but the hunch here is that the user will call this endpoint lesser than a window refresh. 
 
**Breaking Changes**

### DIFF
--- a/src/routes/v2/authenticatedSites/collectionPages.ts
+++ b/src/routes/v2/authenticatedSites/collectionPages.ts
@@ -1,3 +1,4 @@
+/* eslint-disable import/prefer-default-export */
 import autoBind from "auto-bind"
 import express, { NextFunction, Request, Response } from "express"
 
@@ -19,7 +20,7 @@ import UserWithSiteSessionData from "@root/classes/UserWithSiteSessionData"
 import { CollectionPageService } from "@services/fileServices/MdPageServices/CollectionPageService"
 import { SubcollectionPageService } from "@services/fileServices/MdPageServices/SubcollectionPageService"
 
-class CollectionPagesRouter {
+export class CollectionPagesRouter {
   collectionPageService: CollectionPageService
 
   subcollectionPageService: SubcollectionPageService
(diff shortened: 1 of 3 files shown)


src/routes/v2/authenticatedSites/collectionPages.ts
```diff
@@ -130,61 +170,84 @@ class CollectionPagesRouter {
...
-    res.status(200).send("OK")
+    let deleteResp
+    if (subcollectionName) {
+      deleteResp = await this.subcollectionPageService.delete(
+        userWithSiteSessionData,
+        {
+          fileName: pageName,
+          collectionName,
+          subcollectionName,
+          /**
```
kishore03109, line 227: Note: the sha is later [referenced in Github Service](https://github.com/isomerpages/isomercms-backend/blob/fd30cbe2377c2254aeb3ce3bbdd4ebe72919e95e/src/services/db/GitHubService.js#L208), which computes the sha if it does not exist. Thus, this key should be marked optional

src/routes/v2/authenticatedSites/collectionPages.ts
```diff
@@ -54,20 +66,37 @@ class CollectionPagesRouter {
           collectionName,
           content: pageBody,
           frontMatter,
+          shouldIgnoreCheck: false,
```
kishore03109, line 69: @alexanderleegs hey, tagging you here to just double confirm this should be the correct value

kishore03109:
Closing for a better solution

PR #756, opened 2023-05-08
fix(githhub service): get call to github to prevent race condition

kishore03109:
Copying from @alexanderleegs's write up in slack: 
## Problem

issue with releasing the refetch behaviour change - specifically for concurrent editing of pages in similar locations with different users, there is the risk of race conditions. E.g. User A and B both want to modify folder 1; User A edits a page in folder 1, but before he submits the operation, user B deletes the folder - we suspect this will cause a floating page in the deleted folder 1 that is inaccessible via CMS. This issue is already present in production currently, but it's mostly mitigated by our refetch behaviour, because retrieving updated data on refocus will immediately stop users from making these changes to recently updated files/folders.

Closes [insert issue #]

## Solution

Our proposed solution involves adding guards to create and update operations in the base GithubService - other operations (e.g. deleting an already removed file) will trigger an error from Github itself already, so only create and update endpoints have the potential to cause errors. We would have to make an additional query for each create and update endpoint to check that the file/directory still exists before we can make the change, so this increases the number of API calls we're making, but we feel that this is still an overall improvement over refetching on focus since there are a lot more GET queries than CREATE/UPDATE. We can monitor the overall change in token usage after pushing all the changes to see if this will cause any issues

## Tests

To be honest, quite worried about this PR since the surface area of bugs is quite large. Manual testing done so far:
1) CRUD operations for files,folders 
2) unable to replicate bug in the problem stated in this PR 
3) Creating resource rooms folders
4) Creating more deeply nested media folders
5) Moving pages

Please feel free to put any notes about flows that I might have missed out here!

### DIFF
--- a/src/services/db/GitHubService.js
+++ b/src/services/db/GitHubService.js
@@ -106,6 +106,11 @@ class GitHubService {
         !isCreatingTopLevelDirectory && !isCreatingNewResourceFolder
 
       if (checkDirectoryExist) {
+        /**
+         * When we are creating a new resource post or creating a new subDirectory,
+         * we create a _posts and a .keep folder respectively. However, we still need to check if
+         * parent directory still exists.
+         */
         const isCreatingSubDirectory = fileName === ".keep"
         const isCreatingPostResource = directoryName.endsWith("_posts")
         if (!directoryName) {
@@ -113,6 +118,7 @@ class GitHubService {
         }
         let pathToCheck = directoryName
         if (isCreatingSubDirectory || isCreatingPostResource) {
+          // get parent directory
           pathToCheck = directoryName.split("/").slice(0, -1).join("/")
         }
         // this is to check if the file path still exists, else this will throw a 404
--- a/src/services/db/GitHubService.js
+++ b/src/services/db/GitHubService.js
@@ -94,6 +94,23 @@ class GitHubService {
     const { accessToken, siteName, isomerUserId: userId } = sessionData
     try {
       const endpoint = this.getFilePath({ siteName, fileName, directoryName })
+      const isCreatingTopLevelDirectory = fileName === "collection.yml"
+      const isCreatingNewResourceFolder = fileName === "index.html"
+      const checkDirectoryExist =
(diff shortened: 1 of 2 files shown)


src/services/db/GitHubService.js
```diff
@@ -94,6 +94,23 @@ class GitHubService {
     const { accessToken, siteName, isomerUserId: userId } = sessionData
     try {
       const endpoint = this.getFilePath({ siteName, fileName, directoryName })
+      const isCreatingTopLevelDirectory = fileName === "collection.yml"
+      const isCreatingNewResourceFolder = fileName === "index.html"
```
harishv7, line 98: just curious, how do we know that if filename===index.html means a new resource folder is created?

alexanderleegs: It's our invariant for repo structure currently, but that might change in future - I think I'm okay with this for now, but we should leave a comment here about the invariants we currently use, so that we can modify this in future if necessary

harishv7: okay, let's leave comments for future context

seaerchin:
tests r failing @__@

src/services/db/GitHubService.js
```diff
@@ -94,6 +94,23 @@ class GitHubService {
...
+      const checkDirectoryExist =
+        !isCreatingTopLevelDirectory && !isCreatingNewResourceFolder
+
+      if (checkDirectoryExist) {
+        const isCreatingSubDirectory = fileName === ".keep"
+        const isCreatingPostResource = directoryName.endsWith("_posts")
+
+        let pathToCheck = directoryName
+        if (isCreatingSubDirectory || isCreatingPostResource) {
+          pathToCheck = directoryName.split("/").slice(0, -1).join("/")
```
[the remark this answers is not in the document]
kishore03109: done

src/services/db/GitHubService.js
kishore03109: comment added!

harishv7 (approved):

PR #757, opened 2023-05-08
chore(tracer.ts): changed samplerate to 1

seaerchin:
## Problem
`dd-trace` did not fully sample, so we might have missed traces. this is sub-optimal when we want to determine the # of email/github requests hitting our backend api.

## Solution
see [here](https://docs.datadoghq.com/tracing/trace_collection/library_config/nodejs/) - we just don't let the dd agent decide the sample rate.

### DIFF
--- a/src/utils/tracer.ts
+++ b/src/utils/tracer.ts
@@ -1,5 +1,7 @@
 import tracer from "dd-trace"
 
-tracer.init()
+tracer.init({
+  sampleRate: 1,
+})
 
 export default tracer


src/utils/tracer.ts
```diff
@@ -1,5 +1,7 @@
 import tracer from "dd-trace"
 
-tracer.init()
+tracer.init({
+  sampleRate: 1,
```
harishv7, line 4: sample rate = 1 means? every 1ms?

seaerchin: from the docs 
> Controls the ingestion sample rate (between 0.0 and 1.0) between the Agent and the backend.

seaerchin: they provide more extensive documentation [here](https://docs.datadoghq.com/tracing/trace_pipeline/ingestion_mechanisms/?tab=nodejs)

kishore03109 (approved):

harishv7 (approved):

PR #763, opened 2023-05-09
[IS-104] fix(errorhandler): tweak serialisation of error object

seaerchin:
## Problem
refer [here](https://isomeropengov.atlassian.net/browse/IS-104?atlOrigin=eyJpIjoiOTJlMGY5ZDk3MTQ3NDdkN2EzNmFjZjNkMTI1MzUwZjAiLCJwIjoiaiJ9) for overview. brief summary is that auth tokens are being logged and written into our logs. 

Closes IS-104

## Solution
1. for the error object, we check if there exists an authorization header. if there is, both `request` + `config` is removed. this is heavy handed but required because the library (`serialize-errors`) resolves circular references by [walking down teh properties](https://github.com/sindresorhus/serialize-error/blob/main/index.js#L95), which imght have more references to the header. a brief copy-pasting of the attached log + searching for `token ghp` revealed 15 references, of which some were deeply nested, so this was preferred. 
2. the test setup is **not** ideal. the jest mock axios exposes a `mockError` method which mocks a failed (4xx or 5xx) response from the external server but i could not get this to work well with axios, so a object conforming to the axios error interface was created instead. `jest-mock-axios` uses the internal promise queue (see [here](https://github.com/knee-cola/jest-mock-axios/blob/master/lib/mock-axios.ts#L261)) to mock a failed response but the internal promise queue appears to be empty from the error message @__@

## Testing
Prior to testing, we need to change a few things
1. in `axios-utils`, change `validateStatus` so that 404s are recognised as an error. 
2. change `isValidE2E` to `true` in `verifyAccess` in `AuthenticationMiddlewareService` 
3. go to a repo and rename the contact us file to `contactus` (note the lack of dash) 
4. query this endpoint: `/v2/sites/<site-name>/contactUs`
5. log `sanitizedErr` in `errorHandler.js`

`sanitizedErr` should be short and **not contain raw bytes**

### DIFF
--- a/src/middleware/errorHandler.js
+++ b/src/middleware/errorHandler.js
@@ -7,17 +7,15 @@ const logger = require("@logger/logger")
 
 function errorHandler(err, req, res, next) {
   if (!err) return next()
-  const containsSensitiveInfo = _.isEmpty(err.request)
+  const containsSensitiveInfo = !_.isEmpty(err.config?.headers?.Authorization)
   const serialisedErr = serializeError(err)
-  // NOTE: If the error is an axios error or a network error, it will have a response.
-  // We only take the `body` section because `request` + `config`
-  // might contain the auth token used, which is sensitive
+  // NOTE: If the error is an axios error or a network error,
+  // the token used (if any) will be present on `request` + `config`.
+  // Hence, we omit both of these.
   const sanitisedErr = containsSensitiveInfo
     ? _.omit(serialisedErr, ["request", "config"])
     : serialisedErr
-  const errMsg = `${new Date()}: ${JSON.stringify(
-    serializeError(sanitisedErr)
-  )}`
+  const errMsg = `${new Date()}: ${JSON.stringify(sanitisedErr)}`
 
   // set locals, only providing error in development
   res.locals.message = err.message
--- a/src/middleware/errorHandler.js
+++ b/src/middleware/errorHandler.js
@@ -1,4 +1,5 @@
 // Import dependencies
+const _ = require("lodash")
 const { serializeError } = require("serialize-error")
 
 // Import logger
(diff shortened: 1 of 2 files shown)


src/middleware/errorHandler.js
```diff
@@ -1,12 +1,21 @@
...
-  const errMsg = `${new Date()}: ${JSON.stringify(serializeError(err))}`
+  const containsSensitiveInfo = !_.isEmpty(err.config?.headers?.Authorization)
+  const serialisedErr = serializeError(err)
+  // NOTE: If the error is an axios error or a network error,
+  // the token used (if any) will be present on `request` + `config`.
+  // Hence, we omit both of these.
+  const sanitisedErr = containsSensitiveInfo
+    ? _.omit(serialisedErr, ["request", "config"])
+    : serialisedErr
+  const errMsg = `${new Date()}: ${JSON.stringify(sanitisedErr)}`
```
kishore03109, line 18: Style: iirc, our cloudwatch already has a date time attached when we view it from the logs.. if thats the case, do you think we can omit this line then?

kishore03109 (approved):
Mostly LGTM pending: 
[Nit] adding (possibly) redundant datetime

[Tests] understand that there were issues with integrating with mockError, have you done a trivial manual testing (could just make invalid GET request and see if this code indeed works) 
if yes, can ignore this comment!

src/middleware/errorHandler.js
seaerchin: good point but this is in our timezone vs cloudwatch is in UTC :sadge: because of this, i'd rather keep so i dun have to translate timezones 
Screenshot 2023-05-09 at 6 06 13 PM

kishore03109: Screenshot 2023-05-09 at 6 15 49 PM
Hmm no leh I think displays both

seaerchin:
> Mostly LGTM pending: [Nit] adding (possibly) redundant datetime
> 
> [Tests] understand that there were issues with integrating with mockError, have you done a trivial manual testing (could just make invalid GET request and see if this code indeed works) if yes, can ignore this comment!

added testing steps but yea checked locally - we should aim to provide manual tests going forward so that the on-call can copy paste onto the release PR adn verify functionality; thanks for pointing this out!

PR #768, opened 2023-05-16
fix(sites): unwrap result instead of returning

seaerchin:
## Problem
Previously we weren't unwrapping the `Result` when we give a resp back to frontend, which results in an `Object object`. This leads to a error when we click the link to go to the live site.

Closes [IS-199]

## Solution
1. unwrap result

## Tests
1. go to the CMS
2. login via email
3. go to any site
4. make an edit
5. request a review and approve the PR (or just toggle state in db also ok)
6. publish the site
7. click on `go to live site` in the modal
8. you should be directed to the live site

[IS-199]: https://isomeropengov.atlassian.net/browse/IS-199?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/routes/v2/authenticated/sites.ts
+++ b/src/routes/v2/authenticated/sites.ts
@@ -92,15 +92,10 @@ export class SitesRouter {
     { userWithSiteSessionData: UserWithSiteSessionData }
   > = async (req, res) => {
     const { userWithSiteSessionData } = res.locals
-    const possibleSiteUrl = await this.sitesService.getSiteUrl(
-      userWithSiteSessionData
-    )
-
-    // Check for error and throw
-    if (possibleSiteUrl instanceof BaseIsomerError) {
-      return res.status(404).json({ message: possibleSiteUrl.message })
-    }
-    return res.status(200).json({ siteUrl: possibleSiteUrl })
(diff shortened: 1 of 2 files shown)


src/routes/v2/authenticated/sites.ts
```diff
@@ -92,15 +92,10 @@ export class SitesRouter {
     { userWithSiteSessionData: UserWithSiteSessionData }
   > = async (req, res) => {
     const { userWithSiteSessionData } = res.locals
-    const possibleSiteUrl = await this.sitesService.getSiteUrl(
-      userWithSiteSessionData
-    )
-
-    // Check for error and throw
-    if (possibleSiteUrl instanceof BaseIsomerError) {
```
kishore03109, line 100: Nit: Remove import for `BaseIsomerError`

kishore03109 (approved):

src/routes/v2/authenticated/sites.ts
seaerchin: done!

PR #770, opened 2023-05-18
feat(dynamo db): integrate with dynamo db

kishore03109:
## Problem

This is the first PR in the effort to remove the usage of queues in our backend. 

Doesn't fully close IS-186, but am putting a feature flag in the spirit of merging some work into prod first. 

## Solution

The flag `ARE_QUEUES_DEPRECIATED` in `InfraService.ts` will determine whether or not to use the existing queues or use the dynamodbs vs queues for the site launch process. Idea is after the integration work in the infra side + enough testing, we can swap swap the value of the flag for one site launch process, and then if there are no issues for a while in production, we can remove this feature flag + queues related code altogether. 

**Breaking Changes**

  - Details ...

**Reviewer Notes**
I am not sure if its only me but the typing given in the AWS SDK for dynamo DB is abit off ._. 
I have given a screenshot in the related comment below + added test cases to explicitly codify the expected behaviour that we expect from the `dynamoDBDocClient.send(new ScanCommand(params))` call. Open to feedback on this. 

*Steps to replicate the error thrown by ts* 

1. Add this code in `server.js`:
```
import DynamoDBService from "@root/services/infra/DynamoDBService"
const dynamoDbService = new DynamoDBService(new DynamoDBDocClient())
const dynamoDBTests = async () => {
  // put item in dynamoDb
  const createResult = await dynamoDbService.createItem({
    repoName: "my-repo",
    appId: "my-app",
    primaryDomainSource: "example.com",
    primaryDomainTarget: "myapp.example.com",
    domainValidationSource: "example.com",
    domainValidationTarget: "myapp.example.com",
    requestorEmail: "john@example.com",
    agencyEmail: "agency@example.com",
    githubRedirectionUrl: "https://github.com/my-repo",
    redirectionDomain: [
      {
        source: "example.com",
        target: "myapp.example.com",
        type: "A",
      },
    ],
    status: { state: "pending", message: "PENDING_DURING_SITE_LAUNCH" },
  })

  const updateResult = await dynamoDbService.getAllCompletedLaunches()
  console.log("result: ", updateResult.$metadata.httpStatusCode)
  // delete item from dynamoDB
}
dynamoDBTests()
  .then(() => {
    logger.info("Successfully connected to DynamoDB")
  })
  .catch((err) => {
    logger.error(`Unable to connect to DynamoDB: ${err}`)
  })

```
2. Log over the returned entry in DynamoDBService.ts
```
async getAllCompletedLaunches(): Promise<SiteLaunchMessage[]> {
    const entries = ((await this.dynamoDBClient.getAllItems(this.TABLE_NAME))
      .Items as unknown) as SiteLaunchMessage[]
    const entriesWithSDKTypeCast : Record<string, AttributeValue>[] | undefined= (
      await this.dynamoDBClient.getAllItems(this.TABLE_NAME)
    ).Items

    console.log("entries", entries)
    console.log("entriesWithSDKTypeCast", entriesWithSDKTypeCast)
```

Note that the returned object is not of the declared type `Record<string, AttributeValue>[] | undefined`
Screenshot 2023-05-22 at 9 25 28 AM

- `STEP_FUNCTIONS_ARN` : arn for step functions //todo add in 1pw post-approval of this PR
- ARE_QUEUES_DEPRECATED: Whether the queues are deprecated //todo add in 1pw post-approval of this PR

**Tests**

added a file `DynamoDBService.spec.ts` for tests

### DIFF
--- a/microservices/site-launch/shared/types.ts
+++ b/microservices/site-launch/shared/types.ts
@@ -72,7 +72,7 @@ export function isSiteLaunchMessage(obj: unknown): obj is SiteLaunchMessage {
           message.status.state === "failure" ||
           message.status.state === "pending") &&
         typeof message.status.message === "string" &&
-        Object.values(SiteLaunchLambdaStatus).includes(
+        Object.keys(SiteLaunchLambdaStatus).includes(
           message.status.message as SiteLaunchLambdaStatus
         ))) &&
     (typeof message.statusMetadata === "undefined" ||
--- a/microservices/site-launch/shared/types.ts
+++ b/microservices/site-launch/shared/types.ts
@@ -40,7 +40,7 @@ export interface SiteLaunchMessage {
 }
 
 export function isSiteLaunchMessage(obj: unknown): obj is SiteLaunchMessage {
-  if (typeof obj !== "object" || obj === null) {
+  if (!obj) {
     return false
   }
 
--- a/microservices/site-launch/shared/types.ts
+++ b/microservices/site-launch/shared/types.ts
@@ -38,3 +38,44 @@ export interface SiteLaunchMessage {
   status?: SiteLaunchStatus
   statusMetadata?: string
 }
+
+export function isSiteLaunchMessage(obj: unknown): obj is SiteLaunchMessage {
+  if (typeof obj !== "object" || obj === null) {
+    return false
+  }
+
+  const message = obj as SiteLaunchMessage
+
+  return (
+    typeof message.repoName === "string" &&
+    typeof message.appId === "string" &&
+    typeof message.primaryDomainSource === "string" &&
+    typeof message.primaryDomainTarget === "string" &&
+    typeof message.domainValidationSource === "string" &&
+    typeof message.domainValidationTarget === "string" &&
+    typeof message.requestorEmail === "string" &&
+    typeof message.agencyEmail === "string" &&
+    (typeof message.githubRedirectionUrl === "undefined" ||
+      typeof message.githubRedirectionUrl === "string") &&
+    (typeof message.redirectionDomain === "undefined" ||
+      (Array.isArray(message.redirectionDomain) &&
+        message.redirectionDomain.every(
+          (rd) =>
+            typeof rd.source === "string" &&
+            typeof rd.target === "string" &&
+            typeof rd.type === "string"
+        ))) &&
+    (typeof message.status === "undefined" ||
+      (typeof message.status === "object" &&
+        typeof message.status.state === "string" &&
+        (message.status.state === "success" ||
+          message.status.state === "failure" ||
+          message.status.state === "pending") &&
+        typeof message.status.message === "string" &&
+        Object.values(SiteLaunchLambdaStatus).includes(
+          message.status.message as SiteLaunchLambdaStatus
+        ))) &&
+    (typeof message.statusMetadata === "undefined" ||
+      typeof message.statusMetadata === "string")
+  )
+}
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -201,9 +201,9 @@ const config = convict({
         format: "required-string",
         default: "",
       },
-      areQueuesDeprecated: {
+      ffDeprecateSiteQueues: {
         doc: "Whether the queues are deprecated",
-        env: "ARE_QUEUES_DEPRECATED",
+        env: "FF_DEPRECATE_SITE_QUEUES",
         format: "required-boolean",
         default: false,
       },
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -185,7 +185,7 @@ const config = convict({
         doc: "Amazon Resource Name (ARN) of the Step Functions state machine",
         env: "STEP_FUNCTIONS_ARN",
         format: "required-string",
-        default: "",
+        default: "SiteLaunchStepFunctions-dev",
       },
     },
     sqs: {
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -180,6 +180,14 @@ const config = convict({
         default: "site-launch",
       },
     },
+    stepFunctions: {
+      stepFunctionsArn: {
+        doc: "Amazon Resource Name (ARN) of the Step Functions state machine",
+        env: "STEP_FUNCTIONS_ARN",
+        format: "required-string",
+        default: "",
+      },
+    },
     sqs: {
       incomingQueueUrl: {
         doc: "URL of the incoming SQS queue",
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -137,13 +137,13 @@ const config = convict({
     },
   },
   aws: {
+    region: {
+      doc: "AWS region",
+      env: "AWS_REGION",
+      format: "required-string",
+      default: "ap-southeast-1",
+    },
     amplify: {
-      region: {
-        doc: "AWS region",
-        env: "AWS_REGION",
-        format: "required-string",
-        default: "ap-southeast-1",
-      },
       accountNumber: {
         doc: "AWS account number (microservices)",
         env: "AWS_ACCOUNT_NUMBER",
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -201,7 +201,7 @@ const config = convict({
         format: "required-string",
         default: "",
       },
-      featureFlag: {
+      featureFlags: {
         shouldDeprecateSiteQueues: {
           doc: "Whether the queues are deprecated",
           env: "FF_DEPRECATE_SITE_QUEUES",
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -201,6 +201,12 @@ const config = convict({
         format: "required-string",
         default: "",
       },
+      areQueuesDeprecated: {
+        doc: "Whether the queues are deprecated",
+        env: "ARE_QUEUES_DEPRECATED",
+        format: "required-boolean",
+        default: false,
+      },
     },
   },
   github: {
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -201,11 +201,13 @@ const config = convict({
         format: "required-string",
         default: "",
       },
-      ffDeprecateSiteQueues: {
-        doc: "Whether the queues are deprecated",
-        env: "FF_DEPRECATE_SITE_QUEUES",
-        format: "required-boolean",
-        default: false,
+      featureFlag: {
+        shouldDeprecateSiteQueues: {
+          doc: "Whether the queues are deprecated",
+          env: "FF_DEPRECATE_SITE_QUEUES",
+          format: "required-boolean",
+          default: false,
+        },
       },
     },
   },
--- a/src/services/infra/DynamoDBClient.ts
+++ b/src/services/infra/DynamoDBClient.ts
@@ -1,15 +1,11 @@
-import { DynamoDBClient } from "@aws-sdk/client-dynamodb"
+import { DynamoDBClient, ScanCommandOutput } from "@aws-sdk/client-dynamodb"
 import {
   DynamoDBDocumentClient,
   PutCommand,
-  GetCommand,
-  UpdateCommand,
   DeleteCommand,
-  UpdateCommandInput,
   PutCommandOutput,
-  UpdateCommandOutput,
   DeleteCommandOutput,
-  GetCommandOutput,
+  ScanCommand,
 } from "@aws-sdk/lib-dynamodb"
 import autoBind from "auto-bind"
 
@@ -52,7 +48,7 @@ export default class DynamoDBDocClient {
 
   private withLogger = async <T>(
     promise: Promise<T>,
-    type: "create" | "get" | "delete" | "update"
+    type: "create" | "scan" | "delete" | "update"
   ): Promise<T> => {
     try {
       return await promise
@@ -82,38 +78,13 @@ export default class DynamoDBDocClient {
     )
   }
 
-  getItem = async (
-    tableName: string,
-    key: string
-  ): Promise<GetCommandOutput> => {
+  getAllItems = async (tableName: string): Promise<ScanCommandOutput> => {
     const params = {
       TableName: tableName,
-      Key: { appId: key },
-    }
-
-    return this.withLogger(
-      this.dynamoDBDocClient.send(new GetCommand(params)),
-      "get"
-    )
-  }
-
-  updateItem = async ({
-    TableName,
-    Key,
-    UpdateExpression,
-    ExpressionAttributeNames,
-    ExpressionAttributeValues,
-  }: UpdateParams): Promise<UpdateCommandOutput> => {
-    const params: UpdateCommandInput = {
-      TableName,
-      Key,
-      UpdateExpression,
-      ExpressionAttributeNames,
-      ExpressionAttributeValues,
     }
     return this.withLogger(
-      this.dynamoDBDocClient.send(new UpdateCommand(params)),
-      "update"
+      this.dynamoDBDocClient.send(new ScanCommand(params)),
+      "scan"
     )
   }
 
--- a/src/services/infra/DynamoDBClient.ts
+++ b/src/services/infra/DynamoDBClient.ts
@@ -39,7 +39,7 @@ export default class DynamoDBDocClient {
     }
 
     this.dynamoDBDocClient = DynamoDBDocumentClient.from(
-      new DynamoDBClient({ region: config.get("aws.amplify.region") }),
+      new DynamoDBClient({ region: config.get("aws.region") }),
       { marshallOptions, unmarshallOptions }
     )
 
--- a/src/services/infra/DynamoDBService.ts
+++ b/src/services/infra/DynamoDBService.ts
@@ -1,9 +1,13 @@
+import { AttributeValue } from "@aws-sdk/client-dynamodb"
 import { DeleteCommandOutput } from "@aws-sdk/lib-dynamodb"
 import autoBind from "auto-bind"
 
 import { config } from "@config/config"
 
-import { SiteLaunchMessage } from "@root/../microservices/site-launch/shared/types"
+import {
+  SiteLaunchMessage,
+  isSiteLaunchMessage,
+} from "@root/../microservices/site-launch/shared/types"
 
 import DynamoDBClient from "./DynamoDBClient"
 
@@ -23,18 +27,19 @@ export default class DynamoDBService {
   }
 
   async getAllCompletedLaunches(): Promise<SiteLaunchMessage[]> {
-    const entries = ((await this.dynamoDBClient.getAllItems(this.TABLE_NAME))
-      .Items as unknown) as SiteLaunchMessage[]
+    const entries = ((
+      await this.dynamoDBClient.getAllItems(this.TABLE_NAME)
+    ).Items?.filter(isSiteLaunchMessage) as unknown) as SiteLaunchMessage[]
 
-    const successEntries =
+    const completedEntries =
       entries?.filter(
         (entry) =>
           entry.status?.state === "success" || entry.status?.state === "failure"
       ) || []
 
     // Delete after retrieving the items
-    Promise.all(successEntries.map((entry) => this.deleteItem(entry)))
-    return successEntries
+    Promise.all(completedEntries.map((entry) => this.deleteItem(entry)))
+    return completedEntries
   }
 
   async deleteItem(message: SiteLaunchMessage): Promise<DeleteCommandOutput> {
--- a/src/services/infra/DynamoDBService.ts
+++ b/src/services/infra/DynamoDBService.ts
@@ -1,82 +1,45 @@
-import {
-  DeleteCommandOutput,
-  GetCommandOutput,
-  UpdateCommandOutput,
-} from "@aws-sdk/lib-dynamodb"
+import { DeleteCommandOutput } from "@aws-sdk/lib-dynamodb"
 import autoBind from "auto-bind"
 
 import { config } from "@config/config"
 
 import { SiteLaunchMessage } from "@root/../microservices/site-launch/shared/types"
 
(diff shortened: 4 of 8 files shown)


src/services/infra/DynamoDBService.ts
```diff
@@ -1,82 +1,45 @@
...
   async createItem(message: SiteLaunchMessage): Promise<void> {
-    await this.dynamoDBClient.createItem(this.TABLE_NAME, MOCK_LAUNCH)
+    await this.dynamoDBClient.createItem(this.TABLE_NAME, message)
   }
 
-  async getItem(message: SiteLaunchMessage): Promise<GetCommandOutput> {
-    return this.dynamoDBClient.getItem(this.TABLE_NAME, MOCK_LAUNCH.appId)
-  }
+  async getAllSuccessOrFailureLaunches(): Promise<SiteLaunchMessage[]> {
+    const entries = ((await this.dynamoDBClient.getAllItems(this.TABLE_NAME))
```
kishore03109, line 26: Screenshot 2023-05-18 at 8 27 55 AM
Somehow the typing that the SDK defines the object to have the [DynamoDB types](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.NamingRulesDataTypes.html), but the return value already has returned the object of our desired shape O.o 
Since we are the only ones adding the items to the table, the risk of the shape being different from expected is low. 
Open to feedback on this !

src/services/infra/DynamoDBService.ts
```diff
@@ -1,82 +1,45 @@
...
 
   async createItem(message: SiteLaunchMessage): Promise<void> {
-    await this.dynamoDBClient.createItem(this.TABLE_NAME, MOCK_LAUNCH)
+    await this.dynamoDBClient.createItem(this.TABLE_NAME, message)
   }
 
-  async getItem(message: SiteLaunchMessage): Promise<GetCommandOutput> {
-    return this.dynamoDBClient.getItem(this.TABLE_NAME, MOCK_LAUNCH.appId)
-  }
+  async getAllSuccessOrFailureLaunches(): Promise<SiteLaunchMessage[]> {
```
seaerchin, line 25: wdyt of changing the name to `getAllCompletedLaunches`?

src/services/infra/DynamoDBService.ts
```diff
@@ -1,82 +1,45 @@
...
+    const entries = ((await this.dynamoDBClient.getAllItems(this.TABLE_NAME))
+      .Items as unknown) as SiteLaunchMessage[]
+
+    const successEntries =
```
seaerchin, line 29:
```suggestion
    const completedEntries =
```

src/services/infra/DynamoDBService.ts
```diff
@@ -1,82 +1,45 @@
...
-    await this.dynamoDBClient.createItem(this.TABLE_NAME, MOCK_LAUNCH)
+    await this.dynamoDBClient.createItem(this.TABLE_NAME, message)
   }
 
-  async getItem(message: SiteLaunchMessage): Promise<GetCommandOutput> {
-    return this.dynamoDBClient.getItem(this.TABLE_NAME, MOCK_LAUNCH.appId)
-  }
+  async getAllSuccessOrFailureLaunches(): Promise<SiteLaunchMessage[]> {
+    const entries = ((await this.dynamoDBClient.getAllItems(this.TABLE_NAME))
+      .Items as unknown) as SiteLaunchMessage[]
```
seaerchin, line 27: this is a dangerous cast for 2 reasons: 

1. the SDK typing itself says that the `Items` property can be `undefined`, which isn't immediately obvious here and might lead to errors down the line.
2. we might use some properties that we think will always be present but isn't. we should instead just do a typeguard (and check existence of the required properties in the typeguard) and map over the given array.

src/services/infra/InfraService.ts
```diff
@@ -52,6 +54,7 @@ type CreateSiteParams = {
   isEmailLogin: boolean
 }
 
+const ARE_QUEUES_DEPRECIATED = false
```
seaerchin, line 57: if this is meant for feature flagging, then it should be an env var. this is because setting this to `true` still involves a code deployment to our backend (during which we might as well just change the actual code).

changing it to env vars allows us to deploy the required code without an associated code change (ie, we can deploy first and delete later)

src/services/infra/DynamoDBService.ts
```diff
@@ -1,82 +1,45 @@
...
-      // A map of expression attribute values used in the update expression,
-      // in this case mapping ":state" to the value of the Launch status state and ":message" to the value of the Launch status message
-      ExpressionAttributeValues: {
-        ":state": MOCK_LAUNCH.status.state,
-        ":message": MOCK_LAUNCH.status.message,
-      },
-    }
-    return this.dynamoDBClient.updateItem(updateParams)
+    // Delete after retrieving the items
+    Promise.all(successEntries.map((entry) => this.deleteItem(entry)))
```
seaerchin, line 36: this might be dangerous - because we send the email out later on, failure in email sending leads to data loss and we won't know until the agency writes in

src/services/infra/DynamoDBClient.ts
```diff
@@ -82,38 +78,13 @@ export default class DynamoDBDocClient {
...
-      Key,
-      UpdateExpression,
-      ExpressionAttributeNames,
-      ExpressionAttributeValues,
     }
     return this.withLogger(
-      this.dynamoDBDocClient.send(new UpdateCommand(params)),
-      "update"
+      this.dynamoDBDocClient.send(new ScanCommand(params)),
+      "scan"
```
harishv7, line 87: hmm why are we using scan? Is this is DDB Scan command? I think we should not use scan as much as possible

https://dynobase.dev/dynamodb-scan/

src/services/infra/InfraService.ts
harishv7, line 57: nit: ARE_QUEUES_DEPRECATED

harishv7: Yes that's right. Feature flags modifications shouldn't require a full re-deploy. Hence it shld be in the env vars

src/services/infra/DynamoDBService.ts
seaerchin: how do we replicate this behaviour (or if difficult, could you just attach an ss of the `Item` itself so that we can see in entirety the properties)

src/services/infra/DynamoDBClient.ts
kishore03109: Hey! Yup, we shouldn't use scan as much as possible due to it being slow. However, in our use case, we
1. Use DynamoDB to store state of launches during the launch process, prior to being updated in our RDS DB 
2. The amount of entries is small (on our biggest launch dates we only had ~30 launches in a day), which translates to 30 rows

The alternative to this if we dont use the `scan` command is that we store the key in our RDS-DB, which will involve a DB migration to have this functionality, which feels over-engineered for our particular use case! 

As such, I think we should stick scanning our entire DB, wdty?

src/services/infra/InfraService.ts
kishore03109: faed0d0c6cd27a6c2a5c9aa02c5c28acbf797d6c

kishore03109: Sure, modified to use env vars instead here faed0d0c6cd27a6c2a5c9aa02c5c28acbf797d6c

src/services/infra/DynamoDBService.ts
kishore03109: faed0d0c6cd27a6c2a5c9aa02c5c28acbf797d6c

src/services/infra/DynamoDBService.ts
kishore03109: Have updated the PR desc to include the steps, but the TLDR when logging the result over is:

Screenshot 2023-05-22 at 9 28 07 AM

For context, we should instead expect an object of a shape something similar to this:

```
{
  "appId": {
    "S": "my-app"
  },
  "agencyEmail": {
    "S": "agency@example.com"
  },
  "domainValidationSource": {
    "S": "example.com"
  },
  "domainValidationTarget": {
    "S": "myapp.example.com"
  },
  "githubRedirectionUrl": {
    "S": "https://github.com/my-repo"
  },
  "primaryDomainSource": {
    "S": "example.com"
  },
  "primaryDomainTarget": {
    "S": "myapp.example.com"
  },
  "PRIMARY_KEY": {
    "S": "my-app"
  },
  "redirectionDomain": {
    "L": [
      {
        "M": {
          "source": {
            "S": "example.com"
          },
          "target": {
            "S": "myapp.example.com"
          },
          "type": {
            "S": "A"
          }
        }
      }
    ]
  },
  "repoName": {
    "S": "my-repo"
  },
  "requestorEmail": {
    "S": "john@example.com"
  },
  "status": {
    "M": {
      "message": {
        "S": "PENDING_DURING_SITE_LAUNCH"
      },
      "state": {
        "S": "pending"
      }
    }
  }
}
```

src/services/infra/DynamoDBService.ts
kishore03109: 5391e60cd0610e17e5a95a1470dbaae55fc2fc8c

src/services/infra/DynamoDBService.ts
kishore03109: sure, I have added a type guard here: 5391e60cd0610e17e5a95a1470dbaae55fc2fc8c

src/services/infra/DynamoDBService.ts
kishore03109: Hmm, I dont think it's the responsibility of the dynamo db to store failed launches after they have been consumed by our backend. 

As a first cut we are also envisioning them to write in when there is a failure, what do you feel about treating this as an "unrecoverable" error for now, and in the future we can envision recovery options such as retires if the emails are unable to be sent?

src/services/infra/DynamoDBClient.ts
harishv7: given is < 100 rows, I think it's fine for this use case then

src/services/infra/InfraService.ts
harishv7: shall we impose a standard to prefix all feature flags with a FF_ARE_QUEUES_DEPRECATED?

This will allow us to quickly understand an env var is a feature flag toggle

harishv7: nit: the flag should be something like a verb instead of a question

like `DISABLE_SITE_LAUNCH_QUEUES` or `DEPRECATE_SITE_QUEUES`

kishore03109: 0b0dbb1e70d6a38aacf079e41cdee38bebe4a9c4

microservices/site-launch/shared/types.ts
```diff
@@ -38,3 +38,44 @@ export interface SiteLaunchMessage {
...
+
+export function isSiteLaunchMessage(obj: unknown): obj is SiteLaunchMessage {
+  if (typeof obj !== "object" || obj === null) {
+    return false
+  }
+
+  const message = obj as SiteLaunchMessage
+
+  return (
+    typeof message.repoName === "string" &&
```
seaerchin, line 50: actually what i had in mind was just checking `!!message.<required_property>` but this works too HAHA thx for ur hard work

src/config/config.ts
```diff
@@ -180,6 +180,14 @@ const config = convict({
         default: "site-launch",
       },
     },
+    stepFunctions: {
+      stepFunctionsArn: {
+        doc: "Amazon Resource Name (ARN) of the Step Functions state machine",
+        env: "STEP_FUNCTIONS_ARN",
+        format: "required-string",
+        default: "",
```
seaerchin, line 188: sanity check - this can default to an empty string? should we just default to test value or does the step function actually hit an upstream api

src/config/config.ts
```diff
@@ -193,6 +201,12 @@ const config = convict({
         format: "required-string",
         default: "",
       },
+      ffDeprecateSiteQueues: {
+        doc: "Whether the queues are deprecated",
```
seaerchin, line 205: docs should aim to convey contextual/implicit info that isn't obvious from the name. side-note, i think the `ff` prefix is for the raw env var? here, we could group it under a separate tag (ie, `featureFlag`) if that makes more sense

```suggestion
      shouldDeprecateSiteQueues: {
        doc: "Whether queues should be used instead of dynamoDb",
```

src/services/infra/InfraService.ts
```diff
@@ -52,6 +54,7 @@ type CreateSiteParams = {
   isEmailLogin: boolean
 }
 
+const DEPRECATE_SITE_QUEUES = config.get("aws.sqs.ffDeprecateSiteQueues")
```
seaerchin, line 57: take note that this has to change if we change the key above

src/services/infra/DynamoDBService.ts
```diff
@@ -1,82 +1,50 @@
...
-      // A map of expression attribute values used in the update expression,
-      // in this case mapping ":state" to the value of the Launch status state and ":message" to the value of the Launch status message
-      ExpressionAttributeValues: {
-        ":state": MOCK_LAUNCH.status.state,
-        ":message": MOCK_LAUNCH.status.message,
-      },
-    }
-    return this.dynamoDBClient.updateItem(updateParams)
+    // Delete after retrieving the items
+    Promise.all(completedEntries.map((entry) => this.deleteItem(entry)))
```
seaerchin, line 41: can't see the previous comments here but my concern was that this flow: 

1. retrieve entries
2. delete entries
3. send email to ops 

over 

1. retrieve entries
2. send email to ops
3. delete entries where email sent successfully

means that we dk failures till the user writes in whereas if we delete after email sending, we can always check the db to see if there are any failures and take action early

microservices/site-launch/shared/types.ts
```diff
@@ -38,3 +38,44 @@ export interface SiteLaunchMessage {
   status?: SiteLaunchStatus
   statusMetadata?: string
 }
+
+export function isSiteLaunchMessage(obj: unknown): obj is SiteLaunchMessage {
+  if (typeof obj !== "object" || obj === null) {
```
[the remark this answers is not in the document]
kishore03109: fe1a8309116f1560f9a8d63c31d220dc36ed1e34

src/config/config.ts
kishore03109: 0694a084

src/config/config.ts
seaerchin: i think you misunderstood me here, what i meant was whether we should give a different default value if it's not provided, rather than an empty string. in any case, the default value is just meant to be informative, which is why i was curious as to why an empty string over something like `MOCK_STEP_FUNCTION`

src/services/infra/DynamoDBService.ts
seaerchin: hmm, you might be misunderstanding me here - my concern isn't that the emails aren't sent, but that **failed launches can be silently ignored**. 

ie, if the email sending fails, no one will know until the agency writes in about their failed launch whereas if we don't delete until email sending is successful, we can check the db (at the end) to see if there are any failed launches. because most failed launches are human errors, my thinking was that just checking the db at the end then telling the user to restart the process might be a better flow than waiting for them to write in

kishore03109: > ie, if the email sending fails, no one will know until the agency writes in about their failed launch whereas if we don't delete until email sending is successful, we can check the db (at the end) to see if there are any failed launches. because most failed launches are human errors, my thinking was that just checking the db at the end then telling the user to restart the process might be a better flow than waiting for them to write in

Hey yup, you are right to say that this flow may allow launches to fail silently! However, in this case, I would opt for keeping our DB table lean as: 

1. Operationally, agencies + ops tend to be quite anxious about site launch, and typically contact us within 4 hours if the site is not up. It is unlikely that a "www" site is due for launch and eng is not aware of it by the end of the day. 
2. I do think it is going to be unnecessary overhead to ask an on-call to periodically check dynamo DB on the days of site launch. 
3. To check for the cause of failure, eng can look at the SOT for launches, ie amplify console and ask the user to restart the process. 
4. There needs to be a way to mark a successful consumption of the row due to the existence of multiple BE instances. 

I think can chat offline uh text is hard HAHA

src/config/config.ts
kishore03109: Opps yes, thanks for the sanity check! makes sense to put a default value - in this case we can provide the default step functions for this! done here: 90b36c58da73f13922b26d381e89ec2b518290ed

src/services/infra/DynamoDBService.ts
```diff
@@ -1,82 +1,50 @@
...
 export default class DynamoDBService {
   private readonly dynamoDBClient: DynamoDBClient
 
   private readonly TABLE_NAME: string
 
-  constructor() {
-    this.dynamoDBClient = new DynamoDBClient()
+  constructor(dynamoDBClient: DynamoDBClient) {
+    this.dynamoDBClient = dynamoDBClient
     this.TABLE_NAME = config.get("aws.dynamodb.siteLaunchTableName")
```
harishv7, line 21: should this be part of the arguments to the constructor so that we can send in mock values if required for testing?

src/services/infra/InfraService.ts
```diff
@@ -52,6 +54,9 @@ type CreateSiteParams = {
   isEmailLogin: boolean
 }
 
+const DEPRECATE_SITE_QUEUES = config.get(
+  "aws.sqs.featureFlag.shouldDeprecateSiteQueues"
```
harishv7, line 58: nit: can we call the key `featureFlags` plural

src/services/infra/InfraService.ts
```diff
@@ -373,8 +386,12 @@ export default class InfraService {
         message.redirectionDomain = [redirectionDomainObject]
       }
 
-      this.queueService.sendMessage(message)
-
+      if (DEPRECATE_SITE_QUEUES) {
+        this.dynamoDBService.createItem(message)
```
harishv7, line 390: don't we need awaits here?

src/services/infra/InfraService.ts
```diff
@@ -373,8 +386,12 @@ export default class InfraService {
...
+        this.stepFunctionsService.triggerFlow(message)
```
harishv7, line 391: await?

src/services/infra/InfraService.ts
```diff
@@ -373,8 +386,12 @@ export default class InfraService {
...
+      } else {
+        this.queueService.sendMessage(message)
```
harishv7, line 393: await?

src/services/infra/InfraService.ts
```diff
@@ -387,7 +404,9 @@ export default class InfraService {
   }
 
   siteUpdate = async () => {
-    const messages = await this.queueService.pollMessages()
+    const messages = DEPRECATE_SITE_QUEUES
+      ? await this.dynamoDBService.getAllCompletedLaunches()
```
harishv7, line 408: if any of this throws, are the errors caught appropriately?

src/services/infra/StepFunctionsService.ts
```diff
@@ -0,0 +1,22 @@
+import { StepFunctions } from "aws-sdk"
+
+import { SiteLaunchMessage } from "../../../microservices/site-launch/shared/types"
+
+export default class StepFunctionsService {
+  private client: StepFunctions
+
+  constructor(private stateMachineArn: string) {
+    this.client = new StepFunctions({ region: "ap-southeast-1" })
```
harishv7, line 9: should the region come from env vars?

src/services/infra/DynamoDBService.ts
kishore03109: good point! 5af08260e793cc6874eb1d7728178d47648b7b54

src/services/infra/InfraService.ts
kishore03109: okie 5af08260e793cc6874eb1d7728178d47648b7b54

src/services/infra/InfraService.ts
kishore03109: bbe1d9ce0f4b01966bf1242f60ecc56975d4b511

src/services/infra/InfraService.ts
kishore03109: bbe1d9ce0f4b01966bf1242f60ecc56975d4b511

src/services/infra/StepFunctionsService.ts
kishore03109: yes, modified here ca073d14d428b2416659ed0b217655809bbf1214

src/services/infra/InfraService.ts
kishore03109: using a trivial try-catch wrapper here. Understand that we have an error handling portion to standardise the error formatting, will modify that when the scaffolding work for the error formatting is up! 0fa79124d71a6e8ec04bc0c698aff6b18f0260dd

src/services/infra/DynamoDBService.ts
seaerchin: marking this as resolved - even though we have silent failures, the db is just acting as a message broker and it doesn't handle anything beyond message storage + delivery. 

we'll check elsewhere (amp console) for launch failures and recover there instead

src/services/infra/StepFunctionsService.ts
seaerchin: @harishv7 out of curiosity - why inject it from env vars? my understanding is that it's a constant (as we aren't gonna change `region`) so i'd be more inclined to export it from a `constants` file

harishv7: @seaerchin I am fine with a constants file too. I suggested env vars, because we are building this open source, for people to try this, they would need to know exactly what are the **config** they need to provide. Which would be better to come from an env-var or an explicit `config.js` file

harishv7 (approved):

kishore03109:
@seaerchin Hey, let me know if you have any concerns regarding this PR via slack, happy to discuss / create new tickets if needed!

PR #776, opened 2023-05-24
Fix/facebook pixel

alexanderleegs:
## Problem

This PR fixes an issue where users with facebook-pixel values (pre-cms) could not edit their settings page. The issue was that the facebook-pixel value is always a 15 digit number, and sites which edited via github would have entered it as such - we would thus be reading and attempting to write facebook-pixel values as a number, while our request schema specified a string.

The fix for this is to parse the value as a string on read, in order to account for users who may have inserted the `facebook-pixel` value as a number (e.g. `facebook-pixel: 12345`).

This PR is to be reviewed in conjunction with PR #[1282](https://github.com/isomerpages/isomercms-frontend/pull/1282) on the isomercms-frontend repo.

### DIFF
--- a/src/routes/v2/authenticatedSites/__tests__/Settings.spec.js
+++ b/src/routes/v2/authenticatedSites/__tests__/Settings.spec.js
@@ -92,7 +92,10 @@ describe("Settings Router", () => {
       homepage,
     })
 
-    const updatedFbPixelValue = `${configContent["facebook-pixel"]}test`
+    const updatedFbPixelValue = parseInt(
+      `${configContent["facebook-pixel"]}0`,
+      10
+    )
     const updatedTitleValue = `${configContent.title}test`
     const updatedFaq = `${footerContent.faq}test`
     const updatedLogo = `${navigationContent.logo}test`
--- a/src/routes/v2/authenticatedSites/__tests__/Settings.spec.js
+++ b/src/routes/v2/authenticatedSites/__tests__/Settings.spec.js
@@ -92,10 +92,7 @@ describe("Settings Router", () => {
       homepage,
     })
 
-    const updatedFbPixelValue = parseInt(
-      `${configContent["facebook-pixel"]}0`,
-      10
-    )
+    const updatedFbPixelValue = `${configContent["facebook-pixel"]}0`
     const updatedTitleValue = `${configContent.title}test`
     const updatedFaq = `${footerContent.faq}test`
     const updatedLogo = `${navigationContent.logo}test`
--- a/src/routes/v2/authenticatedSites/__tests__/Settings.spec.js
+++ b/src/routes/v2/authenticatedSites/__tests__/Settings.spec.js
@@ -92,7 +92,7 @@ describe("Settings Router", () => {
       homepage,
     })
 
-    const updatedFbPixelValue = `${configContent["facebook-pixel"]}0`
+    const updatedFbPixelValue = `123456789012340`
     const updatedTitleValue = `${configContent.title}test`
     const updatedFaq = `${footerContent.faq}test`
     const updatedLogo = `${navigationContent.logo}test`
--- a/src/services/fileServices/YmlFileServices/ConfigYmlService.js
+++ b/src/services/fileServices/YmlFileServices/ConfigYmlService.js
@@ -20,10 +20,10 @@ class ConfigYmlService {
     const sanitisedContent = sanitizedYamlParse(unparsedContent)
     const sanitisedFacebookPixel =
       !!sanitisedContent["facebook-pixel"] &&
-      parseInt(sanitisedContent["facebook-pixel"], 10)
+      sanitisedContent["facebook-pixel"].toString()
     const content = {
       ...sanitisedContent,
-      "facebook-pixel": sanitisedFacebookPixel || null,
+      "facebook-pixel": sanitisedFacebookPixel || "",
     }
     return { content, sha }
   }
--- a/src/services/fileServices/YmlFileServices/ConfigYmlService.js
+++ b/src/services/fileServices/YmlFileServices/ConfigYmlService.js
@@ -19,7 +19,7 @@ class ConfigYmlService {
     )
     const sanitisedContent = sanitizedYamlParse(unparsedContent)
     const sanitisedFacebookPixel =
-      sanitisedContent["facebook-pixel"] &&
+      !!sanitisedContent["facebook-pixel"] &&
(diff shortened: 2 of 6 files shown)


src/services/fileServices/YmlFileServices/ConfigYmlService.js
```diff
@@ -17,7 +17,13 @@ class ConfigYmlService {
         fileName: CONFIG_FILE_NAME,
       }
     )
-    const content = sanitizedYamlParse(unparsedContent)
+    const sanitisedContent = sanitizedYamlParse(unparsedContent)
+    const content = {
+      ...sanitisedContent,
+      "facebook-pixel": sanitisedContent["facebook-pixel"]
+        ? parseInt(sanitisedContent["facebook-pixel"], 10)
+        : null,
```
seaerchin, line 25: note that `parseInt` can give `NaN`, which has to be handled appropriately (or are we letting the frontend do that?)

src/types/configYml.ts
```diff
@@ -3,4 +3,5 @@ import { ProdPermalink, StagingPermalink } from "./pages"
 export type ConfigYmlData = {
   staging?: StagingPermalink
   prod?: ProdPermalink
+  "facebook-pixel": number | null
```
seaerchin, line 6: why's the type altered here?

src/services/fileServices/YmlFileServices/ConfigYmlService.js
kishore03109: Curious, what are your thoughts on disruptively modifying their repo if it cannot be parsed? That way we can have more assurance in the future that is has to be either an int or a string that can be parsed into an int

src/types/configYml.ts
kishore03109: Hmmm rather than allowing for `null`, could we just make this key optional instead?

src/routes/v2/authenticatedSites/__tests__/Settings.spec.js
```diff
@@ -92,7 +92,10 @@ describe("Settings Router", () => {
       homepage,
     })
 
-    const updatedFbPixelValue = `${configContent["facebook-pixel"]}test`
+    const updatedFbPixelValue = parseInt(
+      `${configContent["facebook-pixel"]}0`,
```
kishore03109, line 96: basic qn, why got "0" at the back ah?

src/validators/RequestSchema.js
```diff
@@ -249,7 +249,7 @@ const UpdateSettingsRequestSchema = Joi.object().keys({
     ),
   }),
   favicon: Joi.string(),
-  "facebook-pixel": Joi.string().allow(""),
+  "facebook-pixel": Joi.number().unsafe().allow(null),
```
kishore03109, line 252: Why arent we just using 
`Joi.string(),Joi.number()` instead ah? 
eg something like how the docs [define the access token](https://joi.dev/api/?v=17.9.1#example)

src/routes/v2/authenticatedSites/__tests__/Settings.spec.js
alexanderleegs: This is to represent a changed value! it used to have `test` appended to it, but since it's now of type number the value needs to still be a number

src/validators/RequestSchema.js
alexanderleegs: A 15 digit number is too large to be verified using Joi.number() alone! .unsafe allows for larger numbers, see https://github.com/hapijs/joi/blob/master/API.md#numberunsafeenabled

src/services/fileServices/YmlFileServices/ConfigYmlService.js
alexanderleegs: Handled in https://github.com/isomerpages/isomercms-backend/pull/776/commits/7fec5ed2d56cb54455a094b8f1dec4f2d01c4557

I think we should be disruptively modifying their repo - this change causes any future change to settings to also modify the type of facebook-pixel where possible. However, we still won't be able to guarantee that this is always true - it's possible for sites to never touch settings and thus not make this change.

src/types/configYml.ts
alexanderleegs: This is just for the purposes of testing! `ConfigYmlData` doesn't have the full typing of data available

src/validators/RequestSchema.js
kishore03109: Hmm wouldnt it then be concerning if we are handling numbers > Number.MAX_SAFE_INTEGER? should we treat it as a string instead?

src/services/fileServices/YmlFileServices/ConfigYmlService.js
```diff
@@ -17,7 +17,14 @@ class ConfigYmlService {
         fileName: CONFIG_FILE_NAME,
       }
     )
-    const content = sanitizedYamlParse(unparsedContent)
+    const sanitisedContent = sanitizedYamlParse(unparsedContent)
+    const sanitisedFacebookPixel =
+      !!sanitisedContent["facebook-pixel"] &&
+      sanitisedContent["facebook-pixel"].toString()
```
seaerchin, line 23: thoughts on maintaining the invariant  (`[0-9]{15}`) either here (on read) or on write? if we do here on read, we must be sure that the `facebook-pixel` property is only ever gotten here.

alexanderleegs: I think we should maintain the invariant on write, but not read! Reason for this is that users can still access github to modify it directly + they might have configured it wrongly in the past but never known - we don't want these to cause errors, and the user is informed via the FE settings not being able to save without modifying it

changed in https://github.com/isomerpages/isomercms-backend/pull/776/commits/5ad709e628452f372d0d982e070581892b016ccc

src/validators/RequestSchema.js
alexanderleegs: This is true, but Kish made a good point that we might want to treat it as a string instead - given that it's an id, we should be treating it as a string instead of a number, i've reverted it to string in https://github.com/isomerpages/isomercms-backend/pull/776/commits/058b761ba760f8489dc97c9c7b32290c430b1148 and parsed int to string instead!

kishore03109 (approved):

kishore03109:
@alexanderleegs seems that tests are failing, please dismiss my review if it involves code change apart from just changing the relevant `.spec` fles

PR #778, opened 2023-05-24
Feat(token): Improve token selection logic

caando:
## Problem

The current token selection algorithm uses random selection, which causes the following problems:
- Difficulty in monitoring token usage as worst token might deviate from overall system usage
- All tokens have close hourly reset time, resulting in usage spike just before reset and dips right after reset
- Does not maximise overall token capacity
- Cannot be easily scaled upward during peak usage

Closes IS-173

## Solution

1. For active tokens pool, pick the token with the earliest reset time thats below threshold.
2. Switch to reserve tokens pool when all active tokens have reached threshold. Each instance would pick only 1 token until its exhausted.
3. Switch back to active token pool after an hour and all active tokens have been reset. 

More details presented [here](https://www.notion.so/opengov/Github-Tokens-Selection-RFC-IS173-af89ac51fb53455a82f7be4c18178483).

**Breaking Changes**

  - This PR requires is_reserved:boolean and reset_time:datetime columns to be added to access_token table in database
  - db migration script should be executed

**Improvements**:

- Allows for monitoring of system capacity through number of exhausted active and reserved tokens
- Optimises overall system token capacity
- Additional reserved tokens can be added during peak periods by insert into database

## Tests

Test for tokens are being inserted into requests and response correctly.
1. Ensure that there is a valid active token in access_tokens table.
2. Login via email user and make edits to `a-test-v4` page. 
3. Ensure that no errors occurred.

Unit tests are added to test TokenServices and its components. 

## Deploy Notes

**Database Schema Changes**
- Run database migration script.


src/errors/NoAvailableTokenError.ts
```diff
@@ -0,0 +1,13 @@
+import { BaseIsomerError } from "@errors/BaseError"
+
+export default class NoAvailableTokenError extends BaseIsomerError {
+  timestamp: Date
+
+  constructor(
+    message = "Unable to select token for request",
+    timestamp = new Date()
```
kishore03109, line 8: Sanity check here: we already have a time stamp in our logger methods, do you still want this explictedly here?

src/services/api/AxiosInstance.ts
```diff
@@ -25,9 +27,11 @@ const requestFormatter = async (axiosConfig: AxiosRequestConfig) => {
...
 
   if (isEmailLoginUser) {
-    const accessToken = await getAccessToken()
-    axiosConfig.headers = {
-      Authorization: `token ${accessToken}`,
+    const accessToken = await tokenServiceInstance
+      .getAccessToken()
+      .unwrapOr("null")
+    if (axiosConfig.headers) {
+      axiosConfig.headers.Authorization = `token ${accessToken}`
```
kishore03109, line 34: From eslint: Assignment to property of function parameter 'axiosConfig'

src/services/api/AxiosInstance.ts
```diff
@@ -25,9 +27,11 @@ const requestFormatter = async (axiosConfig: AxiosRequestConfig) => {
     authMessage === "token undefined"
 
   if (isEmailLoginUser) {
-    const accessToken = await getAccessToken()
-    axiosConfig.headers = {
-      Authorization: `token ${accessToken}`,
+    const accessToken = await tokenServiceInstance
+      .getAccessToken()
+      .unwrapOr("null")
```
kishore03109, line 32: Hmm I think we shouldnt use null at all, what do you feel about just resolving it to ""

src/services/db/TokenService.ts
```diff
@@ -0,0 +1,340 @@
...
+export function queryActiveTokens(
+  tokenDB: ModelStatic<AccessToken>
+): ResultAsync<TokenData[], DatabaseError> {
+  return fromPromise(
+    tokenDB.findAll({
+      where: {
+        isReserved: false,
+      },
+    }),
+    (error) =>
```
kishore03109, line 61:
```suggestion
    (_) =>
```

src/services/db/TokenService.ts
```diff
@@ -0,0 +1,340 @@
...
+import NoAvailableTokenError from "@errors/NoAvailableTokenError"
+import TokenParsingError from "@errors/TokenParsingError"
+
+import { AccessToken } from "@database/models"
+
+// Env vars
+export const GITHUB_TOKEN_LIMIT = 5000
+export const GITHUB_TOKEN_THRESHOLD = 4000 // allowed uses
+export const GITHUB_RESET_INTERVAL = 60 * 60 // seconds
+export const ACTIVE_TOKEN_ALERT_1 = 0.6
```
kishore03109, line 25: Hmm lets be more intentional in our naming, could we use something like 
`ACTIVE_TOKEN_WARN_LEVEL` and `ACTIVE_TOKEN_ALARM_LEVEL` rather than `alert1` and `alert2`?

src/services/db/TokenService.ts
```diff
@@ -0,0 +1,340 @@
...
+  }
+}
+
+export function parseResponseTokenData(
+  response: AxiosResponse
+): Result<ResponseTokenData, TokenParsingError> {
+  // response.config.headers.Authorization format: token ghp_********************************
+  if (
+    typeof response.config?.headers?.Authorization !== "string" ||
+    response.config?.headers?.Authorization.length !== 46 ||
```
kishore03109, line 110: sanity check here: are we sure that github tokens are always a lenght of 46?

src/services/db/TokenService.ts
```diff
@@ -0,0 +1,340 @@
...
+    response.config?.headers?.Authorization.slice(0, 6) !== "token " ||
+    Number.isNaN(+response.headers?.[GITHUB_TOKEN_REMAINING_HEADER]) ||
+    Number.isNaN(+response.headers?.[GITHUB_TOKEN_RESET_HEADER])
```
kishore03109, line 113: style: shall we shift this to a function or something like 

isGithubToken = ...
if (!isGithubToken) {
    return err(...)
}

src/services/db/TokenService.ts
```diff
@@ -0,0 +1,340 @@
...
+  ) {
+    return err(new TokenParsingError(response))
+  }
+  const token: string = response.config.headers.Authorization.slice(6)
+
+  const remainingRequests = +response.headers[GITHUB_TOKEN_REMAINING_HEADER]
```
kishore03109, line 119: basic qn: what is the point of the plus here ya

src/services/db/TokenService.ts
```diff
@@ -0,0 +1,340 @@
...
+
+import logger from "@logger/logger"
+
+import DatabaseError from "@errors/DatabaseError"
+import NoAvailableTokenError from "@errors/NoAvailableTokenError"
+import TokenParsingError from "@errors/TokenParsingError"
+
+import { AccessToken } from "@database/models"
+
+// Env vars
```
kishore03109, line 21: I think this comment is misplaced right?

src/services/db/TokenService.ts
```diff
@@ -0,0 +1,340 @@
...
+  private switchToReserve() {
+    logger.info("switching to using reserved tokens")
+    this.useReservedTokens = true
+    setTimeout(() => {
+      this.useReservedTokens = false
+    }, GITHUB_RESET_INTERVAL * 1000) // 1 hour timeout
+  }
+
+  getAccessToken(): ResultAsync<string, NoAvailableTokenError | DatabaseError> {
+    return this.ensureInitialization().andThen((isInitialised) =>
```
kishore03109, line 290: Hey `isInitialised` is not used

kishore03109 (commented):
Good job in your first PR! 👏 
It's great that you have added tests for the `TokenService.ts` as well

src/services/db/TokenService.ts
caando: Make sense!

src/services/db/TokenService.ts
caando: Make sense! Will do so.

src/services/db/TokenService.ts
caando: It's the [unary operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Unary_plus) which converts a string to number. Given that its not immediately clear what it does, I should switch to parseInt instead for readability.

src/services/db/TokenService.ts
caando: okay, switching to `_`

src/services/db/TokenService.ts
caando: Will do so!

src/services/api/AxiosInstance.ts
caando: I am not too sure how to handle this error case. Currently leaning towards letting the request be made without throwing an error. Perhaps `"No Available Token"` will be more verbose?

src/services/db/TokenService.ts
caando: Yes, will remove this comment.

src/errors/NoAvailableTokenError.ts
caando: I think it make sense to still have a timestamp for when TokenService fails to provide token, because not all errors will be logged. It can be used to get downtime in application without having to go through logs.

src/services/db/TokenService.ts
```diff
@@ -0,0 +1,340 @@
...
+    useLogger.info(
+      `${exhaustedTokensCount}/${activeTokensData.length} active tokens exhausted`
+    )
+  } else if (
+    exhaustedTokensCount >=
+    activeTokensData.length * ACTIVE_TOKEN_ALERT_1
+  ) {
+    useLogger.info(`${ACTIVE_TOKEN_ALERT_1}% of access token capacity reached`)
+    useLogger.info(
+      `${exhaustedTokensCount}/${activeTokensData.length} active tokens exhausted`
```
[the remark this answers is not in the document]
caando: Oh oops forgot to convert decimal to percentage. Thanks for spotting it!

src/services/db/TokenService.ts
caando: I am unable to find any documentation that github tokens will always be length 40. However, from the tokens that we currently have in our database and past experiences with using github tokens, I am fairly confident that it always have length 40. If a token of length not 40 is inserted in the database it can be amended.

alexanderleegs (commented):
lgtm, please resolve kishore's points before merging!

src/services/api/AxiosInstance.ts
caando: Changed to create new object with Authorization instead of assignment.

src/errors/NoAvailableTokenError.ts
kishore03109: > because not all errors will be logged.

wait I am confused here, if the errors are not logged, how will you know that the error occurred in the first place ya?

src/services/api/AxiosInstance.ts
kishore03109: Hmm is there a reason why you are not inclined to throw/log an error here and exit the function immediately? 
This does not seem to me that this the below codes should not be run if we dont get an access code, wdyt

src/services/db/TokenService.ts
kishore03109: Arh, your flow of thought makes sense, and I agree, I am fairly confident that the length of a github token will prob remain at 40. 
However, when dealing with production code, it might not wise to depend on behaviour defined by external parties when it is not explicitly defined in their documentation unless there is strong reason to do so. I would rather have lesser developer overhead of ensuring that the tokens are of this length when adding them to our db!

src/services/db/TokenService.ts
kishore03109: woah TIL

src/errors/NoAvailableTokenError.ts
caando: Just for clarification, all possible errors that `TokenService` can encounter has already been logged on creation. For the `NoAvailableTokenError` specifically, since it is tied to the token limits of the service, I think it will be beneficial to have knowledge of which requests has failed and its timestamp in application. The logs will have another matching timestamp which provides the means of monitoring externally.

src/services/db/TokenService.ts
```diff
@@ -0,0 +1,366 @@
...
+export const GITHUB_TOKEN_REMAINING_HEADER = "x-ratelimit-remaining"
+export const GITHUB_TOKEN_RESET_HEADER = "x-ratelimit-reset"
+const ACTIVE_TOKEN_WARN_LEVEL = 0.6
+const ACTIVE_TOKEN_ALARM_LEVEL = 0.8
+const GITHUB_TOKEN_LENGTH = 40
+
+export type MaybeResetTime = Result<number, null>
+export const NoResetTime: MaybeResetTime = err(null)
+
+export type TokenData = {
```
seaerchin, line 33: don't have to do it here, but 1 thing i noticed was that there's a clear semantic difference between `ReservedTokens` and `ActiveTokens`, which wasn't really clear. 

we can instead expose a `transitionToReserved` function that takes in `ActiveToken` and returns a `ReservedToken` and does some db stuff under the hood too

seaerchin (commented):
good work on this PR!

seaerchin (commented):
good work on this PR!

src/services/db/TokenService.ts
```diff
@@ -0,0 +1,366 @@
...
+) {
+  let exhaustedTokensCount = 0
+  activeTokensData.forEach((tokenData) => {
+    if (
+      tokenData.remainingRequests <=
+      GITHUB_TOKEN_LIMIT - GITHUB_TOKEN_THRESHOLD
+    ) {
+      exhaustedTokensCount += 1
+    }
+  })
```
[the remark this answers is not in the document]
caando: Makes sense!

src/services/db/TokenService.ts
```diff
@@ -0,0 +1,366 @@
...
+export const GITHUB_TOKEN_THRESHOLD = 4000 // allowed uses
+export const GITHUB_RESET_INTERVAL = 60 * 60 // seconds
+export const GITHUB_TOKEN_REMAINING_HEADER = "x-ratelimit-remaining"
+export const GITHUB_TOKEN_RESET_HEADER = "x-ratelimit-reset"
+const ACTIVE_TOKEN_WARN_LEVEL = 0.6
+const ACTIVE_TOKEN_ALARM_LEVEL = 0.8
+const GITHUB_TOKEN_LENGTH = 40
+
+export type MaybeResetTime = Result<number, null>
+export const NoResetTime: MaybeResetTime = err(null)
```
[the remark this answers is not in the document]
caando: yes!

src/services/db/TokenService.ts
```diff
@@ -0,0 +1,366 @@
...
+      this.updateTokens(
+        tokenData.token,
+        tokenData.remainingRequests,
+        tokenData.resetTime
+      )
+    )
+  }
+
+  updateTokens(token: string, remainingRequests: number, resetTime: number) {
+    const findActive = this.activeTokensData.find(
```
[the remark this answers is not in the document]
caando: that name makes sense!

src/services/db/TokenService.ts
```diff
@@ -0,0 +1,366 @@
...
+    const activeToken = selectActiveToken(activeTokenData)
+    if (activeToken.isOk()) {
+      return okAsync([activeToken.value, false])
+    }
+    logger.info("active tokens capacity reached")
+  }
+  return selectReservedToken(reservedToken, tokenDB).andThen(
+    (newReservedToken) => {
+      if (newReservedToken.isOk()) {
+        return okAsync([newReservedToken.value, true] as [
```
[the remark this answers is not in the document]
caando: Thanks!

src/services/api/AxiosInstance.ts
caando: My previous concerns were that cancelling the request at interceptor might cause unintended behaviour. As per Harish comments, it has been change to only make the request if `isOk`.

src/services/api/AxiosInstance.ts
```diff
@@ -25,10 +27,9 @@ const requestFormatter = async (axiosConfig: AxiosRequestConfig) => {
     authMessage === "token undefined"
 
   if (isEmailLoginUser) {
-    const accessToken = await getAccessToken()
-    axiosConfig.headers = {
-      Authorization: `token ${accessToken}`,
-    }
+    const accessToken = await tokenServiceInstance
+      .getAccessToken()
+      .unwrapOr("unable to get github token")
```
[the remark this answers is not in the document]
caando: Makes sense, it has been changed to `AxiosRequestConfig` is only return if there is a valid accessToken.

src/services/db/TokenService.ts
caando: That's true. Perhaps this check isn't suitable at the moment and I'll drop it until I can get further confirmation that tokens will be length 40.

src/services/db/TokenService.ts
```diff
@@ -0,0 +1,366 @@
...
+  left: TokenData,
+  right: TokenData,
+  nowEpochSecondsUTC: number
+): boolean {
+  return (
+    left.resetTime
+      .andThen((resetTime) =>
+        resetTime < nowEpochSecondsUTC ? NoResetTime : ok(resetTime)
+      )
+      .unwrapOr(Number.MAX_VALUE) <
```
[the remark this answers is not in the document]
caando: Yes! I'll break it down into cases instead of defaulting `NO_RESET_TIME` to `Number.MAX_VALUE`.

src/services/db/TokenService.ts
```diff
@@ -0,0 +1,366 @@
...
+  const now: Date = new Date()
+  const nowEpochSecondsUTC: number =
+    now.getTime() / 1000 + (now.getTimezoneOffset() * 60) / 1000
+  activeTokensData.forEach((activeTokenData) => {
+    // Choose earliest non-null reset time from tokens that has not exceeded  threshold
+    if (
+      activeTokenData.remainingRequests >
+        GITHUB_TOKEN_LIMIT - GITHUB_TOKEN_THRESHOLD ||
+      activeTokenData.resetTime.isErr() ||
+      activeTokenData.resetTime.value < nowEpochSecondsUTC
```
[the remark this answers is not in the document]
caando: Yep, makes sense!

src/services/db/TokenService.ts
```diff
@@ -0,0 +1,366 @@
...
+      resetTime: NoResetTime,
+    }))
+  )
+}
+
+type LoggerType = typeof logger
+
+export function activeUsageAlert(
+  activeTokensData: TokenData[],
+  useLogger: LoggerType
```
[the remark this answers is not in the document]
caando: Resolving this by using `logger` directly and `jest.spyOn` in tests.

src/services/db/TokenService.ts
```diff
@@ -0,0 +1,366 @@
...
+export function occupyReservedToken(reservedToken: AccessToken) {
+  const resetTime = new Date()
+  resetTime.setSeconds(resetTime.getSeconds() + GITHUB_RESET_INTERVAL)
+  reservedToken.update("resetTime", resetTime)
+
+  // set reset time to null after reset time
+  setTimeout(() => {
+    reservedToken.update("resetTime", null)
+  }, GITHUB_RESET_INTERVAL * 1000)
+}
```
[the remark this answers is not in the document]
caando: There isn't a good way to set a timer without `setTimeout`

src/services/db/TokenService.ts
caando: Introduced semantic differences between `ActiveTokenData` and `ReservedTokenData`.

caando:
Other issues or considerations

kishore03109 (commented):
As discussed, could we look at all the tokens at startup and mark all unreserved tokens that have yet to reset the time (ie > 1 hour but the `reset-time` is not null) to be marked as null?

QiluXie:
@caando Good job on this PR, description is very easy to read! 
Looks like all comments are resolved and it's good to go

QiluXie (approved):

PR #782, opened 2023-05-29
feat(e2e): create new e2e email users

seaerchin:
## Problem
Our e2e user account at present is only valid for github login; hence, we have no way of testing email logins

Closes IS-206

## Solution
1. piggyback on the verification bypass in `AuthenticationMiddlewareService`

**notes**
- maybe want to shard out the `cypress` object into `github` + `email`

### DIFF
--- a/src/middleware/authentication.ts
+++ b/src/middleware/authentication.ts
@@ -24,13 +24,17 @@ export class AuthenticationMiddleware {
     autoBind(this)
   }
 
-  verifyAccess(req: RequestWithSession, res: Response, next: NextFunction) {
+  async verifyAccess(
+    req: RequestWithSession,
+    res: Response,
+    next: NextFunction
+  ) {
     const { cookies, originalUrl: url, session } = req
     const {
       isomerUserId,
       email,
       ...rest
-    } = this.authenticationMiddlewareService.verifyAccess({
+    } = await this.authenticationMiddlewareService.verifyAccess({
       cookies,
       url,
       userInfo: session.userInfo,
--- a/src/middleware/authentication.ts
+++ b/src/middleware/authentication.ts
@@ -10,6 +10,7 @@ interface RequestWithSession extends Request {
   session: Session & SessionData
 }
 
+// eslint-disable-next-line import/prefer-default-export
 export class AuthenticationMiddleware {
   private readonly authenticationMiddlewareService: AuthenticationMiddlewareService
 
@@ -26,20 +27,18 @@ export class AuthenticationMiddleware {
   verifyAccess(req: RequestWithSession, res: Response, next: NextFunction) {
     const { cookies, originalUrl: url, session } = req
     const {
-      accessToken,
-      githubId,
       isomerUserId,
       email,
+      ...rest
     } = this.authenticationMiddlewareService.verifyAccess({
       cookies,
       url,
       userInfo: session.userInfo,
     })
     const userSessionData = new UserSessionData({
-      accessToken,
-      githubId,
       isomerUserId,
       email,
+      ...rest,
     })
     res.locals.userSessionData = userSessionData
     return next()
--- a/src/services/middlewareServices/AuthenticationMiddlewareService.ts
+++ b/src/services/middlewareServices/AuthenticationMiddlewareService.ts
@@ -39,13 +39,11 @@ const GENERAL_ACCESS_PATHS = [
   "/v2/auth/whoami",
 ]
 
-const E2E_ADMIN_EMAIL = "admin@e2e.gov.sg"
-const E2E_COLLAB_EMAIL = "collab@e2e.gov.sg"
-
 interface E2eCookie {
   isomercmsE2E: string
   e2eUserType: string
   site: string
+  email: string
 }
 
 type UnverifiedSession =
@@ -57,7 +55,10 @@ export type VerifyAccessProps = UnverifiedSession & {
   // NOTE: Either both properties are present on the cookie
   // or none are present.
   // We disallow having 1 or the other.
-  cookies: RequireAllOrNone<E2eCookie, "e2eUserType" | "isomercmsE2E" | "site">
+  cookies?: RequireAllOrNone<
+    E2eCookie,
+    "e2eUserType" | "isomercmsE2E" | "site" | "email"
+  >
   url: string
 }
 
@@ -73,12 +74,12 @@ const getUserType = (userType: string): TestUserTypes => {
 
 const generateE2eEmailUser = async (
   role: CollaboratorRoles,
-  site?: string
+  site?: string,
+  email?: string
 ): Promise<SessionDataProps> => {
   const [user] = await User.findOrCreate({
     where: {
-      email:
-        role === CollaboratorRoles.Admin ? E2E_ADMIN_EMAIL : E2E_COLLAB_EMAIL,
+      email,
       contactNumber: "1235678",
     },
   })
@@ -113,13 +114,14 @@ const generateGithubUser = (): SessionDataProps => ({
 
 const extractE2eUserInfo = async (
   userType: TestUserTypes,
-  site?: string
+  site?: string,
+  email?: string
 ): Promise<SessionDataProps> => {
   switch (userType) {
     case "Email admin":
-      return generateE2eEmailUser(CollaboratorRoles.Admin, site)
+      return generateE2eEmailUser(CollaboratorRoles.Admin, site, email)
     case "Email collaborator":
-      return generateE2eEmailUser(CollaboratorRoles.Contributor, site)
+      return generateE2eEmailUser(CollaboratorRoles.Contributor, site, email)
     case "Github user":
       return generateGithubUser()
     default: {
@@ -134,11 +136,13 @@ export default class AuthenticationMiddlewareService {
     cookies,
     url,
   }: Omit<VerifyAccessProps, "userInfo">): TestUserTypes | false {
-    const { isomercmsE2E, e2eUserType, site } = cookies
+    if (!cookies) return false
+
+    const { isomercmsE2E, e2eUserType } = cookies
     const urlTokens = url.split("/") // urls take the form "/v1/sites/<repo>/<path>""
 
     // NOTE: If the cookie is not set, this is an actual user.
-    if (!isomercmsE2E || !e2eUserType || !site) return false
+    if (!isomercmsE2E || !e2eUserType) return false
 
     // NOTE: Cookie is set but wrong, implying someone is trying to figure out the secret
     if (isomercmsE2E !== E2E_TEST_SECRET) throw new AuthError("Bad credentials")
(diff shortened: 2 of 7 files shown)


src/services/middlewareServices/AuthenticationMiddlewareService.ts
```diff
@@ -10,63 +11,135 @@ import { AuthError } from "@errors/AuthError"
...
+  private verifyE2E({
+    cookies,
+    url,
+  }: Omit<VerifyAccessProps, "userInfo">): TestUserTypes | false {
+    const { isomercmsE2E, e2eUserType } = cookies
     const urlTokens = url.split("/") // urls take the form "/v1/sites/<repo>/<path>""
 
-    if (!isomercmsE2E) return false
+    // NOTE: If the cookie is not set, this is an actual user.
+    if (!isomercmsE2E || !e2eUserType) return false
```
kishore03109, line 101: Just to check my understanding here ya, actually this can be simpified to 
if (!isomercmsE2E) return false;
right, since `e2eUserType` will be present as long as `e2eUserType` is present?

src/services/middlewareServices/AuthenticationMiddlewareService.ts
```diff
@@ -10,63 +11,135 @@ import { AuthError } from "@errors/AuthError"
...
+    const userType = getUserType(e2eUserType)
+
     // General access paths are allowed
-    if (GENERAL_ACCESS_PATHS.includes(url)) return true
+    if (GENERAL_ACCESS_PATHS.includes(url)) return userType
 
-    // Throw an error if accessing a repo other than e2e-test-repo
+    // Throw an error if accessing a repo other than
+    // the allowed repos for each respective user type
     const repo = urlTokens[3]
```
kishore03109, line 115: Am confused here, where does this `3` come from ah

src/services/middlewareServices/AuthenticationMiddlewareService.ts
```diff
@@ -10,63 +11,135 @@ import { AuthError } from "@errors/AuthError"
...
+  }: Omit<VerifyAccessProps, "userInfo">): TestUserTypes | false {
+    const { isomercmsE2E, e2eUserType } = cookies
     const urlTokens = url.split("/") // urls take the form "/v1/sites/<repo>/<path>""
 
-    if (!isomercmsE2E) return false
+    // NOTE: If the cookie is not set, this is an actual user.
+    if (!isomercmsE2E || !e2eUserType) return false
 
+    // NOTE: Cookie is set but wrong, implying someone is trying to figure out the secret
     if (isomercmsE2E !== E2E_TEST_SECRET) throw new AuthError("Bad credentials")
```
kishore03109, line 104: Just a sanity check, the value for `isomercmsE2E` will be the same for all e2e test users right?

src/services/middlewareServices/AuthenticationMiddlewareService.ts
```diff
@@ -80,7 +153,12 @@ export default class AuthenticationMiddlewareService {
       const accessToken = retrievedToken
         ? jwtUtils.decryptToken(retrievedToken)
         : ""
-      return { accessToken, githubId, isomerUserId, email }
+      return {
+        accessToken,
+        githubId: githubId ?? "",
```
kishore03109, line 158: Hmm why are we returning an empty string here ya?

kishore03109 (commented):
Hey, I have a concern regarding the implementation of this :( 
At the moment, if we want to add more e2e users to test more sophisticated flows, we would need to keep adding more env vars. Newer devs would also have to be aware that when they run e2e, they would need to create the new e2e users and manually update the env var for this. 

How do we feel about naming the users intentionally, eg `e2eEmailFlowAdmin1`, `e2eEmailFlowCollab1` and then performing the `findOrCreate()` from our `UsersService` to get the user ID? 
that way we dont have to touch the env vars, and we could leverage on this to create multiple users for our e2e test suite in the future

src/services/middlewareServices/AuthenticationMiddlewareService.ts
seaerchin: that's true at _compile_ time but because we don't validate it, we have to do this check for both.

src/services/middlewareServices/AuthenticationMiddlewareService.ts
seaerchin: from teh comment in the code, 

`// urls take the form "/v1/sites/<repo>/<path>"`

so the [3] refers to `repo` (we split by `/` so the output array is `['', 'v1', 'sites', '<repo>', '<path>']`)

src/services/middlewareServices/AuthenticationMiddlewareService.ts
seaerchin: yes

src/services/middlewareServices/AuthenticationMiddlewareService.ts
seaerchin: the type definition is `SessionDataProps = IsomerUserProps | GithubUserProps`, which expands into 

```ts
type IsomerUserProps = {
  isomerUserId: string
  email: string
}

type GithubUserProps = IsomerUserProps & {
  githubId: string
  accessToken: string
}

type ConcreteProps = {
  githubId?: string
  accessToken?: string
  isomerUserId: string
  email: string
}
```

seaerchin: i can remove the `""` on `githubId` but we need it for `email`

seaerchin:
> Hey, I have a concern regarding the implementation of this :( At the moment, if we want to add more e2e users to test more sophisticated flows, we would need to keep adding more env vars. Newer devs would also have to be aware that when they run e2e, they would need to create the new e2e users and manually update the env var for this.
> 
> How do we feel about naming the users intentionally, eg `e2eEmailFlowAdmin1`, `e2eEmailFlowCollab1` and then performing the `findOrCreate()` from our `UsersService` to get the user ID? that way we dont have to touch the env vars, and we could leverage on this to create multiple users for our e2e test suite in the future

this is a good point! however, i'm not inclined to make the change here - this is because i'm not exactly sure how we'll get + use the `isomerUserId`, which makes it difficult to change. however, i'll first change it from returning a fixed object (in `extractE2eUserInfo`) to using a function to get the desired shape so that we can more easily change this in the future if required. will file this as an isseu

seaerchin:
filed [here](https://isomeropengov.atlassian.net/browse/IS-235)

src/services/middlewareServices/AuthenticationMiddlewareService.ts
kishore03109: hmm if `githubId` not required, what do we feel about making them optional in the interface?

seaerchin: more inclined to raise a separate PR for this if required tbh

src/services/middlewareServices/__tests__/AuthenticationMiddlewareService.spec.ts
alexanderleegs, line 1: Do we also need a test for the non-e2e test case?

src/services/middlewareServices/AuthenticationMiddlewareService.ts
```diff
@@ -10,63 +11,141 @@ import { AuthError } from "@errors/AuthError"
...
+    {
+      isomercmsE2E: string
+      e2eUserType: string
+    },
+    "e2eUserType" | "isomercmsE2E"
+  >
   url: string
 }
 
+type TestUserTypes = "Email admin" | "Email collaborator" | "Github user"
```
alexanderleegs, line 56: Nit: would an enum here make it easier to ensure we're referencing the right value?

src/services/middlewareServices/AuthenticationMiddlewareService.ts
kishore03109: nah is fine, may I get your help to add a small comment summarising this limitation?

kishore03109 (approved):
lgtm, we can rethink this implementation if we need to test with more accounts then

src/services/middlewareServices/AuthenticationMiddlewareService.ts
```diff
@@ -10,63 +11,180 @@ import { AuthError } from "@errors/AuthError"
...
+  // but repo is the underlying github reference
+  name: "e2e email test site",
+  repo: "e2e-email-test-repo",
+}
 const E2E_TEST_SECRET = config.get("cypress.e2eTestSecret")
-const E2E_TEST_GH_TOKEN = config.get("cypress.e2eTestGithubToken")
-const E2E_TEST_USER = "e2e-test"
+
+export const E2E_TEST_GH_TOKEN = config.get("cypress.e2eTestGithubToken")
+export const E2E_TEST_USER = "e2e-test"
```
kishore03109, line 34: Nit: now that we have two users, can we rename this person to `E2E_TEST_GITHUB_USER`

src/services/middlewareServices/AuthenticationMiddlewareService.ts
kishore03109: hmm +1 for not using enums in ts tho :(

seaerchin: hmm, i can do a `const` object + type manipulation bah if we just want a nice aliasing model like `UserTypes.Email.Admin`

src/services/middlewareServices/AuthenticationMiddlewareService.ts
seaerchin: sure

src/services/middlewareServices/AuthenticationMiddlewareService.ts
harishv7: Yea that would be good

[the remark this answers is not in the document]
seaerchin: done!

src/services/middlewareServices/AuthenticationMiddlewareService.ts
seaerchin: done

seaerchin:
> > this is a good point! however, i'm not inclined to make the change here - this is because i'm not exactly sure how we'll get + use the isomerUserId, which makes it difficult to change. however, i'll first change it from returning a fixed object (in extractE2eUserInfo) to using a function to get the desired shape so that we can more easily change this in the future if required. will file this as an issue
> 
> just checking in, this PR already addresses this right? since now got no more env var for this + there is code to find or create here?

yeap! i wanted to write a few specs out first to see whether there was any complications w/ user creation etc. should've marked it as draft maybe and that was why i didn't request a review but now it uses the db cos the user management is still done thru db ultimately

src/services/middlewareServices/__tests__/AuthenticationMiddlewareService.spec.ts
seaerchin: ideally yes, but i lazy cos it's not directly part of this PR ._. wad do we feel about filing an issue?

src/middleware/authentication.ts
```diff
@@ -23,23 +24,25 @@ export class AuthenticationMiddleware {
...
+  ) {
     const { cookies, originalUrl: url, session } = req
     const {
-      accessToken,
-      githubId,
       isomerUserId,
       email,
-    } = this.authenticationMiddlewareService.verifyAccess({
+      ...rest
+    } = await this.authenticationMiddlewareService.verifyAccess({
```
kishore03109, line 37: quick clarification: why are we awaiting here ya

kishore03109 (approved):
code wise lgtm, just have one clarification

PR #784, opened 2023-05-29
ref(logger): refactor logger impl to be separate from interface

seaerchin:
## Problem
Previously, our loggers were a thin wrapper over `CloudWatchLogger` and adding an external logging source was difficult. This was because our loggers depended on a concrete implementation rather than an interface. 

Closes IS-209

## Solution
1. separate the logging interface from the logger 
    - this can be further extended when we refactor our errors, so that instead of taking `string`, the method takes `string | InternalError`
2. add a main logger that exposes the `Logger` interface + allows us to inject loggers into it
    - this also exposes a `useFormatter` method to allow people to customise their log formats (idk if this should be on main or sub loggers ._. feel free to discuss) 
    - this handles calling our sub loggers 
3. adds a new `consoleLogger` in that is a simple object. prefixes w/ the level + has colours so it's obv.

## Tests
1. hit an endpoint on our backend 
2. there should be logs recorded, with colours

**notes**
1. `init` function on `cloudWatchLogger` was copied over as-is and this wasn't called previously **but is now.**
2. `formatter` was chosen to be placed on the main logger - we can change this if needed? 
3. `const defaultImport = require("module")` had to be changed to `const defaultImport = require("module").default` due to how default imports work in es6 vs node (see [here](https://stackoverflow.com/questions/40294870/module-exports-vs-export-default-in-node-js-and-es6)) for details.


src/logger/cloudwatch.logger.ts
```diff
@@ -0,0 +1,91 @@
+import AWS from "aws-sdk"
+import Bluebird from "bluebird"
+import winston from "winston"
+import WinstonCloudwatch from "winston-cloudwatch"
+
+import { consoleLogger } from "./console.logger"
+
+// AWS
+const AWS_REGION_NAME = "ap-southeast-1"
```
kishore03109, line 9: nit: shall we use env var for this?

src/logger/cloudwatch.logger.ts
```diff
@@ -0,0 +1,91 @@
...
+AWS.config.update({ region: AWS_REGION_NAME })
+const awsMetadata = new AWS.MetadataService()
+const metadataRequest = Bluebird.promisify<string, string>(
+  awsMetadata.request.bind(awsMetadata)
+)
+
+// Constants
+// TODO: Check this env var as it is not in example
+const LOG_GROUP_NAME = `${process.env.AWS_BACKEND_EB_ENV_NAME}/nodejs.log`
```
kishore03109, line 18: Hey I dont understand this comment :( 
this is in reference so some example in the documentation ah?

src/logger/cloudwatch.logger.ts
```diff
@@ -0,0 +1,91 @@
...
+
+// Retrieve EC2 instance since that is the cloudwatch log stream name
+async function getEc2InstanceId(): Promise<string> {
+  let id: string
+  try {
+    id = await metadataRequest("/latest/meta-data/instance-id").timeout(1000)
```
kishore03109, line 24: [Perf] hmm i think we should env var this. worried that this would introduce a network call for every log that we make, which seems expensive 

Also to check in, this `id` wont change for a specific env right?

src/logger/console.logger.ts
```diff
@@ -0,0 +1,15 @@
+import { Logger } from "./logger.types"
+
+// eslint-disable-next-line import/prefer-default-export
+export const consoleLogger: Logger = {
+  info: (message: string | Record<string, unknown>): void => {
+    // NOTE: This adds RGB to our console logs
+    console.log("\x1b[32m%s\x1b[0m", `[INFO]: ${message}`)
```
kishore03109, line 7: [Nit, style]: could we name the colours? 
eg.
```
const redColour = \x1b[32m%s\x1b[0m
console.log(redColour, `[INFO]: ${message}`)
```

src/utils/media-utils.ts
```diff
@@ -57,7 +57,7 @@ export const getMediaFileInfo = async ({
       }
     } catch (err) {
       // If an error occurs while retrieving the image, we log but continue to return generic values
-      logger.error(err)
+      logger.error(JSON.stringify(err))
```
kishore03109, line 60: Hmm what do we feel about wrapping a `JSON.stringify()` inside our logger codes, so the callers dont have to worry about converting JSON objects to string but can directly call the logger.error()?

kishore03109 (commented):
[Tests]: as a quick sanity check to confirm that this is working could we push to to staging now then check tomorrow just see if visually looks ok?

src/logger/cloudwatch.logger.ts
seaerchin: i actually dk too - i tried asking @alexanderleegs but he didn't know too HAHAHAHA

seaerchin: for context, this was in the old `logger.js` and i copied this over because it seemed important

src/logger/console.logger.ts
seaerchin: sounds good!

src/logger/cloudwatch.logger.ts
seaerchin: this is actually legacy code that i preserved for the sake of compat - this function is also ever called in the constructor so it doesn't cause a perf hit. i didn't change the `timeout` because i didn't know if there was any additional effects - it was already there in the [original PR](https://github.com/isomerpages/isomercms-backend/pull/69/files) and there wasn't any comments wrt it .__.

src/utils/media-utils.ts
seaerchin: not inclined to do this - we define a message format (`Loggable`) that our log functions have to adhere to. This `JSON.stringify` is required because the error type here is `unknown`; if we were to allow `unknown` (and hence, shift this away from the client), it defeats the purpose of having a structured log 

i thought about exposing a `coerce = JSON.stringify` but given that it does nothing, i didn't add it in in the end

src/logger/cloudwatch.logger.ts
kishore03109: nit so can dismiss this but 
1) we already have an env var for the region
2) we could move to a more consistent code base that way

src/logger/cloudwatch.logger.ts
kishore03109: arh ok if just in constructor should be fine bah, i thought was per every `logger.<logger-type>()` call

src/utils/media-utils.ts
kishore03109: Screenshot 2023-05-31 at 1 42 19 PM
Note that we do log the raw errors at multiple locations, so this would need to be changed at number call sites ya

src/logger/cloudwatch.logger.ts
seaerchin: cc @harishv7 @alexanderleegs should this be env var or const? idm either (since our env var is now handled thru `convict`) but cos now codebase inconsistent, so will raise another PR to do this (make all either env var or ref the `const`)

src/logger/cloudwatch.logger.ts
```diff
@@ -0,0 +1,91 @@
...
+    } catch (err) {
+      consoleLogger.error(`${err}`)
+      consoleLogger.error(`Failed to initiate CloudWatch logger`)
+    }
+  }
+
+  // this method is used to log non-error messages, replacing console.log
+  async info(logMessage: string | Record<string, unknown>) {
+    try {
+      await this._logger.info(`${logMessage}`)
```
kishore03109, line 69: seems like these awaits (note there are multiple) have no effect ya, shall we remove them?

src/logger/logger.ts
```diff
@@ -0,0 +1,60 @@
+import moment from "moment-timezone"
+
+import { config } from "@root/config/config"
+
+import CloudWatchLogger from "./cloudwatch.logger"
+import { consoleLogger } from "./console.logger"
+import { Formatter, Loggable, Logger, LogMethod } from "./logger.types"
+
+const NODE_ENV = config.get("env")
+const useCloudwatchLogger = NODE_ENV === "prod" || NODE_ENV === "vapt"
```
kishore03109, line 10: Sanity check here: should we use this for our staging env?

kishore03109 (commented):
`npm run build` fails :(

src/logger/cloudwatch.logger.ts
harishv7: @seaerchin Either is fine, but it's should be exposed as a configurable parameter. Since we alr have it in env, it is good to standardise towards using that

src/utils/media-utils.ts
seaerchin: interesting - this means we have a missing step in our ci (and also tests for these)

package.json
```diff
@@ -99,6 +99,7 @@
     "@tsconfig/recommended": "^1.0.1",
     "@types/aws-lambda": "^8.10.106",
     "@types/bcrypt": "^5.0.0",
+    "@types/bluebird": "^3.5.38",
```
harishv7, line 102: side qn: why do we need bluebird?

src/logger/cloudwatch.logger.ts
harishv7: Oh this comment is because me and Alex tried to find where this log exists, but we could never find it.

Might need to look again and if we conclude this is not being used anywhere, can remove

cc @alexanderleegs

src/logger/cloudwatch.logger.ts
seaerchin: good catch, refactored the file!

src/logger/logger.ts
seaerchin: good catch, somehow i missed teh `staging` when i copied (i checked teh deleted `logger.js`)

seaerchin:
> `npm run build` fails :(

hmm - we actually lack a build step in `ci`; will be adding this.

src/logger/cloudwatch.logger.ts
seaerchin: sure, gonna raise another PR for this

src/logger/logger.types.ts
```diff
@@ -0,0 +1,13 @@
+export type Loggable = string | Record<string, unknown>
+
+export interface LogMethod {
+  (message: Loggable): void
+}
+
+export type Formatter = (message: string) => string
+
+export interface Logger {
+  info: LogMethod
```
harishv7, line 10: can we add debug & fatal as well?

package.json
seaerchin: `cloudwatch.logger` uses `Bluebird.promisify`

src/logger/cloudwatch.logger.ts
seaerchin: this wasn't actually used until this PR - this was used in `init`, which was never called till now. i decided to call it because it seemed to be setting transport defaults and cloudwatch log settings ._.

src/logger/logger.types.ts
seaerchin: sure, but i do have a few questions:

1. should any of those be optional? we might wish to make debug optional cos prod/staging loggers might not require those (but take note, also, that `winston-cloudwatch` has `debug` but not `fatal`)
2. what colour should we use? thinking blue for debug, bg red for fatal

harishv7: > should any of those be optional?
Debug & fatal can be both optional

> what colour should we use on console logger? thinking blue for debug, bg red for fatal
yea that's what we usually use, sounds good!

src/logger/console.logger.ts
```diff
@@ -0,0 +1,38 @@
...
+// where %s is where the message gets injected
+const COLORS = {
+  FOREGROUND: {
+    RED: "\x1b[31m%s\x1b[0m",
+    GREEN: "\x1b[32m%s\x1b[0m",
+    YELLOW: "\x1b[33m%s\x1b[0m",
+    BLUE: "\x1b[34m%s\x1b[0m",
+  },
+  BACKGROUND: {
+    RED: "\x1b[41m%s\x1b[0m",
```
harishv7, line 16: why is this a background.red?

harishv7 (approved):

src/logger/console.logger.ts
seaerchin: like we agreed above, fatal should be background red, no? we already use foreground red for `error` and this is more severe, so this is bg red.

PR #785, opened 2023-05-30
Fix: make email lowercase

alexanderleegs:
## Problem

This PR fixes an issue with our email verification process for github users. Previously, for users who were logging in via Github for the first time, they would need to verify and associate their email to their isomer user. However, this input was not parsed in any way, so users would be able to include capitalisation in their associated email. This caused issues when users later tried to login via email - as we forcefully lowercase the email in this case, users were no longer able to access their original isomer user and a new user was created for them instead.

In order to make our behaviour more consistent, we're opting to lowercase all references to email. We will also need to convert some existing isomer user entries - these users have multiple accounts mapped to the same email with a different casing, and we should ideally combine these users once this PR has been merged.

### Tests

- [ ] Create a fresh github user with no attached email user, and associate an email with capitalisation via the github login flow
  - [ ] Check that the user's associated email in the DB is in lowercase
- [ ] Go to a website with a different user in email login, and try to associate the email in the above step (but capitalised in places) - this should still work

### After merging
- [ ] Run through the list of user entries, and search for duplicate entries using ```SELECT id, email
FROM "users"
WHERE LOWER(TRIM(email)) IN 
(
SELECT LOWER(TRIM(email))
FROM "users"
GROUP BY LOWER(TRIM(email))
HAVING count(*) > 1
)``` - we should combine the users in this list and their associated sites

### DIFF
--- a/src/routes/formsgSiteCreation.ts
+++ b/src/routes/formsgSiteCreation.ts
@@ -89,6 +89,16 @@ export class FormsgRouter {
       }
       let foundOwner
       if (isEmailLogin) {
+        if (!ownerEmail) {
+          const err = `An owner email is required for email login`
+          await this.sendCreateError(
+            requesterEmail,
+            repoName,
+            submissionId,
+            err
+          )
+          return res.sendStatus(200)
+        }
         foundOwner = await this.usersService.findOrCreateByEmail(ownerEmail)
       }
       // 3. Use service to create site
--- a/src/routes/v2/authenticated/users.ts
+++ b/src/routes/v2/authenticated/users.ts
@@ -68,13 +68,14 @@ export class UsersRouter {
     const { email, otp } = req.body
     const { userSessionData } = res.locals
     const userId = userSessionData.isomerUserId
+    const parsedEmail = email.toLowerCase()
 
-    const isOtpValid = await this.usersService.verifyEmailOtp(email, otp)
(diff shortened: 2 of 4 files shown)


kishore03109 (approved):
[Non-nit, operations]: While I am ok with the code changes, I think its important to get a rough gauge of the number of 'duplicate' emails that we have to manually remediate. ie if there is 5 vs 50 that we would have to remediate right after release. 

raising this up as this PR would effectively make them lose access to repos that are tied to their capitalised repos right after this release, and we would prob remediate this quickly for these users right after release! 

(unless we have some script to run to pull users, delete them and tie their existing repos to one userid, then this concern is not valid)

alexanderleegs:
> [Non-nit, operations]: While I am ok with the code changes, I think its important to get a rough gauge of the number of 'duplicate' emails that we have to manually remediate. ie if there is 5 vs 50 that we would have to remediate right after release.
> 
> raising this up as this PR would effectively make them lose access to repos that are tied to their capitalised repos right after this release, and we would prob remediate this quickly for these users right after release!
> 
> (unless we have some script to run to pull users, delete them and tie their existing repos to one userid, then this concern is not valid)

There's 24 total - these users are already affected in the existing state! This is because our existing login flow automatically lowercases the user's entry, so any user tagged to an email with uppercase characters are currently already unable to log in with the already created user

PR #786, opened 2023-05-30
Fix: check if resources name is empty when creating new resource room

alexanderleegs:
This PR fixes an issue where sites attempting to create a new resource room will fail if their `config.yml` has a `resource_room` param, even if it is blank. An additional check has been added to ensure that the param exists.

### DIFF
--- a/src/services/directoryServices/ResourceRoomDirectoryService.js
+++ b/src/services/directoryServices/ResourceRoomDirectoryService.js
@@ -51,7 +51,7 @@ class ResourceRoomDirectoryService {
       sessionData
     )
     // If resource room already exists, throw error
-    if ("resources_name" in configContent)
+    if ("resources_name" in configContent && configContent.resources_name)
       throw new ConflictError("Resource room already exists")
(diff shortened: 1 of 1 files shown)


src/services/directoryServices/ResourceRoomDirectoryService.js
```diff
@@ -51,7 +51,7 @@ class ResourceRoomDirectoryService {
       sessionData
     )
     // If resource room already exists, throw error
-    if ("resources_name" in configContent)
+    if ("resources_name" in configContent && configContent.resources_name)
```
harishv7, line 54: is this to account for the case where `configContent.resources_name === ""`?

kishore03109 (approved):

src/services/directoryServices/ResourceRoomDirectoryService.js
alexanderleegs: yup it is! the scamshield repo was unable to create a resource room because of this

PR #787, opened 2023-05-31
chore(aws region): shift from const -> env var

seaerchin:
## Problem
`AWS_REGION` was previously referred to as either `const` in code or env var. this streamlines it to getting it from env vars (through convict). 

## Solution
1. use `config.get(aws.region)`
    - `logger` in `microservices` not migrated due to it drawing in imports

### DIFF
--- a/src/logger/cloudwatch.logger.ts
+++ b/src/logger/cloudwatch.logger.ts
@@ -3,7 +3,7 @@ import Bluebird from "bluebird"
 import winston from "winston"
 import WinstonCloudwatch from "winston-cloudwatch"
 
-import { config } from "@root/config/config"
+import { config } from "@config/config"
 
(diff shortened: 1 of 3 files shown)


src/logger/cloudwatch.logger.ts
```diff
@@ -3,6 +3,8 @@ import Bluebird from "bluebird"
 import winston from "winston"
 import WinstonCloudwatch from "winston-cloudwatch"
 
+import { config } from "@root/config/config"
```
harishv7, line 6: nit: can we import from @config/config?

kishore03109 (approved):

src/logger/cloudwatch.logger.ts
seaerchin: gg my auto add import failed me hais

PR #788, opened 2023-05-31
ci(ci): add build step

seaerchin:
## Problem
last time no `build` step so ts type errors can be ignored if no tests.

## Solution
1. add ci step for build
    - not doing `lint -> build -> test` but `lint -> test` and `lint -> build` because it's faster + we can test w/ failing builds if the tests don't touch the files that hte build fails on


kishore03109 (approved):
Thanks for this!

weird that we have this for FE but not BE

PR #792, opened 2023-06-07
feat: introduce types for isomer error

harishv7:
## Problem

This PR introduces the base types for Isomer Errors

Closes IS-210

## Solution

In contrast to the description of task, instead of overloading the logger's error method, we decided to treat logger and error distinctly

Current logger accept a Loggable which is a string or Record<string, unknown> which is a structured log format.

All errors and other types which can be logged will therefore provide a utility function that convert them to one of the above - string or a Record

Hence instead of the logger figuring out how to convert IsomerError and potentially other types in future, it will only accept fixed and predictable types

**Breaking Changes**

  - Details ...

### DIFF
--- a/src/errors/BaseError.ts
+++ b/src/errors/BaseError.ts
@@ -1,7 +1,11 @@
 import { IsomerInternalError } from "./IsomerError"
 
 class BaseIsomerError extends Error implements IsomerInternalError {
+  name: string
+
   status: number
+
+  message: string
 
   code = "BaseError"
 
--- a/src/errors/BaseError.ts
+++ b/src/errors/BaseError.ts
@@ -1,7 +1,11 @@
-class BaseIsomerError extends Error {
+import { IsomerInternalError } from "./IsomerError"
+
+class BaseIsomerError extends Error implements IsomerInternalError {
   status: number
 
-  isIsomerError: true
+  code = "BaseError"
+
+  isIsomerError = true
 
   constructor(status = 500, message = "Something went wrong") {
     super()
@@ -9,7 +13,6 @@ class BaseIsomerError extends Error {
     this.name = this.constructor.name
     this.status = status
     this.message = message
-    this.isIsomerError = true
   }
 }
 
--- a/src/errors/BaseError.ts
+++ b/src/errors/BaseError.ts
@@ -1,7 +1,7 @@
 class BaseIsomerError extends Error {
   status: number
 
-  isIsomerError: boolean
+  isIsomerError: true
 
   constructor(status = 500, message = "Something went wrong") {
     super()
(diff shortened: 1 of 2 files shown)


src/errors/BaseError.ts
```diff
@@ -1,14 +1,20 @@
 class BaseIsomerError extends Error {
-  constructor(status, message) {
+  status: number
+
+  isIsomerError: boolean
```
seaerchin, line 4:
```suggestion
  isIsomerError: true
```

src/errors/BaseError.ts
```diff
@@ -1,14 +1,20 @@
 class BaseIsomerError extends Error {
```
seaerchin, line 1: shouldn't this `implement IsomerInternalError`

src/errors/IsomerError.ts
```diff
@@ -0,0 +1,24 @@
+/* External representation of Isomer errors - passed to frontend client */
+export interface IsomerExternalError {
+  code: string
+  message: string
+}
+
+/* Internal representation of Isomer errors - passed to frontend client */
```
seaerchin, line 7:
```suggestion
/* Internal representation of Isomer errors - passed to log stream */
```

src/errors/BaseError.ts
kishore03109: hmm wait quick question, do we not want to benefit from inheritance here? why are we preferring `implement` over `extends` ah

harishv7: was planning to add in future pr. added now

src/errors/BaseError.ts
```diff
@@ -0,0 +1,23 @@
...
+    this.status = status
+    this.message = message
+  }
+}
+
+module.exports = {
+  BaseIsomerError,
+}
+
+export { BaseIsomerError }
```
kishore03109, line 27: Nit, ESLint: Prefer default export.

kishore03109 (commented):
Just for clarity, we would then need to have the same definition in the FE for the `InternalIsomerError` correct?

src/errors/BaseError.ts
seaerchin: @kishore03109 - we `extends` a `class` but `implement` an `interface` (which is what we have here)

src/errors/BaseError.ts
harishv7: sure i can update this in upcoming PRs

harishv7:
> Just for clarity, we would then need to have the same definition in the FE for the `InternalIsomerError` correct?

@kishore03109 FE only receives `IsomerExternalError` and yes, FE will maintain the same contract and hence definition of the external error type

seaerchin (approved):

kishore03109 (approved):

PR #799, opened 2023-06-14
fix(requestschema): change to allow for 16 digits for fb pixel

seaerchin:
## Problem
see [here](https://github.com/isomerpages/isomercms-frontend/pull/1306)

## Solution
1. change request schema

### DIFF
--- a/src/routes/v2/authenticatedSites/__tests__/Settings.spec.js
+++ b/src/routes/v2/authenticatedSites/__tests__/Settings.spec.js
@@ -113,5 +113,19 @@ describe("Settings Router", () => {
       expect(mockSettingsService.retrieveSettingsFiles).toHaveBeenCalled()
       expect(mockSettingsService.updateSettingsFiles).toHaveBeenCalled()
     })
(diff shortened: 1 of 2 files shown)


src/routes/v2/authenticatedSites/__tests__/Settings.spec.js
```diff
@@ -113,5 +113,19 @@ describe("Settings Router", () => {
       expect(mockSettingsService.retrieveSettingsFiles).toHaveBeenCalled()
       expect(mockSettingsService.updateSettingsFiles).toHaveBeenCalled()
     })
+    it("should allow for facebook pixel values of length 16", async () => {
```
kishore03109, line 116: Hmm if we were to add test cases, should we not add test cases for both cases (ie 15 + 16?)?

seaerchin: the default passing test is 15 digits alr

alexanderleegs (approved):

kishore03109 (approved):

PR #802, opened 2023-06-16
IS-212 - Migrate collection page service to new error format

harishv7:
## Problem

We are restructuring the error handling as per discussed RFC, starting with CollectionPageService

Closes IS-212

## Solution

This PR introduces named parameters for Base Isomer Error. Eventually all the errors should have this, but base error seemed to have the smallest surface area for a initial PR. But because of this change, all "child" classes of the base error will have a small change in their constructors where `super(...)` is called.

We are introducing new fields to the existing error classes - `meta`, `code`, `isV2err`, `componentCode` and `fileCode`.

The error handler is also doing a conditional check to convert the internal representation to external facing. For now, the external rep is just a combination of the `componentCode-fileCode-code`.

As the initial PR, I have modified CollectionPageService.js to this new format.

**Breaking Changes**

  - Details ...

## Tests

Run `npm run test` to ensure all tests are passing.

CollectionPageService is used within the workspace folder when you try to create or rename a folder. Example: http://localhost:3000/sites/{some-site}/folders/example-folder

The errors are thrown when the filename has special characters in them. However, these are already caught on FE. To simulate the BE response you need to copy a successful network request as a cURL and use Postman/Insomnia to import that and manually send a invalid file name in the body

Screenshot 2023-06-15 at 10 31 26 PM

The return response's code should be of format `componentCode-fileCode-code`.

In the event where a invalid filename request is sent from FE, returning the new external rep format will be handled by FE. As tested (by disabling validation on FE) - check this PR: https://github.com/isomerpages/isomercms-frontend/pull/1312

## Deploy notes
To deploy together with FE PR - https://github.com/isomerpages/isomercms-frontend/pull/1312

### DIFF
--- a/src/errors/BadRequestError.js
+++ b/src/errors/BadRequestError.js
@@ -2,8 +2,8 @@
 const { BaseIsomerError } = require("@errors/BaseError")
 
 class BadRequestError extends BaseIsomerError {
-  constructor(message) {
-    super(400, message)
+  constructor(message, meta = {}) {
+    super({ status: 400, code: "BadRequestError", message, meta })
   }
 }
 
--- a/src/errors/BadRequestError.js
+++ b/src/errors/BadRequestError.js
@@ -1,14 +1,14 @@
 // Import base error
 const { BaseIsomerError } = require("@errors/BaseError")
 
-const { ComponentTypes } = require("./IsomerError")
+const { ComponentTypes, FileCodes } = require("./IsomerError")
 
 class BadRequestError extends BaseIsomerError {
   constructor(
     message,
     meta = {},
     componentCode = ComponentTypes.Other,
-    fileCode = "000"
+    fileCode = FileCodes.Undefined
   ) {
     super({
       status: 400,
--- a/src/errors/BadRequestError.js
+++ b/src/errors/BadRequestError.js
@@ -1,9 +1,23 @@
 // Import base error
 const { BaseIsomerError } = require("@errors/BaseError")
 
+const { ComponentTypes } = require("./IsomerError")
+
 class BadRequestError extends BaseIsomerError {
-  constructor(message, meta = {}) {
-    super({ status: 400, code: "BadRequestError", message, meta })
+  constructor(
+    message,
+    meta = {},
+    componentCode = ComponentTypes.Other,
+    fileCode = "000"
+  ) {
+    super({
+      status: 400,
+      code: "BadRequestError",
+      message,
+      meta,
+      componentCode,
+      fileCode,
+    })
   }
 }
 
--- a/src/errors/BaseError.ts
+++ b/src/errors/BaseError.ts
@@ -1,4 +1,4 @@
-import { IsomerInternalError } from "./IsomerError"
+import { ComponentTypes, IsomerInternalError } from "./IsomerError"
 
 class BaseIsomerError extends Error implements IsomerInternalError {
   name: string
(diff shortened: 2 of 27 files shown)


src/errors/BadRequestError.js
```diff
@@ -1,9 +1,23 @@
...
+const { ComponentTypes, FileCodes } = require("./IsomerError")
+
 class BadRequestError extends BaseIsomerError {
-  constructor(message) {
-    super(400, message)
+  constructor(
+    message,
+    meta = {},
+    componentCode = ComponentTypes.Other,
+    fileCode = FileCodes.Undefined
```
seaerchin, line 11: as mentioned earlier, will probably shift these into a separate interface and get `BadRequestError` to implement that

src/services/fileServices/MdPageServices/PageService.ts
```diff
@@ -44,13 +44,13 @@ const FRONTEND_URL = process.env.FRONTEND_URL || "http://localhost:3000"
...
 const withErrorHandler = (promise: Promise<PageInfo>) =>
-  ResultAsync.fromPromise(promise, () => new BaseIsomerError()).andThen(
+  ResultAsync.fromPromise(promise, () => new BaseIsomerError({})).andThen(
     (data) =>
       // NOTE: This is a fail-safe check as `yaml.parse`
       // doesn't guarantee existence.
       data?.content?.frontMatter?.permalink
         ? ok(data)
-        : err(new BaseIsomerError())
+        : err(new BaseIsomerError({}))
```
seaerchin, line 53: actually maybe we want to set the initial argument as `Object | undefined` if we have defaults for them/they are safe to omit

src/errors/BaseError.ts
```diff
@@ -7,16 +7,35 @@ class BaseIsomerError extends Error implements IsomerInternalError {
...
+
+  fileCode: string // a selection from FileCodes
+
+  constructor({
+    status = 500,
+    code = "BaseError",
+    message = "Something went wrong",
+    meta = {},
+    componentCode = ComponentTypes.Other,
+    fileCode = "000",
```
seaerchin, line 28: as discussed offline, shifting this away from `BaseError` so non-v2 errors don't inherit this

src/errors/BaseError.ts
```diff
@@ -7,16 +7,35 @@ class BaseIsomerError extends Error implements IsomerInternalError {
...
+  }) {
     super()
     Error.captureStackTrace(this, this.constructor)
     this.name = this.constructor.name
+    this.code = code
```
seaerchin, line 33: fyi, in the downstream, this was changed to

`this.code = this.constructor.name` - seems to be identical to what we have at present

seaerchin (approved):
lgtm; maybe want to consider giving `undefined` as arg to `BaseError` cos no big diff between `BaseError()` and `BaseError({})`

src/errors/BaseError.ts
harishv7: agreed!

src/errors/BadRequestError.js
harishv7: yes agreed

src/services/fileServices/MdPageServices/PageService.ts
harishv7: ideally would prefer this to be explicit as this is a new format introduced

src/errors/BaseError.ts
harishv7: 👍🏼

PR #803, opened 2023-06-16
fix(email): log mail status after reasonable delay

caando:
## Problem

Postman API send mail post request to endpoint `/transactional/email/send` returns status code 201 indicating the request has been created, but does not guarantee that the email will be sent and delivered. 

Closes IS-135

## Solution

After a reasonable delay, verify with Postman that the mail has been delivered. If otherwise, log the status of the email to be tracked. 

**Improvements**:

- After a 1 minute delay, send a get request to endpoint `/transactional/email/{id}` for the status of the sent email.
- Log the status of the email. If the email has yet to be delivered, log the full email data from Postman.

**Notes**:
This change will cause `Users.spec.ts` to encounter warnings for remaining asynchronous operations after the tests are complete. This is because `MailClient.ts` has a separate routine for verification of email statuses which has a delay of 1 minute. Waiting for that long will cause the tests to timeout. useFakeTimers() have been experimented with, but behaviour seem to be buggy and cause all tests to fail. Hence, am currently sticking with the warning and letting the tests pass.

### DIFF
--- a/src/services/utilServices/MailClient.ts
+++ b/src/services/utilServices/MailClient.ts
@@ -10,7 +10,7 @@ const MAIL_VERIFICATION_DELAY = 60000 // 60 seconds
 type MailData = {
   id: string
   recipient: string
-  status: string
+  status: "UNSENT" | "ACCEPTED" | "SENT" | "BOUNCED" | "DELIVERED" | "OPENED"
 }
 
 class MailClient {
@@ -48,9 +48,9 @@ class MailClient {
   }
 
   async verifyMail(sendMailData: MailData): Promise<void> {
-    const verfyEndpoint = `${POSTMAN_API_URL}/transactional/email/${sendMailData.id}`
+    const verifyEndpoint = `${POSTMAN_API_URL}/transactional/email/${sendMailData.id}`
     await new Promise((res) => setTimeout(res, MAIL_VERIFICATION_DELAY))
-    const verifyMailResponse = await axios.get<MailData>(verfyEndpoint, {
+    const verifyMailResponse = await axios.get<MailData>(verifyEndpoint, {
       headers: {
         Authorization: `Bearer ${this.POSTMAN_API_KEY}`,
       },
@@ -60,7 +60,12 @@ class MailClient {
 
   logEmailStatus(verifyMailData: MailData) {
     const mailStatus = verifyMailData.status
+    let unknownMailStatus: never
     switch (mailStatus) {
+      case "DELIVERED":
+      case "OPENED":
+        logger.info(`Email delivered to ${verifyMailData.recipient}`)
+        break
       case "UNSENT":
         logger.error(
           `Email to ${verifyMailData.recipient} not accepted: ${verifyMailData}`
@@ -82,7 +87,10 @@ class MailClient {
         )
         break
       default:
-        logger.info(`Email delivered to ${verifyMailData.recipient}`)
+        unknownMailStatus = mailStatus
+        logger.error(
+          `Email to ${verifyMailData.recipient} encounter unknown status: ${unknownMailStatus}`
+        )
         break
     }
   }
--- a/src/services/utilServices/MailClient.ts
+++ b/src/services/utilServices/MailClient.ts
@@ -7,11 +7,25 @@ import logger from "@logger/logger"
 const POSTMAN_API_URL = "https://api.postman.gov.sg/v1"
 const MAIL_VERIFICATION_DELAY = 60000 // 60 seconds
 
+const MailStatus = {
+  Unsent: "UNSENT",
+  Accepted: "ACCEPTED",
+  Sent: "SENT",
+  Bounced: "BOUNCED",
+  Delivered: "DELIVERED",
+  Opened: "OPENED",
+} as const
+
+type MailStatusKeys = typeof MailStatus[keyof typeof MailStatus]
+
 type MailData = {
   id: string
   recipient: string
-  status: "UNSENT" | "ACCEPTED" | "SENT" | "BOUNCED" | "DELIVERED" | "OPENED"
+  status: MailStatusKeys
+  error_code: string
 }
+
+const BLACKLISTED_RECIPIENT_ERROR_CODE = "Error 400: Blacklisted recipient"
 
 class MailClient {
   POSTMAN_API_KEY: string
@@ -67,26 +81,32 @@ class MailClient {
     const mailStatus = verifyMailData.status
     let unknownMailStatus: never
     switch (mailStatus) {
-      case "DELIVERED":
-      case "OPENED":
+      case MailStatus.Delivered:
+      case MailStatus.Opened:
         logger.info(`Email delivered to ${verifyMailData.recipient}`)
         break
-      case "UNSENT":
-        logger.error(
-          `Email to ${verifyMailData.recipient} not accepted: ${verifyMailData}`
-        )
+      case MailStatus.Unsent:
+        if (verifyMailData.error_code === BLACKLISTED_RECIPIENT_ERROR_CODE) {
+          logger.error(
+            `Email to blacklisted recipient ${verifyMailData.recipient} not accepted: ${verifyMailData}`
+          )
+        } else {
+          logger.error(
+            `Email to ${verifyMailData.recipient} not accepted: ${verifyMailData}`
+          )
+        }
         break
-      case "ACCEPTED":
+      case MailStatus.Accepted:
         logger.error(
           `Email to ${verifyMailData.recipient} not sent: ${verifyMailData}`
         )
         break
-      case "SENT":
+      case MailStatus.Sent:
         logger.error(
           `Email to ${verifyMailData.recipient} not delivered: ${verifyMailData}`
         )
         break
-      case "BOUNCED":
+      case MailStatus.Bounced:
         logger.error(
           `Email to ${verifyMailData.recipient} rejected by recipient's mail server: ${verifyMailData}`
         )
--- a/src/services/utilServices/MailClient.ts
+++ b/src/services/utilServices/MailClient.ts
@@ -79,7 +79,6 @@ class MailClient {
 
   logEmailStatus(verifyMailData: MailData) {
     const mailStatus = verifyMailData.status
-    let unknownMailStatus: never
     switch (mailStatus) {
       case MailStatus.Delivered:
       case MailStatus.Opened:
@@ -111,12 +110,13 @@ class MailClient {
           `Email to ${verifyMailData.recipient} rejected by recipient's mail server: ${verifyMailData}`
         )
         break
-      default:
-        unknownMailStatus = mailStatus
+      default: {
+        const unknownMailStatus: never = mailStatus
         logger.warn(
           `Email to ${verifyMailData.recipient} encounter unknown status: ${unknownMailStatus}`
         )
         break
+      }
     }
   }
 }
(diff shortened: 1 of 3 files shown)


src/services/utilServices/MailClient.ts
```diff
@@ -5,6 +5,13 @@ import { config } from "@config/config"
 import logger from "@logger/logger"
 
 const POSTMAN_API_URL = "https://api.postman.gov.sg/v1"
+const MAIL_VERIFICATION_DELAY = 60000 // 60 seconds
+
+type MailData = {
+  id: string
+  recipient: string
+  status: string
```
seaerchin, line 13: nit: this should be a sum type - `UNSENT | BOUNCE | (rest of status)`

src/services/utilServices/MailClient.ts
```diff
@@ -28,16 +35,57 @@ class MailClient {
...
       })
+      this.verifyMail(sendMailResponse.data)
     } catch (err) {
       logger.error(`Error occurred when sending email to ${recipient}: ${err}`)
       throw new Error("Failed to send email.")
     }
   }
+
+  async verifyMail(sendMailData: MailData): Promise<void> {
+    const verfyEndpoint = `${POSTMAN_API_URL}/transactional/email/${sendMailData.id}`
```
seaerchin, line 51:
```suggestion
    const verifyEndpoint = `${POSTMAN_API_URL}/transactional/email/${sendMailData.id}`
```

src/services/utilServices/MailClient.ts
```diff
@@ -28,16 +35,57 @@ class MailClient {
...
+        break
+      case "SENT":
+        logger.error(
+          `Email to ${verifyMailData.recipient} not delivered: ${verifyMailData}`
+        )
+        break
+      case "BOUNCED":
+        logger.error(
+          `Email to ${verifyMailData.recipient} rejected by recipient's mail server: ${verifyMailData}`
+        )
```
seaerchin, line 82: why is accepted/sent an `error`?

src/services/utilServices/MailClient.ts
```diff
@@ -28,16 +35,57 @@ class MailClient {
...
+        break
+      default:
+        logger.info(`Email delivered to ${verifyMailData.recipient}`)
+        break
```
seaerchin, line 86: we can use a `never` variable if we go the sum type route: 
```ts
const err: never = mailStatus
throw new err(...)
```

this will get caught at compile time as `never` cannot be assigned to a concrete type - we're making use of a trick here to get it

src/services/utilServices/__tests__/MailClient.spec.ts
```diff
@@ -10,7 +10,7 @@ import {
 } from "@fixtures/identity"
 import _MailClient from "@services/utilServices/MailClient"
 
-const mockEndpoint = "https://api.postman.gov.sg/v1/transactional/email/send"
+const mockEndpoint = "https://api.postman.gov.sg/v1/transactional/email/"
```
seaerchin, line 13: doesn't this actually hit the endpoint?

src/services/utilServices/__tests__/MailClient.spec.ts
seaerchin, line 1: can we add 1 more spec to check that throwing in `verifyMail` doesn't cause the main thread to die?

src/services/utilServices/MailClient.ts
caando: `ACCEPTED` means the email is not sent. Sent does not guarantee that the email has been successfully delivered. The ideal state after 1 min should be `DELIVERED` or `OPENED` as `SENT` suggests that the email is invalid and there is no email server handling the email.

src/services/utilServices/MailClient.ts
caando: will fix 👍

src/services/utilServices/MailClient.ts
caando: will do 👍

src/services/utilServices/__tests__/MailClient.spec.ts
caando: There are now multiple endpoints `https://api.postman.gov.sg/v1/transactional/email/send` and `https://api.postman.gov.sg/v1/transactional/email/send/{id}`. I will be appending `send` and `{id}` when appropriate

src/services/utilServices/__tests__/MailClient.spec.ts
caando: okay, will do 👍

src/services/utilServices/MailClient.ts
caando: okay, listing all valid status in switch and changing default to unknown status.

src/services/utilServices/MailClient.ts
```diff
@@ -28,16 +35,57 @@ class MailClient {
     }
 
     try {
-      await axios.post(endpoint, email, {
+      const sendMailResponse = await axios.post<MailData>(sendEndpoint, email, {
         headers: {
           Authorization: `Bearer ${this.POSTMAN_API_KEY}`,
         },
       })
+      this.verifyMail(sendMailResponse.data)
```
kishore03109, line 43: Should we not await this for error handling purposes?

kishore03109 (commented):
From my previous convo with Postman, they mentioned that they will tell us if an email is blacklisted in the first API call itself. Can you also add a logger to state `xxx@blah.gov.sg is blacklisted`? this way it would be easier for the oncall to quickly tell if an email has been blacklisted or not! 

CICD is also failing, please do add the necessary test cases to capture this new flow accurately

src/services/utilServices/__tests__/MailClient.spec.ts
seaerchin: no, what i mean is if this is the actual endpoint and we fail to `mockResolvedValueOnce` (or `mockXxx`) correctly, it will actually hit the network, no?

src/services/utilServices/MailClient.ts
caando: Due to the delay being set to 1 minute, and that sendMail is used in `formsgSiteCreation.ts`, `formsgSiteLaunch.ts`, `UsersServices.ts` and `InfraServices.ts` for otps and other emails with await calls, I am not sure if awaiting for 1 minute will delay any important response from being sent. In terms of error handling, since the response for those errors would be to log them, I think there is no need to elevate the error. Wdyt?

src/services/utilServices/__tests__/MailClient.spec.ts
caando: I think axios.post is a mocked function so it should not send any requests out. `mockEndpoint` is used to check if `expect(mockAxios.post).toHaveBeenCalledWith` has been called with the right endpoints: `${mockEndpoint}send` and `${mockEndpoint}1`. I'm not sure if this is what you meant?

caando:
@kishore03109 
> From my previous convo with Postman, they mentioned that they will tell us if an email is blacklisted in the first API call itself. Can you also add a logger to state `xxx@blah.gov.sg is blacklisted`? this way it would be easier for the oncall to quickly tell if an email has been blacklisted or not!

>However, there are two ways you will be informed if a recipient is on the blacklist:
>1. In our current setup, during your API call to our system, we will check if the recipient is on the blacklist. If the recipient is on the blacklist, we will return an error message indicating that the recipient is on the blacklist. However, to scale our system, we plan to change this behavior in the future, so we advise against relying on this behavior.
>2. You can query an email by its ID using this [API endpoint](https://github.com/api-guide/programmatic-email-api/get-email-by-id-api). If the email was not sent because the recipient was on the blacklist, this will be indicated in the errorCode field of the response.

It seems that they are intending to remove the feedback from first post api call. I will instead check for the errorCode on verification. What are your opinions on this?

src/services/utilServices/MailClient.ts
```diff
@@ -5,6 +5,13 @@ import { config } from "@config/config"
 import logger from "@logger/logger"
 
 const POSTMAN_API_URL = "https://api.postman.gov.sg/v1"
+const MAIL_VERIFICATION_DELAY = 60000 // 60 seconds
+
+type MailData = {
+  id: string
+  recipient: string
+  status: "UNSENT" | "ACCEPTED" | "SENT" | "BOUNCED" | "DELIVERED" | "OPENED"
```
harishv7, line 13: can abstract to const MailDataStatus = { Unsent: "UNSENT"...}

harishv7: can further abstract to something like:

```
const Roles = {
  Admin: "admin",
  Writer: "writer",
  Reader: "reader"
} as const;

// Convert object key in a type
type RoleKeys = typeof Roles[keyof typeof Roles]

declare function hasAccess(role: RoleKeys): void;
```
https://dev.to/ivanzm123/dont-use-enums-in-typescript-they-are-very-dangerous-57bh

src/services/utilServices/MailClient.ts
harishv7: yea we should not be awaiting here as it may block other operations and the response from `sendMail` will be pending

src/services/utilServices/__tests__/MailClient.spec.ts
seaerchin: ^oops u right

src/services/utilServices/MailClient.ts
seaerchin: supernit: can we don't use `logger.error` for those but rather `warn`? ie, we use `error` only for end error states and for transitions, we just `warn`. this isn't really a big deal but just a good to have so that in the event where we crawl through logs, we can sift through less

src/services/utilServices/MailClient.ts
caando: done!

src/services/utilServices/MailClient.ts
caando: okay, changing to warning instead

seaerchin:
could you rebase on top of develop/merge develop into this branch? got extra + conflicts

src/services/utilServices/__tests__/MailClient.spec.ts
```diff
@@ -24,10 +24,30 @@ const generateEmail = (recipient: string, subject: string, body: string) => ({
 
 describe("Mail Client", () => {
   afterEach(() => mockAxios.reset())
+  jest.useFakeTimers()
```
kishore03109, line 27: Hey I am confused here, are we still intending to use `jest.useFakeTimers()`? or is this buggy to the point where we shouldnt use it? 
Sanity check here: should we use this after `sendMailResponse` rather than for all test methods?

kishore03109: the CICD passes with the use of `jest.useFakeTimers()` here, do check in if this behaviour is still buggy ya

caando: Every test in `Users.spec.ts` involving MailClient except for 1 will freeze and timeout upon `jest.useFakeTimers()`.

src/services/utilServices/MailClient.ts
```diff
@@ -28,16 +49,76 @@ class MailClient {
...
+    // and will return undefined for verifyMailResponse. In order to pass the test, we
+    // need to check if it is a valid response. This should not happen in application.
+    if (verifyMailResponse) {
+      this.logEmailStatus(verifyMailResponse.data)
+    }
+  }
+
+  logEmailStatus(verifyMailData: MailData) {
+    const mailStatus = verifyMailData.status
+    let unknownMailStatus: never
```
seaerchin, line 82: nit: assign in line in the default case for locality of information

seaerchin (approved):

src/services/utilServices/MailClient.ts
caando: Will have eslint warning for declaration in switch case

caando: will result in eslint warning for declaration of variable in switch case

PR #804, opened 2023-06-19
refactor(errors): migrate from js to ts

seaerchin:
## Problem
This PR migrates errors from js -> ts, under the new proposal (see [here](https://www.notion.so/opengov/Structuring-Errors-in-Isomer-75cd8adf68e443618e4fcd4e7ede18a5?pvs=4)). 

Closes IS-211

## Solution
1. add `code` property to the base error - will be inherited by our errors so it's less onerous on devs
2. change errors from js -> ts 
    - this was done by renaming the file, then changing the `super` call to take an object. only `BadRequestError` has a change
3. add `LegacyErrors` - should use as a marker. when can't find in codebase, we know migration is done
4. stricter type enforcement on our error types using template literal types + `const obj` - we want devs to go through our constants to get the values or pass **only the correct values**. if we don't use `as const`/template literal, this enforcement isn't done as the type is `string`
5. add a very simple test spec for errors - just testing the `code` property works as expected


src/errors/BadRequestError.ts
```diff
@@ -0,0 +1,30 @@
...
+  extends BaseIsomerError
+  implements IdentifiableError {
+  componentCode: PossibleComponentTypes
+
+  fileCode: PossibleFileCodes
+
+  constructor(
+    // NOTE: This is technically optional but we want to
+    // preserve argument ordering for back-compat
+    // with our JS files.
```
kishore03109, line 18: Hmm I dont understand this comment, what does the argument being optional relate to argument ordering here ya?

src/errors/BaseError.ts
```diff
@@ -15,32 +15,16 @@ class BaseIsomerError extends Error implements IsomerInternalError {
...
-    componentCode = ComponentTypes.Other,
-    fileCode = "000",
-  }) {
+  constructor({ status = 500, message = "Something went wrong", meta = {} }) {
     super()
     Error.captureStackTrace(this, this.constructor)
     this.name = this.constructor.name
-    this.code = code
     this.status = status
+    this.code = this.name
```
kishore03109, line 23: if `this.code` = `this.name`, why do we have 2 variables here ah? 

[nit]: could we rename `code` to `errorType` instead?

src/errors/IsomerError.ts
```diff
@@ -20,26 +7,75 @@ export const ComponentTypes = {
...
     ...error,
   }),
+  isIdentifiable: <E extends BaseIsomerError>(
+    err: E
+  ): err is E & IdentifiableError =>
+    ((err as unknown) as IdentifiableError).componentCode !== undefined &&
+    ((err as unknown) as IdentifiableError).fileCode !== undefined,
+}
+
+interface LegacyRepresentation {
```
kishore03109, line 65:
```suggestion
interface LegacyErrorRepresentationInputs {
```

src/errors/BadRequestError.ts
seaerchin: oops, this comment was written when i set `message` to a default arg. will remove

src/errors/BaseError.ts
seaerchin: tbh, this is just to keep parity with the doc - because implementing engineers will (probably) refer to the doc, i opted to keep the `code` argument. honestly, not a big deal - i'm open to removing it and we refer to it as `name` instead.

re: `code` vs `errorType` - sure, wdyt about just keeping as `name`? i think this avoids the most important issue, which is conflation of properties between teh internal/external rep

src/errors/IsomerError.ts
seaerchin: would prefer to have it as `LegacyErrorRepresentation` - this is not really an `Input` more an interface that existing errors fulfill.

src/errors/BaseError.ts
kishore03109: ye prefer to change doc when needed than keep two variables that does the same thing 

> re: code vs errorType - sure, wdyt about just keeping as name? i think this avoids the most important issue, which is conflation of properties between teh internal/external rep
sure

kishore03109 (approved):

PR #806, opened 2023-06-21
Feat/is 21 repo privatisation

alexanderleegs:
## Problem

This PR introduces a repo privatisation feature. To be reviewed in conjunction with PR #[1316](https://github.com/isomerpages/isomercms-frontend/pull/1316) on the isomercms-frontend repo.

This PR adds new fields to the `sites` and `deployments` table - `isPrivate` and `encryptedPassword`, `encryptionIv`, `passwordDate` respectively. The new endpoints have been added to the settings router and related services, with the deployment modification logic added to `DeploymentsService`.

## Tests
See frontend for related tests - for private repos, the `is_private` param for the `sites` entry should be `true` and the `encrypted_password`, `encryption_iv` and `password_date`  in `deployments` should contain the appropriate entries

## Deploy Notes

**Migrations will need to be run on staging/prod before merging this PR.**

- `SITE_PASSWORD_SECRET_KEY` : Encryption key for stored passwords, should match the equivalent key on the backend
- `NETLIFY_ACCESS_TOKEN`: this is the access token we use to access our account. We should be using the one generated for isomeradmin


src/database/migrations/20230615070344-add-password-and-iv.js
```diff
@@ -0,0 +1,58 @@
...
+          }
+        ),
+        queryInterface.addColumn(
+          "deployments", // name of Source model
+          "password_date", // name of column we're adding
+          {
+            type: Sequelize.DATE,
+            allowNull: true,
+            transaction: t,
+          }
```
seaerchin, line 31: could i get more information/context on what `encryption_iv + password_date` implies? i was under the impression that we'd only need `encrypted_password`?

src/services/configServices/SettingsService.js
```diff
@@ -116,6 +154,37 @@ class SettingsService {
...
+      if (privatiseRepoRes.isErr()) return privatiseRepoRes
+      try {
+        await this.sitesService.update({
+          id,
+          isPrivate: enablePassword,
+        })
+      } catch (err) {
+        return errAsync(err)
+      }
+    }
```
seaerchin, line 179: just wanted to note that we don't need the `try/catch` cos we still throw upstream :sadge:

src/services/configServices/SettingsService.js
```diff
@@ -43,6 +50,37 @@ class SettingsService {
...
+      return okAsync({
+        encryptedPassword: "",
+        iv: "",
+        isAmplifySite: true,
+      })
+
+    const deploymentInfo = await this.deploymentsService.getDeploymentInfoFromSiteId(
+      id
+    )
+    if (deploymentInfo.isErr()) return deploymentInfo
```
kishore03109, line 75: Sanity check here, shouldnt we reuturn some sort of error here instead?

src/routes/v2/authenticatedSites/settings.js
```diff
@@ -77,11 +80,58 @@ class SettingsRouter {
...
 
+  async getRepoPassword(req, res, next) {
+    const { userWithSiteSessionData } = res.locals
+
+    const passwordRes = await this.settingsService.getEncryptedPassword(
+      userWithSiteSessionData
+    )
+
+    if (passwordRes.isErr()) {
+      throw passwordRes.error
```
kishore03109, line 91: sanity check here: should we return some sort of 404 rather than thorwing an error here?

src/routes/v2/authenticatedSites/settings.js
```diff
@@ -77,11 +80,58 @@ class SettingsRouter {
...
+
+    return res.status(200).send(passwordRes.value)
+  }
+
+  async updateRepoPassword(req, res, next) {
+    const { body } = req
+    const { userWithSiteSessionData } = res.locals
+
+    const { error } = UpdateRepoPasswordRequestSchema.validate(req.body)
+    if (error) throw new BadRequestError(error.message)
```
kishore03109, line 102: Similar comment as above, should we be returning 404 rather than throwing here?

src/services/identity/DeploymentClient.ts
```diff
@@ -61,6 +64,11 @@ class DeploymentClient {
       this.amplifyClient.send(new CreateBranchCommand(options))
     ) as ResultAsync<CreateBranchCommandOutput, AmplifyError>
 
+  sendUpdateApp = (options: UpdateBranchCommandInput) =>
+    wrap(
+      this.amplifyClient.send(new UpdateBranchCommand(options))
+    ) as ResultAsync<UpdateBranchCommandOutput, AmplifyError>
```
seaerchin, line 70:
```suggestion
    wrap(this.amplifyClient.send(new CreateAppCommand(options)))

  sendCreateBranch = (options: CreateBranchCommandInput) =>
    wrap(this.amplifyClient.send(new CreateBranchCommand(options)))

  sendUpdateApp = (options: UpdateBranchCommandInput) =>
    wrap(this.amplifyClient.send(new UpdateBranchCommand(options)))
```

src/utils/crypto-utils.ts
```diff
@@ -0,0 +1,20 @@
...
+    config.get("aws.amplify.passwordSecretKey"),
+    "hex"
+  )
+  const decipher = createDecipheriv(
+    ALGORITHM,
+    SECRET_KEY,
+    Buffer.from(iv, "hex")
+  )
+  let decrypted = decipher.update(encryptedPassword, "hex", "utf8")
+  decrypted += decipher.final("utf8")
```
kishore03109, line 18: why do we need to do an update + final here ah?

src/services/identity/DeploymentClient.ts
seaerchin: just need to change
```ts
const wrap = <T>(promise: Promise<T>)
```

src/services/identity/DeploymentsService.ts
```diff
@@ -115,6 +118,83 @@ class DeploymentsService {
           .map(() => amplifyInfo)
       })
   }
+
+  getDeploymentInfoFromSiteId = async (repoId: string) => {
```
kishore03109, line 122: could we rename `repoId` to `siteId`? we have a `repos` and a `sites` table, might be better to make a distinction here

src/services/identity/DeploymentClient.ts
```diff
@@ -88,6 +96,28 @@ class DeploymentClient {
...
   })
+
+  generateUpdatePasswordInput = (
+    appId: string,
+    password: string
+  ): UpdateBranchCommandInput => {
+    if (password) {
+      return {
+        appId,
+        branchName: "staging",
```
seaerchin, line 107: nit: prefer to have these shifted out cos they remain teh same

src/services/identity/DeploymentsService.ts
```diff
@@ -115,6 +118,83 @@ class DeploymentsService {
...
+    return okAsync(deploymentInfo)
+  }
+
+  updateAmplifyPassword = async (
+    repoName: string,
+    encryptedPassword: string,
+    iv: string,
+    enablePassword: boolean
+  ) => {
+    const deploymentInfo = await this.repository.findOne({
```
kishore03109, line 139: hey, this can be made simpler right?
eg:
```
    const deploymentInfo = await this.repository.findOne({
      where: {
        name: repoName,
      },
    })
```

src/services/identity/DeploymentsService.ts
```diff
@@ -115,6 +118,83 @@ class DeploymentsService {
...
+            },
+          ],
+        },
+      ],
+    })
+    if (!deploymentInfo)
+      return errAsync(`Deployment for ${repoName} does not exist`)
+    const { id, hostingId: appId } = deploymentInfo
+    let updateAppInput
+    if (!enablePassword) {
```
seaerchin, line 160: could we just branch here and hide the rest of the logic into 2 separate methods? 

very confusing to figure out where/what is used, especially because the purpose of the two is diametrically opposed (1 sets password, 1 removes)

src/utils/crypto-utils.ts
```diff
@@ -0,0 +1,20 @@
+import { createDecipheriv } from "crypto"
+
+import { config } from "@config/config"
+
+const ALGORITHM = "aes-256-cbc"
```
seaerchin, line 5:
```suggestion
const DECRYPTION_ALGORITHM = "aes-256-cbc"
```

src/services/identity/DeploymentsService.ts
kishore03109, line 160: clarification, why is this called `enablePassword` ah? wouldnt this function only be called iff the repo has enabled password in the first place? 
do you mean `defaultPassword`?

kishore03109: or is this trying to __**remove**__ existing password?

src/services/identity/DeploymentClient.ts
```diff
@@ -88,6 +96,28 @@ class DeploymentClient {
       JEKYLL_ENV: branchName === "master" ? "production" : "staging",
     },
   })
+
+  generateUpdatePasswordInput = (
```
kishore03109, line 100: nit: could we separate out the concerns of with two functions of
`generateUpdatePasswordInput`
and 
`generateDeletePasswordInput`?

src/utils/crypto-utils.ts
```diff
@@ -0,0 +1,20 @@
...
+    "hex"
+  )
+  const decipher = createDecipheriv(
+    ALGORITHM,
+    SECRET_KEY,
+    Buffer.from(iv, "hex")
+  )
+  let decrypted = decipher.update(encryptedPassword, "hex", "utf8")
+  decrypted += decipher.final("utf8")
+  return decrypted
```
seaerchin, line 19: sorry, but this is abit confusing to me - it seems to me that we are decrypting first and thereafter, converting from `hex` to `utf8` before calling a finalizer of `utf8` and appending it to the updated value (in effect, `decrypted === decipher<utf8> * 2)` isn't it?)  

am i misunderstanding something? shouldn't it be either 

```ts
// update -> final
decipher.update(...)
return decipher.final(...)

// just return straight
return decipher.final(...)
```

why have a repeated output string?

src/utils/crypto-utils.ts
```diff
@@ -0,0 +1,20 @@
+import { createDecipheriv } from "crypto"
+
+import { config } from "@config/config"
+
+const ALGORITHM = "aes-256-cbc"
+
+export const decryptPassword = (encryptedPassword: string, iv: string) => {
```
seaerchin, line 7: **highly** suggest typing + branding the return and enforcing this on downstream callers. doesn't have to be here if the logic here is correct.

```ts
// type-def of decrypted password, prevents assigning string to it
type DecryptedPassword = Brand<string, "DecryptedPassword">

export const decryptPassword = (): DecryptedPassword
```

src/validators/RequestSchema.js
```diff
@@ -275,6 +275,12 @@ const UpdateSettingsRequestSchema = Joi.object().keys({
   logo: Joi.string().allow(""),
 })
 
+const UpdateRepoPasswordRequestSchema = Joi.object().keys({
+  encryptedPassword: Joi.string(),
+  iv: Joi.string(),
+  enablePassword: Joi.boolean(),
+})
```
seaerchin, line 282: is this all or nothing? ie, do we accept partial inputs like `encryptedPassword + iv` only

seaerchin:
some comments; also, should we update 1pw?

seaerchin (commented):
some comments; also, should we update 1pw env vars?

src/utils/crypto-utils.ts
alexanderleegs: This is the usage as given by the [documentation](https://nodejs.org/docs/latest-v18.x/api/crypto.html#class-decipher) - i'm not sure what to call the intermediate steps either tbh, so I used their naming. But .update alone doesn't actually provide the final decrypted string, and .final is needed to cut off the decipher usage, but doesn't actually provide any decryption

src/validators/RequestSchema.js
alexanderleegs: No partial input - always specify whether we're enabling or disabling the password

src/services/configServices/SettingsService.js
alexanderleegs: This returns the wrapped error from `getDeploymentInfoFromSiteId` - we're not modifying it here

src/routes/v2/authenticatedSites/settings.js
alexanderleegs: Hmm this is thrown so that the error can be handled by our error handler instead - should we be catching individual errors here to return a specific error?

src/database/migrations/20230615070344-add-password-and-iv.js
alexanderleegs: `encryption_iv` is the initialisation vector used for encryption - this is needed so that similar patterns of text when encrypted don't give rise to similar ciphertexts! This information needs to be stored to retrieve the original unencrypted value

`password_date` is an additional field which will be helpful for a future feature - we eventually want to be able to tell when a repo password has not been changed for a certain amount of time, so this is useful for that, though it's currently unused

src/services/configServices/SettingsService.js
```diff
@@ -43,6 +50,37 @@ class SettingsService {
     }
   }
 
+  async getEncryptedPassword(sessionData) {
```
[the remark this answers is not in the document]
alexanderleegs: It's a js file, I don't think we can enforce the return type :(

src/services/identity/DeploymentsService.ts
alexanderleegs: repoName only exists in `Repo` though, so we need to do the joins!

src/services/identity/DeploymentsService.ts
alexanderleegs: Hmm it's meant to be a boolean indicating whether the site contains a password or not - i'm open to what we can rename this to make this clearer!

src/routes/v2/authenticatedSites/settings.js
```diff
@@ -77,11 +80,58 @@ class SettingsRouter {
     return next()
   }
 
+  async getRepoPassword(req, res, next) {
```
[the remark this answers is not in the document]
alexanderleegs: https://github.com/isomerpages/isomercms-backend/pull/806/commits/66ed5a9584195cd3a206713d4b847c65caef43e0

src/services/identity/DeploymentClient.ts
alexanderleegs: Split into 2 methods following kishore's request: https://github.com/isomerpages/isomercms-backend/pull/806/commits/592ab87a469d24dd1564872dd7d79bfba4ee5c6b

src/services/identity/DeploymentClient.ts
alexanderleegs: Sure, changed in https://github.com/isomerpages/isomercms-backend/pull/806/commits/592ab87a469d24dd1564872dd7d79bfba4ee5c6b

src/services/identity/DeploymentsService.ts
alexanderleegs: Yeah that makes sense, https://github.com/isomerpages/isomercms-backend/pull/806/commits/f13e3f214b5fcc692fdc2f8d486959f52d0c2fa5

src/utils/crypto-utils.ts
alexanderleegs: changed in https://github.com/isomerpages/isomercms-backend/pull/806/commits/d1f2659caecc368efa5bca1ee9f6389486359533!

src/config/config.ts
```diff
@@ -171,6 +171,12 @@ const config = convict({
         format: "required-boolean",
         default: true,
       },
+      passwordSecretKey: {
+        doc: "Secret key used to encrypt password",
+        env: "SITE_PASSWORD_SECRET_KEY",
+        format: String,
+        default: "",
```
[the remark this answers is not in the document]
alexanderleegs: Good catch, https://github.com/isomerpages/isomercms-backend/pull/806/commits/0c5622f80b20899ea829d9f79e186b9a5bc8c6b0

src/services/identity/DeploymentClient.ts
alexanderleegs: Actually I've swapped this back - wrapping alone doesn't seem to provide us with the correct return type, which is breaking some existing functionality

src/utils/crypto-utils.ts
alexanderleegs: Updated in https://github.com/isomerpages/isomercms-backend/pull/806/commits/cea7e0e6abd457d5305c03b41508533f3c30d815!

src/routes/v2/authenticatedSites/settings.js
alexanderleegs: Had a chat with Chin about this - given that it's difficult to step through to retrieve all the possible errors here now, I'm leaning towards leaving this as `throw` and fixing it after the mvp is out (by refactoring to ts) - do you have concerns about this?

src/services/configServices/__tests__/SettingsService.spec.js
```diff
@@ -91,6 +116,68 @@ describe("Settings Service", () => {
     })
   })
 
+  describe("getEncryptedPassword", () => {
```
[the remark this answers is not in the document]
alexanderleegs: Will have custom tests for crypto-utils in the associated test PR!

src/services/configServices/__tests__/SettingsService.spec.js
```diff
@@ -373,4 +460,136 @@ describe("Settings Service", () => {
       )
     })
   })
+
+  describe.only("updatePassword", () => {
```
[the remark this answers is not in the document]
alexanderleegs: Removed in https://github.com/isomerpages/isomercms-backend/pull/807/commits/833df6758aa7d357d6c848032bb189ad6c0d3848 - Change made in the associated test PR! This should not have been in this PR but I accidentally pushed haha

src/services/configServices/__tests__/SettingsService.spec.js
```diff
@@ -373,4 +460,136 @@ describe("Settings Service", () => {
...
+        mockDeploymentsService.updateAmplifyPassword
+      ).toHaveBeenLastCalledWith(
+        mockUserWithSiteSessionData.siteName,
+        encryptedPassword,
+        iv,
+        enablePassword
+      )
+    })
+
+    it("removes password for previously private amplify repos", async () => {
```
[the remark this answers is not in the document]
alexanderleegs: Yep that's a good point, made the change in the associated test pr https://github.com/isomerpages/isomercms-backend/pull/807/commits/833df6758aa7d357d6c848032bb189ad6c0d3848

src/services/configServices/SettingsService.js
```diff
@@ -43,6 +52,38 @@ class SettingsService {
     }
   }
 
+  async getPassword(sessionData) {
+    const { siteName } = sessionData
+    const siteInfo = await this.sitesService.getBySiteName(siteName)
+    if (siteInfo.isErr()) {
+      // Missing site indicating netlify site - return special result
+      return okAsync({
+        password: "",
```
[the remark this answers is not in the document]
alexanderleegs: Yeah that makes sense, have updated it in https://github.com/isomerpages/isomercms-backend/pull/806/commits/af96a205ec03ffd0ecf0b4ad511aa583192138d6

kishore03109 (commented):
sadly cicd failign ya

src/services/identity/DeploymentsService.ts
alexanderleegs: https://github.com/isomerpages/isomercms-backend/pull/806/commits/1703033f5753b37cbeb243c9fcf05557df4474b4

CHANGELOG.md
```diff
@@ -93,14 +104,7 @@ Generated by [`auto-changelog`](https://github.com/CookPete/auto-changelog).
...
 
-#### [v0.24.1](https://github.com/isomerpages/isomercms-backend/compare/v0.24.0...v0.24.1)
-
-> 27 April 2023
-
-- fix(review request service): use repo name rather than site name [`a323551`](https://github.com/isomerpages/isomercms-backend/commit/a323551feda7b077a7d5d6c2d11ae841ece9bf8d)
-- fix(test cases): fix failing tests [`c843865`](https://github.com/isomerpages/isomercms-backend/commit/c84386549a28602e1af10cd2eda9f056212afaa7)
-
-#### [v0.24.0](https://github.com/isomerpages/isomercms-backend/compare/v0.23.1...v0.24.0)
+#### [v0.24.1](https://github.com/isomerpages/isomercms-backend/compare/v0.23.1...v0.24.1)
```
kishore03109, line 107: hey i think these got committed by accident?

kishore03109 (commented):
@alexanderleegs not sure why but there are sadly alot of extra commits 
Could I get your help to rebase this?

CHANGELOG.md
harishv7: @alexanderleegs i think need to rebase

alexanderleegs: Yup fixed on rebase!

alexanderleegs:
> @alexanderleegs not sure why but there are sadly alot of extra commits Could I get your help to rebase this? [Style] When we are making stacked diffs, shall we we only merge them in when the parent PR is approved? abit hard to reason about a larger pr in one sitting :(

Sorry I thought the base PR was also approved already!

src/routes/v2/authenticatedSites/settings.js
kishore03109: @alexanderleegs am ok with this, but are we sure that there is is no errors meant for internal consumption thrown back to the end user?

src/services/db/GitHubService.js
```diff
@@ -487,6 +490,35 @@ class GitHubService {
       throw err
     }
   }
+
+  async changeRepoPrivacy(sessionData, shouldMakePrivate) {
+    const { siteName, isomerUserId } = sessionData
+    const endpoint = `${siteName}`
```
kishore03109, line 496: sanity check here, wouldnt endpoint refer to `repos/<repoName>` and `siteName` refer to just `repoName`? should we just use the variable `siteName` instead?

kishore03109 (commented):
mostly ok pending 2 concerns:
1. this pr contains throw <result>.error. Have  we made sure that the errors thrown back to the users are not meant for internal consumption? Eg, an axios error from github should **_not_** be returned to the user. 
2. This is a stacked diff, have we made sanity checks on the child diff? Specifically ensuring that 
 - all non-email sites **_dont_** have access to this 
 - all the email sites is able to privatise their repos?

kishore03109 (approved):

src/services/db/GitHubService.js
alexanderleegs: This is referring to the endpoint still! The actual endpoint to call is `/repos/<orgname>/<repoName>`, but our axios instance prefixes with `/repos/<orgname>/`, so the remaining endpoint is just `<repoName>` (referred to as `siteName` here

src/routes/v2/authenticatedSites/settings.js
alexanderleegs: Verified offline - the thrown isomer errors have user facing error messages, and any other error gets caught by our error handler to throw a generic error and message

PR #808, opened 2023-06-22
Fix: step functions env var

kishore03109:
## Problem

the step functions arn was not injected in.

Additionally, the env vars are not present in prod :sadge:, but this should have be caught by config.

## Solution

changing the default values to warns devs if the end var does not exist. 

**Breaking Changes**

  - Details ...

- `FF_DEPRECATE_SITE_QUEUES = false`
- `STEP_FUNCTIONS_ARN = found in 1pw`

### DIFF
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -206,7 +206,7 @@ const config = convict({
           doc: "Whether the queues are deprecated",
           env: "FF_DEPRECATE_SITE_QUEUES",
           format: "required-boolean",
-          default: null,
+          default: false,
         },
       },
     },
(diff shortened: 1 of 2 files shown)


alexanderleegs (approved):

src/config/config.ts
```diff
@@ -206,7 +206,7 @@ const config = convict({
           doc: "Whether the queues are deprecated",
           env: "FF_DEPRECATE_SITE_QUEUES",
           format: "required-boolean",
-          default: false,
+          default: null,
```
alexanderleegs, line 209: I think it's okay to keep false as the default, though we should still add it to our env vars, but i'm fine either way!

PR #815, opened 2023-06-27
Is 236 keep files not recognised as media

caando:
## Problem

`.keep` files are not being parsed properly because it exist in all folders despite not being whitelisted as a filetype. 

Closes IS-236

## Solution

Add a new placeholder filetype that handles all `.keep` files. 

**Bug Fixes**:

- `ReviewRequestService.ts` parses files in the order priority of config, media and page files. 
- Since `.keep` is not a recognised a valid media filetype, it gets treated as a page file. 
- Therefore the file has the wrong directory when making queries to Github API.


alexanderleegs (commented):
I think this might need some frontend changes as well! The `placeholder` type isn't handled properly by the review request screen currently

src/types/dto/review.ts
```diff
@@ -29,7 +35,15 @@ export interface EditedMediaDto extends BaseEditedItemDto {
   type: "file" | "image"
 }
 
-export type EditedItemDto = EditedPageDto | EditedConfigDto | EditedMediaDto
+export interface EditedPlaceholderDto extends BaseEditedItemDto {
+  type: "placeholder"
+}
+
+export type EditedItemDto =
```
seaerchin, line 42: if we do want to return this to the frontend, we probably need an associated PR on the frontend that either hides the placeholder files or does something useful with them.

i'm of hte mind to filter them uot atm

seaerchin (approved):
lgtm but **please do not merge until associated FE PR is merged**

caando:
Requesting for re-review as review request endpoints have been changed to filter placeholder files from `change_files` count and list of files.

src/services/review/ReviewRequestService.ts
```diff
@@ -393,9 +407,17 @@ export default class ReviewRequestService {
         const {
           title,
           body,
-          changed_files,
           created_at,
         } = await this.apiService.getPullRequest(siteName, pullRequestNumber)
+        const { files } = await this.apiService.getCommitDiff(siteName)
+        const filterPlaceholderFiles = files
```
seaerchin, line 413:
```suggestion
        const displayedFiles = files
```

src/services/review/ReviewRequestService.ts
```diff
@@ -393,9 +407,17 @@ export default class ReviewRequestService {
...
+          .filter((file) => {
+            const extractPlaceholderFileResult = extractPathInfo(
+              file.filename
+            ).andThen((pathInfo) => this.extractPlaceholderInfo(pathInfo))
+            return extractPlaceholderFileResult.isErr()
+          })
```
seaerchin, line 419: why not `extractPathInfo` -> `isPlaceholderFile`?

src/services/review/ReviewRequestService.ts
```diff
@@ -711,10 +733,15 @@ export default class ReviewRequestService {
         // Refer here for details; https://docs.github.com/en/rest/commits/commits#compare-two-commits
         // Note that we need a triple dot (...) between base and head refs
         this.compareDiff(userWithSiteSessionData, stagingLink).map(
-          (changedItems) => ({
-            ...rest,
-            changedItems,
-          })
+          (changedItems) => {
+            const filterPlaceholderFiles = changedItems.filter(
```
seaerchin, line 737:
```suggestion
            const displayedFiles = changedItems.filter(
```

src/services/review/__tests__/ReviewRequestService.spec.ts
seaerchin, line 1: it seems like we don't have tests for checking if it still returns `placeholder` files at present? are we planning to add it anytime soon?

src/types/dto/review.ts
```diff
@@ -29,7 +35,15 @@ export interface EditedMediaDto extends BaseEditedItemDto {
...
+  | EditedPageDto
+  | EditedConfigDto
+  | EditedMediaDto
+  | EditedPlaceholderDto
```
seaerchin, line 47: is this meant to be here? i thought we shifted over to processing and not returning to FE? (the `filter` stuff in `reviewRequestService`)

src/services/review/ReviewRequestService.ts
caando: okay

src/services/review/ReviewRequestService.ts
caando: makes sense

src/types/dto/review.ts
caando: placeholder file changes should still show up on `compareDiff` and `compareDiffWithMappings`, creating new type of `DisplayedEditedItemDto` without `EditedPlaceholderDto` when returning to FE

src/services/review/__tests__/ReviewRequestService.spec.ts
caando: adding tests for `listValidReviewRequests` and `getFullReviewRequest`

seaerchin (approved):

PR #818, opened 2023-06-28
Is 219 be add routes to start of site launch process to be called from fe

kishore03109:
## Problem

Add in routes for site launch. To be reviewed in conjunction with [fe pr](https://github.com/isomerpages/isomercms-frontend/pull/1325)

## Solution

Add in the routes 
`launchSite` and `launchInfo` to enable this functionality. 

**Breaking Changes**

  - Details ...

**Features**:

This PR introduces some semantic changes for the end state for the site that will be stored in the db. 
Previously, all site launches used end with a `site_status` = `LAUNCHED` and the `job_status` = `RUNNING`.
Moving forward, we want to make it such that at the end state, we want to change this `site_status` = `LAUNCHED` and `job_status` = `READY`.  The problem I am trying to solve here is that I need a way to store in the DB that the launch process has started, but has yet to complete, as this leads to a visual change of what we display to the user. What this means is that there will be some manual remediation for the already launched sites to modify their states to `job_status` from `RUNNING` to `READY`. Open to alternatives to solve this!  

CC. @seaerchin @harishv7 
I want to introduce a util method for the error handling stuff. I noticed that there are quite a number of functions that require me to throw an error, and log the exact same message. I have introduced this in [here](https://github.com/isomerpages/isomercms-backend/blob/65848b0ea0814b404bb879b4eb205dd328d38f43/src/utils/error-utils.ts), but am not sure if this is the best place to put this/if there is already an util function for this. 

**Note to reviewers** 
I am not too sure if I am [logging this error message properly](https://github.com/isomerpages/isomercms-backend/pull/818/files#diff-b26e8847cd2412d8929843091c09863ff22a59ce92dca045c4ca228e3324e8e8R59-R63). The reason for this dilemma is that an error thrown here is not necessarily an error. It could mean that the site launch process has not started, or already ended. 
I am still opting for logging the error here as if in the case of an genuine error, at least we can refer to the logs for this rather than being confused about why that occurred in the first place 

## Tests

I will be exploring introducing e2e tests for this but don't have bandwidth to do this yet. 
for now a manual test is shown [here](https://github.com/isomerpages/isomercms-frontend/assets/42832651/46f411ff-19de-49aa-a302-cc9a0f2a42d0).


src/services/infra/InfraService.ts
```diff
@@ -229,6 +260,159 @@ export default class InfraService {
     return ok(dnsRecord)
   }
 
+  launchSiteFromCms({
```
seaerchin, line 263: out of curiosity, why prefer object notation over the positional one

src/services/infra/InfraService.ts
```diff
@@ -229,6 +260,159 @@ export default class InfraService {
...
+            primaryDomain,
+            subDomainSettings
+          ),
+          (error) => new SiteLaunchError(`Error launching site: ${error}`)
+        ).andThen((siteLaunch) => {
+          if (siteLaunch.isOk()) {
+            return okAsync(siteLaunch.value)
+          }
+          return errAsync(siteLaunch.error)
+        })
```
seaerchin, line 320: idt we need this right?

src/services/infra/InfraService.ts
```diff
@@ -229,6 +260,159 @@ export default class InfraService {
...
+    siteName,
+    primaryDomain,
+    useWww,
+    email,
+  }: LaunchSiteFromCMSParams): ResultAsync<
+    SiteLaunchCreateParams,
+    AmplifyError | SiteLaunchError | MissingUserEmailError
+  > {
+    // prepare site launch params
+    return ResultAsync.fromPromise(
```
seaerchin, line 273: not a blocker, but i feel like if the precondition to this function is to have a valid user email, we might want to strengthen the params from `email: string` -> `user: User`

src/services/infra/InfraService.ts
```diff
@@ -229,6 +260,159 @@ export default class InfraService {
...
+              // an empty array but we defensively check for this
+              return errAsync(
+                new SiteLaunchError(`No DNS records found for ${siteName}`)
+              )
+            }
+            return okAsync({
+              siteStatus: "LAUNCHING",
+              dnsRecords: records.dnsRecords,
+              siteUrl: records.siteUrl,
+            } as SiteLaunchDto)
```
seaerchin, line 360: we can avoid the cast by hinting to the compiler what the return type of `okAsync` is supposed to be: 
```ts
okAsync<SiteLaunchDto>({
                siteStatus: "NOT_LAUNCHED",
              })
```

src/services/infra/InfraService.ts
```diff
@@ -229,6 +260,159 @@ export default class InfraService {
...
+          })
+        )
+
+        // 2. If not, get the source of truth from the db
+        .orElse(() =>
+          this.sitesService.getBySiteName(siteName).andThen((site) => {
+            if (!site) {
+              errAsync(
```
seaerchin, line 368: missing return

src/services/infra/InfraService.ts
```diff
@@ -229,6 +260,159 @@ export default class InfraService {
...
+              dnsRecords: records.dnsRecords,
+              siteUrl: records.siteUrl,
+            } as SiteLaunchDto)
+          })
+        )
+
+        // 2. If not, get the source of truth from the db
+        .orElse(() =>
+          this.sitesService.getBySiteName(siteName).andThen((site) => {
+            if (!site) {
```
seaerchin, line 367: we shouldn't need this check - i think `getBySiteName` already handles the missing case for us

src/services/infra/InfraService.ts
```diff
@@ -229,6 +260,159 @@ export default class InfraService {
...
+                new MissingSiteError(
+                  `Site launch status for ${siteName} is unknown`
+                )
+              )
+            }
+            if (site.siteStatus !== SiteStatus.Launched) {
+              return okAsync({
+                siteStatus: "NOT_LAUNCHED",
+              } as SiteLaunchDto)
+            }
```
seaerchin, line 378: why are we opting to default to `NOT_LAUNCHED` instead of the site's `siteStatus`? wouldn't that be more accurate

src/services/infra/InfraService.ts
```diff
@@ -229,6 +260,159 @@ export default class InfraService {
...
+    siteName: string
+  ): ResultAsync<
+    DnsResultsForSite,
+    MissingSiteError | AmplifyError | SiteLaunchError
+  > =>
+    this.sitesService
+      .getBySiteName(siteName)
+      .andThen(() => this.launchesService.getDNSRecords(siteName))
+
+  getSiteLaunchStatus(
```
seaerchin, line 335: feels like there should be a function called `getSiteStatusFromDnsRecord` that does `DnsResultForSite` -> `SiteLaunchDto` because that seems to be a repeated pattern

seaerchin: on a side note, `DnsResultForSite` sounds weird - wdyt about `SiteDnsResults` ._.

src/services/infra/InfraService.ts
```diff
@@ -229,6 +260,159 @@ export default class InfraService {
...
+                siteStatus: "NOT_LAUNCHED",
+              } as SiteLaunchDto)
+            }
+
+            // 3. If site is either launched or in the
+            //    process of being launched, get the dns records
+            return (
+              this.getGeneratedDnsRecords(siteName)
+                .andThen((records) => {
+                  if (records.dnsRecords.length === 0) {
```
seaerchin, line 385: from the comment here, it sounds like we need to check the `SiteStatus` for it being in the process of being launched also? the earlier check only covers `LAUNCHED`

src/services/infra/InfraService.ts
```diff
@@ -422,7 +606,7 @@ export default class InfraService {
             updateSiteLaunchParams = {
               id: site.value.id,
               siteStatus: SiteStatus.Launched,
-              jobStatus: JobStatus.Running,
+              jobStatus: JobStatus.Ready,
```
seaerchin, line 609: could i get more context on this chnage?

src/types/siteInfo.ts
```diff
@@ -8,3 +10,20 @@ export type SiteInfo = {
...
+export interface DnsResultsForSite {
+  dnsRecords: DNSRecord[]
+  siteUrl: string
+}
+
+export interface SiteLaunchDto {
+  siteStatus: "LAUNCHED" | "NOT_LAUNCHED" | "LAUNCHING"
+  dnsRecords?: DNSRecord[] // only present iff siteStatus is LAUNCHED
+  siteUrl?: string
+}
```
seaerchin, line 29: can we make different types for this? 

```ts
export type SiteLaunchDto = LaunchedSiteDto 
| SiteLaunchingDto
| SiteNotLaunchedDto
```

allows the inference to work better so that only if we assert `siteDto.siteStatus === "LAUNCHED"` then `dnsRecords` is present.

src/utils/error-utils.ts
```diff
@@ -0,0 +1,9 @@
+import logger from "@root/logger/logger"
+
+export default function createErrorAndLog<T extends Error>(
+  ErrorType: new (errorMessage: string) => T,
+  message: string
+): T {
+  logger.error(message)
+  return new ErrorType(message)
+}
```
seaerchin, line 9: sadly this doesn't lend itself well to extension because errors are very likely to take additional arguments. 

if we want to always ensure errors are logged at initialisation, we could always do `logger.error(message)` but this might be verbose and overkill, especially when errors are expected.

kishore03109: hmm true, but that could be solved with 1c4bb481442be753945a061fc33687acff938dd2 right? 

> if we want to always ensure errors are logged at initialisation, we could always do logger.error(message) but this might be verbose and overkill, especially when errors are expected.

nah ya I want this to be flexible to dev, I dont think logging every error should be the way to go here.

src/routes/v2/authenticated/sites.ts
```diff
@@ -141,7 +182,18 @@ export class SitesRouter {
...
     )
-
+    router.get(
+      "/:siteName/launchInfo",
+      attachSiteHandler,
+      this.authorizationMiddleware.verifySiteMember,
+      attachReadRouteHandlerWrapper(this.getSiteLaunchInfo)
+    )
+    router.post(
+      "/:siteName/launchSite",
```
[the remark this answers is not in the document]
harishv7: yea routes lets use `-` for separator

src/constants/constants.ts
```diff
@@ -56,3 +56,4 @@ export const ISOMER_ADMIN_REPOS = [
 
 export const INACTIVE_USER_THRESHOLD_DAYS = 60
 export const GITHUB_ORG_REPOS_ENDPOINT = `https://api.github.com/orgs/${ISOMER_GITHUB_ORG_NAME}/repos`
+export const REDIRECTION_SERVER_IP = "18.136.36.203"
```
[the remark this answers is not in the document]
harishv7: hmm i would prefer not hardcoding IP addresses. these should come from .env

src/services/identity/LaunchesService.ts
```diff
@@ -146,6 +165,84 @@ export class LaunchesService {
...
+      })
+      .andThen((launchRecord) =>
+        ResultAsync.fromPromise(
+          this.redirectionsRepository.findOne({
+            where: { launchId: launchRecord.id },
+          }),
+          () =>
+            // There are legitimate cases where this error is thrown (ie the site has no redirection record)
+            new AmplifyError(`Failed to get redirection record for ${repoName}`)
+        ).andThen((redirectionRecord) => {
```
harishv7, line 223: nit: is it possible to flatten this out instead of chaining multiple levels?

harishv7: was thinking that we will primarily use the isOk isErr constructs and write the error handling in usual JS/TS way of if, else instead of the method chaining constructs from neverthrow

src/services/infra/InfraService.ts
```diff
@@ -229,6 +260,159 @@ export default class InfraService {
...
+          this.getGeneratedDnsRecords(siteName).andThen((records) => {
+            if (records.dnsRecords.length === 0) {
+              // will not occur, as the function should not return
+              // an empty array but we defensively check for this
+              return errAsync(
+                new SiteLaunchError(`No DNS records found for ${siteName}`)
+              )
+            }
+            return okAsync({
+              siteStatus: "LAUNCHING",
```
harishv7, line 357: this is supposed to be namespaced right Status.Launching?

src/services/infra/InfraService.ts
```diff
@@ -229,6 +260,159 @@ export default class InfraService {
...
+                    // error instead, but we defensively check for this
+                    return okAsync({
+                      siteStatus: "LAUNCHING",
+                    } as SiteLaunchDto)
+                  }
+                  return okAsync({
+                    siteStatus:
+                      // status is only successful iff job is ready
+                      site.jobStatus === JobStatus.Ready
+                        ? "LAUNCHED"
```
harishv7, line 396: SiteStatus.Launched

src/services/infra/__tests__/DynamoDBService.spec.ts
```diff
@@ -41,7 +41,7 @@ const mockFailureLaunch: SiteLaunchMessage = {
   },
 }
 
-const tableName = "site-launch"
+const tableName = "ddb-name"
```
harishv7, line 44: why do we have a change in table name here?

src/errors/SiteLaunchError.ts
```diff
@@ -0,0 +1,11 @@
+import { BaseIsomerError } from "@root/errors/BaseError"
+
+export default class SiteLaunchError extends BaseIsomerError {
+  constructor(siteName: string) {
+    super({
+      status: 500,
+      code: "InfraError",
+      message: `The given file: ${siteName} was not a config file!`,
```
harishv7, line 8: @seaerchin @kishore03109 should we starting writing new errors as per V2 format similar to `CollectionPageService`?

example
```
 const err = new BadRequestError(
        `Special characters not allowed when renaming files. Given name: ${newFileName}`,
        { fileName: newFileName },
        ComponentTypes.Service,
        FileCodes.CollectionPageService
      )
      err.isV2Err = true
      throw err
    }
```

if so things like `meta` can contain the `siteName` and if it is compatible with the new interface it can expose the new methods in `IsomerError.ts`

```
/* Namespace for Error functions */
export const IsomerError = {
  toExternalRepresentation: (e: IsomerInternalError): IsomerExternalError => ({
    name: e.code,
    code: getExternalCode(e.componentCode, e.fileCode, e.code),
    message: e.message,
    isV2Err: true,
  }),
  getLog: (error: IsomerInternalError): Record<string, unknown> => ({
    ...error,
  }),
}

```

src/routes/v2/authenticated/sites.ts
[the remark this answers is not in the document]
kishore03109: hmm I think maintaining consistency here might be more beneficial. Since the rest of the authenticated sites need to use `camelCasing` and changing things at the API level can increase the surface area for bugs, am concerned about making this change at all (even if we agree I think this should be tacked in a separate PR ya). Thoughts?

src/services/identity/LaunchesService.ts
```diff
@@ -3,14 +3,29 @@ import {
   GetDomainAssociationCommandOutput,
   SubDomainSetting,
 } from "@aws-sdk/client-amplify"
```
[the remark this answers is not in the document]
kishore03109: Following naming conventions similar to `DeploymentsService` and `UsersService` in the same folder

src/services/identity/LaunchesService.ts
kishore03109: hmm am quite confused here ah, how can I consume an async neverthrow without using chaining like this? 
I think it also might be beneficial if we clarified on slack whether we want to encourage/discourage chaining as a team!

src/services/infra/InfraService.ts
kishore03109: Personally tend to use object notation for more than 2 args. Even though it is more verbose, it makes the code alot clearer. Too many instances for myself when I debug only to find out the issue was the ordering of the args, would advocate for this esp since we still have js code :sadge:

src/services/infra/__tests__/DynamoDBService.spec.ts
kishore03109: oh this is due to the var in .env.test using `ddb-name`

src/services/identity/LaunchesService.ts
```diff
@@ -146,6 +165,84 @@ export class LaunchesService {
...
+  getDNSRecords = (
+    repoName: string
+  ): ResultAsync<DnsResultsForSite, AmplifyError | SiteLaunchError> => {
+    let dnsRecords: DNSRecord[] = []
+    let siteUrl: string
+    return ResultAsync.fromPromise(this.getSiteId(repoName), () =>
+      createErrorAndLog(AmplifyError, `Failed to get siteId for ${repoName}`)
+    )
+      .andThen((siteId) => {
+        if (siteId.isOk()) {
```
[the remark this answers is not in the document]
kishore03109: I am ok to make this change but in a separate PR as this would change the contract for other methods that is not within the scope of the PR + after we get clarification if as a team are ok with chaining constructs/if we just want to use `isOk`/ `isErr`, ie effectively returning a `Promise<Result>`

src/constants/constants.ts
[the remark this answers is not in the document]
kishore03109: sure 186f45c96c39eb55259b37d9e3d581421d8ea8a9

src/services/infra/DynamoDBService.ts
```diff
@@ -32,10 +35,38 @@ export default class DynamoDBService {
...
+        `Something went wrong when querying DynamoDB: ${error}`
+      )
+    ).andThen((entries) => {
+      const entry = entries.find((e) => e.repoName === repoName)
+      if (entry) return okAsync(entry.status?.state ?? "pending")
+      return errAsync(
+        createErrorAndLog(
+          MissingSiteError,
+          `No site found for ${repoName} in DynamoDB`
+        )
```
alexanderleegs, line 63: I think we could update the error message to reflect the possible reasons as well (already ended, haven't started)? So this is easier for engineers to debug in future

alexanderleegs (commented):
The build seems to be failing due to type errors! Might need to configure `createErrorAndLog` to accept `AmplifyError` and `SiteLaunchError`

src/errors/SiteLaunchError.ts
seaerchin: yes, but not inclined to do it here. this PR is already well scoped and the errors don't directly impact functionality. i think we should not let perfect be the enemy of good and as long as the PR contains good code and has relevant test coverage, should be good to go. 

the error can go in as part of another ticket.

src/routes/v2/authenticated/sites.ts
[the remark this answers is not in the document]
seaerchin: if the rest of the routes in this router use `camelCase` then ok. otherwise, still prefer kebab.

src/services/identity/LaunchesService.ts
seaerchin: i think we have to consume `ResultAsync` using `andThen` - this is by design because `await` will turn it into a `Promise` and escape from our error handling patterns

src/services/infra/InfraService.ts
seaerchin: can. as a side note, usually people do main args as positional and options as object. this is because you want consumers to have an easier time writing the method (object very annoying) but when there are options, be precise about the options

src/utils/error-utils.ts
seaerchin: sadly, i don't actually think that works ._. it's important to think about whether the increased verbosity is worth the trade off (here, the verbosity allows us flexibility). 

i think in this case, the LOC saved is probably like 1-2 lines, which isn't much. i think if we do want the method, and i can see why, we might wish to scope it to only a very specific file or module because we can't enforce the error shape that the util method wants. 

to show the benefits of the util method, consider the 2 cases below - 
```ts
const err = new err(createErrorAndLog(...)) // 1 line

const error = new SomeExampleError(...)
logger.error(err.message)
return err(error) // 3 lines
```

src/errors/SiteLaunchError.ts
kishore03109: ea88911320ab365c68528e1a4b6b66038db942b0
following the inspiration from `BadRequestError` from `CollectionPageService`

src/services/identity/LaunchesService.ts
```diff
@@ -146,6 +165,84 @@ export class LaunchesService {
...
+            () =>
+              createErrorAndLog(
+                AmplifyError,
+                `Failed to get launch record for ${repoName}`
+              )
+          )
+        }
+        return errAsync(
+          createErrorAndLog(
+            AmplifyError,
```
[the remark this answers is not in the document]
kishore03109: yup, apologies for this, shouldnt have an error now

src/services/identity/LaunchesService.ts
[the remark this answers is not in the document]
harishv7: I see. Okay, personally prefer LaunchService as LaunchesService sounds weird plus the responsibility of the service is for "a" site launch

But leaving this up to you what you want to opt for

src/types/siteInfo.ts
kishore03109: 14c038ff941fdc02c43c696236b65257c57b665c
oops this comment is wrong, as it could be in the `LAUNCHED` status but the records are still being generated

src/services/infra/DynamoDBService.ts
kishore03109: yup is in the roadmap, making more meaningful error messages in a separate pr bah

src/services/identity/LaunchesService.ts
```diff
@@ -146,6 +165,84 @@ export class LaunchesService {
     return ok(siteId)
   }
 
+  getDNSRecords = (
+    repoName: string
+  ): ResultAsync<DnsResultsForSite, AmplifyError | SiteLaunchError> => {
+    let dnsRecords: DNSRecord[] = []
+    let siteUrl: string
```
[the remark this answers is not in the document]
kishore03109: sure fc35535967274c7c8f6e4efcc29512bdb6ffc8d0

src/utils/error-utils.ts
kishore03109: removing this bit here to unblock this PR, think this warrents a separate discussion in another thread! 3f2ead83ed79d6df958523103ce0767f1e6f35df

src/services/identity/LaunchesService.ts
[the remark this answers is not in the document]
seaerchin: sure - file a ticket for it! i'd hope for the ticket to be worked on soon to prevent the signature from propagating

src/services/infra/InfraService.ts
seaerchin: my bad - thought it was returning `Result`, not `ResultAsync`

src/services/infra/InfraService.ts
seaerchin: i think my concern here is that this feels like a router responsibility (to validate parameters for services) rather than a service responsibility. 

i envision routers doing validation etc prior to passing on to the service, so that services can be strict about their pre-conditions. hence, i'd prefer this to be separated! (only saw 1 call site)

we should also wrap the methods for `userService` to return a `Result` but out of scope

src/services/infra/InfraService.ts
```diff
@@ -1,15 +1,39 @@
...
+  RedirectionTypes,
+  REDIRECTION_SERVER_IP,
+} from "@root/constants"
+import MissingSiteError from "@root/errors/MissingSiteError"
+import MissingUserEmailError from "@root/errors/MissingUserEmailError"
+import SiteLaunchError from "@root/errors/SiteLaunchError"
 import logger from "@root/logger/logger"
 import { AmplifyError } from "@root/types/amplify"
+import {
+  DnsResultsForSite as SiteDnsResults,
```
seaerchin, line 33: could we do a rename on the type?

seaerchin (approved):
Lgtm, please 

1. rename type
2. shift validation out of router 
3. file ticket to shift services from `Promise<Result>` to `ResultAsync`

src/services/infra/InfraService.ts
seaerchin: dun see this alr so resolving

src/services/identity/LaunchesService.ts
seaerchin: shouldn't be a blocker if the code is working - we can iterate on this as a ticket next time.

seaerchin: side note is that when we write as a `Promise<Result>` and we do `isOk + await`, we can resolve both of those using `andThen` ._.

src/config/config.ts
```diff
@@ -371,6 +371,14 @@ const config = convict({
       default: false,
     },
   },
+  redirectionServer: {
+    elasticIp: {
+      doc: "Elastic IP of the redirection server",
+      env: "REDIRECTION_SERVER_ELASTIC_IP",
+      format: String,
```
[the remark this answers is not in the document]
kishore03109: we have a sane default here of '18.136.36.203', so it not really required!

src/services/infra/InfraService.ts
kishore03109: opps sorry forgot to refactor this, done here 198e2a257522b08f6296a7c4bd21a135e67ed325

src/errors/IsomerError.ts
```diff
@@ -23,6 +23,7 @@ export const ComponentTypes = {
 }
 
 export const FileCodes = {
+  InfraServiceError: "002",
```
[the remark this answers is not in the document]
kishore03109: okie 12c4fcf328625a8ee72e57521635961e48b860ce

src/services/identity/LaunchesService.ts
```diff
@@ -165,89 +152,86 @@ export class LaunchesService {
     return ok(siteId)
   }
 
-  getDNSRecords = (
+  getDNSRecords = async (
```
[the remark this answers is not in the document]
kishore03109: filed ticket https://isomeropengov.atlassian.net/browse/IS-288?atlOrigin=eyJpIjoiOGJjZjgwZTkwY2QyNDc4MDgyYmU4YTQzY2JmMTY0NDAiLCJwIjoiaiJ9

src/services/identity/LaunchesService.ts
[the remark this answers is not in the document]
kishore03109: agreed, filed ticket [here](https://isomeropengov.atlassian.net/browse/IS-288?atlOrigin=eyJpIjoiOGJjZjgwZTkwY2QyNDc4MDgyYmU4YTQzY2JmMTY0NDAiLCJwIjoiaiJ9)

src/services/infra/InfraService.ts
kishore03109: > i think my concern here is that this feels like a router responsibility (to validate parameters for services) rather than a service responsibility

just curious here, shoudnt our mental model be router should verify params rather than validate since this validation of parameters for services here requires a call to our repositories?

PR #823, opened 2023-07-05
feat(error-utils): add error utils

kishore03109:
## Problem
This is an exploratory PR that tries to solve a repeated pattern in our code base. Relatively trivial since it only impact number of lines of code

## Solution

add a util function


kishore03109:
seems like there are multiple instances where the logged error is different from the error message, which defeats the usefulness of this PR. Closing this

PR #828, opened 2023-07-07
Feat/privatise netlify sites

alexanderleegs:
This PR modifies our privatisation flow to also include netlify repos. Previously, we realised that sites we migrated to amplify still retained their associated netlify repo (as a backup), and we were building to both netlify and amplify. This means that it is theoretically possible for a user to access content on a privatised repo via the link to the netlify site, if the user was able to obtain it.

Video for full user flow is on https://github.com/isomerpages/isomercms-frontend/pull/1316

To fix this, this PR introduces functionality to also modify the build on netlify. Specifically, we take the following actions on the netlify build, for any site which builds from the staging branch of the desired repo:

- We change the password to the one first provided by the user
- We disable automated builds for this repo in future

As we're keeping the netlify repo only for emergency uses, we can simply reenable the builds if we ever need to use it.

As we do not store the link to the netlify repo anywhere in either the github repo or our db, the only way to locate the netlify repo is to retrieve the full list of netlify repos we control, and filter this down via inspecting the connected repo. As this is a slow operation, I've opted to make this as an async operation that we do not await, and call only when sites are changed from a public to private repo, in order to minimize the number of times this is called. However, this means that the password on the netlify site may not match up with the password on the amplify site - I feel like this is acceptable as we are still able to access the netlify password via the netlify UI, and the netlify site is unknown to the users.

## New environment variables
`NETLIFY_ACCESS_TOKEN`: this is the access token we use to access our account. We should be using the one generated for isomeradmin


src/utils/netlify-utils.ts
```diff
@@ -0,0 +1,95 @@
...
+
+type UpdateNetlifySiteDetails = {
+  password: string
+  build_settings: {
+    stop_builds: boolean
+  }
+}
+
+// Retrieves all netlify sites which build from the given repo's staging branch
+const getNetlifySiteDetails = async (repoName: string) => {
```
kishore03109, line 27: could we define our return type expectedly here? 
just to clarify ya, why are we returning an array here? wouldnt we only have 1 staging site?

src/utils/netlify-utils.ts
```diff
@@ -0,0 +1,95 @@
...
+
+const GITHUB_ORG_NAME = config.get("github.orgName")
+const NETLIFY_ACCESS_TOKEN = config.get("netlify.accessToken")
+
+type NetlifySiteDetails = {
+  id: string
+  password: string
+  build_settings: {
+    repo_path: string
+    repo_branch: string
```
kishore03109, line 15: should we assign this to `staging` since that is the only branch we care about here?

src/utils/netlify-utils.ts
```diff
@@ -0,0 +1,95 @@
...
+    const allSites = await axios.get<NetlifySiteDetails[]>(endpoint, {
+      headers,
+    })
+    const relatedSites = allSites.data.filter(
+      (site) =>
+        site.build_settings.repo_path === repoPath &&
+        site.build_settings.repo_branch === "staging"
+    )
+    return relatedSites
+  } catch (err: any) {
```
kishore03109, line 44: nit: `err: unknown`

src/services/configServices/SettingsService.js
```diff
@@ -182,6 +183,10 @@ class SettingsService {
       } catch (err) {
         return errAsync(err)
       }
+      if (enablePassword) {
```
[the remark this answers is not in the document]
alexanderleegs: Hey this is already limited to public -> private! It's nested within the `isPrivate !== enablePassword` check, would it be clearer if i moved this outside

src/utils/netlify-utils.ts
alexanderleegs: Hmm this is the response from netlify, which can contain branches other than those building on `staging`, so I think we'll need to maintain this info!

src/utils/netlify-utils.ts
alexanderleegs: sure, defined in https://github.com/isomerpages/isomercms-backend/pull/828/commits/b8ecf27e5f57094f67e0bceb3967748599f533bb

I've opted to return an array just in case we have multiple repos building from the same repo and staging branch - just to be safe!

src/utils/netlify-utils.ts
alexanderleegs: Changed in https://github.com/isomerpages/isomercms-backend/pull/828/commits/66f5014956123616a85c3afe4f7f897221555d69

src/utils/netlify-utils.ts
```diff
@@ -0,0 +1,95 @@
...
+  build_settings: {
+    repo_path: string
+    repo_branch: string
+  }
+}
+
+type UpdateNetlifySiteDetails = {
+  password: string
+  build_settings: {
+    stop_builds: boolean
```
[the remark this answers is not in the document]
alexanderleegs: Sure, changed in https://github.com/isomerpages/isomercms-backend/pull/828/commits/44e386da7894360aa26f3be9e7add2e03b1f8681!

src/config/config.ts
```diff
@@ -385,6 +385,14 @@ const config = convict({
       default: "18.136.36.203",
     },
   },
+  netlify: {
+    accessToken: {
+      doc: "Access token for netlify actions",
+      env: "NETLIFY_ACCESS_TOKEN",
+      format: "required-string",
+      default: "",
```
seaerchin, line 393: pls give a sane default sir - i think this might fail on cold start

src/utils/netlify-utils.ts
```diff
@@ -0,0 +1,98 @@
...
+    })
+    const relatedSites = allSites.data.filter(
+      (site) =>
+        site.build_settings.repo_path === repoPath &&
+        site.build_settings.repo_branch === "staging"
+    )
+    return relatedSites
+  } catch (_err: unknown) {
+    const err = _err as AxiosError
+    if (err.message)
```
seaerchin, line 48: pls do a check

src/utils/netlify-utils.ts
```diff
@@ -0,0 +1,98 @@
...
+) => {
+  const endpoint = `https://api.netlify.com/api/v1/sites/${repoId}`
+  const headers = {
+    Authorization: `Bearer ${NETLIFY_ACCESS_TOKEN}`,
+    "Content-Type": "application/json",
+  }
+  try {
+    await axios.patch(endpoint, settings, { headers })
+  } catch (err: any) {
+    if (err.message)
```
seaerchin, line 70: pls dun use `any` - use `unknown` instead and check for existence of `message` property

src/utils/netlify-utils.ts
```diff
@@ -0,0 +1,98 @@
...
+  }
+  const repoPath = `${GITHUB_ORG_NAME}/${repoName}`
+  try {
+    const allSites = await axios.get<NetlifySiteDetails[]>(endpoint, {
+      headers,
+    })
+    const relatedSites = allSites.data.filter(
+      (site) =>
+        site.build_settings.repo_path === repoPath &&
+        site.build_settings.repo_branch === "staging"
```
seaerchin, line 43: what about for prod builds?

src/utils/netlify-utils.ts
```diff
@@ -0,0 +1,98 @@
...
+) => {
+  const netlifySiteDetails = await getNetlifySiteDetails(repoName)
+
+  await Promise.all(
+    netlifySiteDetails.map(async (netlifySite) => {
+      // We always stop the builds - we're only using netlify as a backup for these amplify sites
+      const { id } = netlifySite
+      const updatedSettings = {
+        build_settings: {
+          stop_builds: true as const,
```
seaerchin, line 91:
```suggestion
          stop_builds: true,
```

nit: idt we need `as const` - the assertion is for compile time constants that we want ts to infer the strictest type for

src/utils/netlify-utils.ts
seaerchin, line 1: can we shift this to `services/api` or `services/db` where appropriate

src/config/config.ts
alexanderleegs: We want this to fail if nothing is provided though, so we should leave this blank?

src/utils/netlify-utils.ts
alexanderleegs: Prod is okay - we're fine with keeping prod up to date, since the issue is with embargoed information on staging

src/utils/netlify-utils.ts
alexanderleegs: https://github.com/isomerpages/isomercms-backend/pull/828/commits/681c43c2d84c144513b1ca76e8b1acc00a53b79e

src/utils/netlify-utils.ts
alexanderleegs: https://github.com/isomerpages/isomercms-backend/pull/828/commits/681c43c2d84c144513b1ca76e8b1acc00a53b79e

src/utils/netlify-utils.ts
alexanderleegs: Moved in https://github.com/isomerpages/isomercms-backend/pull/828/commits/944d2d3343d9b19679d148bdf42b8cd81ad32c17

src/utils/netlify-utils.ts
seaerchin: oh true good point

src/services/configServices/SettingsService.js
```diff
@@ -167,6 +168,11 @@ class SettingsService {
     }
     const { id, isPrivate } = siteInfo.value
     if (!isPrivate && !enablePassword) return okAsync("")
+    if (!isPrivate && enablePassword) {
+      // Additionally, we perform an async operation here to swap the netlify repo to private and disable builds,
+      // only when changing from public to private
+      privatiseNetlifySite(siteName, password)
```
harishv7, line 174: can we add a comment on why we are not awaiting this

src/config/config.ts
seaerchin: eh i derped, forgot this was supposed to be in 1PW/`.env` ._.

src/services/configServices/SettingsService.js
alexanderleegs: Added in 0ec15298!

src/services/db/__tests__/GitHubService.spec.ts
```diff
@@ -2,7 +2,7 @@ import { ConflictError } from "@errors/ConflictError"
 import { NotFoundError } from "@errors/NotFoundError"
 import { UnprocessableError } from "@errors/UnprocessableError"
 
-import validateStatus from "@utils/axios-utils"
+import { validateStatus } from "@utils/axios-utils"
```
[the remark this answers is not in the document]
alexanderleegs: Either is okay - I think vscode just defaults to `@root/utils`, I've modified all of them to use `@utils` instead in 0ec15298

src/services/api/NetlifyService.ts
```diff
@@ -0,0 +1,112 @@
...
+        )}`
+      )
+  }
+}
+
+export const privatiseNetlifySite = async (
+  repoName: string,
+  password: string
+) => {
+  const netlifySiteDetails = await getNetlifySiteDetails(repoName)
```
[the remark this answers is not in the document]
alexanderleegs: It's possible, but i'm not sure if any user repos do this - but i thought it was better to be safe and catch any potential edge case

src/services/api/NetlifyService.ts
```diff
@@ -0,0 +1,112 @@
...
+) => {
+  const endpoint = `https://api.netlify.com/api/v1/sites/${repoId}`
+  const headers = {
+    Authorization: `Bearer ${NETLIFY_ACCESS_TOKEN}`,
+    "Content-Type": "application/json",
+  }
+  const settings = {
+    password,
+    build_settings: {
+      stop_builds: true,
```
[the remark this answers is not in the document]
alexanderleegs: hmm, would it be better if the method name was changed then? https://github.com/isomerpages/isomercms-backend/pull/828/commits/6346f28a0054ea4ffc8ad303d972faf073b0eb9f

src/services/api/NetlifyService.ts
seaerchin, line 1: nit: is it possible to call this `NetlifyApi` - i think of services as performing some biz logic. this is just a wrapper over raw axios calls

src/services/api/NetlifyService.ts
```diff
@@ -0,0 +1,113 @@
...
+
+const GITHUB_ORG_NAME = config.get("github.orgName")
+const NETLIFY_ACCESS_TOKEN = config.get("netlify.accessToken")
+
+type NetlifySiteDetails = {
+  id: string
+  password: string
+  build_settings: {
+    repo_path: string
+    repo_branch: string
```
seaerchin, line 17: nit: can we do a manual rename or use `lodash`'s `camelCase` function to retain the naming convention? this type seems like it's gonna be used outside of this file so it might load to the snake_case being spread.

seaerchin (approved):
lgtm but please do at least rename the file + `camelCase` the type. 

we shouldn't be mixing `camelCase` and `snake_case`, especially when they have a chance to leak

src/services/api/NetlifyService.ts
alexanderleegs: As discussed offline, this is contained within the file - the only export is for a `() => Promise<void>` method so it will not leak!

src/services/api/NetlifyService.ts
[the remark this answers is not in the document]
harishv7: got it

harishv7 (approved):

src/services/api/NetlifyService.ts
alexanderleegs: https://github.com/isomerpages/isomercms-backend/pull/828/commits/88fcc22c35a7c1c32ed212b75d7c4135b2390ce3

PR #829, opened 2023-07-07
fix(config): change arn from required-string to string

seaerchin:
## Problem
ARN was `required-string` previously but had an empty default, leading to crashes on startup. this changes it to be a `string` instead.

## Solution
1. `required-string` -> `string`

### DIFF
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -184,8 +184,8 @@ const config = convict({
       stepFunctionsArn: {
         doc: "Amazon Resource Name (ARN) of the Step Functions state machine",
         env: "STEP_FUNCTIONS_ARN",
-        format: "string",
-        default: "",
+        format: "required-string",
+        default: "mock-arn",
       },
     },
     sqs: {
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -184,7 +184,7 @@ const config = convict({
       stepFunctionsArn: {
         doc: "Amazon Resource Name (ARN) of the Step Functions state machine",
         env: "STEP_FUNCTIONS_ARN",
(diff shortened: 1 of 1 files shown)


kishore03109 (commented):
not sure where the best place to discuss this should be, can move to slack if you like. 
TLDR, this string is meant to be required, esp in prod and staging envs, but not so much when developing and not touching site launch infra (eg say devs are changing some github calls (ie, any non site launch related dev work), then polling the ddb or triggering the step functions shouldnt be of a concern to them)

that being said, there is a pain point here, dev work that doesnt revolve around site launch shouldnt be concerned about site launch infra stuff.

perhaps a workaround here could be to have a flag (eg INTERACT_WITH_SITE_LAUNCH_INFRA) -> which would be set to true for prod and staging, and conditionally true when developing only for site launch stuff. But this sol does feel quite janky, and doesnt feel to be in the right layer of abstraction here, wdyt?

seaerchin:
@kishore03109 in that case then i'll swap to `required-string` + some nonsense default. wdyt?

kishore03109:
@seaerchin hmm I am trying to avoid that case because if we put some nonsense string, when we actually depreciate the site launch queues, this would lead to the `StepFunctions` service throwing an error.

In the most ideal case, each one of us would have our own infra stack that our local BE interact with, but for now the leaner solution is that we only use dev-stack for this! Thus, the current values for stepfunctions and ddb are from the ones in the `dev` stack from our infra. Open to suggestions to streamline this!

seaerchin:
@kishore03109 i think there's no need for this PR, forgot to check 1pw for env var.

PR #830, opened 2023-07-07
Feat/IS-169 add site lastUpdated with caching

caando:
## Problem

`lastUpdated` field for site info has been removed from backend and frontend dashboard for email login users due to excessive token use.  Querying `lastUpdated` for each individual site cost 1 token use each which is too expensive for commonly visited such as dashboard. 

Closes IS-169

## Solution

`lastUpdated` along with other repo info can be queried in bulk through Github endpoint. Each query allow a maximum of 100 repos and cost only 1 token use. It would take only 10 token uses to get repo info of all sites under `isomerpages` (pageCount 10 for up to 1000 repos). Hence, we can query all repos info at regular intervals and store in a cache to be served on request. Currently, its set to refresh at once per 5 minutes, and with a maximum of 5 instances, which will lead to 10 * 12 * 5 = 600 token uses per hour. This is 600 / 50000 = 1.2% of the total token capacity, which does not contribute significantly to reaching token limits.

**Breaking Changes**

## Tests

Visiting the sites dashboard as an email login user should show `lastUpdated` information. 

## Note

The primary motivation for this solution is reduce load on the database instance. As more users are onboarded onto email-based login, it can be expected that sites read queries and updates can take up significant database CPU usage. However, if in the case that database size is scaled up in the future, this solution can be dropped in favour of directly reading and writing to database.

### DIFF
--- a/src/routes/v2/authenticated/sites.ts
+++ b/src/routes/v2/authenticated/sites.ts
@@ -70,7 +70,7 @@ export class SitesRouter {
     { userWithSiteSessionData: UserWithSiteSessionData }
   > = async (req, res) => {
     const { userWithSiteSessionData } = res.locals
-    const lastUpdated = await this.sitesService.getLastUpdated(
+    const lastUpdated = this.sitesService.getLastUpdated(
       userWithSiteSessionData
     )
     return res.status(200).json({ lastUpdated })
--- a/src/services/identity/SitesCacheService.ts
+++ b/src/services/identity/SitesCacheService.ts
@@ -14,8 +14,6 @@ import type { GitHubRepositoryData, RepositoryData } from "@root/types/repoInfo"
 // Changes are polled at fixed intervals, stored in cache, and served
 // to avoid long load time from live querying.
 
-const REPO_DATA_REFRESH_INTERVAL = 60 * 1000 // 1 minutes
-
 export async function getAllRepoData(
   accessToken: string | undefined
 ): Promise<RepositoryData[]> {
@@ -76,10 +74,13 @@ export async function getAllRepoData(
 export class SitesCacheService {
   private repoDataCache: RepositoryData[]
 
-  constructor() {
+  private refreshInterval: number
+
+  constructor(refreshInterval: number) {
     this.repoDataCache = []
+    this.refreshInterval = refreshInterval
     this.renewCache()
-    setInterval(() => this.renewCache(), REPO_DATA_REFRESH_INTERVAL)
+    setInterval(() => this.renewCache(), this.refreshInterval)
   }
 
   private async renewCache() {
--- a/src/services/identity/SitesCacheService.ts
+++ b/src/services/identity/SitesCacheService.ts
@@ -9,7 +9,12 @@ import {
 import { genericGitHubAxiosInstance } from "@root/services/api/AxiosInstance"
 import type { GitHubRepositoryData, RepositoryData } from "@root/types/repoInfo"
 
-const CACHE_REFRESH_INTERVAL = 60 * 1000 // 1 minutes
+// The SitesCacheService is responsible for storing information about
+// sites that is not updated live such as lastUpdated and previewImage.
+// Changes are polled at fixed intervals, stored in cache, and served
+// to avoid long load time from live querying.
+
+const REPO_DATA_REFRESH_INTERVAL = 60 * 1000 // 1 minutes
 
 export async function getAllRepoData(
   accessToken: string | undefined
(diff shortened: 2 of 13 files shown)


src/services/identity/SitesCacheService.ts
```diff
@@ -0,0 +1,93 @@
...
+  return _.flatten(allSites)
+}
+
+export class SitesCacheService {
+  private repoDataCache: RepositoryData[]
+
+  constructor() {
+    this.repoDataCache = []
+    this.renewCache()
+    setInterval(() => this.renewCache(), REPO_DATA_REFRESH_INTERVAL)
```
harishv7, line 82: Am curious if we use setInterval for a multi-instance setup, each instance will keep running this right?

In our default case of 2 instances, this will be then leading to 2 x 60 x 10 = 1200 token use?

src/services/identity/SitesCacheService.ts
```diff
@@ -0,0 +1,93 @@
...
+} from "@root/constants"
+import { genericGitHubAxiosInstance } from "@root/services/api/AxiosInstance"
+import type { GitHubRepositoryData, RepositoryData } from "@root/types/repoInfo"
+
+// The SitesCacheService is responsible for storing information about
+// sites that is not updated live such as lastUpdated and previewImage.
+// Changes are polled at fixed intervals, stored in cache, and served
+// to avoid long load time from live querying.
+
+const REPO_DATA_REFRESH_INTERVAL = 60 * 1000 // 1 minutes
```
harishv7, line 17: can this be injected into the constructor of the service?

src/services/identity/SitesCacheService.ts
kishore03109: should we make a safer interval here that is closer to our business req? 
ie, since our last updated is to the nearest date (please correct me if I am wrong about this), could we just shift to a daily interval instead?

src/services/identity/SitesCacheService.ts
caando: Okay 👍

src/services/identity/SitesCacheService.ts
caando: @harishv7 Yea each instance will be running the interval. Perhaps once every minutes is too frequent. How about once every 5 minutes instead, leading to 2 x 12 x 10 = 240 uses per hour.

caando: @kishore03109 I think this interval, lets call it refresh rate, is different from the business requirements of how precise the date is. Consider if the user makes an edit today, if the refresh rate is set to once per day, the user would still see the previous `lastUpdated` date until tomorrow, in which case the user will see `lastUpdated` yesterday.

harishv7: @caando A couple of questions/comments:

1. Can I check why did we decide to use a in-memory cache over writing to the DB? Then when fetching repo info, it will be present in the same call right?

2. My concern is not really the rate, but that we are making unnecessary duplicate calls from instances. The  way to do this on a distributed system is by using a "lock" and only execute the fetches when the lock is available. If the lock is not available then it means some other instance is already fetching the data and stores it to the DB.

3. Previously,I rmb that when adding back the field, we will enforce our own definition of a last-updated at - example when a site is published. Referencing IS-169:

> We can come up with our own definition of updated at. Example, if a user pushes merges a PR then we will consider the site as updated. Then we can introduce a new column in our table to track this and hence add it back to the frontend

Are we still doing this or is the scope changed?

harishv7: @caando thanks for explaining! Regarding 

> Multiple instances is a good point and I overlooked the fact that instances might not be synchronised, so depending on which instance is serving the user, he might see different lastUpdated info.

can we confirm with @NatMaeTan if that's an acceptable tradeoff?

harishv7: > Periodically fetching and storing in-memory allows us to set a consistent limit regardless of the traffic period at a relatively low cost.

The only issue I have with this is that the total is number of instances x interval frequency. While number of instance is dynamic, our token pool is not.

Consider the case where our traffic spikes. At this point we actually need more tokens to serve requests. However, as we scale, the increase in number of instances actually causes more token usage purely from this setInterval - tokens which could be otherwise used to serve the spike.

However, I believe this would be an unlikely scenario given the history of traffic spikes. @alexanderleegs @seaerchin @kishore03109 do we have such instances before where we had to auto-scale to multiple instances?

If its an unlikely scenario, we can then use a low frequency to fetch this data.

caando: As discussed yesterday, the maximum number of instances CMS can be scaled up to is 5. If we limit the interval to be 5mins, the maximum token uses per hour is 12x10x5=600, which is 600/50000=1.2% of our total capacity at the current moment. I think this usage is insignificant compared to user expenditure.

src/routes/v2/authenticated/sites.ts
```diff
@@ -63,7 +63,7 @@ export class SitesRouter {
     { userWithSiteSessionData: UserWithSiteSessionData }
   > = async (req, res) => {
     const { userWithSiteSessionData } = res.locals
-    const lastUpdated = await this.sitesService.getLastUpdated(
+    const lastUpdated = this.sitesService.getLastUpdated(
```
kishore03109, line 66: hey sanity check here, is this intended?

src/services/identity/SitesCacheService.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  constructor(refreshInterval: number) {
+    this.repoDataCache = []
+    this.refreshInterval = refreshInterval
+    this.renewCache()
+    setInterval(() => this.renewCache(), this.refreshInterval)
+  }
+
+  private async renewCache() {
+    this.repoDataCache = await getAllRepoData(undefined)
```
kishore03109, line 87: why is the access token here undefined? this wouldnt update all of our repos and only update the public repos correct?

kishore03109 (commented):
Mostly lgtm, just pending 2 clarifications ya

src/routes/v2/authenticated/sites.ts
caando: Yes! The function no longer returns a promise

src/services/identity/SitesCacheService.ts
caando: It is undefined so as to use the production tokens in the DB instead of supplying user's tokens. The prod tokens are inserted at the axios interceptors step so we simply need to not supply the headers.

kishore03109 (approved):

kishore03109:
rebase before merging ya

PR #832, opened 2023-07-11
Fix: siteLaunchEmailNotification

kishore03109:
## Problem

When site launch fails, it contains error message that is useful for us to consume. This might confuse end users if they were to receive such an email. 
Ideally, if there are any error messages, the user should be given some info if it is recoverable, else it should ping us for manual remediation. 

## Solution

As a first cut, the failure email heads straights to admin@isomer.gov.sg, so that we are able to take a look at it during launch. 

**Breaking Changes**

  - Details ...

### DIFF
--- a/src/services/infra/InfraService.ts
+++ b/src/services/infra/InfraService.ts
@@ -22,6 +22,7 @@ import {
   JobStatus,
   RedirectionTypes,
   REDIRECTION_SERVER_IP,
+  ISOMER_ADMIN_EMAIL,
 } from "@root/constants"
 import MissingSiteError from "@root/errors/MissingSiteError"
 import MissingUserEmailError from "@root/errors/MissingUserEmailError"
@@ -614,10 +615,8 @@ export default class InfraService {
       emailDetails = failureEmailDetails
     }
 
-    await mailer.sendMail(
-      message.requestorEmail,
-      emailDetails.subject,
-      emailDetails.body
-    )
+    const targetEmail = isSuccess ? message.requestorEmail : ISOMER_ADMIN_EMAIL
+
+    await mailer.sendMail(targetEmail, emailDetails.subject, emailDetails.body)
   }
 }
--- a/src/constants/constants.ts
+++ b/src/constants/constants.ts
@@ -57,3 +57,4 @@ export const ISOMER_ADMIN_REPOS = [
 export const INACTIVE_USER_THRESHOLD_DAYS = 60
 export const GITHUB_ORG_REPOS_ENDPOINT = `https://api.github.com/orgs/${ISOMER_GITHUB_ORG_NAME}/repos`
 export const REDIRECTION_SERVER_IP = config.get("redirectionServer.elasticIp")
+export const ISOMER_ADMIN_EMAIL = "admin@isomer.gov.sg"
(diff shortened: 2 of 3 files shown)


src/services/infra/InfraService.ts
```diff
@@ -614,10 +615,8 @@ export default class InfraService {
...
     }
 
-    await mailer.sendMail(
-      message.requestorEmail,
-      emailDetails.subject,
-      emailDetails.body
-    )
+    const targetEmail = isSuccess ? message.requestorEmail : ISOMER_ADMIN_EMAIL
+
+    await mailer.sendMail(targetEmail, emailDetails.subject, emailDetails.body)
```
alexanderleegs, line 620: Should we still send out an email that the process failed with a generic message though? Otherwise the requestor might be stuck waiting for a response when they should actually be trying again

kishore03109: @alexanderleegs 
There is no functionality to try again in the first cut of site launch feature! I think messages are useful if we provide actionable steps for them! (actually will check in with nat on this as well) cc. @NatMaeTan

kishore03109:
To @NatMaeTan 
during failure, do you want users to receive a generic error email, or just send to isomer team can already?

NatMaeTan:
> To @NatMaeTan during failure, do you want users to receive a generic error email, or just send to isomer team can already?

Sending them an email would be good comms, but also is there actionable for them to do? Or just standby for Isomer team?

src/services/infra/InfraService.ts
seaerchin: won't this imply that our users don't receive a failure email? under our old behaviour, users receive all emails (success + failure) but now they only receive success. 

if the concern is just about information clarity, maybe we can just send them a templated email that their site launch failed and we are aware of the issue + pls contact our helpdesk email

kishore03109: Hey @seaerchin this is a fair point!
The thing is here is that we dont differentiate a failed launch and a timeout launch, both are considered failures. My concern here is that there might be false positives that is being sent. 

Since this needs to be out by next week's release for next week's trial, and the next trial after this is not for another month, am opting for this flow for now and to be iterated on.

seaerchin: i'm fine with this if

1. it isn't a regression in behaviour (or this has been communicated to end user) -> given that prev site launches were manual and now is automatic, manual -> automatic users might be confused wrt why they aren't receiving any emails
2. @shazlithebestie signs off on this **and** emails the users on actual failure to inform them.

kishore03109: > automatic users might be confused wrt why they aren't receiving any emails
for now, there is only one user that will go through this 
> @shazlithebestie signs off on this and emails the users on actual failure to inform them.
nope, I am taking point on this, im handling the site launch for the user that we are trialing this feature out with! Ops has no implications

seaerchin: Sgtm

seaerchin: from clarification offline, the manual site launch process is done through a form. the email that is inputted in that form is prod ops emails so end user never sees it.

seaerchin (approved):

PR #835, opened 2023-07-14
chore(types): add warning message about changing the shape here

kishore03109:
## Problem

There is an duplicate of type definitions in this repo and infra. While this is not ideal, this was chosen to keep the infra repo private. There lies a risk when devs work on this feature that might change this shape and break corresponding infra. 

## Solution

Add a comment to warn dev not to change type def.


kishore03109:
@seaerchin thanks for raising your concern yesterday! While there was a warning message for the infra side, it didn't exist in our be side. 
Let me know if this is ok!

seaerchin (approved):
Lgtm

PR #836, opened 2023-07-14
ref(errors): migrate existing legacy errors to ts

seaerchin:
## Problem
Currently, our errors are in `.js`, which impacts our ability to make guarantees about the shape of the errors. This is an issue because we want to display error messages on our frontend based on what we return from our backend. In order to do this, we convert to `.ts` and guarantee the shape by extending from `BaseIsomerError`, which has a guarantee of said shape.

## Solution
1. Migrate from `.js` -> `.ts`


src/errors/ForbiddenError.ts
```diff
@@ -0,0 +1,13 @@
+// Import base error
+import { BaseIsomerError } from "@errors/BaseError"
+
+// eslint-disable-next-line import/prefer-default-export
```
kishore03109, line 4: Hmm if we feel as a team this is not impt eslint rule, could we change this at the global settings at a separate pr?

src/services/identity/__tests__/AuthService.spec.ts
```diff
@@ -15,11 +14,6 @@ const AuthService = new _AuthService({
 })
 
 describe("Auth Service", () => {
-  const mocksessionData = {
```
kishore03109, line 18: hmm why are we removing this ah

src/services/identity/__tests__/AuthService.spec.ts
```diff
@@ -57,11 +51,8 @@ describe("Auth Service", () => {
 
   it("should call axios successfully and bubble the error when the status is not 403 or 404", async () => {
     // Arrange
-    const expected = {
-      response: { status: "400" },
-    }
     mockGitHubService.checkHasAccess.mockRejectedValueOnce(
-      new BadRequestError(expected)
+      new BadRequestError("")
```
kishore03109, line 55: could I get context on why this returns a different shape now?

src/utils/json.ts
```diff
@@ -4,5 +4,6 @@ import { UnprocessableError } from "@root/errors/UnprocessableError"
 
 export const safeJsonParse = Result.fromThrowable(
   JSON.parse,
-  () => new UnprocessableError()
+  () =>
+    new UnprocessableError("The given string could not be parsed into json!")
```
kishore03109, line 8: Hmm may I get context here? why are we adding this param here ah

src/errors/ForbiddenError.ts
seaerchin: shure

src/services/identity/__tests__/AuthService.spec.ts
seaerchin: not used

src/services/identity/__tests__/AuthService.spec.ts
seaerchin: it still accepts the same shape - last time no type safety cos in `js` so it accepts object, but `message` is a `string`

src/utils/json.ts
seaerchin: `message` is a required `string`

harishv7 (approved):

kishore03109 (approved):

PR #840, opened 2023-07-18
Fix/restrict privatisation to email login

alexanderleegs:
This PR adds an additional restriction to the repo privatisation feature such that only email login users can see the feature - this is due to the prerequisite of restricting github access before being able to convert repos to private, as private repos are unable to set additional branch protection rules, meaning the PR approval requirement can be bypassed. To be reviewed in conjunction with PR #[1344](https://github.com/isomerpages/isomercms-frontend/pull/1344) on the isomercms-frontend repo.

### DIFF
--- a/src/routes/v2/authenticatedSites/settings.js
+++ b/src/routes/v2/authenticatedSites/settings.js
@@ -20,7 +20,8 @@ const { isPasswordValid } = require("@root/validators/validators")
 const { SettingsService } = require("@services/configServices/SettingsService")
 
 class SettingsRouter {
-  constructor({ settingsService }) {
+  constructor({ settingsService, authorizationMiddleware }) {
+    this.authorizationMiddleware = authorizationMiddleware
     this.settingsService = settingsService
     // We need to bind all methods because we don't invoke them from the class directly
     autoBind(this)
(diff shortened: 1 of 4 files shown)


src/routes/v2/authenticatedSites/settings.js
```diff
@@ -129,10 +130,12 @@ class SettingsRouter {
     router.post("/", attachRollbackRouteHandlerWrapper(this.updateSettingsPage))
     router.get(
       "/repo-password",
```
kishore03109, line 132: Extremely basic qn: where exactly in the FE are we calling this ah? I couldnt find any corresponding one, the only enpoint that the corresponding [FE pr](https://github.com/isomerpages/isomercms-frontend/pull/1344) seems to call is `/sites/<siteName>/settings`, is this intended?

kishore03109 (commented):
while the code changes lgtm, I am slightly confused about how this is called in the FE and if we are parsing the response of a string `OK` correctly when the user is not an email based login. Happy to hop onto a huddle to unblock you on this

src/routes/v2/authenticatedSites/settings.js
alexanderleegs: The corresponding FE PR is only to restrict the privatisation route on the FE to email login as well! the call site is located in the base PR, https://github.com/isomerpages/isomercms-frontend/pull/1316

kishore03109 (approved):

PR #847, opened 2023-07-20
fix: review request placeholder files

alexanderleegs:
This PR fixes an issue where compareDiff was not filtering placeholder files. This affected any review requests which involved placeholder files, as our frontend does not handle that specific file type. To fix this, we moved the filter from `compareFullReviewRequest` to `compareDiff`, which is called by both relevant endpoints.

## Tests 
- [ ] create a review request when the only file changes are placeholder files which are hidden
- [ ] FE should block creation of review request
- [ ] created a review request when there are file changes other than placeholder files, but is subsequently removed, leaving only placeholder file changes in the review request
- [ ] FE shows 0 file changes but does not block closing and merging for review request
- [ ] server should not throw any error

### DIFF
--- a/src/routes/v2/authenticated/review.ts
+++ b/src/routes/v2/authenticated/review.ts
@@ -527,7 +527,6 @@ export class ReviewsRouter {
           stagingLink
         )
       )
-    console.log(possibleReviewRequest)
 
     // NOTE: This method call (`getFullReviewRequest`) should be
     // in the result also but as it returns a promise at the moment,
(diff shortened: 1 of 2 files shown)


kishore03109:
@alexanderleegs @caando sorry may I get your help to dot down the edge cases that yall tested already?

caando:
@kishore03109 we tested this 2 cases:

Test: attempt to create a review request when the only file changes are placeholder files which are hidden
Response: FE blocks creation of review request

Test: created a review request when there are file changes other than placeholder files, but is subsequently removed, leaving only placeholder file changes in the review request
Response: FE shows 0 file changes but does not block closing and merging for review request

Neither cases led to a server error

kishore03109 (approved):

PR #859, opened 2023-07-25
feat(stats): track gh calls

seaerchin:
## Problem
Adds metric tracking for `pa-corp`.

## Solution
1. add `statsD` call for new endpoint
2. get the trial site as a `const` in code - **not very tied to this**, we could do either env var or db table instead of this but my understanding is that the trial site is unlikely to change + we are implementing ld soon, which makes the extra effort not worth.

see the data [here](https://app.datadoghq.com/dashboard/8tj-6k5-p2v/isomercms?fullscreen_end_ts=1690257331702&fullscreen_paused=false&fullscreen_section=overview&fullscreen_start_ts=1689652531702&fullscreen_widget=116425520844967&from_ts=1689647396729&to_ts=1690252196729&live=true) 

## Tests
- [ ] go to `pa-corp` 
- [ ] make an edit
- [ ] refresh
- [ ] go back to datadog 
- [ ] number shuold have changed (might have to wait abit)

### DIFF
--- a/src/services/api/AxiosInstance.ts
+++ b/src/services/api/AxiosInstance.ts
@@ -1,5 +1,6 @@
 import axios, { AxiosRequestConfig, AxiosResponse } from "axios"
 import { setupCache } from "axios-cache-interceptor"
+import _ from "lodash"
 
 import { config } from "@config/config"
 
@@ -12,7 +13,19 @@ import { tokenServiceInstance } from "@services/db/TokenService"
 
 import { statsService } from "../infra/StatsService"
 
-const GITHUB_EXPERIMENTAL_TRIAL_SITE = "pa-corp"
+const GITHUB_EXPERIMENTAL_TRIAL_SITES = ["pa-corp"]
+
+const REPOS_SUBSTRING = "repos/isomerpages"
+const extractRepoNameFromGithubUrl = (url: string): string => {
+  const idx = url.search(REPOS_SUBSTRING)
+  // NOTE: Should not hit here because we check that the url contains the site already
+  if (idx === -1) return ""
+  const ignoredLength = REPOS_SUBSTRING.length
+  return _.takeWhile(
+    url.slice(idx + ignoredLength + 1),
+    (char) => char !== "/"
+  ).join("")
+}
 
 const getIsEmailUserFromAuthMessage = (
   authMessage?: string | number | boolean
@@ -79,14 +92,15 @@ const respHandler = (response: AxiosResponse) => {
 }
 
 const githubApiInterceptor = (resp: AxiosResponse) => {
+  const fullUrl = `${resp.config.baseURL || ""}${resp.config.url || ""}`
   if (
     resp.status !== 304 &&
-    resp.config.url?.includes(GITHUB_EXPERIMENTAL_TRIAL_SITE) &&
+    _.some(GITHUB_EXPERIMENTAL_TRIAL_SITES, (site) => fullUrl.includes(site)) &&
     resp.config.method
   ) {
     statsService.incrementGithubApiCall(
       resp.config.method,
-      GITHUB_EXPERIMENTAL_TRIAL_SITE
+      extractRepoNameFromGithubUrl(fullUrl)
     )
   }
   return resp
--- a/src/services/api/AxiosInstance.ts
+++ b/src/services/api/AxiosInstance.ts
@@ -1,6 +1,5 @@
 import axios, { AxiosRequestConfig, AxiosResponse } from "axios"
 import { setupCache } from "axios-cache-interceptor"
-import { err } from "neverthrow"
 
 import { config } from "@config/config"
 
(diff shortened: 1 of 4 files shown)


kishore03109 (approved):
[Non-nit, feel quite strongly abt this]: can we please not commit to agency websites in production for testing? Lets use a test repo for this ya

[nit] To facilitate this ^  I suggest moving `GITHUB_EXPERIMENTAL_TRAIL_SITE` into the `constants.ts` file and make it into an array and then we can map this over. 

Since you have already tested this I am ok with the code as is, but this **_should_** not be the test case during our production

seaerchin:
@kishore03109 to be clear, we are releasing the github changes to `pa-corp`, so i did this to be sure that it works as expected. re: array vs single const - no strong opinion, @harishv7 will ld come out anytime soon/do we have plans to shift to more than 1 site?

we could potentially extend this to every site and just remove the `if` condition tbh, but i'd have to check w tim on billing cos there was an incident in the past

kishore03109:
@seaerchin could we skip this test for release? 
If we want to test functionality, how do you feel about env-varing it, test in some `a-test-v4` and the change to `pa-corp`? 

Mentioning this because have had agencies raise concern about us commenting on their repos before

seaerchin:
> @seaerchin could we skip this test for release? If we want to test functionality, how do you feel about env-varing it, test in some `a-test-v4` and the change to `pa-corp`?
> 
> Mentioning this because have had agencies raise concern about us commenting on their repos before either that or can just force push after to clean git history also can

ya, tested alr - feel free to skip for release. as mentioned in the PR description, i don't really mind either putting env var or table but didn't due to the reasons stated.

kishore03109:
@seaerchin sure, can j make a note in the pr desc that no need to test for release and merge alr bah

src/services/api/AxiosInstance.ts
```diff
@@ -73,6 +78,20 @@ const respHandler = (response: AxiosResponse) => {
   return response
 }
 
+const githubApiInterceptor = (resp: AxiosResponse) => {
+  if (
+    resp.status !== 304 &&
```
harishv7, line 83: Anything other than 304 means that a API call quota has been used right?

src/services/api/AxiosInstance.ts
```diff
@@ -11,6 +10,15 @@ import tracer from "@utils/tracer"
 import { customHeaderInterpreter } from "@root/utils/headerInterpreter"
 import { tokenServiceInstance } from "@services/db/TokenService"
 
+import { statsService } from "../infra/StatsService"
+
+const GITHUB_EXPERIMENTAL_TRIAL_SITE = "pa-corp"
```
harishv7, line 15: Is this possible to be an array? if we put as an array of sites, we can still track on per site level metric right, i.e. filter githubApiCalls made by sites?

src/services/infra/StatsService.ts
```diff
@@ -105,6 +106,15 @@ export class StatsService {
       version: Versions.V2,
     })
   }
+
+  incrementGithubApiCall = (method: Method, site: string) => {
+    this.statsD.increment("users.github.api", {
+      site,
```
harishv7, line 112: So on DD we will be able to filter by site to githubApiCallsMade ?

seaerchin: yes - i've an `if` statement just in case of billing cos there was a prev incident before

src/services/api/AxiosInstance.ts
seaerchin: yes

src/services/api/AxiosInstance.ts
seaerchin: yes

kishore03109 (commented):
sanity check here, have we tested this with a test site with the new changes?

harishv7:
> @kishore03109 to be clear, we are releasing the github changes **only** to `pa-corp`, so i did this to be sure that it works as expected. re: array vs single const - no strong opinion, @harishv7 will ld come out anytime soon/do we have plans to shift to more than 1 site?
> 
> we could potentially extend this to every site and just remove the `if` condition tbh, but i'd have to check w tim on billing cos there was an incident in the past

@seaerchin we plan to roll out to more sites upon evaluating the success of the MVP

seaerchin:
> sanity check here, have we tested this with a test site with the new changes?

yes, see the dashboard [here](https://app.datadoghq.com/dashboard/8tj-6k5-p2v/isomercms?fullscreen_end_ts=1690439768187&fullscreen_paused=false&fullscreen_section=overview&fullscreen_start_ts=1689834968187&fullscreen_widget=116425520844967&tpl_var_env%5B0%5D=local&from_ts=1689834770831&to_ts=1690439570831&live=true). also note that this was done from anotehr site (`chin-test`) to check functionality for multi-repo support

kishore03109 (approved):

PR #860, opened 2023-07-25
Feat/sgid experimental integration

alexanderleegs:
This PR introduces new endpoints to support sgid login. To be reviewed in conjunction with PR[#1369](https://github.com/isomerpages/isomercms-frontend/pull/1369) on the isomercms-frontend repo.

This PR introduces a new `SgidAuthRouter`, which introduces 2 new routes. GET `/v2/auth/sgid/auth-url` returns the sgID authentication url, and GET `/api/v3/auth/sgid/auth-redirect` takes two query parameters (`code` and `state`), performs the access token exchange and data retrieval with SGID, and sets the user's session.

A new db table (`sgidLogin`) has also been added to store details of user sessions.

SGID dev docs: https://docs.id.gov.sg/integrations-with-sgid/typescript-javascript

## New Environment Variables
- `SGID_CLIENT_ID`: sgID client ID provided during client registration
- `SGID_CLIENT_SECRET`: sgID client secret provided during client registration
- `SGID_PRIVATE_KEY`: sgID client private key provided during client registration
- `SGID_REDIRECT_URI`: URL to redirect to after authentication with sgID
Note that these new env vars are located in AWS SSM instead of elastic beanstalk, as we would hit the character limit otherwise.


lamkeewei (commented):
Left some comments! Not an expert here, so good to get another opinion!

src/routes/v2/sgidAuth.ts
```diff
@@ -0,0 +1,145 @@
...
+    never,
+    { redirectUrl: string } | ResponseErrorBody,
+    never,
+    never,
+    never
+  > = async (req, res) => {
+    this.sgidAuthService
+      .createSgidRedirectUrl()
+      .map((url) => {
+        res.json({ redirectUrl: url })
```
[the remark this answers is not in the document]
alexanderleegs: Thanks for catching this, fixed in https://github.com/isomerpages/isomercms-backend/pull/860/commits/d41bdf62f528a17a84415cdc430b979ff016ed76!

src/services/utilServices/SgidAuthService.ts
```diff
@@ -0,0 +1,110 @@
...
+    } catch (error) {
+      logger.error(`Error while creating sgid redirect url: ${error}`)
+      return err(new SgidCreateRedirectUrlError())
+    }
+  }
+
+  retrieveSgidAccessToken(authCode: string, state: string) {
+    return ResultAsync.fromPromise(
+      this.sgidLoginRepository.findOne({
+        where: { state },
```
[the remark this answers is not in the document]
alexanderleegs: Yup you're right, fixed in https://github.com/isomerpages/isomercms-backend/pull/860/commits/d41bdf62f528a17a84415cdc430b979ff016ed76

src/services/utilServices/SgidAuthService.ts
```diff
@@ -0,0 +1,117 @@
...
+    try {
+      const { url, nonce } = this.sgidClient.authorizationUrl({
+        state: "",
+        codeChallenge,
+        scope: ["openid", SGID_WORK_EMAIL_SCOPE],
+      })
+
+      return ResultAsync.fromPromise(
+        this.sgidLoginRepository.create({
+          state: sessionId,
```
lamkeewei, line 45: Technically the state that you pass in the authorization URL should match the `state` that you're storing. Perhaps rather than having the `sessionId` be the `state`, just have it as the `id`?

src/services/utilServices/SgidAuthService.ts
```diff
@@ -0,0 +1,117 @@
...
+        logger.error(`Error while querying sgid login database: ${error}`)
+        return new DatabaseError()
+      }
+    )
+      .andThen((loginDetails) => {
+        if (!loginDetails) {
+          logger.error(`Unable to find sgid login details`)
+          return errAsync(new DatabaseError())
+        }
+
```
lamkeewei, line 75: Could be good to verify `state` matches too. This is assuming that you decouple `state` from `sessionId`.

src/routes/v2/sgidAuth.ts
```diff
@@ -0,0 +1,157 @@
...
+        ) {
+          loginStatusCode = "500"
+        } else if (err instanceof AuthError) {
+          loginStatusCode = "401"
+        } else {
+          loginStatusCode = "500"
+        }
+      })
+
+    return res.redirect(`${FRONTEND_URL}/ogp-login?status=${loginStatusCode}`)
```
lamkeewei, line 140: `loginStatusCode` feels like a weird anti-pattern though. Shouldn't the frontend take the http response status code then redirect accordingly?

src/routes/v2/sgidAuth.ts
```diff
@@ -0,0 +1,157 @@
...
+
+    // Exchange the authorization code and code verifier for the access token, then use the access token to retrieve user's email
+    await this.sgidAuthService
+      .retrieveSgidAccessToken(authCode, sessionId)
+      .andThen(({ accessToken, sub }) =>
+        // We can immediately process the access token - we only need to retrieve the email from sgid
+        this.sgidAuthService.retrieveSgidUserEmail(accessToken, sub)
+      )
+      .andThen((email) => {
+        if (!email)
```
lamkeewei, line 98: **Optional:** Not sure if it's worthwhile to do an explicit checked for `endsWith('@open.gov.sg')`. Understand that we're relying on the fact that only OGP emails are available for now.

kwajiehao:
As discussed with @alexanderleegs, can consider moving to storing the `codeVerifier` in a cookie (with proper security config) so that you don't need to maintain additional infra. But will leave it to you to decide whether you want to do it in this PR or in subsequent ones

src/services/utilServices/SgidAuthService.ts
alexanderleegs: After chatting with Jie Hao, we've opted to remove the db store completely - https://github.com/isomerpages/isomercms-backend/pull/860/commits/22ce8f90bb6325d20df7d4305f7688f7ff036821

src/services/utilServices/SgidAuthService.ts
alexanderleegs: See above!

lamkeewei (commented):
@alexanderleegs Changes lgtm overall! But didn't test this on staging! Will let another Isomer engineer to approve after testing.

QiluXie (approved):
Tested on staging for several logins and worked well, lgtm! Thanks @alexanderleegs & @harishv7

kwajiehao:
lgtm!

PR #861, opened 2023-07-26
fix(reviewRequestService): upsert in case of concurrent api calls

kishore03109:
[wip i fixing test cases]
## Problem
[Credits to @caando for code changes] 
Currently, when calling the the route `/viewed` twice, this might lead to errors thrown in the backend. While the logic is sane, the reason why the errors occur is because there is a race condition when two threads attempt to create at the same time. 
Screenshot 2023-07-26 at 10 16 01 AM

## Solution

`Upsert` rather than `create`. Note that the `comments.spec.ts` already needed these file changes to work. 

**Limitations**

~~This not the ideal way to solve race conditions. The ideal method is to create a lock and then keep the function `markAllReviewRequestsAsViewed` atomic. However, at this point of time, that might be over-engineering here as the only impact that could happen with unlikely concurrent calls to `markAllReviewRequestsAsViewed` is the dashboard does not turn blue to mark it as unread.~~ 

Following Chin's suggestion of using transactions, this limitation is no longer valid. 

## Tests 

e2e for notifications pass (failed without these changes) 
Screenshot 2023-07-27 at 2 19 31 PM

**Misc** 
There is another issue of our `/viewed` being called twice, which is why this bug was noticed in the e2e testing in the first place. While that might incur separate effort to investigate, this PR intends to quick fix this regardless as this can still occur say when a user has 2 tabs open and refreshes them simultaneously.

### DIFF
--- a/src/services/review/ReviewRequestService.ts
+++ b/src/services/review/ReviewRequestService.ts
@@ -512,7 +512,7 @@ export default class ReviewRequestService {
       // Using map here to allow creations to be done concurrently
       // But we do not actually need the result of the view creation
       requestIdsToMarkAsViewed.map(async (requestId) =>
-        this.reviewRequestView.create({
+        this.reviewRequestView.upsert({
           reviewRequestId: requestId,
           siteId: site.id,
           userId,
--- a/src/services/review/ReviewRequestService.ts
+++ b/src/services/review/ReviewRequestService.ts
@@ -10,6 +10,7 @@ import {
   combine,
 } from "neverthrow"
 import { Op, ModelStatic } from "sequelize"
+import { Sequelize } from "sequelize-typescript"
 
 import UserWithSiteSessionData from "@classes/UserWithSiteSessionData"
 
@@ -31,6 +32,7 @@ import { NotFoundError } from "@root/errors/NotFoundError"
 import PageParseError from "@root/errors/PageParseError"
 import PlaceholderParseError from "@root/errors/PlaceholderParseError"
 import RequestNotFoundError from "@root/errors/RequestNotFoundError"
+import logger from "@root/logger/logger"
 import {
   BaseEditedItemDto,
   CommentItem,
@@ -102,6 +104,8 @@ export default class ReviewRequestService {
 
   private readonly configService: ConfigService
 
+  private readonly sequelize: Sequelize
+
   constructor(
     apiService: typeof ReviewApi,
     users: ModelStatic<User>,
@@ -110,7 +114,8 @@ export default class ReviewRequestService {
     reviewMeta: ModelStatic<ReviewMeta>,
     reviewRequestView: ModelStatic<ReviewRequestView>,
     pageService: PageService,
-    configService: ConfigService
+    configService: ConfigService,
+    sequelize: Sequelize
   ) {
     this.apiService = apiService
     this.users = users
@@ -120,6 +125,7 @@ export default class ReviewRequestService {
     this.reviewRequestView = reviewRequestView
     this.pageService = pageService
     this.configService = configService
+    this.sequelize = sequelize
   }
 
   compareDiff = (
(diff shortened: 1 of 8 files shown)


src/services/review/ReviewRequestService.ts
```diff
@@ -512,7 +512,7 @@ export default class ReviewRequestService {
       // Using map here to allow creations to be done concurrently
       // But we do not actually need the result of the view creation
       requestIdsToMarkAsViewed.map(async (requestId) =>
-        this.reviewRequestView.create({
+        this.reviewRequestView.upsert({
```
seaerchin, line 515: hmm, maybe the correct thing to do would be to use a transaction here. wdyt?

kishore03109: [ok]([5a7a311](https://github.com/isomerpages/isomercms-backend/pull/861/commits/5a7a311d962761db16e86d73811e18fa098f8b06))

alexanderleegs (approved):
lgtm - let's still file an issue for the root cause of the issue though

src/services/review/ReviewRequestService.ts
```diff
@@ -481,47 +487,57 @@ export default class ReviewRequestService {
...
-          reviewRequestId: requestId,
-          siteId: site.id,
-          userId,
-          // This field represents the user opening the review request
-          // itself, which the user has not done so yet at this stage.
-          lastViewedAt: null,
-        })
-      )
-    )
+        await Promise.all(
```
seaerchin, line 520: actually i just realised this is because of duplicates right? we can just call `_.uniq(requestIdsToMarkAsViewed`. transaction seems abit heavy here tbh because the view status isn't super important (we don't really care if the notification is mistakenly shown as not viewed). also, i think it'll still run into the issue of duplicates, no?

src/services/review/__tests__/ReviewRequestService.spec.ts
seaerchin, line 1: would be godo to have a test here to ensure this bug is fixed + doesn't have regression issues

src/services/review/ReviewRequestService.ts
seaerchin, line 520: just realised that this is cos of repeated attempts to create - in this case, the `transaction` might not help at all actually because we'd still try to create the repeated entry. 

maybe we should just use `_.unique(requestIdsToMarkAsViewed)` to prevent dups and accept hte race condition of having some views being outdated (actually seen but presented as not), as it's relatively minor.

src/services/review/__tests__/ReviewRequestService.spec.ts
seaerchin, line 1: should write a simple spec to prevent regression

src/services/review/ReviewRequestService.ts
kishore03109: > in this case, the transaction might not help at all actually because we'd still try to create the repeated entry.

hmm i am confused here, why would it try to create a repeated entry ya? there is still a race condition here, but in this case we are "safely" failing here right due to a rollback? so rather than making our server return 500, we are increasing the stability of our be here no? 

> maybe we should just use _.unique(requestIdsToMarkAsViewed) to prevent dups and accept hte race condition of having some views being outdated (actually seen but presented as not), as it's relatively minor.

hmm wait ah this is due to duplicate api calls leh, how would this help here ah?

src/services/review/__tests__/ReviewRequestService.spec.ts
kishore03109: [a1c8782](https://github.com/isomerpages/isomercms-backend/pull/861/commits/a1c8782098acd092d9d5111f69879b5513723078)

src/services/review/ReviewRequestService.ts
seaerchin: i think the error stems from an attempt to create when there's an existing entry + a unique constraint. this is **not** due to concurrency issues and you can replicate this just by having the `requestIds` to be identical and do the creation sequentially.

because the error is due to duplicates in `requestIds`, the fix is to either remove the duplicates or allow the update but have it do nth.

kishore03109: wait we do have an initial call here right, in line 493? 
```
 const requestsViewed = await this.reviewRequestView.findAll({
          where: {
            siteId: site.id,
            userId,
          },
          transaction,
        })
```
we get the list of requestsViews, diff it using the current open review requests, and iff there are active requests that **_have not been_** viewed yet, create a entry in the `reviewRequestViews`?

src/services/review/__tests__/ReviewRequestService.spec.ts
seaerchin: actually because the transaction here is mocked, it mihgt not be very useful. we might wish to write an integration spec and verify that it fails initially (without the fix) and passes later on with the fix

src/services/review/ReviewRequestService.ts
seaerchin: if you've tried that this solution works on local, then please merge it. i might have a misconception, so it shouldn't block the PR

src/services/review/__tests__/ReviewRequestService.spec.ts
kishore03109: ^ ok, i tackle this in a separate pr cuz at least the changes should be merged in by tmr's release uh

PR #862, opened 2023-07-26
feat(axios): add method to github api logs

caando:
## Problem

Currently our logs of requests to GitHub api only has the endpoint and not the request method. This presents difficulties in identifying which api its calling and whether its a read/write request.

Closes IS-366

## Solution

Add request method to GitHub api logs. 

**Breaking Changes**

### DIFF
--- a/src/services/api/AxiosInstance.ts
+++ b/src/services/api/AxiosInstance.ts
@@ -37,8 +37,11 @@ const requestFormatter = async (axiosConfig: AxiosRequestConfig) => {
           },
         },
       })
+      const requestMethod = axiosConfig.method
+        ? axiosConfig.method
+        : "undefined method"
       logger.info(
-        `Email login user made call to Github API: ${axiosConfig.url}`
+        `Email login user made ${requestMethod} call to Github API: ${axiosConfig.url}`
       )
       return {
         ...axiosConfig,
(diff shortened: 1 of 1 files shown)


src/services/api/AxiosInstance.ts
```diff
@@ -37,8 +37,11 @@ const requestFormatter = async (axiosConfig: AxiosRequestConfig) => {
           },
         },
       })
+      const requestMethod = axiosConfig.method
+        ? axiosConfig.method
+        : "undefined method"
```
QiluXie, line 42: If `axiosConfig` is `undefined`, accessing its property `method` could result in a type error. Could we make it more robust and concise like:
`const requestMethod = axiosConfig?.method ?? "undefined method";`

src/services/api/AxiosInstance.ts
```diff
@@ -57,7 +60,12 @@ const requestFormatter = async (axiosConfig: AxiosRequestConfig) => {
       },
     },
   })
-  logger.info(`Github login user made call to Github API: ${axiosConfig.url}`)
+  const requestMethod = axiosConfig.method
+    ? axiosConfig.method
+    : "undefined method"
```
QiluXie, line 65: Similarly as the previous comment, could we modify this to:
`const requestMethod = axiosConfig?.method ?? "undefined method";`

QiluXie (changes requested):

src/services/api/AxiosInstance.ts
caando: simplified ternary operator to `??`

src/services/api/AxiosInstance.ts
caando: simplified ternary operator to `??`

QiluXie (approved):
lgtm

PR #872, opened 2023-08-01
chore: update site creation to block squash and merge

alexanderleegs:
This PR introduces a minor change to site creation to add in the blocking of squash and merge - the reason for this is to prevent users from accidentally using squash and merge to merge prs, which would cause issues in future when attempting to merge staging to master again. See https://github.com/isomerpages/isomer-tooling/pull/47 for the one off fix for existing repos


kishore03109 (approved):
LOL i just commented on the other pr to do this

PR #873, opened 2023-08-02
feat(siteLaunch): restrict to admin

kishore03109:
## Problem

No access controls for admins, can be reviewed with [fe pr](https://github.com/isomerpages/isomercms-frontend/pull/1384).

## Solution

Only admins can call the BE route.

### DIFF
--- a/src/services/infra/InfraService.ts
+++ b/src/services/infra/InfraService.ts
@@ -265,6 +265,11 @@ export default class InfraService {
     SiteLaunchCreateParams,
     AmplifyError | SiteLaunchError | MissingUserEmailError
   > {
+    const isAdmin = email.endsWith(".gov.sg")
+
+    if (!isAdmin) {
+      return errAsync(new SiteLaunchError("Not authorized to launch site"))
+    }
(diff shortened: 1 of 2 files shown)


src/services/infra/InfraService.ts
```diff
@@ -265,6 +265,11 @@ export default class InfraService {
     SiteLaunchCreateParams,
     AmplifyError | SiteLaunchError | MissingUserEmailError
   > {
+    const isAdmin = email.endsWith(".gov.sg")
+
+    if (!isAdmin) {
+      return errAsync(new SiteLaunchError("Not authorized to launch site"))
+    }
```
alexanderleegs, line 272: Just checking, is there a reason why we don't attach the `checkIsSiteAdmin` middleware to this instead (so checking for site admin, rather than just checking the email)

kishore03109: O.o Derp moment  
ae851f77f071e97bcfb6d1feb249b509de3f0f62

alexanderleegs (approved):
lgtm

PR #874, opened 2023-08-02
feat(GGs): Add drop-in replacement RepoService

dcshzj:
## Problem

Closes IS-351 and IS-405.

## Solution

**Breaking Changes**

**Features**:

- This PR adds a new RepoService that aims to be a drop-in replacement of the legacy GitHubService
    - This RepoService **extends** (instead of implements) the GitHubService so that we can progressively TypeScript-ify GitHubService (and eventually replace it). This allows existing methods to continue being called, keeps RepoService backward compatible and methods can be re-implemented later (such as `read` and `readDirectory` in this PR).
    - Eventually, this RepoService will be independent from GitHubService, and other classes can potentially use both concurrently. RepoService will only have methods that are specific to an Isomer site repository (mainly just CRUD operations), whereas GitHubService will only have methods that require interfacing with GitHub (such as privatising repos).
- Both `git checkout` and `git pull` involve writing to disk, which causes issues in terms of concurrency due to file locks. Hence, they have been modified to only checkout when necessary, and the known concurrency issue of `git pull` is safely recovered from.

This PR is the second of a 2-part PR. The eventual architecture should look something like this:

IS-320

## Tests

1. Unit tests
2. End-to-end testing to ensure no regression

## Deploy Notes

- `WHITELISTED_GIT_SERVICE_REPOS` : A comma-separated list of repositories that will use the GitFileSystemService instead of calling GitHub directly.

[IS-320]: https://isomeropengov.atlassian.net/browse/IS-320?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -74,12 +74,6 @@ const config = convict({
       format: "required-string",
       default: "",
     },
-    whitelistRepos: {
-      doc: "Comma-separated list of whitelisted repos for local Git service",
-      env: "WHITELISTED_GIT_SERVICE_REPOS",
-      format: String,
-      default: "",
-    },
   },
   mutexTableName: {
     doc: "Name of the DynamoDB table used for mutexes",
@@ -408,6 +402,14 @@ const config = convict({
       doc: "URL to redirect to after authentication with sgID",
       env: "SGID_REDIRECT_URI",
       format: "required-string",
+      default: "",
+    },
+  },
+  featureFlags: {
+    whitelistRepos: {
+      doc: "Comma-separated list of whitelisted repos for local Git service",
+      env: "WHITELISTED_GIT_SERVICE_REPOS",
+      format: String,
       default: "",
     },
   },
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -406,7 +406,7 @@ const config = convict({
     },
   },
   featureFlags: {
-    whitelistRepos: {
+    ggsWhitelistedRepos: {
       doc: "Comma-separated list of whitelisted repos for local Git service",
       env: "WHITELISTED_GIT_SERVICE_REPOS",
       format: String,
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -74,6 +74,12 @@ const config = convict({
       format: "required-string",
       default: "",
     },
+    whitelistRepos: {
+      doc: "Comma-separated list of whitelisted repos for local Git service",
+      env: "WHITELISTED_GIT_SERVICE_REPOS",
+      format: String,
+      default: "",
+    },
   },
   mutexTableName: {
     doc: "Name of the DynamoDB table used for mutexes",
--- a/src/services/db/GitFileSystemService.ts
+++ b/src/services/db/GitFileSystemService.ts
@@ -315,69 +315,75 @@ export default class GitFileSystemService {
     repoName: string,
     directoryPath: string
   ): ResultAsync<GitDirectoryItem[], GitFileSystemError> {
-    return ResultAsync.fromPromise(
-      fs.promises.stat(`${EFS_VOL_PATH}/${repoName}/${directoryPath}`),
-      (error) => {
-        if (error instanceof Error) {
-          return new GitFileSystemError(error.message)
-        }
-
-        logger.error(`Error when getting ${directoryPath} stats: ${error}`)
-        return new GitFileSystemError("An unknown error occurred")
-      }
-    )
-      .andThen((stats) => {
-        if (!stats.isDirectory()) {
-          return errAsync(
-            new GitFileSystemError(
-              `Path "${directoryPath}" is not a directory in repo "${repoName}"`
-            )
-          )
-        }
-        return okAsync(true)
-      })
-      .andThen(() =>
-        ResultAsync.fromPromise(
-          fs.promises.readdir(`${EFS_VOL_PATH}/${repoName}/${directoryPath}`, {
-            withFileTypes: true,
-          }),
-          (error) => {
-            if (error instanceof Error) {
-              return new GitFileSystemError(error.message)
-            }
-
-            logger.error(`Error when reading ${directoryPath}: ${error}`)
-            return new GitFileSystemError("An unknown error occurred")
+    return this.pull(repoName).andThen(() =>
+      ResultAsync.fromPromise(
+        fs.promises.stat(`${EFS_VOL_PATH}/${repoName}/${directoryPath}`),
+        (error) => {
+          if (error instanceof Error) {
+            return new GitFileSystemError(error.message)
           }
-        )
-      )
-      .andThen((directoryContents) => {
-        const resultAsyncs = directoryContents.map((directoryItem) => {
-          const isDirectory = directoryItem.isDirectory()
-          const { name } = directoryItem
-          const path = directoryPath === "" ? name : `${directoryPath}/${name}`
-          const type = isDirectory ? "dir" : "file"
 
-          return this.getGitBlobHash(repoName, path)
-            .orElse(() => okAsync(""))
-            .andThen((sha) => {
-              const result: GitDirectoryItem = {
-                name,
-                type,
-                sha,
-                path,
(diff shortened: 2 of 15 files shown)


src/config/config.ts
```diff
@@ -74,6 +74,12 @@ const config = convict({
       format: "required-string",
       default: "",
     },
+    whitelistRepos: {
```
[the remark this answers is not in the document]
harishv7: actually there can be a new section for feature flags like this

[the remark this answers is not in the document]
dcshzj: That's a good point, done in 3e03720931a86d79d9fa58afdddcdaa035f8cfff and 5caac04e98505980c54d736178c9447feff05525.

src/services/db/RepoService.ts
```diff
@@ -0,0 +1,245 @@
...
+      return result.value
+    }
+
+    return await super.read(sessionData, {
+      fileName,
+      directoryName,
+    })
+  }
+
+  // TODO: This is no longer used, remove it
```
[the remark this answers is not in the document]
dcshzj: I was not able to find any usages of this function anywhere, and I think it was removed in PR #740 without the function being removed. I added this todo here in case I missed out something + felt that it would be out of scope of this PR.

src/services/db/GitFileSystemService.ts
```diff
@@ -200,17 +235,48 @@ export default class GitFileSystemService {
...
+          this.git.cwd(`${EFS_VOL_PATH}/${repoName}`).pull(),
+          (error) => {
+            // Full error message 1: Your configuration specifies to merge
+            // with the ref 'refs/heads/staging' from the remote, but no
+            // such ref was fetched.
+            // Full error message 2: error: cannot lock ref
+            // 'refs/remotes/origin/staging': is at <new sha> but expected <old sha>
+            // Full error message 3: Cannot fast-forward your working tree.
+            // Full error message 4: Need to specify how to reconcile divergent branches.
+            // These are known errors that can be safely ignored
```
harishv7, line 249: just to confirm the above errors does not mean git pull has failed?

Will the git pull succeed in updating the local file contents for the above scenarios?

src/services/db/RepoService.ts
```diff
@@ -0,0 +1,247 @@
...
+  ): Promise<GitDirectoryItem[]> {
+    if (this.isRepoWhitelisted(sessionData.siteName)) {
+      logger.info("Reading directory from local Git file system")
+      const result = await this.gitFileSystemService.listDirectoryContents(
+        sessionData.siteName,
+        directoryName
+      )
+
+      if (result.isErr()) {
+        throw result.error
```
harishv7, line 164: do we want to throw errors? in some places we use neverthrow, but some places we throw. is it because of the way upstream calls reposervice?

src/services/db/RepoService.ts
```diff
@@ -0,0 +1,247 @@
...
+
+  mergePullRequest(siteName: string, pullRequestNumber: number) {
+    return ReviewApi.mergePullRequest(siteName, pullRequestNumber)
+  }
+
+  approvePullRequest(siteName: string, pullRequestNumber: number) {
+    return ReviewApi.approvePullRequest(siteName, pullRequestNumber)
+  }
+
+  async getComments(siteName: string, pullRequestNumber: number) {
```
harishv7, line 81: why do we need to re define the methods from super class? When extending, these methods will be available on RepoService right?

Is it for future modifications?

src/services/db/GitFileSystemService.ts
dcshzj: These errors mean that _one_ of the git pulls have failed, but overall it would have succeeded in updating the local file contents.

Due to the number of read calls made when loading a page, it is possible that multiple git pulls are running concurrently. The first git pull would have succeeded eventually, as it is the one that has acquired the necessary locks to update the local files. However, it is possible that subsequent git pulls happening at the same time would throw errors (like the 4 here) as they would have "seen" the local directory in an inconsistent state (the first git pull is in progress). Hence, it is safe for us to ignore these errors, as the first git pull would have succeeded.

src/services/db/RepoService.ts
dcshzj: Yup, the way the upstream calls RepoService is that they expect errors to be thrown, so we would have to use this way for now until the various services are updated to handle neverthrow.

src/services/db/GitFileSystemService.ts
```diff
@@ -112,24 +113,58 @@ export default class GitFileSystemService {
...
+          this.git.cwd(`${EFS_VOL_PATH}/${repoName}`).checkout(BRANCH_REF),
+          (error) => {
+            if (error instanceof GitError) {
+              return new GitFileSystemError(error.message)
+            }
+
+            logger.error(`Error when checking out ${BRANCH_REF}: ${error}`)
+            return new GitFileSystemError("An unknown error occurred")
+          }
+        ).andThen(() => okAsync<true, never>(true))
```
kishore03109, line 142: [nit] moving forward, we can just `okAsync<true>(true)`

src/services/db/GitFileSystemService.ts
kishore03109: in realtion to: `Full error message 4: Need to specify how to reconcile divergent branches.`
what is our behvaiour when there is a conflict? do we force merge?

src/services/db/RepoService.ts
dcshzj: Yup, this is for future modifications. Eventually, we should move towards having all of these functions having type definitions, but for now it's just an explicit passthrough to GitHubService.

src/services/db/GitFileSystemService.ts
dcshzj: Hmm there shouldn't be a conflict because based on this MVP, as only reads are being performed. A conflict will only occur if the local copy has additional changes beyond the original HEAD commit that Git cannot recover from.

src/services/db/RepoService.ts
```diff
@@ -0,0 +1,247 @@
...
+
+    return await super.readDirectory(sessionData, {
+      directoryName,
+    })
+  }
+
+  async update(
+    sessionData: any,
+    { fileContent, sha, fileName, directoryName }: any
+  ): Promise<any> {
```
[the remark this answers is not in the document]
dcshzj: These are methods that have not been typed yet, since we are only serving reads as part of the MVP. It is `any` now since downstream usages are mostly in JavaScript so they will not be affected by the types returned here.

src/services/db/GitFileSystemService.ts
dcshzj: Done in bd30514d4857b761be2ddef3f4ba2db0cc6b2188!

dcshzj:
@isomerpages/iso-engineers Going to stop this PR at just supporting reads, commits and pushes (IS-351 and IS-405). Existing functionality should not be affected if `WHITELISTED_GIT_SERVICE_REPOS` is an empty string.

src/services/db/GitFileSystemService.ts
```diff
@@ -112,6 +115,39 @@ export default class GitFileSystemService {
       })
   }
 
+  // Ensure that the repository is in the BRANCH_REF branch
+  ensureCorrectBranch(repoName: string): ResultAsync<true, GitFileSystemError> {
```
harishv7, line 119: this function's job to ensure a repo is on a correct branch. shouldn't it take in the 2 dependencies - repoName and branchName as 2 arguments and allow upstream to call `ensureCorrectBranch(repoName, branchName)`

harishv7 (approved):

src/services/db/GitFileSystemService.ts
dcshzj: Hmm for this one, we are making sure that the branch is at `staging`, as the CMS does not work on any other branch when performing changes. Hence, it only takes in one argument in this case.

PR #877, opened 2023-08-03
feat: add platform predeploy hooks

harishv7:
## Problem

This PR contains all the scripts we need to run on an infra-level to make CMS BE ready to integrate with our Git File System Service.

Closes IS-317, IS-319, IS-337, IS-375, IS-390

## Solution

This PR contains pre-deploy hooks which we instruct Elastic Beanstalk to perform before starting app deployment.

Note: There are 3 different types of hooks - `prebuild`, `predeploy` and `postdeploy`. We are using the predeploy hooks as we need to setup these config before our app starts serving requests.

There are 5 scripts corresponding to the following actions:

### Mount EFS

This fetches a file system ID & mount location from SSM and mounts the corresponding EFS volume. 

Note that we have decided to perform EFS creation manually outside the use of scripts/hooks as it is a 1-time operation

### Fetch SSH Keys

Next, we will be using SSH authentication for GitHub. So we create a SSH pub/private keypair. We store these as secure strings in SSM.

We fetch these SSH keys on predeploy and copy it to the `.ssh/` of `webapp` user.

The user we add to here is important. There are 3 types of users: `root`, `ec2-user` and `webapp`.

When our app runs, it runs as `webapp` user. When EB runs the predeploy or other hooks, it runs as `root`. Hence, it is important to ensure our files are able to be accessed by `webapp` user.

### Add to SSH Config

Next, we add to the SSH config to use the SSH keys for auth to Github.com as the host.

### Add Github to Known Hosts

Often when cloning for first time, we will encounter the prompt whether to trust the host. On an automated environment, we won’t be able to provide this manual input.

Hence, we add it to the `known_hosts` file ahead of time.

### Setup Git profile

To perform Git CLI commands, we need to setup the git user profile/config. We do this via the `.gitconfig` file for the `webapp` user's directory.

**Breaking Changes**

- [] Yes - this PR contains breaking changes
  - Details ...

## Tests

We will need to test a few things manually:
1. If EFS is mounted successfully
2. Whether 2 instances are able to see/edit the shared files in EFS
3. If SSH configs are set correctly within `.ssh/`
4. Whether we are able to git clone, git pull, git push from isomerpages

## Deploy Notes

**Note: This PR should only be merged when we are ready to rollout GGS**

**Dependencies**:
- We will need to create an EFS volume in the staging and prod VPCs beforehand
- We will need to set the EFS ID into the SSM parameters

**New SSM parameters**:
1. `EFS_ID` - exposed on SSM to mount to the correct EFS vol
    1. `STAGING_EFS_ID`
    2. `PROD_EFS_ID`
2.  `STAGING_GIT_USER_NAME`
    1. `STAGING_GIT_USER_EMAIL`
    2. `PROD_GIT_USER_NAME`
    3. `PROD_GIT_USER_EMAIL`
3. SSH keys - exposed on SSM to be fetched at deployment time
    1. `STAGING_SSH_PUBLIC_KEY`
    2. `STAGING_SSH_PRIVATE_KEY`
    3. `PROD_SSH_PUBLIC_KEY`
    5. `PROD_SSH_PRIVATE_KEY`
  
- `EFS_VOL_PATH`: This should have been already added from @dcshzj 's PRs. This should point to `/efs/repos`.

### DIFF
--- a/.platform/hooks/predeploy/01_mount_efs.sh
+++ b/.platform/hooks/predeploy/01_mount_efs.sh
@@ -1,0 +1,75 @@
+#!/bin/bash
+
+# Install AWS CLI if it's not installed yet
+which aws >/dev/null 2>&1
+if [ $? -ne 0 ]; then
+    echo "Installing AWS CLI..."
+    curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip"
+    unzip awscliv2.zip
+    sudo ./aws/install
+    rm awscliv2.zip
+    rm -rf aws
+    echo "AWS CLI installed successfully."
+fi
+
+ENV_TYPE=$(/opt/elasticbeanstalk/bin/get-config environment -k SSM_PREFIX)
+
+AWS_REGION="ap-southeast-1"
+FILE_SYSTEM_ID_PARAM_NAME="${ENV_TYPE}_EFS_ID"
+MOUNT_DIRECTORY_PARAM_NAME="${ENV_TYPE}_EFS_MOUNT_DIR"
+
+FILE_SYSTEM_ID=$(aws ssm get-parameter --name $FILE_SYSTEM_ID_PARAM_NAME --region $AWS_REGION --query 'Parameter.Value' --output text)
+MOUNT_DIRECTORY=$(aws ssm get-parameter --name $MOUNT_DIRECTORY_PARAM_NAME --region $AWS_REGION --query 'Parameter.Value' --output text)
+
+yum install -y amazon-efs-utils
+
+EFS_MOUNT_DIR=$MOUNT_DIRECTORY
+EFS_FILE_SYSTEM_ID=$FILE_SYSTEM_ID
+
+echo "Mounting EFS filesystem ${EFS_FILE_SYSTEM_ID} to directory ${EFS_MOUNT_DIR} ..."
+
+echo 'Stopping NFS ID Mapper...'
+service rpcidmapd status &> /dev/null
+if [ $? -ne 0 ] ; then
+    echo 'rpc.idmapd is already stopped!'
+else
+    service rpcidmapd stop
+    if [ $? -ne 0 ] ; then
+        echo 'ERROR: Failed to stop NFS ID Mapper!'
+        exit 1
+    fi
+fi
+
+echo 'Checking if EFS mount directory exists...'
+if [ ! -d ${EFS_MOUNT_DIR} ]; then
+    echo "Creating directory ${EFS_MOUNT_DIR} ..."
+    mkdir -p ${EFS_MOUNT_DIR}
+    if [ $? -ne 0 ]; then
+        echo 'ERROR: Directory creation failed!'
+        exit 1
+    fi
+else
+    echo "Directory ${EFS_MOUNT_DIR} already exists!"
+fi
+
+mountpoint -q ${EFS_MOUNT_DIR}
+if [ $? -ne 0 ]; then
+    echo "mount -t efs -o tls ${EFS_FILE_SYSTEM_ID}:/ ${EFS_MOUNT_DIR}"
+    mount -t efs -o tls ${EFS_FILE_SYSTEM_ID}:/ ${EFS_MOUNT_DIR}
+    if [ $? -ne 0 ] ; then
+        echo 'ERROR: Mount command failed!'
+        exit 1
+    fi
+    chmod 777 ${EFS_MOUNT_DIR}
+    runuser -l  ec2-user -c "touch ${EFS_MOUNT_DIR}/it_works"
+    if [[ $? -ne 0 ]]; then
+        echo 'ERROR: Permission Error!'
+        exit 1
+    else
+        runuser -l  ec2-user -c "rm -f ${EFS_MOUNT_DIR}/it_works"
+    fi
+else
+    echo "Directory ${EFS_MOUNT_DIR} is already a valid mountpoint!"
+fi
+
+echo 'EFS mount complete.'
--- a/.platform/hooks/predeploy/01_mount_efs.sh
+++ b/.platform/hooks/predeploy/01_mount_efs.sh
@@ -1,5 +1,29 @@
 #!/bin/bash
 
+# Note the following has been adapted from https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/services-efs.html
+
+###################################################################################################
+#### Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+####
+#### Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file
+#### except in compliance with the License. A copy of the License is located at
+####
+####     http://aws.amazon.com/apache2.0/
+####
+#### or in the "license" file accompanying this file. This file is distributed on an "AS IS"
+#### BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#### License for the specific language governing permissions and limitations under the License.
+###################################################################################################
+
+###################################################################################################
+#### This configuration file mounts an Amazon EFS file system to a specified directory.
+#### Both the directory and the EFS file system ID to be mounted is retrieved from AWS SSM.
+####
+#### If your environment and file system are in a custom VPC, you must configure the VPC to allow
+#### DNS resolution and DNS host names. See this topic in the VPC User Guide for more information:
+####    http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/vpc-dns.html
+###################################################################################################
+
 # Install AWS CLI if it's not installed yet
 which aws >/dev/null 2>&1
 if [ $? -ne 0 ]; then
--- a/.platform/hooks/predeploy/02_fetch_ssh_keys.sh
+++ b/.platform/hooks/predeploy/02_fetch_ssh_keys.sh
@@ -11,8 +11,8 @@ aws configure set default.region ap-southeast-1
 
 echo "Fetching keys"
 # Note we write to webapp user directory which runs our app
-aws ssm get-parameter --name $SSH_PUBLIC_KEY_PARAM_NAME --with-decryption --query "Parameter.Value" --output text > /home/webapp/.ssh/github.pub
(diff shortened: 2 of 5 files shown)


.platform/hooks/predeploy/02_fetch_ssh_keys.sh
```diff
@@ -0,0 +1,22 @@
...
+
+echo "Fetching keys"
+# Note we write to webapp user directory which runs our app
+aws ssm get-parameter --name $SSH_PUBLIC_KEY_PARAM_NAME --with-decryption --query "Parameter.Value" --output text > /home/webapp/.ssh/github.pub
+aws ssm get-parameter --name $SSH_PRIVATE_KEY_PARAM_NAME --with-decryption --query "Parameter.Value" --output text > /home/webapp/.ssh/github
+
+# Set the permissions for the keys
+echo "Setting permissions"
+chmod 600 /home/webapp/.ssh/github.pub
+chmod 600 /home/webapp/.ssh/github
```
QiluXie, line 20: Since the files are created by `root` user, we also need to change the ownership of the files so the permissions would work for `webapp` user. Otherwise, `webapp` user may not be able to access them, right? 
we can use `chown` to change the ownership of the files.

.platform/hooks/predeploy/04_add_github_to_known_hosts.sh
```diff
@@ -0,0 +1,31 @@
...
+    # If the fingerprints don't match, output a warning and exit with an error
+    echo "WARNING: The server's SSH key fingerprint doesn't match the official GitHub fingerprint." >> /tmp/setup-github-known-hosts.txt
+    exit 1
+fi
+
+# Remove the temporary public key file
+rm github_rsa.pub
+
+# all the above would have been performed as root
+# so we need to allow webapp user to access so that it can add to the known_hosts etc.
```
QiluXie, line 30: nit: the comment can be more clear by mentioning the `chown` command is changing the ownership of all files under `/home/webapp/.ssh`

.platform/hooks/predeploy/03_add_keys_to_ssh_config.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+# add github as host and ssh key to ssh config
+cat << EOF > /home/webapp/.ssh/config
+Host github.com
+  IdentityFile /home/webapp/.ssh/github
+  User git
+EOF
```
QiluXie, line 8: nit: would it be better to change the owner of `config` file to webapp if it's created by this script? 
It's probably okay if we don't do it as the default permission should allow read by other users.

.platform/hooks/predeploy/05_set_git_profile.sh
```diff
@@ -0,0 +1,23 @@
...
+aws configure set default.region ap-southeast-1
+
+# Fetch git user's name and email from SSM
+GIT_USER_NAME=$(aws ssm get-parameter --name $GIT_USER_NAME_PARAM_NAME --with-decryption --query "Parameter.Value" --output text)
+GIT_USER_EMAIL=$(aws ssm get-parameter --name $GIT_USER_EMAIL_PARAM_NAME --with-decryption --query "Parameter.Value" --output text)
+
+# Write the configuration to .gitconfig
+echo "[user]" >> /home/webapp/.gitconfig
+echo "  name = $GIT_USER_NAME" >> /home/webapp/.gitconfig
+echo "  email = $GIT_USER_EMAIL" >> /home/webapp/.gitconfig
```
QiluXie, line 19: Also for the file `.gitconfig` ownership and permission, if the file is created in the first run of the script, the owner might be `root` user and we need to change the ownership to `webapp` so it can access it (and set up permission).

QiluXie, line 19: Separately, if we run this multiple times, it will keep appending the 3 lines to `.gitconfig`. Is it better to check if they already exist before appending them? 
Or maybe overwriting every time we rung the script?

QiluXie (changes requested):
Thanks for writing all these scripts @harishv7 !! 
Left some comments mainly regarding error handling and file ownership & permission.

.platform/hooks/predeploy/01_mount_efs.sh
```diff
@@ -0,0 +1,75 @@
...
+service rpcidmapd status &> /dev/null
+if [ $? -ne 0 ] ; then
+    echo 'rpc.idmapd is already stopped!'
+else
+    service rpcidmapd stop
+    if [ $? -ne 0 ] ; then
+        echo 'ERROR: Failed to stop NFS ID Mapper!'
+        exit 1
+    fi
+fi
```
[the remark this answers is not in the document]
kishore03109: +1

.platform/hooks/predeploy/04_add_github_to_known_hosts.sh
```diff
@@ -0,0 +1,31 @@
...
+    cat github_rsa.pub > /home/webapp/.ssh/known_hosts
+    echo "GitHub's public key added to known_hosts." >> /tmp/setup-github-known-hosts.txt
+else
+    # If the fingerprints don't match, output a warning and exit with an error
+    echo "WARNING: The server's SSH key fingerprint doesn't match the official GitHub fingerprint." >> /tmp/setup-github-known-hosts.txt
+    exit 1
+fi
+
+# Remove the temporary public key file
+rm github_rsa.pub
```
kishore03109, line 27: hmm why is this temp? + why do we need to remove them?

kishore03109 (commented):
I understand we need to rush this to staging by eod, as long as the code is sound (manual testing has worked), I just have functional qns
1. What is our sad path flow? Say this script fails, will this lead to an easy rollback?  (A bit of context to my qn is that I have heard other teams not being able to rollback cuz their elastic deployment kept failing, as long as there is a way to gracefully exit, I think this is ok

.platform/hooks/predeploy/01_mount_efs.sh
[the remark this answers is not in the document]
harishv7: @kishore03109 @QiluXie I will strongly suggest not to modify this file as it was recommended by AWS and is what we use for redir server

Ref
https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/services-efs.html
https://github.com/isomerpages/isomer-redirection/blob/staging/.ebextensions/mount-efs.config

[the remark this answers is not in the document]
kishore03109: oh hmm shall we add this context in comment at the top to warn other devs?

.platform/hooks/predeploy/02_fetch_ssh_keys.sh
harishv7: Yes we do that but at the end of all ssh operations. See script 4, we do a recursive owner change

```
# all the above would have been performed as root
# so we need to allow webapp user to access so that it can add to the known_hosts etc.
chown -R webapp:webapp /home/webapp/.ssh
```

.platform/hooks/predeploy/03_add_keys_to_ssh_config.sh
harishv7: we should change it, and as mentioned is in script 04

.platform/hooks/predeploy/04_add_github_to_known_hosts.sh
```diff
@@ -0,0 +1,31 @@
...
+# Check if the server's fingerprint matches the official fingerprint
+# Note: This check is important to prevent any MITM attacks
+if [ "$SERVER_FINGERPRINT" = "$OFFICIAL_FINGERPRINT" ]; then
+    # If the fingerprints match, add the public key to the known_hosts file
+    cat github_rsa.pub > /home/webapp/.ssh/known_hosts
+    echo "GitHub's public key added to known_hosts." >> /tmp/setup-github-known-hosts.txt
+else
+    # If the fingerprints don't match, output a warning and exit with an error
+    echo "WARNING: The server's SSH key fingerprint doesn't match the official GitHub fingerprint." >> /tmp/setup-github-known-hosts.txt
+    exit 1
```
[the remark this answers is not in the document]
harishv7: actually the github_rsa.pub can just stay as it's a public key, but for cleanliness we can remove. have added

.platform/hooks/predeploy/04_add_github_to_known_hosts.sh
harishv7: is just a temp public key of github which we receive from ssh-keyscan

once we compare and verify that the fingerprint is correct, we write it to .github/known_hosts.

after that this file serves no purpose and hence we remove it

.platform/hooks/predeploy/04_add_github_to_known_hosts.sh
harishv7: sure, added

.platform/hooks/predeploy/05_set_git_profile.sh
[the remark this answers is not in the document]
harishv7: added error handling for the variables

the file writing failing is a very unlikely scenario though, so i have left it

harishv7: good point, I added the overwrite

harishv7: good point, have added logic to update to webapp user

.platform/hooks/predeploy/01_mount_efs.sh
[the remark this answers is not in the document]
harishv7: okay added

QiluXie (commented):
@harishv7 Thanks for resolving most of the comments, only a few left, will approve it once they are resolved.

harishv7:
> @harishv7 Thanks for resolving most of the comments, only a few left, will approve it once they are resolved.

@QiluXie I have fixed the issues mentioned in all files except 01_mount_efs as this is a default script from AWS as added in the code comments. Let me know if I missed anything else

QiluXie:
> > @harishv7 Thanks for resolving most of the comments, only a few left, will approve it once they are resolved.
> 
> @QiluXie I have fixed the issues mentioned in all files except 01_mount_efs as this is a default script from AWS as added in the code comments. Let me know if I missed anything else

Got it. Sounds good. Yeah, I was referring to the comments in `01_mount_efs`. It should be fine since it's provided by AWS as you explained. Thanks!

QiluXie (approved):

harishv7:
Note: Will merge this PR when we are ready to release GGS and pre-requisite infra setup has been completed

PR #883, opened 2023-08-08
fix(authservice): update import

seaerchin:
## Problem
previously, no `.default` added on import leading to errors (due to `export default` interaction wtih `require`)

## Solution
add `.default` to `require` import


QiluXie (approved):
lgtm

PR #884, opened 2023-08-08
fix(GGs): return sha of commit made instead of path to repository

dcshzj:
## Problem

The `commit` function in GitFileSystemService mistakenly returned the path to the directory instead of the original intended SHA of the commit made.

## Solution

**Breaking Changes**

**Bug Fixes**:

- The `commit` function was adjusted to return the SHA of the latest commit made, instead of returning the path to the repository

## Tests

1. Unit tests

## Deploy Notes

*None*

### DIFF
--- a/src/services/db/GitFileSystemService.ts
+++ b/src/services/db/GitFileSystemService.ts
@@ -358,7 +358,7 @@ export default class GitFileSystemService {
             }
           )
         )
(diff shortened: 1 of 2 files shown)


dcshzj:
@alexanderleegs @harishv7 My bad! FYI the commit SHA should be returned when you commit, instead of the path to the repository.

harishv7 (approved):

PR #889, opened 2023-08-10
feat(GGs): implement update operation

dcshzj:
## Problem

Closes IS-400.

## Solution

**Breaking Changes**

**Features**:

- The update operation using the local file system is now supported.
- In the event of an error, the file system may potentially be in an inconsistent state. A new rollback function was added to support resetting the file system to the last working state.
- Pushes will happen asynchronously so that the user can continue working on the CMS while the backend pushes to GitHub.

## Tests

1. Unit tests
2. E2E tests to verify that there is no regressions

## Deploy Notes

*None*

### DIFF
--- a/src/errors/GitFileSystemNeedRollbackError.ts
+++ b/src/errors/GitFileSystemNeedRollbackError.ts
@@ -1,0 +1,11 @@
+import { BaseIsomerError } from "./BaseError"
+
+export default class GitFileSystemNeedRollbackError extends BaseIsomerError {
+  constructor(message: string) {
+    super({
+      status: 500,
+      code: "GitFileSystemNeedRollbackError",
+      message,
+    })
+  }
+}
--- a/src/services/db/GitFileSystemService.ts
+++ b/src/services/db/GitFileSystemService.ts
@@ -521,7 +521,14 @@ export default class GitFileSystemService {
     oldSha: string,
     userId: SessionDataProps["isomerUserId"]
   ): ResultAsync<string, GitFileSystemError | NotFoundError | ConflictError> {
-    return this.getFilePathStats(repoName, filePath)
+    let oldStateSha = ""
+
+    return this.getLatestCommitOfBranch(repoName, BRANCH_REF)
+      .andThen((latestCommit) => {
+        oldStateSha = latestCommit
+        return okAsync(true)
+      })
+      .andThen(() => this.getFilePathStats(repoName, filePath))
       .andThen((stats) => {
         if (!stats.isFile()) {
           return errAsync(
@@ -574,7 +581,7 @@ export default class GitFileSystemService {
       })
       .orElse((error) => {
         if (error instanceof GitFileSystemNeedRollbackError) {
-          return this.rollback(repoName, oldSha).andThen(() =>
+          return this.rollback(repoName, oldStateSha).andThen(() =>
             errAsync(new GitFileSystemError(error.message))
           )
         }
--- a/src/services/db/GitFileSystemService.ts
+++ b/src/services/db/GitFileSystemService.ts
@@ -171,6 +171,27 @@ export default class GitFileSystemService {
     )
   }
 
+  // Get filesystem stats of a file or directory
+  getFilePathStats(
+    repoName: string,
+    filePath: string
+  ): ResultAsync<fs.Stats, NotFoundError | GitFileSystemError> {
+    return ResultAsync.fromPromise(
+      fs.promises.stat(`${EFS_VOL_PATH}/${repoName}/${filePath}`),
+      (error) => {
+        if (error instanceof Error && error.message.includes("ENOENT")) {
+          return new NotFoundError("File/Directory does not exist")
+        }
+        if (error instanceof Error) {
+          return new GitFileSystemError(error.message)
+        }
+
+        logger.error(`Error when reading ${filePath}: ${error}`)
+        return new GitFileSystemError("An unknown error occurred")
+      }
+    )
+  }
+
   // Reset the state of the local Git repository to a specific commit
   rollback(
     repoName: string,
@@ -432,25 +453,12 @@ export default class GitFileSystemService {
     repoName: string,
     directoryPath: string
   ): ResultAsync<GitDirectoryItem[], GitFileSystemError | NotFoundError> {
-    return ResultAsync.fromPromise(
-      fs.promises.stat(`${EFS_VOL_PATH}/${repoName}/${directoryPath}`),
-      (error) => {
-        if (error instanceof Error && error.message.includes("ENOENT")) {
-          return new NotFoundError("Directory does not exist")
-        }
-        if (error instanceof Error) {
-          return new GitFileSystemError(error.message)
-        }
-
-        logger.error(`Error when getting ${directoryPath} stats: ${error}`)
-        return new GitFileSystemError("An unknown error occurred")
-      }
-    )
+    return this.getFilePathStats(repoName, directoryPath)
       .andThen((stats) => {
         if (!stats.isDirectory()) {
           return errAsync(
             new GitFileSystemError(
-              `Path "${directoryPath}" is not a directory in repo "${repoName}"`
+              `Path "${directoryPath}" is not a valid directory in repo "${repoName}"`
             )
           )
         }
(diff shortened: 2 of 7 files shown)


src/errors/GitFileSystemNeedRollbackError.ts
```diff
@@ -0,0 +1,11 @@
+import { BaseIsomerError } from "./BaseError"
```
harishv7, line 1: Reg PR description
>Pushes will happen synchronously so that the user can continue working on the CMS while the backend pushes to GitHub.

You mean async right?

harishv7: nit: better to name branches with ticket nums, can copy paste from jira
Screenshot 2023-08-10 at 1 13 19 PM

src/services/db/GitFileSystemService.ts
```diff
@@ -473,8 +499,78 @@ export default class GitFileSystemService {
...
+    repoName: string,
+    filePath: string,
+    fileContent: string,
+    oldSha: string,
+    userId: SessionDataProps["isomerUserId"]
+  ): ResultAsync<string, GitFileSystemError | NotFoundError | ConflictError> {
+    return ResultAsync.fromPromise(
+      fs.promises.stat(`${EFS_VOL_PATH}/${repoName}/${filePath}`),
+      (error) => {
+        if (error instanceof Error && error.message.includes("ENOENT")) {
```
harishv7, line 513: I find repeated checks for this ENOENT and not found error, may can abstract out to a util that checks

src/services/db/GitFileSystemService.ts
```diff
@@ -473,8 +499,78 @@ export default class GitFileSystemService {
...
+                "File has been changed recently, please try again"
+              )
+            )
+          }
+          return okAsync(sha)
+        })
+      )
+      .andThen(() =>
+        ResultAsync.fromPromise(
+          fs.promises.writeFile(
```
harishv7, line 538: writing to a file and immediately reading does it give back the updated file?

do we need to make an explicit flush to the disk
https://stackoverflow.com/questions/42277513/how-to-make-fs-appendfile-method-of-node-js-flush-to-disk-right-away

src/services/db/GitFileSystemService.ts
```diff
@@ -350,11 +374,13 @@ export default class GitFileSystemService {
               .commit(commitMessage),
             (error) => {
               if (error instanceof GitError) {
-                return new GitFileSystemError(error.message)
+                return new GitFileSystemNeedRollbackError(error.message)
```
harishv7, line 377: do we need a new orElse branch to call `rollback`?

src/services/db/GitFileSystemService.ts
dcshzj: Good point, done in 49aff43f0a98e202aab77ae65dd8d64a103d26b6.

src/errors/GitFileSystemNeedRollbackError.ts
dcshzj: Yup oops I meant async, adjusted the description! Also noted on the branch name, thanks!

src/services/db/GitFileSystemService.ts
dcshzj: I checked around (documentation is a little sparse on this) but it seems like the `writeFile` method is a "higher level abstraction", which means that it [handles opening and closing](https://stackoverflow.com/a/21177492) of the underlying file descriptors. When the file descriptor is closed, the OS will handle the flushing of the contents to disk, so no explicit flushing to the disk would be necessary.

So in this theory, immediately reading from the file should return the updated contents. However, because we are using EFS, there might be some slight network latency, which may give a small window of opportunity where another user retrieves an older version of the file. However, I think that it might be a very small chance of it happening, and in this case the user will still be getting an older SHA, which would give them a warning about overriding changes when they try to save the file.

src/services/db/GitFileSystemService.ts
dcshzj: Hmm I don't have a preference for this. My original intention was to let the respective functions (create, update, delete) handle the rollback, as it is possible that something bad happens during the `fs.promises` step. So it is possible to rollback here in the commit function, but the original calling functions would have also performed a rollback themselves, so I avoided adding it here.

kishore03109: ... we are re-implementing transactions here arent we :( 

+ 1, we should have a guaranteed rollback incase fs.promises throws t

src/services/db/GitFileSystemService.ts
```diff
@@ -473,8 +507,75 @@ export default class GitFileSystemService {
...
+          userId,
+          `Update file: ${fileName}`
+        )
+      })
+      .orElse((error) => {
+        if (error instanceof GitFileSystemNeedRollbackError) {
+          return this.rollback(repoName, oldSha).andThen(() =>
+            errAsync(new GitFileSystemError(error.message))
+          )
+        }
```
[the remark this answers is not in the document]
dcshzj: Oops yup, this will need `getLatestCommitOfBranch` to be implemented. Updated in 07f243a055f32bbf921f2b2aa4dc317fd0ef2aba.

src/services/db/GitFileSystemService.ts
```diff
@@ -168,6 +170,49 @@ export default class GitFileSystemService {
...
+  rollback(
+    repoName: string,
+    commitSha: string
+  ): ResultAsync<true, GitFileSystemError> {
+    logger.warn(`Rolling repo ${repoName} back to ${commitSha}`)
+    return ResultAsync.fromPromise(
+      this.git
+        .cwd(`${EFS_VOL_PATH}/${repoName}`)
+        .reset(["--hard", commitSha])
+        .clean("f"),
```
alexanderleegs, line 204: Does this clean up files recursively from directories? From the [documentation](https://git-scm.com/docs/git-clean), it seems you need to include an additional `d` flag in order to remove files more deeply nested

dcshzj: Ah it doesn't, thanks! Fixed in a4f2b01579932fda6757bb4e6b89161ca97aa19e.

kishore03109: Do we need to clean untracked files for sanity as well?

src/services/db/GitFileSystemService.ts
```diff
@@ -453,12 +487,17 @@ export default class GitFileSystemService {
...
+              combine([okAsync(sha), this.getFilePathStats(repoName, path)])
+            )
+            .andThen((shaAndStats) => {
+              const [sha, stats] = shaAndStats as [string, fs.Stats]
               const result: GitDirectoryItem = {
                 name,
                 type,
                 sha,
                 path,
+                size: type === "dir" ? 0 : stats.size,
```
kishore03109, line 500: Quick qn: How do I reason about the `size` key and why is this significant to the functionality of GGs?

src/services/db/GitFileSystemService.ts
alexanderleegs: Yup we should, `git clean` is being used for this purpose!

src/services/db/GitFileSystemService.ts
dcshzj: The size is the file size (or 0 for directories). It is not particularly for GGs but rather for compatibility with the return type of `GitHubService.readDirectory()`.

src/errors/GitFileSystemNeedRollbackError.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export default class GitFileSystemNeedRollbackError extends BaseIsomerError {
```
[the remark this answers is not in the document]
dcshzj: Fixed in a85859a43d5718d487579a8e8ce92cf9007542d3.

src/services/db/GitFileSystemService.ts
```diff
@@ -473,12 +512,91 @@ export default class GitFileSystemService {
...
+              )
+            )
+          }
+          return okAsync(sha)
+        })
+      )
+      .andThen(() =>
+        ResultAsync.fromPromise(
+          fs.promises.writeFile(
+            `${EFS_VOL_PATH}/${repoName}/${filePath}`,
```
QiluXie, line 556: nit: use `path.join`

src/services/db/RepoService.ts
```diff
@@ -173,9 +177,38 @@ export default class RepoService extends GitHubService {
...
+    }: {
+      fileContent: string
+      sha: string
+      fileName: string
+      directoryName?: string
+    }
+  ): Promise<GitCommitResult> {
+    if (this.isRepoWhitelisted(sessionData.siteName)) {
+      logger.info("Updating file in local Git file system")
+      const filePath = directoryName ? `${directoryName}/${fileName}` : fileName
```
QiluXie, line 195: nit: `path.join`

QiluXie (approved):
lgtm with some non-blocking nit comments

src/services/db/GitFileSystemService.ts
```diff
@@ -9,20 +9,22 @@ import {
   Result,
```
harishv7, line 9: Screenshot 2023-08-11 at 3 48 01 PM
This should not be an user facing error (though this error should not happen on prod)

dcshzj: Hmm currently we return a `GitFileSystemError` which RepoService will throw and be caught by the middleware, which shows this error toast to the user. Do we want to just throw the error without any message?

dcshzj: Adjusted the messages in 61eee9b7183edd03221f8d24330a2978047ffc29.

harishv7 (approved):

PR #890, opened 2023-08-10
IS-398: Write retrieve images from disk and return as base 64

harishv7:
## Problem

Extend logic of converting images to base64 currently used for private repositories to cover all repositories.

Closes IS-398

## Solution

MediaFileService's `read` delegates to the `readMediaFile` function located in RepoService, which calls `readMedia` function in GitFileSystemService for a whitelisted repository, or forwards the request to the `getMediaFileInfo` utility function.

MediaDirectoryService calls RepoService's `readMediaDirectory` and conditionally outputs the media directory contents depending on if it is a whitelisted repo or not.

**Breaking Changes**

  - Details ...

## Tests

- Unit tests for MediaFileService, MediaDirectoryService, RepoService should all pass

- Image from non-whitelisted, public repository should continue to be retrieved from raw.githubusercontent.com
- Image from non-whitelisted, private repository should continue to be retrieved from [raw.githubusercontent.com](http://raw.githubusercontent.com/), and then converted to base64 returned to the user
- Image from whitelisted repository (regardless of public or private) should be retrieved from the local file system, then converted to base64 returned to the user

CMS Flows
- Go to Images section and ensure the images are loading correctly from disk
- Step into nested folders and ensure images are loading fine
- Right click on images -> open in new tab -> ensure image src is in base64

Note for local testing, you need to clone the repo to disk and whitelist it first


src/services/db/RepoService.ts
```diff
@@ -144,6 +151,51 @@ export default class RepoService extends GitHubService {
     })
   }
 
+  async readMediaFile(
+    sessionData: UserWithSiteSessionData,
+    { fileName, directoryName }: { fileName: string; directoryName: string }
+  ): Promise<any> {
```
dcshzj, line 157: nit: Are we able to specify the return type other than `any`?

src/services/directoryServices/MediaDirectoryService.js
```diff
@@ -35,36 +36,32 @@ class MediaDirectoryService {
...
-        if (curr.type === "dir") {
-          return {
-            name: curr.name,
-            type: "dir",
-          }
-        }
-        return getMediaFileInfo({
-          file: curr,
-          siteName,
+    if (isGGSWhitelistedRepo(siteName)) {
```
dcshzj, line 40: Hmm, is it possible to keep the check on whether the repo is whitelisted within RepoService? Because I was thinking RepoService should be the only one that is aware of what to do for whitelisted and non-whitelisted repositories. These downstream services would not have to worry about the underlying storage mechanisms.

src/services/directoryServices/MediaDirectoryService.js
```diff
@@ -35,36 +36,32 @@ class MediaDirectoryService {
...
+      return await this.gitHubService.readMediaDirectory(sessionData, {
+        readFromGithub: false,
```
dcshzj, line 42: Same concerns here. I think we should consider that non-whitelisted repo == read from GitHub and whitelisted repo == don't read from GitHub.

harishv7: I considered that, but this means we need to have BaseDirectoryService within RepoService as the Github flow is dependent on BaseDirectoryService's `list` method

But I guess it makes sense to keep the whitelisted flow internal to RepoService as well

dcshzj: Hmm I think we should be able to cut away the dependency on BaseDirectoryService and utilise the `readDirectory` method in RepoService directly. The return types are similar.

src/services/db/RepoService.ts
QiluXie: I think we can specify the type like:
`Promise<MediaFileOutput, GitFileSystemError | NotFoundError>`

src/services/db/GitFileSystemService.ts
```diff
@@ -400,6 +402,49 @@ export default class GitFileSystemService {
...
+  readMediaFile(
+    siteName: string,
+    directoryName: string,
+    fileName: string
+  ): ResultAsync<MediaFileOutput, GitFileSystemError | NotFoundError> {
+    return this.read(
+      siteName,
+      `${directoryName}/${fileName}`,
+      "base64"
+    ).andThen((file) => {
```
QiluXie, line 433: `file` implicitly has an  `any` type, we can specify `GitFile` as the type:
`andThen((file: GitFile)`

src/services/db/RepoService.ts
```diff
@@ -144,6 +151,51 @@ export default class RepoService extends GitHubService {
...
+      )
+      if (result.isErr()) {
+        throw result.error
+      }
+
+      return result.value
+    }
+
+    // fetch from Github
+    const directoryData = await super.readDirectory(sessionData, {
```
QiluXie, line 180: Should we do a `try catch` to handle potential error from calling `super.readDirectory` here?

src/services/db/RepoService.ts
```diff
@@ -144,6 +151,51 @@ export default class RepoService extends GitHubService {
...
+      directoryName,
+    })
+
+    const mediaType = directoryName.split("/")[0] as MediaType
+    const targetFile = directoryData.find(
+      (fileOrDir: { name: string }) => fileOrDir.name === fileName
+    )
+    const { private: isPrivate } = await super.getRepoInfo(sessionData)
```
QiluXie, line 188: Similarly, should we do a `try catch` to handle potential error from calling `super.getRepoInfo` here?

src/services/db/RepoService.ts
```diff
@@ -172,6 +224,67 @@ export default class RepoService extends GitHubService {
...
+        }
+        if (readMediaInput.readFromGithub) {
+          return getMediaFileInfo({
+            file: curr,
+            siteName,
+            directoryName,
+            mediaType: readMediaInput.directoryInfo.mediaType as MediaType,
+            isPrivate: readMediaInput.directoryInfo.isPrivate,
+          })
+        }
```
QiluXie, line 281: is it possible that all three `if` conditions are not true, and the return type here would be `undefined`, which would not match the promise return type `ReadMediaDirectoryOutput`?

src/services/db/GitFileSystemService.ts
```diff
@@ -400,6 +402,49 @@ export default class GitFileSystemService {
...
+  getMimeType(fileExtension: string): string {
+    switch (fileExtension) {
+      case "svg":
+        return "image/svg+xml"
+      case "ico":
+        return "image/vnd.microsoft.icon"
+      default:
+        return `image/${fileExtension}`
+    }
+  }
```
kishore03109, line 422: Hey we do allow a bigger range of file types, eg. `.png`, `.jpeg`, etc. Are these taken into account as well?

src/services/db/GitFileSystemService.ts
```diff
@@ -369,13 +370,14 @@ export default class GitFileSystemService {
   // Read the contents of a file
   read(
     repoName: string,
-    filePath: string
+    filePath: string,
+    encoding: "utf-8" | "base64" = "utf-8"
```
kishore03109, line 374: nit: Hmm is it a lot of effort to remove the default? would rather enforce callers to define encoding for clarity!

src/services/db/RepoService.ts
```diff
@@ -172,6 +224,67 @@ export default class RepoService extends GitHubService {
...
+    sessionData: UserWithSiteSessionData,
+    readMediaInput: ReadMediaDirectoryInput
+  ): Promise<ReadMediaDirectoryOutput[]> {
+    const { directoryName } = readMediaInput.directoryInfo
+
+    logger.debug(`Reading media directory: ${directoryName}`)
+    const { siteName } = sessionData
+
+    // making as any to support incoming data from MediaDirectoryService
+    let filteredResult: any[] = []
```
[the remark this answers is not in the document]
kishore03109: ^If yes, might be wise to shift out the type + reference it form here as well

src/services/db/RepoService.ts
```diff
@@ -144,6 +151,51 @@ export default class RepoService extends GitHubService {
...
+    { fileName, directoryName }: { fileName: string; directoryName: string }
+  ): Promise<any> {
+    logger.debug(`Reading media file: ${fileName}`)
+    logger.debug(`Reading directoryName: ${directoryName}`)
+    const { siteName } = sessionData
+
+    // fetch from local disk
+    if (this.isRepoWhitelisted(siteName)) {
+      logger.info(
+        `Reading media file from disk. Sitname: ${siteName}, directory name: ${directoryName}, fileName: ${fileName},`
```
alexanderleegs, line 165:
```suggestion
        `Reading media file from disk. Site name: ${siteName}, directory name: ${directoryName}, fileName: ${fileName},`
```

src/services/db/RepoService.ts
alexanderleegs: I think this is okay because it's the same behaviour as existing - any errors get caught by our overall error handler, and we don't need to catch and convert the error we receive at this point to something else

src/services/db/GitFileSystemService.ts
harishv7: Yea they fall into the default case of image/{extension}

ref: https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types

src/services/db/GitFileSystemService.ts
harishv7: i see this as by default you should be always reading as utf-8 for all cases, except the case for images when base64 is needed. Hence I made it optional

If we find ourselves adding more encoding types in future we can make this explicit, but for now base64 seems to be the special case

src/services/db/RepoService.ts
QiluXie: Thanks @alexanderleegs for the explanation, I will look more into our overall error handler logic. Will resolve these two comments then.

src/services/db/GitFileSystemService.ts
alexanderleegs: hmm yeah let's add these in, I think we might have overlooked these in the past

src/services/db/RepoService.ts
```diff
@@ -177,6 +225,56 @@ export default class RepoService extends GitHubService {
...
+    } else {
+      const repoInfo = await super.getRepoInfo(sessionData)
+      isPrivate = repoInfo.private
+      const files = await super.readDirectory(sessionData, {
+        directoryName,
+      })
+      filteredResult = files.filter(filterLogic)
+    }
+
+    return Promise.all(
```
dcshzj, line 261: Can I check if an `await` is needed here?

src/utils/ggs-utils.ts
```diff
@@ -1,8 +0,0 @@
-import config from "@config/config"
-
-const WHITELISTED_GIT_SERVICE_REPOS = config.get(
```
kishore03109: hey may I get context to why this is removed?

src/utils/file-upload-utils.js
```diff
@@ -14,8 +14,10 @@ const ALLOWED_FILE_EXTENSIONS = [
   "pdf",
   "png",
   "jpg",
+  "jpeg",
```
kishore03109, line 17: Hey just a heads up, we maintain this list over in the FE, may I trouble you to add in a corresponding PR in the FE for this?

kishore03109 (approved):
Code wise, mostly ok, but I am concerned about the removal of `gg-utils.ts`. Is this intended? 

If no, please dismiss this review

src/utils/file-upload-utils.js
harishv7: @kishore03109 added here: https://github.com/isomerpages/isomercms-frontend/pull/1413

dcshzj (approved):

PR #891, opened 2023-08-10
IS-404: Write extend repo service to get latest commit of branch

harishv7:
## Problem

Create a function that provides the latest commit information for a given branch. The return type should match the existing function with the same name in GitHubService. However, the latest commit of branch should be retrieved from the local file system for a whitelisted repository, but continue the existing behaviour of retrieving from GitHub API for a non-whitelisted repository.

Note: In addition to the existing properties, the commit SHA should also be returned, to allow for rolling back of changes.

Closes IS-404

## Solution

RepoService contains the logic to call the appropriate function depending on whether the repository is whitelisted or not, whereas GitFileSystemService contains the logic to retrieve the latest commit information from the local file system.

**Breaking Changes**

  - Details ...

## Tests

Ensure the following test suites pass:
- src/integration/NotificationOnEditHandler.spec.ts
- src/integration/Notifications.spec.ts
- src/integration/Privatisation.spec.ts
- src/integration/Reviews.spec.ts
- src/services/db/__tests__/RepoService.spec.ts
- src/services/identity/__tests__/SitesService.spec.ts

CMS Flow:
- Visit site dashboard will trigger API call to /info endpoint
- For a whitelisted repo, this should fetch the info from local disk
- For non-whitelisted repo, this should fetch the info from GitHub

### DIFF
--- a/src/services/db/GitFileSystemService.ts
+++ b/src/services/db/GitFileSystemService.ts
@@ -501,7 +501,7 @@ export default class GitFileSystemService {
     branchName: string
   ): ResultAsync<GitLocalDiskCommitData, GitFileSystemError> {
     return ResultAsync.fromPromise(
-      this.git.cwd(`${EFS_VOL_PATH}/${repoName}`).checkout(branchName).log(),
+      this.git.cwd(`${EFS_VOL_PATH}/${repoName}`).log([branchName]),
       (error) => {
         if (error instanceof GitError) {
           return new GitFileSystemError(error.message)
@@ -511,6 +511,7 @@ export default class GitFileSystemService {
       }
     ).andThen((logSummary) => {
       const possibleCommit = logSummary.latest
+      console.log(`POSSIBLE COMMIT`, possibleCommit)
       if (this.isGitLocalDiskRawCommitData(possibleCommit)) {
         return okAsync({
           author: {
--- a/src/services/db/GitFileSystemService.ts
+++ b/src/services/db/GitFileSystemService.ts
@@ -480,14 +480,14 @@ export default class GitFileSystemService {
   // TODO: Delete a file
   async delete() {}
 
-  isDefaultLogFields(commit: unknown): commit is DefaultLogFields {
+  isDefaultLogFields(logFields: unknown): logFields is DefaultLogFields {
     return (
-      !!commit &&
-      (commit as DefaultLogFields).author_name !== undefined &&
-      (commit as DefaultLogFields).author_email !== undefined &&
-      (commit as DefaultLogFields).date !== undefined &&
-      (commit as DefaultLogFields).message !== undefined &&
-      (commit as DefaultLogFields).hash !== undefined
+      !!logFields &&
+      (logFields as DefaultLogFields).author_name !== undefined &&
+      (logFields as DefaultLogFields).author_email !== undefined &&
+      (logFields as DefaultLogFields).date !== undefined &&
+      (logFields as DefaultLogFields).message !== undefined &&
+      (logFields as DefaultLogFields).hash !== undefined
     )
   }
 
--- a/src/services/db/GitFileSystemService.ts
+++ b/src/services/db/GitFileSystemService.ts
@@ -9,7 +9,7 @@ import {
   Result,
   ResultAsync,
 } from "neverthrow"
-import { GitError, SimpleGit } from "simple-git"
+import { GitError, SimpleGit, DefaultLogFields } from "simple-git"
 
 import { config } from "@config/config"
 
@@ -21,10 +21,7 @@ import { ISOMER_GITHUB_ORG_NAME } from "@constants/constants"
 
 import { SessionDataProps } from "@root/classes"
 import { NotFoundError } from "@root/errors/NotFoundError"
-import {
-  GitHubCommitData,
-  GitLocalDiskRawCommitData,
-} from "@root/types/commitData"
+import { GitHubCommitData } from "@root/types/commitData"
 import type { GitDirectoryItem, GitFile } from "@root/types/gitfilesystem"
 import { IsomerCommitMessage } from "@root/types/github"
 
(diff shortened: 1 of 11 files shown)


src/services/db/GitFileSystemService.ts
```diff
@@ -479,6 +483,50 @@ export default class GitFileSystemService {
...
+      (commit as GitLocalDiskRawCommitData).hash !== undefined
+    )
+  }
+
+  getLatestCommitOfBranch(
+    repoName: string,
+    branchName: string
+  ): ResultAsync<GitLocalDiskCommitData, GitFileSystemError> {
+    return ResultAsync.fromPromise(
+      this.git.cwd(`${EFS_VOL_PATH}/${repoName}`).checkout(branchName).log(),
```
dcshzj, line 504: Are we able to do a `.log()` and specify the branch directly here? Seems like it is possible to do a `git log staging` to get just the log of that branch without needing to checkout that branch.

Asking because checkouts tend to be quite expensive and it will involve modifying the state of the local repo, which may not be ideal.

src/services/db/RepoService.ts
```diff
@@ -203,8 +203,25 @@ export default class RepoService extends GitHubService {
     return await super.getRepoState(sessionData)
   }
 
-  async getLatestCommitOfBranch(sessionData: any, branch: any): Promise<any> {
-    return await super.getLatestCommitOfBranch(sessionData, branch)
+  async getLatestCommitOfBranch(
+    sessionData: any,
+    branchName: any
+  ): Promise<any> {
```
dcshzj, line 209: We can probably type the `sessionData` as UserWithSiteSessionData and the return to be `Promise<GitLocalDiskCommitData>`.

src/types/commitData.ts
```diff
@@ -6,3 +6,16 @@ export type GitHubCommitData = {
   }
   message: string
 }
+
+export type GitLocalDiskCommitData = {
+  sha: string
+} & GitHubCommitData
```
dcshzj, line 12: Hmm I think it would be better if GitHubCommitData has the `sha` attribute added (plus a simple modification to GitHubService, GitHub already provides this field) since we should make GitHubService and GitFileSystemService return the same types as much as possible.

src/types/commitData.ts
```diff
@@ -6,3 +6,16 @@ export type GitHubCommitData = {
...
+
+// returned by Simple Git
+export type GitLocalDiskRawCommitData = {
```
dcshzj, line 15: Just checking, are we able to use [simple-git's typings here](https://github.com/steveukx/git-js/blob/d184c13273abca4b6572c260f9625c19f944d4f7/simple-git/src/lib/tasks/log.ts#L39)?

src/services/db/GitFileSystemService.ts
harishv7: good point, will update!

src/types/commitData.ts
harishv7: Sure!

harishv7: but I think sha can be an optional attribute within GitHubCommitData

src/types/commitData.ts
harishv7: good point, will update to use `DefaultLogFields` type from simple-git

src/services/db/RepoService.ts
harishv7: oops, my bad, missed updating the types

src/services/db/GitFileSystemService.ts
```diff
@@ -479,6 +483,50 @@ export default class GitFileSystemService {
   // TODO: Delete a file
   async delete() {}
 
-  // TODO: Get the latest commit of branch
-  async getLatestCommitOfBranch() {}
+  isGitLocalDiskRawCommitData(
+    commit: unknown
+  ): commit is GitLocalDiskRawCommitData {
+    return (
+      !!commit &&
```
QiluXie, line 490: Should we also check to make sure `commit` is an object before proceeding to avoid unexpected issue? 
```ts
!!commit && typeof commit === `object` && {rest}
```

src/services/db/GitFileSystemService.ts
```diff
@@ -479,6 +483,50 @@ export default class GitFileSystemService {
...
+      (commit as GitLocalDiskRawCommitData).author_name !== undefined &&
+      (commit as GitLocalDiskRawCommitData).author_email !== undefined &&
+      (commit as GitLocalDiskRawCommitData).date !== undefined &&
+      (commit as GitLocalDiskRawCommitData).message !== undefined &&
+      (commit as GitLocalDiskRawCommitData).hash !== undefined
```
QiluXie, line 495: Two comments here:

1. is it better to check and make sure all properties of `commit` are `string` according to the definition of `GitLocalDiskRawCommitData`? 
2. nit: `commit` is being cast on multiple lines here, to make it more concise, we can cast it once and reuse the variables for multiple times. 
Suggestions incorporating the above two poitns:
```suggestion
      const c = commit as GitLocalDiskRawCommitData;
      !!commit &&
      typeof c.author_name === 'string' &&
      typeof c.author_email === 'string' &&
      typeof c.date === 'string' &&
      typeof c.message === 'string' &&
      typeof c.hash === 'string'
```

src/services/db/GitFileSystemService.ts
QiluXie, line 504: nit: better to use `path.join` instead of template literals?

src/services/db/RepoService.ts
QiluXie: We can also specify `branchName` to be `string`

QiluXie: The promise return type should be `Promise<GitHubCommitData>` if we incorporate[ this comment ](https://github.com/isomerpages/isomercms-backend/pull/891/files#r1290821301)

src/services/db/GitFileSystemService.ts
harishv7: done

src/services/db/RepoService.ts
harishv7: done!

src/types/commitData.ts
harishv7: done!

src/types/commitData.ts
harishv7: done!

src/services/db/GitFileSystemService.ts
harishv7: sure!

src/services/db/GitFileSystemService.ts
harishv7: updated!

src/services/db/GitFileSystemService.ts
harishv7: no preference, used template literal to be conssitent with rest of code

QiluXie (approved):
lgtm
Thanks @harishv7 for addressing the comments quickly!

dcshzj (approved):

PR #892, opened 2023-08-11
Feat/is 399 ggs create file

alexanderleegs:
## Problem

This PR introduces creates to our GGS solution. Resolves IS-399. This PR also modifies checks for the existence of paths to use `gitFilePathStats` instead of `fs.existsSync`.

**Features**:

- Refactored safeExistsSync calls in `clone` and `isValidGitRepo` to use `getFilePathStats` instead
- Create operation supported for GGS

Future work:

- Verify image creation - pending image retrieval PR

### DIFF
--- a/src/services/configServices/NetlifyTomlService.js
+++ b/src/services/configServices/NetlifyTomlService.js
@@ -1,3 +1,4 @@
+const { Base64 } = require("js-base64")
 const toml = require("toml")
 
 const { config } = require("@config/config")
--- a/src/services/db/GitFileSystemService.ts
+++ b/src/services/db/GitFileSystemService.ts
@@ -1,6 +1,5 @@
 import fs from "fs"
 
-import { Base64 } from "js-base64"
 import {
   combine,
   err,
--- a/src/services/db/GitFileSystemService.ts
+++ b/src/services/db/GitFileSystemService.ts
@@ -27,7 +27,11 @@ import { SessionDataProps } from "@root/classes"
 import { MediaTypeError } from "@root/errors/MediaTypeError"
 import { MediaFileInput, MediaFileOutput } from "@root/types"
 import { GitHubCommitData } from "@root/types/commitData"
-import type { GitDirectoryItem, GitFile } from "@root/types/gitfilesystem"
+import type {
+  GitCommitResult,
+  GitDirectoryItem,
+  GitFile,
+} from "@root/types/gitfilesystem"
 import type { IsomerCommitMessage } from "@root/types/github"
 import { ALLOWED_FILE_EXTENSIONS } from "@root/utils/file-upload-utils"
 
@@ -438,7 +442,7 @@ export default class GitFileSystemService {
     directoryName: string,
     fileName: string
   ): ResultAsync<
-    { sha: string },
+    GitCommitResult,
     ConflictError | GitFileSystemError | NotFoundError
   > {
     const filePath = directoryName ? `${directoryName}/${fileName}` : fileName
@@ -511,7 +515,7 @@ export default class GitFileSystemService {
           `Create file: ${filePath}`
         )
       )
-      .map((commit) => ({ sha: commit }))
+      .map((commit) => ({ newSha: commit }))
       .orElse((error) => {
         if (error instanceof GitFileSystemNeedsRollbackError) {
           return this.rollback(repoName, oldStateSha).andThen(() =>
--- a/src/services/db/GitFileSystemService.ts
+++ b/src/services/db/GitFileSystemService.ts
@@ -46,18 +46,6 @@ export default class GitFileSystemService {
     this.git = git
   }
 
-  private safeExistsSync(path: string): Result<boolean, GitFileSystemError> {
-    const res = Result.fromThrowable(fs.existsSync, (error) => {
-      if (error instanceof GitError || error instanceof Error) {
-        return new GitFileSystemError(error.message)
-      }
-
-      logger.error(`Error when checking if ${path} exists: ${error}`)
-      return new GitFileSystemError("An unknown error occurred")
-    })
-    return res(path)
-  }
-
   isGitInitialized(repoName: string): ResultAsync<boolean, GitFileSystemError> {
     return ResultAsync.fromPromise(
       this.git.cwd(`${EFS_VOL_PATH}/${repoName}`).checkIsRepo(),
@@ -98,16 +86,22 @@ export default class GitFileSystemService {
 
   // Determine if the folder is a valid Git repository
   isValidGitRepo(repoName: string): ResultAsync<boolean, GitFileSystemError> {
-    return this.safeExistsSync(`${EFS_VOL_PATH}/${repoName}`)
-      .andThen((isFolderExisting) => {
-        if (!isFolderExisting) {
+    return this.getFilePathStats(repoName, "")
+      .andThen((stats) => {
+        if (!stats.isDirectory()) {
           // Return as an error to prevent further processing
           // The function will eventually return false
-          return err<never, false>(false)
+          return errAsync(false)
         }
-        return ok(true)
+        return okAsync(true)
       })
-      .asyncAndThen(() => this.isGitInitialized(repoName))
+      .orElse((error) => {
+        if (error instanceof NotFoundError) {
+          return err(false)
+        }
+        return err(error)
+      })
+      .andThen(() => this.isGitInitialized(repoName))
       .andThen((isGitInitialized) => {
         if (!isGitInitialized) {
           return err<never, false>(false)
@@ -235,9 +229,16 @@ export default class GitFileSystemService {
   clone(repoName: string): ResultAsync<string, GitFileSystemError> {
     const originUrl = `git@github.com:${ISOMER_GITHUB_ORG_NAME}/${repoName}.git`
 
-    return this.safeExistsSync(`${EFS_VOL_PATH}/${repoName}`).asyncAndThen(
-      (isFolderExisting) => {
-        if (!isFolderExisting) {
+    return this.getFilePathStats(repoName, "")
+      .andThen((stats) => ok(stats.isDirectory()))
+      .orElse((error) => {
+        if (error instanceof NotFoundError) {
+          return ok(false)
+        }
+        return err(error)
+      })
+      .andThen((isDirectory) => {
+        if (!isDirectory) {
           return ResultAsync.fromPromise(
             this.git
               .clone(originUrl, `${EFS_VOL_PATH}/${repoName}`)
@@ -277,8 +278,7 @@ export default class GitFileSystemService {
             }
             return okAsync(`${EFS_VOL_PATH}/${repoName}`)
           })
-      }
-    )
+      })
   }
 
   // Pull the latest changes from upstream Git hosting provider
@@ -430,8 +430,7 @@ export default class GitFileSystemService {
     })
   }
 
-  // TODO: Creates either directory or file
-  // ResourceDirectoryService used this to create a directory + file at the same time
+  // Creates a file and the associated directory if it doesn't exist
   create({
     repoName,
     userId,
@@ -442,7 +441,7 @@ export default class GitFileSystemService {
     repoName: string
     userId: string
     content: string
-    directoryName?: string
+    directoryName: string
     fileName: string
   }): ResultAsync<
     { sha: string },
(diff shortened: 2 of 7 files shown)


src/services/db/GitFileSystemService.ts
```diff
@@ -433,9 +427,104 @@ export default class GitFileSystemService {
...
-  // ResourceDirectoryService used this to create a directory + file at the same time
-  create() {}
+  // Creates a file and the associated directory if it doesn't exist
+  create({
+    repoName,
+    userId,
+    content,
+    directoryName,
+    fileName,
+  }: {
```
dcshzj, line 437: Should we standardise to passing function arguments as-is? Or should we pass as object instead like this?

src/services/db/GitFileSystemService.ts
```diff
@@ -433,9 +427,104 @@ export default class GitFileSystemService {
...
+    repoName: string
+    userId: string
+    content: string
+    directoryName: string
+    fileName: string
+  }): ResultAsync<
+    { sha: string },
```
dcshzj, line 444: This should be `GitCommitResult`.

src/services/db/GitFileSystemService.ts
```diff
@@ -433,9 +427,104 @@ export default class GitFileSystemService {
...
+        }
+        return err(error)
+      })
+      .andThen(() => this.getFilePathStats(repoName, filePath))
+      .andThen((stats) => {
+        if (stats.isFile())
+          return err(
+            new ConflictError(
+              `File ${filePath} already exists in repo ${repoName}`
+            )
```
dcshzj, line 485: I think we might potentially need to rollback the earlier creation of the directory if it didn't exist originally.

src/services/db/GitFileSystemService.ts
```diff
@@ -433,9 +427,104 @@ export default class GitFileSystemService {
...
+          (error) => {
+            logger.error(`Error when creating ${filePath}: ${error}`)
+            if (error instanceof Error) {
+              return new GitFileSystemError(error.message)
+            }
+            return new GitFileSystemError("An unknown error occurred")
+          }
+        )
+          .map((res) => res)
+          .mapErr((error) => error)
```
dcshzj, line 507: Hmm can I check what does this part do?

src/services/db/GitFileSystemService.ts
```diff
@@ -433,9 +427,104 @@ export default class GitFileSystemService {
...
+      })
+      .andThen(() =>
+        ResultAsync.fromPromise(
+          fs.promises.writeFile(pathToEfsFile, encodedContent),
+          (error) => {
+            logger.error(`Error when creating ${filePath}: ${error}`)
+            if (error instanceof Error) {
+              return new GitFileSystemError(error.message)
+            }
+            return new GitFileSystemError("An unknown error occurred")
```
dcshzj, line 503: We potentially might need a rollback here.

src/services/db/RepoService.ts
```diff
@@ -113,9 +113,36 @@ export default class RepoService extends GitHubService {
...
+      fileName,
+      directoryName,
+      isMedia = false,
+    }: {
+      content: string
+      fileName: string
+      directoryName: string
+      isMedia?: boolean
+    }
+  ): Promise<{ sha: string }> {
```
dcshzj, line 128: Same here, this should be `GitCommitResult`.

src/services/db/GitFileSystemService.ts
alexanderleegs: If the file already exists, the directory definitely exists, so no need to rollback! Also github doesn't care about directory structure, you can only specify directory structure by pushing files into that directory, so this wouldn't affect us regardless!

dcshzj: Oops yup, you're right! Thanks for clarifying!

src/services/db/GitFileSystemService.ts
```diff
@@ -1,5 +1,6 @@
 import fs from "fs"
 
+import { Base64 } from "js-base64"
```
dcshzj, line 3: Oh yeah, do we still need this?

src/services/db/RepoService.ts
alexanderleegs: The result here needs to be `{ sha: string }` in order to match the result of the `githubservice.create`!

dcshzj: Oh gosh 😭

src/services/db/GitFileSystemService.ts
alexanderleegs: Hmm i think it might be clearer to pass as an object, but let's swap it to passing as in for consistency! Done in https://github.com/isomerpages/isomercms-backend/pull/892/commits/c6190b92eddd62f95ede7100a207565ed8a9e555

src/services/db/GitFileSystemService.ts
alexanderleegs: Changed in https://github.com/isomerpages/isomercms-backend/pull/892/commits/b0b811aec0ce097f1595fbfd772bb64d271c09da

src/services/db/GitFileSystemService.ts
alexanderleegs: Whoops this isn't needed anymore! https://github.com/isomerpages/isomercms-backend/pull/892/commits/b8b22f19298a33df4d10ade42f05416a0d5cfcb6

src/services/db/GitFileSystemService.ts
alexanderleegs: Fixed in https://github.com/isomerpages/isomercms-backend/pull/892/commits/a1dd5c2b37fc75410e1f58d9c9c9d4a4f2cde9a4

src/services/db/GitFileSystemService.ts
alexanderleegs: Not in this PR at least, it might be needed for images? But if work needs to be done on that aspect we can tackle it in a separate PR! Removed in https://github.com/isomerpages/isomercms-backend/pull/892/commits/a4776f561cbe8392a1f208d940543c6406b867c3

src/services/db/RepoService.ts
dcshzj: I think we can leave it as it is right now, but I think we can potentially refactor in the future so that they have the same return types

src/services/configServices/NetlifyTomlService.js
```diff
@@ -1,3 +1,4 @@
+const { Base64 } = require("js-base64")
```
dcshzj, line 1: Hmm this one also :(

dcshzj (approved):

src/services/configServices/NetlifyTomlService.js
alexanderleegs: This one is needed! For some reason it wasn't being imported before

src/services/db/GitFileSystemService.ts
```diff
@@ -82,26 +86,22 @@ export default class GitFileSystemService {
...
-      if (error instanceof GitError || error instanceof Error) {
-        return new GitFileSystemError("Unable to determine if directory exists")
-      }
-
-      return new GitFileSystemError("An unknown error occurred")
-    })
-
-    return safeExistsSync(`${EFS_VOL_PATH}/${repoName}`)
-      .andThen((isFolderExisting) => {
-        if (!isFolderExisting) {
```
harishv7, line 97: why are these removed?

src/services/db/GitFileSystemService.ts
```diff
@@ -228,19 +228,17 @@ export default class GitFileSystemService {
...
-      if (error instanceof GitError || error instanceof Error) {
-        return new GitFileSystemError("Unable to determine if directory exists")
-      }
-
-      return new GitFileSystemError("An unknown error occurred")
-    })
-
-    return safeExistsSync(`${EFS_VOL_PATH}/${repoName}`).asyncAndThen(
-      (isFolderExisting) => {
-        if (!isFolderExisting) {
```
harishv7, line 243: same here, do we not need these anymore?

src/services/db/GitFileSystemService.ts
```diff
@@ -433,9 +430,98 @@ export default class GitFileSystemService {
...
+      .andThen((latestCommit) => {
+        const { sha } = latestCommit
+        if (!sha) {
+          return errAsync(new GitFileSystemError("An unknown error occurred"))
+        }
+        oldStateSha = sha
+        return okAsync(true)
+      })
+      .andThen(() => this.getFilePathStats(repoName, directoryName))
+      .andThen(() => ok(""))
```
harishv7, line 460: what is this empty string for, seems like a redundant step?

src/services/db/GitFileSystemService.ts
```diff
@@ -433,9 +430,98 @@ export default class GitFileSystemService {
...
-  // TODO: Creates either directory or file
-  // ResourceDirectoryService used this to create a directory + file at the same time
-  create() {}
+  // Creates a file and the associated directory if it doesn't exist
+  create(
+    repoName: string,
+    userId: string,
+    content: string,
+    directoryName: string,
+    fileName: string
```
harishv7, line 439: how do we differentiate between a file creation vs directory creation? should we add an explicit isDir flag same as delete?

src/services/db/GitFileSystemService.ts
```diff
@@ -433,9 +430,98 @@ export default class GitFileSystemService {
...
+        if (stats.isFile())
+          return err(
+            new ConflictError(
+              `File ${filePath} already exists in repo ${repoName}`
+            )
+          )
+        return ok("")
+      })
+      .orElse((error) => {
+        if (error instanceof NotFoundError) {
```
harishv7, line 487: where do we get the NotFoundError from?

src/services/db/GitFileSystemService.ts
```diff
@@ -433,9 +430,98 @@ export default class GitFileSystemService {
...
+      })
+      .andThen(() => this.getFilePathStats(repoName, filePath))
+      .andThen((stats) => {
+        if (stats.isFile())
+          return err(
+            new ConflictError(
+              `File ${filePath} already exists in repo ${repoName}`
+            )
+          )
+        return ok("")
```
harishv7, line 484: nit: maybe can use boolean ok(true)

src/services/db/RepoService.ts
```diff
@@ -113,9 +113,36 @@ export default class RepoService extends GitHubService {
...
+    if (this.isRepoWhitelisted(sessionData.siteName)) {
+      logger.info("Writing file to local Git file system")
```
harishv7, line 130: can we log for which repoName is this along with the dirName and fileName?

src/services/db/RepoService.ts
```diff
@@ -113,9 +113,36 @@ export default class RepoService extends GitHubService {
...
-  ): Promise<any> {
+    sessionData: UserWithSiteSessionData,
+    {
+      content,
+      fileName,
+      directoryName,
+      isMedia = false,
+    }: {
+      content: string
+      fileName: string
```
harishv7, line 124: will filename be there for create dir too? if not, is this optional? 
some thinking points - would it be better to have 2 separate functions for createFile createDir or maybe a flag like isDir to make it explicit?

src/services/db/__tests__/GitFileSystemService.spec.ts
```diff
@@ -57,6 +58,140 @@ describe("GitFileSystemService", () => {
...
+        size: "Another fake content".length,
+      }
+
+      const result = await GitFileSystemService.listDirectoryContents(
+        "fake-repo",
+        ""
+      )
+      const actual = result
+        ._unsafeUnwrap()
+        .sort((a, b) => a.name.localeCompare(b.name))
```
harishv7, line 101: why is this sort important?

src/services/db/__tests__/GitFileSystemService.spec.ts
```diff
@@ -677,182 +812,224 @@ describe("GitFileSystemService", () => {
...
         "fake-repo",
-        "fake-dir/fake-file"
+        "fake-user-id",
+        "fake content",
+        "fake-dir",
+        "fake-file"
       )
 
-      expect(result.isErr()).toBeTrue()
+      expect(actual.isErr()).toBeTrue()
```
harishv7, line 935: we can assert the type of error returned with instanceOf

src/services/db/__tests__/GitFileSystemService.spec.ts
```diff
@@ -677,182 +812,224 @@ describe("GitFileSystemService", () => {
...
+      })
+
+      const result = await GitFileSystemService.read(
         "fake-repo",
-        "non-existent-dir"
+        "fake-dir/fake-file"
       )
 
-      expect(result._unsafeUnwrapErr()).toBeInstanceOf(NotFoundError)
+      expect(result.isErr()).toBeTrue()
```
harishv7, line 1032: how about tests to ensure rollbacks are done correctly - i.e. when failures occur along one of the many steps in create flow

src/services/db/GitFileSystemService.ts
alexanderleegs: We've opted to use `gitFilePathStats` instead of `fs.existsSync`, so these have been removed!

src/services/db/GitFileSystemService.ts
alexanderleegs: We never explicitly create a directory within our existing codebase (since github only takes the presence of files within a directory to imply the existence of the directory - previously we'd just create the file directly). Previous "directory creation" operations were actually just creating a `.keep` file in that particular directory path, to ensure it showed up in github, and this behaviour stays the same, so there's no need to have a differentiated argument here

src/services/db/GitFileSystemService.ts
alexanderleegs: This is from `this.getFilePathStats(repoName, filePath)` - if it's not found, the file does not yet exist

src/services/db/RepoService.ts
alexanderleegs: Yup it'll be there as well, create dir operations are just creating a `.keep` file in the desired directory currently

src/services/db/__tests__/GitFileSystemService.spec.ts
alexanderleegs: This is the existing test behaviour - i moved the position of this test because it was failing for some reason haha

I think the sort is to ensure that the retrieval always returns in the same order, regardless of how it's returned

dcshzj: Yup, to add on, jest also compares the order, even if the items are the same. This would ensure that the unit test does not get flaky for no reason.

src/services/db/__tests__/GitFileSystemService.spec.ts
alexanderleegs: Hmm we actually only rollback in either the actual file creation, or when committing the change. I'm not entirely sure how to mock the fs call itself!

src/services/db/GitFileSystemService.ts
alexanderleegs: See above!

src/services/db/GitFileSystemService.ts
alexanderleegs: Whoops it's meant to convert stats into a boolean value but the functionality wasn't implemented fully, this has been modified in https://github.com/isomerpages/isomercms-backend/pull/892/commits/b3d87e344ceb2d146ceaf661b40c9548dcc5102d

src/services/db/GitFileSystemService.ts
alexanderleegs: Changed in https://github.com/isomerpages/isomercms-backend/pull/892/commits/b3d87e344ceb2d146ceaf661b40c9548dcc5102d!

src/services/db/__tests__/GitFileSystemService.spec.ts
harishv7: I see, I believe there is a jest function like strict equals or something to remove impact of order. But this looks good to me!

src/services/db/RepoService.ts
alexanderleegs: https://github.com/isomerpages/isomercms-backend/pull/892/commits/20d3539fadd5f6e1abbe84a363637c69a9172a19

src/services/db/__tests__/GitFileSystemService.spec.ts
alexanderleegs: Fixed in https://github.com/isomerpages/isomercms-backend/pull/892/commits/f0dd458b5ecea65092b685a34ec2e332d591f997

src/services/db/GitFileSystemService.ts
harishv7: i thought GitFileSystemService will decide to create the .keep but actually it comes from the upstream directory services

harishv7 (approved):

src/services/db/__tests__/GitFileSystemService.spec.ts
harishv7: okay, let's focus on the manual testing for rollback and add this in later. I think @dcshzj added a mock-fs library into the spec file which we can use possibly

PR #893, opened 2023-08-14
Feat/siteLaunch/integrationForFinalScreens

kishore03109:
## Problem

Currently the BE routes does not support the final screens for site launch. can be reviewed together with [fe pr](https://github.com/isomerpages/isomercms-frontend/pull/1410) 

## Solution

This PR modifies a failure state to actually be returned to the fe.

### DIFF
--- a/src/services/infra/InfraService.ts
+++ b/src/services/infra/InfraService.ts
@@ -356,21 +356,21 @@ export default class InfraService {
       return errAsync(generatedDnsRecords.error)
     }
 
-    let jobStatus: SiteLaunchStatus
+    let siteLaunchStatus: SiteLaunchStatus
     switch (site.value.jobStatus) {
       case JobStatus.Ready:
-        jobStatus = SiteLaunchStatusObject.Launched
+        siteLaunchStatus = SiteLaunchStatusObject.Launched
         break
       case JobStatus.Failed:
-        jobStatus = SiteLaunchStatusObject.Failure
+        siteLaunchStatus = SiteLaunchStatusObject.Failure
         break
       default:
-        jobStatus = SiteLaunchStatusObject.Launching
+        siteLaunchStatus = SiteLaunchStatusObject.Launching
         break
     }
 
     return okAsync<SiteLaunchDto>({
-      siteLaunchStatus: jobStatus,
+      siteLaunchStatus,
       dnsRecords: generatedDnsRecords.value.dnsRecords,
       siteUrl: generatedDnsRecords.value.siteUrl,
     })
--- a/src/services/infra/InfraService.ts
+++ b/src/services/infra/InfraService.ts
@@ -347,7 +347,7 @@ export default class InfraService {
     }
     if (site.value.siteStatus !== SiteStatus.Launched) {
       return okAsync<SiteLaunchDto>({
-        siteStatus: SiteLaunchStatusObject.NotLaunched,
+        siteLaunchStatus: SiteLaunchStatusObject.NotLaunched,
       })
     }
 
@@ -370,7 +370,7 @@ export default class InfraService {
     }
 
     return okAsync<SiteLaunchDto>({
-      siteStatus: jobStatus,
+      siteLaunchStatus: jobStatus,
       dnsRecords: generatedDnsRecords.value.dnsRecords,
       siteUrl: generatedDnsRecords.value.siteUrl,
     })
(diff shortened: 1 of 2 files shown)


src/services/infra/InfraService.ts
```diff
@@ -355,12 +356,17 @@ export default class InfraService {
...
 
+    let jobStatus: SiteLaunchStatus
+    if (site.value.jobStatus === JobStatus.Ready) {
+      jobStatus = SiteLaunchStatusObject.Launched
+    } else if (site.value.jobStatus === JobStatus.Failed) {
+      jobStatus = SiteLaunchStatusObject.Failure
+    } else {
+      jobStatus = SiteLaunchStatusObject.Launching
+    }
+
```
QiluXie, line 367: Two nit non-blocking comments here:

1. The variable `jobStatus` can be renamed to `siteStatus` to make the code more readable, since there are already `site.value.jobStatus` and `JobStatus`.
2. We can consider refactoring the `if-else` conditions to `switch` on `site.value.jobStatus` to make the code more extendable in case there will be more status in the future.

src/types/siteInfo.ts
```diff
@@ -26,6 +26,7 @@ export const SiteLaunchStatusObject = {
   Launched: "LAUNCHED",
   NotLaunched: "NOT_LAUNCHED",
   Launching: "LAUNCHING",
+  Failure: "FAILURE",
```
QiluXie, line 29: Nit and non-blocking: `Failure` can be renamed to `Failed` and the value could be `FAILED` to make it consistent with other site launch status which are all past or ongoing tense of verb.

QiluXie (commented):
Left nit and non-blocking comments

QiluXie:
@kishore03109 is this PR linked to any Jira ticket?

src/services/infra/InfraService.ts
kishore03109: sure for point 2 f0f01b1350829d8da6127c0b70424a399c9f3d58

Highly against point 1. For the purposes of our repo, our db has a `siteStatus` and a `jobStatus` column that semanitically refers to different things. I think to keep the phrasing consistent, `jobStatus` is used.

kishore03109 (commented):
@QiluXie nope, realised this was needed only during some manual integration testing, I can add this in as a 1 pointer

src/services/infra/InfraService.ts
QiluXie: Thanks. For point 1, I noticed that the function is returning the following:
```ts
 return okAsync<SiteLaunchDto>({
      siteStatus: jobStatus,
      dnsRecords: generatedDnsRecords.value.dnsRecords,
      siteUrl: generatedDnsRecords.value.siteUrl,
    })
```
the value of `jobStatus` is being returned as `siteStatus` property inside the `SiteLaunchDto` object eventually. So the value will be referred by `siteStatus` when the return object is used, is this not consistent with the DB column name?

src/types/siteInfo.ts
kishore03109: sure correcsponging fe pr done b57172550a617e803b96f134c9589de6d85f676f

src/services/infra/InfraService.ts
kishore03109: Fair point. The reason for this is because I didnt want the FE to be concerned about `SiteStatus` + `JobStatus`. ie, FE should only be concerned about whether the site is `LAUNCHED | FAILED | LAUNCHING`. 

While I am against changing the var name `jobStatus`, I am happy to change the `SiteLaunchDto` key value of `siteStatus` to reduce confusion, do you have any suggestions for the naming?

QiluXie: Sure, perhaps we could change it to `siteLaunchStatus`? 
Also since the type of `jobStatus` is already `SiteLaunchStatus` in the line:
`let jobStatus: SiteLaunchStatus`
We could just change var `jobStatus` to `siteLaunchStatus`, I feel `siteLaunchStatus` is different enough from `siteStatus` to avoid any confusion. (understand the value of `siteLaunchStatus` is probably mapped to `jobStatus` in the DB column).

kishore03109: sure a67518834aa2cd7312ac115e930fd88b36b2abee

QiluXie (approved):
lgtm. Thanks @kishore03109 for addressing the comments quickly!

QiluXie:
> @QiluXie nope, realised this was needed only during some manual integration testing, I can add this in as a 1 pointer

Sure, please. It helps to articulate your effort in this sprint, thanks!

PR #895, opened 2023-08-14
Feat/ggs rollback handler

alexanderleegs:
## Problem

This PR introduces rollback functionality for whitelisted ggs repos. Currently, this is being done by the `attachRollbackRouteHandler` wrapper by taking Github as the source of truth before and reverting to the latest commit at the beginning of the failed transaction - this PR extends that method for whitelisted repos to instead take the latest commit of the EFS, and revert the local copy of the repo stored on EFS before force pushing that to github.

It also introduces an optional `force` param to the `push` method in `GitFileSystemService`, in order to support reverting the remote repo to the rolled back state on the EFS.

## Tests

For local testing only

- [ ] Ensure the whitelisted target git repo on your local file system is identical to remote
- [ ] Make the create method return an error (e.g. `return errAsync(new GitFileSystemError("test create error"))`)
- [ ] Attempt to create a collection page (Operation involves an `update` step, which should succeed, then a `create` step, which should fail)
- [ ] Check the local git file system - should be up to date with remote
- [ ] Check the github repo - no new commits should have appeared

### DIFF
--- a/src/middleware/routeHandler.js
+++ b/src/middleware/routeHandler.js
@@ -1,9 +1,22 @@
 const { backOff } = require("exponential-backoff")
+const SimpleGit = require("simple-git")
+
+const { config } = require("@config/config")
 
 const { default: GithubSessionData } = require("@classes/GithubSessionData")
 
 const { lock, unlock } = require("@utils/mutex-utils")
 const { getCommitAndTreeSha, revertCommit } = require("@utils/utils.js")
+
+const GitFileSystemError = require("@root/errors/GitFileSystemError")
+const {
+  default: GitFileSystemService,
+} = require("@services/db/GitFileSystemService")
+
+const WHITELISTED_GIT_SERVICE_REPOS = config.get(
+  "featureFlags.ggsWhitelistedRepos"
+)
+const BRANCH_REF = config.get("github.branchRef")
 
 // Used when there are no write API calls to the repo on GitHub
 const attachReadRouteHandlerWrapper = (routeHandler) => async (
@@ -31,6 +44,8 @@ const attachWriteRouteHandlerWrapper = (routeHandler) => async (
   await unlock(siteName)
 }
 
+const gitFileSystemService = new GitFileSystemService(new SimpleGit())
+
 const attachRollbackRouteHandlerWrapper = (routeHandler) => async (
   req,
   res,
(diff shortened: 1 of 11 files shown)


src/middleware/routeHandler.js
```diff
@@ -40,32 +55,72 @@ const attachRollbackRouteHandlerWrapper = (routeHandler) => async (
...
       siteName,
-      accessToken
+      BRANCH_REF
     )
+    if (result.isErr()) {
+      await unlock(siteName)
+      next(result.err)
+    } else {
+      originalCommitSha = result.value.sha
+      if (!originalCommitSha) {
```
harishv7, line 75: when would this occur?

src/services/db/GitFileSystemService.ts
```diff
@@ -341,7 +333,10 @@ export default class GitFileSystemService {
   }
 
   // Push the latest changes to upstream Git hosting provider
-  push(repoName: string): ResultAsync<string, GitFileSystemError> {
+  push(
+    repoName: string,
+    isForce?: boolean
```
harishv7, line 338: maybe can put as default value isForce = false

src/middleware/routeHandler.js
alexanderleegs: This is mostly just a guard clause, it shouldn't occur normally!

src/services/db/GitFileSystemService.ts
alexanderleegs: Fixed in https://github.com/isomerpages/isomercms-backend/pull/895/commits/5e49f77a80c941a1dac4e3dd3fc857e79f3ef878!

src/middleware/routeHandler.js
```diff
@@ -1,10 +1,23 @@
 const { backOff } = require("exponential-backoff")
```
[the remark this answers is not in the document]
QiluXie: Looks like force push is not being used? 
Shouldn't have this error if it's a force push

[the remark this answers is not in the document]
harishv7: tested now and works fine

src/middleware/routeHandler.js
harishv7: okay!

harishv7 (approved):

PR #902, opened 2023-08-15
fix: use synchronous fs mkdir

harishv7:
## Problem

Closes IS-432

## Solution

Make mkdir synchronous

**Breaking Changes**

  - Details ...

### DIFF
--- a/src/services/db/GitFileSystemService.ts
+++ b/src/services/db/GitFileSystemService.ts
@@ -481,15 +481,10 @@ export default class GitFileSystemService {
       .orElse((error) => {
         if (error instanceof NotFoundError) {
           // Create directory if it does not already exist
-          try {
-            fs.mkdirSync(pathToEfsDir)
-            return ok(true)
-          } catch (mkdirErr) {
-            logger.error(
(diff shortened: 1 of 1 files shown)


dcshzj:
Does using `ResultAsync.fromPromise` work here instead of the try-catch?

harishv7:
> Does using `ResultAsync.fromPromise` work here instead of the try-catch?

Actually realised mkdirSync does not throw any errors, so we don't need the try-catch
https://nodejs.org/api/fs.html#fsmkdirsyncpath-options

dcshzj:
Wait sorry, I meant using `ResultAsync.fromPromise` with `fs.promises.mkdir` instead of using `mkdirSync`. My understanding is that the sync equivalent methods are not suitable for production use so I was thinking if it was due to the try-catch that was causing the original error.

harishv7:
> Wait sorry, I meant using `ResultAsync.fromPromise` with `fs.promises.mkdir` instead of using `mkdirSync`. My understanding is that the sync equivalent methods are not suitable for production use so I was thinking if it was due to the try-catch that was causing the original error.

@dcshzj got it, fixed

dcshzj (approved):

PR #905, opened 2023-08-16
Fix/git file lock

alexanderleegs:
## Problem

This PR resolves issues we currently have with concurrent edits on the same file. As not all git operations are awaited before returning (git push is done async), it becomes possible for a git repo to still be locked when a new user attempts to make a change. However, our existing route handlers only look at the status of our mutex lock, which does not take this into consideration. This PR adds an additional method `hasGitFileLock` to `GitFileSystemService`, and extends our existing write and rollback handlers to check for the existence of the git file lock before proceeding.

This PR has been tested on staging.

To test on local dev:

- [ ] On a whitelisted repo, from the root of the repo, create a `.git/index.lock` file
- [ ] Attempt to perform any operation that modifies the git state of that repo on the cms
- [ ] You should receive a 423 on the frontend, reflected as an error toast `Your page could not be updated. Someone else is currently modifying repo a-test-v4. Please try again later.`

Resolves IS-434.

### DIFF
--- a/src/middleware/routeHandler.js
+++ b/src/middleware/routeHandler.js
@@ -8,8 +8,10 @@ const { default: GithubSessionData } = require("@classes/GithubSessionData")
 const { lock, unlock } = require("@utils/mutex-utils")
 const { getCommitAndTreeSha, revertCommit } = require("@utils/utils.js")
 
-const { ConflictError } = require("@root/errors/ConflictError")
-const GitFileSystemError = require("@root/errors/GitFileSystemError")
+const GitFileSystemError = require("@root/errors/GitFileSystemError").default
+
+const LockedError = require("@root/errors/LockedError").default
+
 const {
   default: GitFileSystemService,
 } = require("@services/db/GitFileSystemService")
@@ -37,7 +39,7 @@ const handleGitFileLock = async (repoName, next) => {
   if (isGitLocked) {
     logger.error(`Failed to lock repo ${repoName}: git file system in use`)
     next(
-      new ConflictError(
+      new LockedError(
         `Someone else is currently modifying repo ${repoName}. Please try again later.`
       )
     )
--- a/src/middleware/routeHandler.js
+++ b/src/middleware/routeHandler.js
@@ -8,15 +8,38 @@ const { default: GithubSessionData } = require("@classes/GithubSessionData")
 const { lock, unlock } = require("@utils/mutex-utils")
 const { getCommitAndTreeSha, revertCommit } = require("@utils/utils.js")
 
+const { ConflictError } = require("@root/errors/ConflictError")
 const GitFileSystemError = require("@root/errors/GitFileSystemError")
 const {
   default: GitFileSystemService,
 } = require("@services/db/GitFileSystemService")
 
+const logger = require("@logger/logger").default
+
 const WHITELISTED_GIT_SERVICE_REPOS = config.get(
   "featureFlags.ggsWhitelistedRepos"
 )
 const BRANCH_REF = config.get("github.branchRef")
+
+const gitFileSystemService = new GitFileSystemService(new SimpleGit())
+
+const isRepoWhitelisted = (siteName) =>
+  WHITELISTED_GIT_SERVICE_REPOS.split(",").includes(siteName)
+
+const handleGitFileLock = async (repoName, next) => {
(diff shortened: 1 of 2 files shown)


src/middleware/routeHandler.js
```diff
@@ -8,16 +8,39 @@ const { default: GithubSessionData } = require("@classes/GithubSessionData")
...
+const logger = require("@logger/logger").default
+
 const WHITELISTED_GIT_SERVICE_REPOS = config.get(
   "featureFlags.ggsWhitelistedRepos"
 )
 const BRANCH_REF = config.get("github.branchRef")
 
+const gitFileSystemService = new GitFileSystemService(new SimpleGit())
+
+const isRepoWhitelisted = (siteName) =>
```
kishore03109, line 26: hmm could we shift this out into a separate place since this logic is exactly the same in `ReposService.ts` ?

kishore03109 (commented):
code wise ok, whats our test plan for this PR?

src/services/db/GitFileSystemService.ts
```diff
@@ -50,6 +50,21 @@ export default class GitFileSystemService {
     this.git = git
   }
 
+  hasGitFileLock(repoName: string): ResultAsync<boolean, GitFileSystemError> {
+    const gitFileLockPath = ".git/index.lock"
+    return this.getFilePathStats(repoName, gitFileLockPath)
```
harishv7, line 55: will this check for repoName/.git/index.lock -> check if `/` is added automatically

src/middleware/routeHandler.js
harishv7: used to have this is a separate utils called ggs-utils.ts, but we removed it because we didn't want any other file to be aware of this special GGS logic apart from RepoService

However, for this routeHandler, I think we don't really have a choice? Unless we inject RepoService as a dependency into routeHandler instead of GitFIleSystemService. So RepoService will be the middleman for the isGitLocked() method. That will be a possible approach

src/services/db/GitFileSystemService.ts
alexanderleegs: Yup, `getFilePathStats` checks for `${EFS_VOL_PATH}/${repoName}/${filePath}`

src/middleware/routeHandler.js
alexanderleegs: I opted for attaching `GitFileSystemService` because it has no dependencies, whereas having `RepoService` instead would mean having to initialise a separate `RepoService` (and its dependencies) inside the route handler!

alexanderleegs:
> code wise ok, whats our test plan for this PR?

We'll need to test this on staging first, since it only gets triggered if the git file system is in place!

QiluXie:
@alexanderleegs this PR is not ready for review yet, right?
Could you turn it to draft if more work is needed.

alexanderleegs:
> @alexanderleegs this PR is not ready for review yet, right? Could you turn it to draft if more work is needed.

Mostly needs testing - not possible to replicate on local so we need to test on staging

src/middleware/routeHandler.js
kishore03109: I am confused here :( we abstract this `isRepoWhitelisted` method, there wouldnt be any dep on `RepoService` no?

src/middleware/routeHandler.js
```diff
@@ -8,16 +8,46 @@ const { default: GithubSessionData } = require("@classes/GithubSessionData")
 const { lock, unlock } = require("@utils/mutex-utils")
```
harishv7, line 8: @alexanderleegs the error I am seeing on FE toast is "Your page could not be updated. Unable to rollback to original state"

Is this the expected error?

Let's also update the test plan with the exact toast error we should see

harishv7: nit: the msg needs a `.` at the back

harishv7 (approved):

src/middleware/routeHandler.js
alexanderleegs: Resolved offline - was a testing error!

The message has been modified in https://github.com/isomerpages/isomercms-backend/pull/905/commits/49dc1a50ccb8a26636818fc1b76dacee4f994fe5

PR #909, opened 2023-08-18
feat(integration): reviews inte test

kishore03109:
## Problem

When the PR #867 was introduced, the test case was mocked so there was no proof of functionality (at the unit test level) + no test cases to actually prevent regression. 

## Solution

Add integration test for this.

## Before & After Screenshots

**BEFORE**:

NOTE: This is run on a codebase where the changes introduced in #867 was _**reverted**_. 
Screenshot 2023-08-18 at 1 14 45 PM

**AFTER**:

Screenshot 2023-08-18 at 1 15 20 PM

## Tests

integration tests

### DIFF
--- a/src/integration/Reviews.spec.ts
+++ b/src/integration/Reviews.spec.ts
@@ -719,6 +719,23 @@ describe("Review Requests Integration Tests", () => {
       // Assert
       expect(actual.statusCode).toEqual(404)
     })
+
+    it("should not throw an error if two seb-sequent calls are being made", async () => {
+      const app = generateRouterForUserWithSite(
+        subrouter,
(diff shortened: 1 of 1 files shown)


alexanderleegs (approved):

src/integration/Reviews.spec.ts
```diff
@@ -719,6 +719,23 @@ describe("Review Requests Integration Tests", () => {
       // Assert
       expect(actual.statusCode).toEqual(404)
     })
+
+    it("should not throw an error if two seb-sequent calls are being made", async () => {
```
dcshzj, line 723: nit: should this be **sub**-sequent?

PR #910, opened 2023-08-18
fix(sl): email adr

kishore03109:
[todo]: double cfm with shaz about this behaviour first
## Problem

SL failure emails currently sent to admin, 

## Solution

send to support instead


alexanderleegs (approved):
lgtm if this has been cleared with Shaz!

PR #911, opened 2023-08-18
Fix/mutex lock

alexanderleegs:
## Problem
This PR fixes issues with our existing mutex lock. It has been tested on both local dev and staging.

Currently, our route handler behaviour seems to suffer from 2 issues that make our mutex locks not work as intended:

- The wrapped route is not awaited, resulting in the existing behaviour where the mutex locks and unlocks while the method is still executing
- Express' `next` does not cut off the remainder of the method, and must be cut off explicitly - in our existing implementation, even if the mutex lock is hit, the actual route still continues to execute

This PR modifies the behaviour of both our write and rollback wrappers to handle this scenario.

In addition, the returned error has been modified from `409` to `423` (Locked) - this is to better reflect the reason for rejecting the request, and also results in a regular error toast on the frontend instead of the special conflict modal that would normally show up for modification of outdated files.

Resolves IS-478

### DIFF
--- a/src/errors/LockedError.ts
+++ b/src/errors/LockedError.ts
@@ -1,0 +1,11 @@
+import { BaseIsomerError } from "./BaseError"
+
+export default class LockedError extends BaseIsomerError {
+  constructor(message: string) {
+    super({
+      status: 423,
+      code: "LockedError",
+      message,
+    })
+  }
+}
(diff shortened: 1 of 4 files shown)


harishv7 (commented):
@alexanderleegs have we retried how this functions with FE with the new 423 error?

alexanderleegs:
> @alexanderleegs have we retried how this functions with FE with the new 423 error?

Yup, it shows up as an error toast with the message we've specified (`Your page could not be updated. Someone else is currently modifying repo a-test-v4. Please try again later.`)

harishv7 (approved):

PR #912, opened 2023-08-22
feat: make tracked sites env var

harishv7:
## Problem

Currently the tracked sites are hard-coded and this means a full-redeploy or hotfix releases are required if we want to add more sites along the way. So, we make this into an env var on BE

Closes IS-489

## Solution

Add to convict + add new env var

**Breaking Changes**

  - Details ...

- `GGS_EXPERIMENTAL_TRACKING_SITES` : list of sites for which we are tracking the GitHub API hits

### DIFF
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -412,6 +412,12 @@ const config = convict({
       format: String,
       default: "",
     },
+    ggsTrackedSites: {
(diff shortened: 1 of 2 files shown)


seaerchin (approved):

harishv7:
Rolled out as hotfix 0.39.1

PR #916, opened 2023-08-23
chore: remove test for concurrent viewed call

alexanderleegs:
Reverts https://github.com/isomerpages/isomercms-backend/pull/909 - now that our mutex lock works, the test throws a 423 error instead, and we're unable to replicate the conditions which we were testing on production anymore (prod instead returns 423 for 2 update calls in quick succession)


QiluXie (approved):
lgtm

harishv7 (approved):

PR #924, opened 2023-08-24
Release/0.40.1

alexanderleegs:
This PR fixes an issue where our calls to `/repos/{owner}/{repo}/commits` from the Github API were not referencing the correct branch. This issue was because we were referring to `ref`, when the expected param is `sha`. This caused issues for repos where the default branch was mistakenly set to `master` or any other non staging branch.

See [link](https://docs.github.com/en/rest/commits/commits?apiVersion=2022-11-28#list-commits) for api docs.

This is a 2-part solution for resolving IS-501 - we have already run a script to modify the default of all existing repos to ensure they are `staging`.

Testing:

- [ ] Visit the workspace of `a-test-v4` (master is set to the default branch)
- [ ] Rename a folder
- [ ] Check that staging history has not been reset to master


harishv7:
@alexanderleegs Followed the test plan with default branch for repo as master, but repo was still reset to master
Screenshot 2023-08-24 at 11 35 42 PM

harishv7 (approved):

PR #927, opened 2023-08-29
IS-312: Inject User Data into GrowthBook for Segmentation

harishv7:
## Problem

We need to have context of user attributes for us to do targeted rollouts/experiments.

Closes IS-312

## Solution

We inject information at both the authentication and authorization middlewares based on what we know about the user at that point in time.

Note: To be reviewed in conjunction with IS-310 which introduces GrowthBook for BE.

**Breaking Changes**

  - Details ...

### DIFF
--- a/src/types/express/request.d.ts
+++ b/src/types/express/request.d.ts
@@ -2,7 +2,8 @@ import { GrowthBook } from "@growthbook/growthbook"
 
 import { FeatureFlags } from "@root/types/featureFlags"
 
-// to make the file a module and avoid the TypeScript error
+// Empty export to make this file a module therefore allow
+// global augmentation (avoiding the related error)
 export {}
 
 declare global {
--- a/src/types/express/request.d.ts
+++ b/src/types/express/request.d.ts
@@ -1,12 +1,14 @@
 import { GrowthBook } from "@growthbook/growthbook"
 
+import { FeatureFlags } from "@root/types/featureFlags"
+
 // to make the file a module and avoid the TypeScript error
 export {}
 
 declare global {
   namespace Express {
     export interface Request {
-      growthbook: GrowthBook
+      growthbook: GrowthBook<FeatureFlags>
     }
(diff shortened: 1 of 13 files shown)


harishv7:
Note: Tests will pass once #928 is merged into this

src/types/express/request.d.ts
```diff
@@ -0,0 +1,14 @@
+import { GrowthBook } from "@growthbook/growthbook"
+
+import { FeatureFlags } from "@root/types/featureFlags"
+
+// to make the file a module and avoid the TypeScript error
```
QiluXie, line 5: nit: added more details in comment for the nature of the error we are trying to avoid here 

```suggestion
// Empty export to make this file a module therefore allow global augmentation (avoiding the related error)
```

src/types/featureFlags.ts
```diff
@@ -3,3 +3,12 @@
...
   samplekey: string
 }
+
+// List of attributes we set in GrowthBook Instance in auth middleware
+export type GrowthBookAttributes = {
+  isomerUserId: string
+  email: string
+  githubId?: string
+  siteName?: string
+  role?: "email" | "admin"
```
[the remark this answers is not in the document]
harishv7: Actually this is not needed now, but only if we need to segment by roles. I will remove this for now and we can evaluate if there is an actual need for it later

src/types/express/request.d.ts
harishv7: updatedd

QiluXie (approved):
lgtm. Thanks @harishv7 for addressing the comments quickly.
Should merge #928 first and make sure the CI testing passed before merging this one.

harishv7:
> lgtm. Thanks @harishv7 for addressing the comments quickly. Should merge #928 first and make sure the CI testing passed before merging this one.

@QiluXie all checks have passed. merging now, thanks

PR #928, opened 2023-08-29
IS-506: Migrate GGS Whitelist Feature Flag to GB

harishv7:
## Problem

We are currently using env vars to rollout GGS implementation which requires a rebuild each time we change something.

Closes IS-506

## Solution

Use GrowthBook feature flag. Introduces a new flag `ggs_whitelisted_repos` with a json structure of `{ "repos": string[] }`. Please check the portal for more info.

Note: To be reviewed in conjunction with IS-310 and IS-312

**Breaking Changes**

  - Details ...

## Tests

- Ensure whitelisted repos are being read in from GB. To do so, can add a simple `console.log` in `isRepoWhitelisted` function of `RepoService`.
- Ensure all unit tests pass especially for `RepoService.spec.ts`

## Deploy Notes

Please remove existing env var of `WHITELISTED_GIT_SERVICE_REPOS` from EB.

**Deprecated environment variables**:

- `WHITELISTED_GIT_SERVICE_REPOS`

**New feature flags**

- `ggs_whitelisted_repos`: ensure correct values are set on GB portal

### DIFF
--- a/src/middleware/routeHandler.js
+++ b/src/middleware/routeHandler.js
@@ -69,10 +69,9 @@ const attachWriteRouteHandlerWrapper = (routeHandler) => async (
     })
   }
 
-  let isGitAvailable = false
-  if (!isRepoWhitelisted(siteName, ggsWhitelistedRepos.repos)) {
-    isGitAvailable = true
-  } else {
+  let isGitAvailable = true
+  // only check git file lock if the repo is whitelisted
+  if (isRepoWhitelisted(siteName, ggsWhitelistedRepos.repos)) {
     isGitAvailable = await handleGitFileLock(siteName, next)
   }
 
--- a/src/middleware/routeHandler.js
+++ b/src/middleware/routeHandler.js
@@ -16,18 +16,14 @@ const {
 
 const logger = require("@logger/logger").default
 
-const WHITELISTED_GIT_SERVICE_REPOS = config.get(
-  "featureFlags.ggsWhitelistedRepos"
-)
 const BRANCH_REF = config.get("github.branchRef")
 
 const gitFileSystemService = new GitFileSystemService(new SimpleGit())
 
-const isRepoWhitelisted = (siteName) =>
-  WHITELISTED_GIT_SERVICE_REPOS.split(",").includes(siteName)
+const isRepoWhitelisted = (siteName, ggsWhitelistedRepos) =>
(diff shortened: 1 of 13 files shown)


src/services/db/RepoService.ts
```diff
@@ -36,8 +33,27 @@ export default class RepoService extends GitHubService {
...
+      repos: [],
+    })
+
+    console.log(
+      `I RECEIVED: ${JSON.stringify(
+        ggsWhitelistedRepos
+      )} to compare with ${repoName}. Returning ${ggsWhitelistedRepos.repos.includes(
+        repoName
+      )}`
+    )
```
QiluXie, line 54: Will this log print out all whitlisted repo names inside `ggsWhitelistedRepos`? Is it necessary?

src/middleware/routeHandler.js
```diff
@@ -64,8 +60,22 @@ const attachWriteRouteHandlerWrapper = (routeHandler) => async (
...
+      repos: [],
+    })
+  }
+
+  let isGitAvailable = false
+  if (!isRepoWhitelisted(siteName, ggsWhitelistedRepos.repos)) {
+    isGitAvailable = true
+  } else {
+    isGitAvailable = await handleGitFileLock(siteName, next)
+  }
```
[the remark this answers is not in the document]
harishv7: great suggestion, thanks!

src/services/db/RepoService.ts
harishv7: oops, this was for local debug. have updated the log statement - need this temporarily in case we want to debug on staging/prod

QiluXie (approved):
lgtm. Thanks @harishv7 for resolving the comments quickly!

alexanderleegs (approved):

alexanderleegs (commented):
Implementation looks fine! It looks like it's only active on write/rollback route handlers though, so if we ever need to modify the content we're passing via GB we'll need to add this to the read route handler as well. We can just add this in as needed in future

harishv7:
> Implementation looks fine! It looks like it's only active on write/rollback route handlers though, so if we ever need to modify the content we're passing via GB we'll need to add this to the read route handler as well. We can just add this in as needed in future

@alexanderleegs sounds good. we can add in when we have a flag that touches the read route handler. for ggs, i have replaced all instances of `isRepoWhitelisted` with the feature flag

PR #933, opened 2023-09-04
feat(template): add ffs as a manual check-in

kishore03109:
## Problem

It is becoming harder for one eng to keep track of all the other feature flags that are avail. It is also well known a lot of production bugs in big tech companies are results for failed roll outs for experiments. (eg, amplify build times vs repo privatisation) 

Closes [insert issue #]

## Solution

Eng to manually tick off that their features are compat with the permutations of the features (or at least spend some time thinking intentionallu about it) 

**Breaking Changes**

  - Details ...
- No - this PR is backwards compatible with ALL of the following feature flags in this [doc](https://www.notion.so/opengov/Existing-feature-flags-518ad2cdc325420893a105e88c432be5)

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -14,7 +14,11 @@ Closes [insert issue #]
 
 - [ ] Yes - this PR contains breaking changes
   - Details ...
-- [ ] No - this PR is backwards compatible
+- No - this PR is backwards compatible with ALL of the following feature flags
+  - [ ] ggs_whitelisted_repos
+  - [ ] styles
+  - [ ] REACT_APP_IS_SITE_PRIVATISATION_ACTIVE
+  - [ ] REACT_APP_SITE_LAUNCH_FEATURE_WHITELISTED_REPOS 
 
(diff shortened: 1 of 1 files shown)


seaerchin (commented):
Can we change this to point to growthbook instead of having it listed manually here?

Intention is to reduce overhead + less changes have to be made when we go to 100%

harishv7:
> Can we change this to point to growthbook instead of having it listed manually here?
> 
> Intention is to reduce overhead + less changes have to be made when we go to 100%

@seaerchin we have some in env and growthbook. hence, we need a consolidated list somewhere

seaerchin:
> @seaerchin we have some in env and growthbook. hence, we need a consolidated list somewhere

yep - what i meant is that the growthbook env vars should just be a link to growthbook. otherwise, we'd need to maintain a duplicate (here, in the PR template and in growthbook also)

kishore03109:
@seaerchin instead linking to doc as sot

dcshzj (approved):

PR #939, opened 2023-09-12
IS-342: Clone repos into EFS vol path

harishv7:
## Problem

We need to ensure new repos/site creations are automatically cloned into the EFS

Closes IS-342

## Solution

We are already creating a local git folder with the base template under `/tmp` and doing a `git push` to create it on github. Instead of creating it in `/tmp`, now it will create within the `EFS_VOL_PATH` and instead of http git remote, it uses ssh protocol.

Note: Had to change from isomorphic-git to using simple-git as we are using SSH auth. SSH auth is not supported in isomorphic git.

Example of site created with this code: https://github.com/isomerpages/harish-test-staging/tree/staging

**Breaking Changes**

  - Details ...

## Tests
1. Create a site using the staging site-creation form
2. Ensure the site is created on GitHub. Default branch should be staging as shown in below screenshot.
3. Both staging and master should have 2 commits - "Initial commit" and "Set URLs" from `isomeradmin` user
4. Visit CMS Staging and you should see the new site there
5. Ensure the repo exists on EFS staging

## Depedency Changes

Removes `isomorphic-git`. No longer required as we are using `simple-git`

### DIFF
--- a/src/services/identity/ReposService.ts
+++ b/src/services/identity/ReposService.ts
@@ -8,6 +8,7 @@ import git from "isomorphic-git"
 import http from "isomorphic-git/http/node"
 import { ResultAsync, errAsync, okAsync } from "neverthrow"
 import { ModelStatic } from "sequelize"
+import { SimpleGit } from "simple-git"
 
 import { config } from "@config/config"
 
@@ -35,6 +36,7 @@ const EFS_VOL_PATH = config.get("aws.efs.volPath")
 
 interface ReposServiceProps {
   repository: ModelStatic<Repo>
+  simpleGit: SimpleGit
 }
 
 type octokitCreateTeamResponseType = GetResponseTypeFromEndpointMethod<
@@ -57,8 +59,11 @@ export default class ReposService {
   // but having this seems to help navigation in some editors.
   private readonly repository: ReposServiceProps["repository"]
 
-  constructor({ repository }: ReposServiceProps) {
+  private readonly simpleGit: SimpleGit
+
+  constructor({ repository, simpleGit }: ReposServiceProps) {
     this.repository = repository
+    this.simpleGit = simpleGit
   }
 
   getLocalRepoPath = (repoName: string) => `${EFS_VOL_PATH}/${repoName}`
@@ -77,7 +82,7 @@ export default class ReposService {
   }): Promise<Repo> => {
     const repoUrl = `https://github.com/isomerpages/${repoName}`
 
-    await this.createRepoOnGithub(repoName)
+    // await this.createRepoOnGithub(repoName)
     if (!isEmailLogin) {
       await this.createTeamOnGitHub(repoName)
     }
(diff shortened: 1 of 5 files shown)


src/services/identity/ReposService.ts
```diff
@@ -219,58 +214,37 @@ export default class ReposService {
...
-      dir,
-      ref: "staging",
-      singleBranch: true,
-      url: SITE_CREATION_BASE_REPO_URL,
-      depth: 1,
-    })
+    fs.mkdirSync(dir)
+    await this.simpleGit
+      .cwd(dir)
+      .clone(SITE_CREATION_BASE_REPO_URL, dir, ["-b", "staging"])
```
seaerchin, line 220: sanity check here, this appears to create a branch called `staging` - will this fail if `staging` already exists? or will the `staging` branch always be created here

src/services/identity/ReposService.ts
```diff
@@ -108,37 +115,25 @@ export default class ReposService {
...
-      dir,
-      remote,
-      remoteRef: "master",
-      onAuth: () => ({ username: "user", password: SYSTEM_GITHUB_TOKEN }),
-    })
+    // 4. Merge these changes into master branch
+    await this.simpleGit.checkout("master").merge(["staging"])
+
+    // 5. Push changes to master branch
+    await this.simpleGit.push("origin", "master")
```
dcshzj, line 133: Simple Git will need to always do a `.cwd(dir)` every time before performing the action.

Same for the rest of the operations in this file too.

src/services/identity/ReposService.ts
```diff
@@ -4,10 +4,9 @@ import { retry } from "@octokit/plugin-retry"
 import { Octokit } from "@octokit/rest"
 // eslint-disable-next-line import/no-extraneous-dependencies
 import { GetResponseTypeFromEndpointMethod } from "@octokit/types"
-import git from "isomorphic-git"
```
dcshzj, line 7: Can we also delete the dependency from package.json? I think this is the only place that `isomorphic-git` is being used.

src/services/identity/ReposService.ts
```diff
@@ -108,37 +115,25 @@ export default class ReposService {
...
-      message: "Set URLs",
-      author: {
-        name: ISOMER_GITHUB_ORGANIZATION_NAME,
-        email: "isomeradmin@users.noreply.github.com",
-      },
-    })
+    await this.simpleGit
+      .cwd(dir)
+      .checkout("staging") // ensure on staging branch
+      .add(".")
```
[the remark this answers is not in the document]
harishv7: @seaerchin why do you think there could be differences? this is in site creation flow, so the repos are brand new is my understanding

src/services/identity/ReposService.ts
```diff
@@ -108,37 +115,25 @@ export default class ReposService {
...
-      remoteRef: "master",
-      onAuth: () => ({ username: "user", password: SYSTEM_GITHUB_TOKEN }),
-    })
+    // 4. Merge these changes into master branch
+    await this.simpleGit.checkout("master").merge(["staging"])
+
+    // 5. Push changes to master branch
+    await this.simpleGit.push("origin", "master")
+
+    // 6. Checkout back to staging branch
```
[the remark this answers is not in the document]
harishv7: oh this is site creation, like for the very first base setup. See this: https://github.com/isomerpages/harish-test-staging/commits/staging the 1st 2 commits

src/services/identity/ReposService.ts
harishv7: this clones from the branch staging of the remote `SITE_CREATION_BASE_REPO_URL`

src/services/identity/ReposService.ts
harishv7: Oh really? Somehow, in my tests after the first time it worked fine. Let me add it in to be cautious

src/services/identity/ReposService.ts
harishv7: good point, will do

src/services/identity/ReposService.ts
dcshzj: Yeah unfortunately we need to add it in because of concurrency issues. There can be some other place using the same simpleGit instance, which causes the working directory to change as a result and throw a series of errors.

dcshzj (approved):

src/services/identity/ReposService.ts
harishv7: sure added

src/services/identity/ReposService.ts
harishv7: done

PR #943, opened 2023-09-19
IS-412: Move fetching SSM params to prebuild

harishv7:
## Problem

Right now we face an issue with fetching envs from SSM:
- We have to fetch envs from SSM in the post deploy step because otherwise EB overwrites the env with the ones we set on the console
- On convict, we have put the params as required string
- However, doing this means whenever we change an env, EB will perform a re-build. However, since this is a re-build, it won't execute the deploy scripts which means the envs gets overriden with the ones from EB console that is missing the SSM params.

Closes IS-412

## Solution

This solution introduces a few changes:
- We move ALL of our envs to SSM except the `SSM_PREFIX` which is required to determine whether it is staging or prod
- We move the fetching of envs from post-deploy previously to now pre-deploy
- We need to keep a list of our envs in the new script for us to iterate and fetch from SSM
- We need to read from a different file other than .env which is getting overridden by EB. For this, we use our existing EFS volume which already exists for GGS solution.

### Why store on EFS?
When multiple instances boot up, the pre-deploy step executes on each of them. This is fine on normal deployment. However, consider the edge case where we want to update envs on SSM. To trigger a rebuild, we need to re-deploy from GH Actions.

However, this takes long due to our the rolling deploy. Instead, for urgent cases, we can expedite by directly running this script and doing a "Restart App Instances" on EB.

EFS enables this as you just need to run the script once instead of once on each instance. However, in the event where multiple instances are booted up at the same time, there will be concurrency issues in overwriting the .env file leading to inconsistent states. To prevent this, each instance writes to a local file and then transfers it to the EFS. A simple locking mechanism is present in the script to lock the folder we are copying into to again prevent a clash bet the 2 instances.

Note that if a param is present on script but not on SSM, current behaviour is to skip this and proceed to next param. While exiting the script and causing the deploy to fail might be a good practice of failing early, this PR's aim is to first achieve parity with our current behaviour of having convict as the checking layer (though this happens at runtime after deploy stage).

### How locking works?

Here's what happens:

1. The file descriptor 200 is associated with the file /efs/isomer/.isomer.lock.
2. If /efs/isomer/.isomer.lock doesn't exist, it's created. If it already exists, it's just opened.
3. flock then tries to acquire a lock on the file descriptor 200, which is linked to the .isomer.lock file.
4. If flock can't acquire the lock (because some other process has it), then the script exits due to || exit 1.

### Benefits

- Have a centralised management of envs for both prod + staging
- No need to manually transfer envs during node env upgrades
- Possibly reduce time taken for changing envs -> all you need to do is "Restart app server"
- Encrypt important/private envs

### Cons
- Maintaining list of envs in the script
- We can't differentiate between required and optional envs in SSM. But this is less of a con as convict is our intended safety check layer for this

**Breaking Changes**

  - This moves our envs to SSM

## Tests
- Ensure staging instances start up fine
- Ensure changes in env on SSM and re-deploy works/reflects fine (see runbook)
- Ensure changes in env on SSM and re-starting app instances works (see runbook)

Tests ran as part of checks for PR:
- Missing env on SSM -> pre-deploy script passes -> convict still catches this -> app fails to start
- Following from above, create the missing env on SSM, attempt a manual restart process (as per runbook)
- Modify the param on SSM and perform re-deploy from GitHub Actions -> ensure new values are captured

## Deploy Notes

Ensure ALL envs are moved to SSM with appropriate prefix.

**New dependencies**:

- `dotenv` : Loads env from specified file path


src/server.js
```diff
@@ -420,6 +420,8 @@ app.use((req, res, next) => {
 // error handler
 app.use(errorHandler)
 
+logger.info(`PRINTING OUT TEST VAR: ${config.get("testVar")}`)
```
harishv7, line 423: still testing, will remove later

.platform/hooks/predeploy/06_fetch_ssm_parameters.sh
```diff
@@ -0,0 +1,116 @@
+#!/bin/bash
```
harishv7, line 1: comment

.platform/hooks/predeploy/06_fetch_ssm_parameters.sh
```diff
@@ -0,0 +1,116 @@
...
+  flock -n 200 || exit 1
+
+  # Copy the local file to EFS
+  echo "Copying local env file to EFS"
+  cp /tmp/isomer/.isomer.env /efs/isomer/.isomer.env
+
+  # Ensure the file on EFS is owned by webapp so it has access
+  chown webapp:webapp /efs/isomer/.isomer.env
+
+) 200>/efs/isomer/.isomer.lock
```
seaerchin, line 105: what does the number indicate here? the [man page](https://man7.org/linux/man-pages/man1/flock.1.html) doesn't seem to have it

seaerchin (approved):
looks ok, have a question about `flock`

.platform/hooks/predeploy/06_fetch_ssm_parameters.sh
harishv7: 200 is just a file descriptor - can be an arbitrary value that is associated with a file

harishv7: updated PR desc

seaerchin: ps, how do we ensure that the fd is always given to flock?

seaerchin (approved):
looks ok, have comment

.platform/hooks/predeploy/06_fetch_ssm_parameters.sh
harishv7: Most commonly used file descriptors are:

0: Standard input (stdin)
1: Standard output (stdout)
2: Standard error (stderr)

Choosing a high number minimises chances of conflict. There is still a super low probability it is unavailable for some reason, but for purpose of deploy script such as this I think we may not need to over-engineer

harishv7: I can add a check to find a available file descriptor using a loop or smth, but yea, let me know if you think its necessary

harishv7: Highly unlikely - tried listing open FDs using `ls -l /proc/$$/fd`

it only has 4 entries, of which 0, 1, 2 and 255 are used

harishv7: @seaerchin to edit my explanation - File descriptors are local to each process. Two different processes can use the same file descriptor number simultaneously without interfering with each other

PR #947, opened 2023-09-26
feat(formsg): clone repo on webhook trigger from forms

seaerchin:
## Problem
See [here](https://isomeropengov.atlassian.net/browse/IS-558?atlOrigin=eyJpIjoiOTQyOTA0ZWIyZDNjNDhmYzhiYzJjMDgxZTY5ZjJmMTYiLCJwIjoiaiJ9)

## Solution
1. add an endpoint at `/formsg/clone-repo`
2. add handler to clone at that endpoint based on the [form's](https://form.gov.sg/6512935d3efe7a0012034f61) reply. The form has validation build in (email domain **must** be `.open.gov.sg` and **must** verify email + repo is a required field)

## New env vars
- `SITE_CLONE_FORM_KEY`: used to decrypt the form sg submissions (already added to 1PW on both `staging` and `prod` env vars; not yet added to `staging` EB yet)

## Notes
this was tested **locally** through updating the `EFS_VOL_PATH` + `originUrl` (use `https` instead of `ssh` as i was facing auth issues) and found working
Screenshot 2023-09-26 at 7 01 04 PM
Screenshot 2023-09-26 at 7 01 37 PM

### DIFF
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -283,6 +283,13 @@ const config = convict({
       format: "required-string",
       default: "",
     },
+    siteCloneFormKey: {
+      doc: "FormSG API key for site clone form",
+      env: "SITE_CLONE_FORM_KEY",
+      sensitive: true,
+      format: "required-string",
+      default: "",
+    },
   },
   postman: {
     apiKey: {
--- a/src/routes/formsgSiteCreation.ts
+++ b/src/routes/formsgSiteCreation.ts
@@ -11,10 +11,12 @@ import { BadRequestError } from "@errors/BadRequestError"
 import { getField } from "@utils/formsg-utils"
 
 import { attachFormSGHandler } from "@root/middleware"
+import GitFileSystemService from "@services/db/GitFileSystemService"
 import UsersService from "@services/identity/UsersService"
 import InfraService from "@services/infra/InfraService"
 import { mailer } from "@services/utilServices/MailClient"
 
+const SITE_CLONE_FORM_KEY = config.get("formSg.siteCloneFormKey")
 const SITE_CREATE_FORM_KEY = config.get("formSg.siteCreateFormKey")
 const REQUESTER_EMAIL_FIELD = "Government E-mail"
 const SITE_NAME_FIELD = "Site Name"
@@ -25,6 +27,7 @@ const LOGIN_TYPE_FIELD = "Login Type"
 export interface FormsgRouterProps {
   usersService: UsersService
   infraService: InfraService
+  gitFileSystemService: GitFileSystemService
 }
 
 export class FormsgRouter {
@@ -32,9 +35,16 @@ export class FormsgRouter {
 
   private readonly infraService: FormsgRouterProps["infraService"]
 
-  constructor({ usersService, infraService }: FormsgRouterProps) {
+  private readonly gitFileSystemService: FormsgRouterProps["gitFileSystemService"]
+
+  constructor({
+    usersService,
+    infraService,
+    gitFileSystemService,
+  }: FormsgRouterProps) {
     this.usersService = usersService
     this.infraService = infraService
+    this.gitFileSystemService = gitFileSystemService
     // We need to bind all methods because we don't invoke them from the class directly
     autoBind(this)
   }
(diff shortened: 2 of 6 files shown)


src/routes/formsgSiteCreation.ts
```diff
@@ -157,9 +166,83 @@ export class FormsgRouter {
...
+    const html = `<p>Isomer site ${githubRepoName} was <b>not</b> cloned successfully. Cloning failed with error: ${message} (Form submission id [${submissionId}])</p>`
+    await mailer.sendMail(requesterEmail, subject, html)
+  }
+
   getRouter() {
     const router = express.Router({ mergeParams: true })
 
+    router.post(
+      "/clone-site",
+      attachFormSGHandler("XO8YmgUbENTEftCEvnZRxM1Nb+wpERTfEj/w+4kPink="),
```
harishv7, line 242: what is this hardcoded string?

harishv7, line 242: also to confirm this middleware verifies the incoming request is actually from form sg right?

src/routes/formsgSiteCreation.ts
```diff
@@ -157,9 +166,83 @@ export class FormsgRouter {
...
+    { data: { submissionId: string } },
+    never,
+    { submission: DecryptedContent }
+  > = async (req, res) => {
+    // 1. Extract arguments
+    const { submissionId } = req.body.data
+    const { responses } = res.locals.submission
+    // NOTE: This is validated by formsg to be of domain `@open.gov.sg`;
+    // hence, not revalidating here
+    const requesterEmail = getField(responses, "Email") as string
```
harishv7, line 181: should we just allow whitelisted emails?

our admin email perhaps?

src/routes/formsgSiteCreation.ts
seaerchin: My bad, form submissions are encoded by default and this is used to decrypt; will shift to env var

src/routes/formsgSiteCreation.ts
seaerchin: To be clear, you want to stricten from allowing `@open.gov.sg` to only `admin@open.gov.sg`? I don't see too much of an issue here but what's the extra benefit ah cos this only prevents against internal attacks

harishv7: just suggesting [least privilege](https://en.wikipedia.org/wiki/Principle_of_least_privilege) principle. Not too much issue, since this is all internal, will let it up to you to decide

src/routes/formsgSiteCreation.ts
harishv7: @seaerchin pls get a new decoding key

seaerchin: done, updated vars in 1pw

src/routes/formsgSiteCreation.ts
seaerchin: ^not doing, the experience of having prod ops enter their own email then receive a rejection email saying it must be admin@isomer.gov.sg seems less than ideal given that this is all internal users anyway (validation of `@open.gov.sg` domain)

src/routes/formsgSiteCreation.ts
```diff
@@ -157,9 +167,83 @@ export class FormsgRouter {
...
+    Record<string, never>,
+    { data: { submissionId: string } },
+    never,
+    { submission: DecryptedContent }
+  > = async (req, res) => {
+    // 1. Extract arguments
+    const { submissionId } = req.body.data
+    const { responses } = res.locals.submission
+    // NOTE: This is validated by formsg to be of domain `@open.gov.sg`;
+    // hence, not revalidating here
```
kishore03109, line 181: Hmm lets revalidate here endsWith("@open.gov.sg")? Since this piece of infra (ie formsg) is not documented and subject to change, it is a quick win to revalidate here! 
PS. there was an vulnerability before where the form fields could have been mutated maliciously, rather stick to best practices here to revalidate!

src/routes/formsgSiteCreation.ts
```diff
@@ -157,9 +167,83 @@ export class FormsgRouter {
     await mailer.sendMail(email, subject, html)
   }
 
+  cloneSiteToEfs: RequestHandler<
```
kishore03109, line 170: Could we shift this to a `formsgSiteClone.ts`? We keep one file -> path

kishore03109 (commented):
This PR has quite a number of infra moving parts -
eg. 
1. formsg
2. efs
3. be 

Could we do a quick sanity check testing on stg before release? 
since this is internal use using post deploy testing is also fine, thoughts?

src/config/config.ts
```diff
@@ -283,6 +283,13 @@ const config = convict({
       format: "required-string",
       default: "",
     },
+    siteCloneFormKey: {
```
harishv7, line 286: This needs to be added to `.platform/hooks/predeploy/06_fetch_ssm_parameters.sh`

harishv7 (changes requested):

src/routes/formsgSiteCreation.ts
seaerchin: sure, sgtm

seaerchin:
> This PR has quite a number of infra moving parts - eg.
> 
> 1. formsg
> 2. efs
> 3. be
> 
> Could we do a quick sanity check testing on stg before release? since this is internal use using post deploy testing is also fine, thoughts?

tbh, i think the only difference between testing on staging vs not, is that the clone step takes place in a different environment (with all of the annoyances that entails); i think i'll just do post-deploy cos it's not open to public anyway.

src/config/config.ts
seaerchin: thanks for the catch!

src/routes/formsgSiteCreation.ts
seaerchin: surev

harishv7 (approved):

kishore03109 (approved):

PR #950, opened 2023-09-28
feat: add new admin endpoint to reset repository

dcshzj:
## Problem

Part of IS-446.

## Solution

**Breaking Changes**

**Features**:

- A new endpoint `/v2/sites/:siteName/admin/resetRepo` has been added to allow for e2e test repos to be easily reset to the specified commit.
    - If the repository is GGS-whitelisted, it will execute the Git commands on the local file system (`git reset --hard` and `git push -f origin`). It also checks beforehand that the commit exists, otherwise a 400 error is returned.
    - If the repository is not GGS-whitelisted, it will call an existing GitHub endpoint to update the repo state, which is an existing functionality used on the frontend to reset repos. GitHub automatically checks if the commit exists and will reject if it does not, but we return a 502 error as we cannot differentiate between this issue vs GitHub-related issues.

**Further notes**

This endpoint is located in a new `admin/` folder to lay the foundations for future admin-related endpoints and hopefully we can move towards having an admin dashboard.

## Tests

Unit tests.

## Deploy Notes

*None*

### DIFF
--- a/src/routes/v2/authenticatedSites/index.js
+++ b/src/routes/v2/authenticatedSites/index.js
@@ -16,6 +16,9 @@ const {
 const { MediaFilesRouter } = require("@routes/v2/authenticatedSites/mediaFiles")
 const { NavigationRouter } = require("@routes/v2/authenticatedSites/navigation")
 const {
+  RepoManagementRouter,
+} = require("@routes/v2/authenticatedSites/repoManagement")
+const {
   ResourceCategoriesRouter,
 } = require("@routes/v2/authenticatedSites/resourceCategories")
 const {
@@ -92,6 +95,7 @@ const getAuthenticatedSitesSubrouter = ({
   notificationOnEditHandler,
   sitesService,
   deploymentsService,
+  repoManagementService,
 }) => {
   const collectionYmlService = new CollectionYmlService({ gitHubService })
   const homepagePageService = new HomepagePageService({ gitHubService })
@@ -196,6 +200,10 @@ const getAuthenticatedSitesSubrouter = ({
   const navigationV2Router = new NavigationRouter({
     navigationYmlService: navYmlService,
   })
+  const repoManagementV2Router = new RepoManagementRouter({
+    repoManagementService,
+    authorizationMiddleware,
+  })
 
   const authenticatedSitesSubrouter = express.Router({ mergeParams: true })
 
@@ -236,6 +244,7 @@ const getAuthenticatedSitesSubrouter = ({
   authenticatedSitesSubrouter.use("/contactUs", contactUsV2Router.getRouter())
   authenticatedSitesSubrouter.use("/homepage", homepageV2Router.getRouter())
   authenticatedSitesSubrouter.use("/settings", settingsV2Router.getRouter())
+  authenticatedSitesSubrouter.use("/admin", repoManagementV2Router.getRouter())
   authenticatedSitesSubrouter.use(notificationOnEditHandler.createNotification)
 
   return authenticatedSitesSubrouter
--- a/src/routes/v2/authenticatedSites/repoManagement.ts
+++ b/src/routes/v2/authenticatedSites/repoManagement.ts
@@ -1,0 +1,71 @@
+import autoBind from "auto-bind"
+import express from "express"
+
+import { BadRequestError } from "@errors/BadRequestError"
+import { ForbiddenError } from "@errors/ForbiddenError"
+
+import { AuthorizationMiddleware } from "@middleware/authorization"
+import { attachWriteRouteHandlerWrapper } from "@middleware/routeHandler"
+
+import UserWithSiteSessionData from "@root/classes/UserWithSiteSessionData"
+import GitFileSystemError from "@root/errors/GitFileSystemError"
+import { attachSiteHandler } from "@root/middleware"
+import type { RequestHandler } from "@root/types"
+import RepoManagementService from "@services/admin/RepoManagementService"
+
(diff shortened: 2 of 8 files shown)


src/routes/v2/authenticatedSites/index.js
```diff
@@ -196,6 +200,10 @@ const getAuthenticatedSitesSubrouter = ({
   const navigationV2Router = new NavigationRouter({
     navigationYmlService: navYmlService,
   })
+  const repoManagementV2Router = new RepoManagementRouter({
+    repoManagementService,
+    authorizationMiddleware,
```
harishv7, line 205: how is the FE cypress run able to authorise itself to this route?

src/routes/v2/authenticatedSites/repoManagement.ts
```diff
@@ -0,0 +1,71 @@
...
+        if (error instanceof BadRequestError) {
+          return res.status(400).json({ message: error.message })
+        }
+        if (error instanceof ForbiddenError) {
+          return res.status(403).json({ message: error.message })
+        }
+        if (error instanceof GitFileSystemError) {
+          return res.status(500).json({ message: error.message })
+        }
+        return res.status(502).json({ message: error.message })
```
harishv7, line 58: this is thrown in case of github related issues?

src/routes/v2/authenticatedSites/repoManagement.ts
```diff
@@ -0,0 +1,71 @@
...
+      })
+  }
+
+  getRouter() {
+    const router = express.Router({ mergeParams: true })
+    router.use(attachSiteHandler)
+    router.use(this.authorizationMiddleware.verifySiteMember)
+
+    router.post("/resetRepo", attachWriteRouteHandlerWrapper(this.resetRepo))
```
harishv7, line 67: This is only meant for E2E repo right? If so, we should nest within /e2e

If this is meant as a generic endpoint to reset repos to a specific commit then this would be fine

src/services/admin/RepoManagementService.ts
```diff
@@ -0,0 +1,53 @@
...
+        if (error instanceof BadRequestError) {
+          return new BadRequestError(error.message)
+        }
+        if (error instanceof GitFileSystemError) {
+          return new GitFileSystemError(error.message)
+        }
+
+        return new GitHubApiError(`Failed to reset repo to commit ${commitSha}`)
+      }
+    ).map(() => undefined)
```
harishv7, line 49: do we need the `map -> undefined`?

dcshzj: Oops you're right! Removed in 9f5ffce.

src/routes/v2/authenticatedSites/index.js
dcshzj: This happens in a series of events:
1. The frontend sets a special cookie that is recognised by the backend via the authentication middleware. When this cookie is present, the authentication middleware will set a series of defaults to "become" a special e2e user. This can be either the GitHub user (which is currently our isomeradmin default user), or the e2e email login user.
2. With this user session set up, the authorization middleware will take over to determine if the user is authorised to access the site. There's no special treatment here, as the special e2e users have access to the e2e test repos.
3. Finally, at this router level, the access would have been granted, so the frontend will be able to enter this route and access the repo management routes.

src/routes/v2/authenticatedSites/repoManagement.ts
dcshzj: Yup, I opted for 502 as we are acting as a gateway, so I felt it was more appropriate to differentiate the error code.

src/routes/v2/authenticatedSites/repoManagement.ts
dcshzj: Yup it is intentional, I'm leaving the option open in the future in case we might ever need this endpoint for some other reason via the CMS frontend. For now however, the service also checks that it is within the e2e test repos, so the functionality is very limited.

src/routes/v2/authenticatedSites/index.js
harishv7: Got it, so is a agreed upon cookie string

src/routes/v2/authenticatedSites/repoManagement.ts
harishv7: Makes sense

src/routes/v2/authenticatedSites/repoManagement.ts
harishv7: sgtm

src/services/middlewareServices/AuthenticationMiddlewareService.ts
```diff
@@ -190,7 +191,8 @@ export default class AuthenticationMiddlewareService {
       (userType === E2E_USERS.Email.Admin ||
         userType === E2E_USERS.Email.Collaborator)
     const isGithubE2eAccess =
-      repo === E2E_TEST_REPO && userType === "Github user"
+      (repo === E2E_TEST_REPO || repo === E2E_NOTGGS_TEST_REPO) &&
```
harishv7, line 194: nit: E2E_NOT_GGS_TEST_REPO

harishv7 (approved):
approved with minor nit

src/services/middlewareServices/AuthenticationMiddlewareService.ts
dcshzj: Done in d7c9eef.

PR #957, opened 2023-10-05
fix(contactus): allow location section to be empty but not contact information

dcshzj:
## Problem

Closes IS-471.

## Solution

**Breaking Changes**

**Bug Fixes**:

- The location section of the contact us frontmatter is now optional. Code changes are mostly just indentation, except for the last line where the `required()` part is removed.

## Tests

Testable only with frontend.

## Deploy Notes

*None*

### DIFF
--- a/src/validators/RequestSchema.js
+++ b/src/validators/RequestSchema.js
@@ -19,22 +19,20 @@ const UpdateContactUsSchema = Joi.object({
       permalink: Joi.string().required(),
       feedback: Joi.string().allow(""),
       agency_name: Joi.string().required(),
-      locations: Joi.array()
-        .items(
-          Joi.object({
-            address: Joi.array().items(Joi.string().allow("")),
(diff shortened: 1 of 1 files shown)


seaerchin (approved):

seaerchin:
just clarifying here that this change allows both `undefined` (ie the key is missing) and an empty array. 

we might wanna check if downstream callers handle the new case (`undefined`) gracefully or if we need to also provide a default argument to avoid blowing up

harishv7 (approved):

dcshzj:
@seaerchin this is handled gracefully on the frontend [under the data sanitisers](https://github.com/isomerpages/isomercms-frontend/blob/develop/src/utils/contact-us/dataSanitisers.js), so the frontend will always expect to have this key available, even as an empty array. If the backend does not provide this key, it will default to being an empty array.

PR #958, opened 2023-10-09
refactor(ff): make site launch flag a bool flag

seaerchin:
## Problem
Right now, our GGS flag adopts a two tiered approach - first, we get the list of repos from growthbook then we validate based on that. 

This is a _whitelist_ approach, where we add sites as they shift to ggs. This is effortful and going forward, we want to make our flag a _blacklist_ approach, where we allow sites by default until we need to blacklist them. 

## Solution
Add new flag on growthbook that returns a boolean value checking if the user type is email and whether site is blacklisted. (it's true if user is email && site isn't blacklisted)

## Deploy notes
**these steps have to be done pre deploy**
- [ ] for new **email login** sites that are not on ggs, migrate them to ggs by cloning them onto efs (can use form atm) 

**these steps have to be done post deploy**
- [ ] remove the flag for ggs whitelisted repo
- [ ] enable the flag for `is_ggs_enabled` on **production**

### DIFF
--- a/src/constants/featureFlags.ts
+++ b/src/constants/featureFlags.ts
@@ -1,5 +1,4 @@
 export const FEATURE_FLAGS = {
-  GGS_WHITELISTED_REPOS: "ggs_whitelisted_repos",
   IS_BUILD_TIMES_REDUCTION_ENABLED: "is_build_times_reduction_enabled",
   IS_GGS_ENABLED: "is_ggs_enabled",
 } as const
--- a/src/constants/featureFlags.ts
+++ b/src/constants/featureFlags.ts
@@ -1,4 +1,5 @@
 export const FEATURE_FLAGS = {
   GGS_WHITELISTED_REPOS: "ggs_whitelisted_repos",
   IS_BUILD_TIMES_REDUCTION_ENABLED: "is_build_times_reduction_enabled",
+  IS_GGS_WHITELISTED: "is_ggs_whitelisted",
 } as const
--- a/src/constants/featureFlags.ts
+++ b/src/constants/featureFlags.ts
@@ -1,5 +1,5 @@
 export const FEATURE_FLAGS = {
   GGS_WHITELISTED_REPOS: "ggs_whitelisted_repos",
   IS_BUILD_TIMES_REDUCTION_ENABLED: "is_build_times_reduction_enabled",
-  IS_GGS_WHITELISTED: "is_ggs_whitelisted",
+  IS_GGS_ENABLED: "is_ggs_enabled",
 } as const
--- a/src/middleware/routeHandler.js
+++ b/src/middleware/routeHandler.js
@@ -104,7 +104,7 @@ const attachRollbackRouteHandlerWrapper = (routeHandler) => async (
   const { growthbook } = req
 
   const shouldUseGitFileSystem = !!growthbook?.getFeatureValue(
-    FEATURE_FLAGS.IS_GGS_WHITELISTED,
+    FEATURE_FLAGS.IS_GGS_ENABLED,
     false
   )
 
--- a/src/middleware/routeHandler.js
+++ b/src/middleware/routeHandler.js
@@ -25,14 +25,13 @@ const gitFileSystemService = new GitFileSystemService(
   new SimpleGit({ maxConcurrentProcesses: MAX_CONCURRENT_GIT_PROCESSES })
 )
 
-const isRepoWhitelisted = (siteName, ggsWhitelistedRepos) => {
+const isRepoWhitelisted = (siteName, growthbook) => {
   // TODO: adding log to simplify debugging, to be removed after stabilising
-  logger.info(
-    `Checking if ${siteName} is GGS whitelisted: ${ggsWhitelistedRepos.includes(
-      siteName
-    )}`
-  )
-  ggsWhitelistedRepos.includes(siteName)
+  logger.info(`Evaluating if ${siteName} is GGS whitelisted.`)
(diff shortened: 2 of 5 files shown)


src/constants/featureFlags.ts
```diff
@@ -1,3 +1,4 @@
 export const FEATURE_FLAGS = {
   GGS_WHITELISTED_REPOS: "ggs_whitelisted_repos",
```
kishore03109, line 2: Should we remove this since this is no longer in use?

src/middleware/routeHandler.js
```diff
@@ -74,23 +73,15 @@ const attachWriteRouteHandlerWrapper = (routeHandler) => async (
...
-      FEATURE_FLAGS.GGS_WHITELISTED_REPOS,
-      {
-        repos: [],
-      }
-    )
-  }
-
   let isGitAvailable = true
+
   // only check git file lock if the repo is whitelisted
```
kishore03109, line 78: blacklisted**

src/services/db/RepoService.ts
```diff
@@ -38,27 +38,19 @@ export default class RepoService extends GitHubService {
...
-    logger.info(
-      `Evaluating if ${repoName} is GGS whitelisted: ${ggsWhitelistedRepos.includes(
-        repoName
-      )}`
+    logger.info(`Evaluating if ${repoName} is GGS whitelisted.`)
+    // NOTE: Growthbook has to be optional
+    // as we cannot guarantee its presence on `sessionData`.
+    // It is present iff there is a `siteHandler` attached.
+    return !!growthbook?.getFeatureValue(
+      FEATURE_FLAGS.IS_GGS_WHITELISTED,
```
kishore03109, line 51: are we using the wrong feature fflag here?

kishore03109 (commented):
How does this play with migrated sites? ie when is the git clone step for these sites going to occur ah?

src/constants/featureFlags.ts
seaerchin: we cannot remove until this feature is released cos prod still uses this flag

src/middleware/routeHandler.js
seaerchin: good catch, gonna update

src/services/db/RepoService.ts
seaerchin: actually no, i just named it wrongly cos i think i got confused between white/blacklist

seaerchin:
> How does this play with migrated sites? ie when is the git clone step for these sites going to occur ah?

there's a form for them to fill up that triggers a clone

kishore03109:
@seaerchin wait so to be clear, post email migration of sites there needs to a manual sending submission of this form by eng/ops? 
cc. @alexanderleegs think we might need to take this into account ya

src/constants/featureFlags.ts
harishv7: let's add a todo above

src/middleware/routeHandler.js
```diff
@@ -25,14 +25,13 @@ const gitFileSystemService = new GitFileSystemService(
   new SimpleGit({ maxConcurrentProcesses: MAX_CONCURRENT_GIT_PROCESSES })
 )
 
-const isRepoWhitelisted = (siteName, ggsWhitelistedRepos) => {
+const isRepoWhitelisted = (siteName, growthbook) => {
```
harishv7, line 28: do we still need a function for this? seems like the sitename is only passed in to log it and we are just calling 1 growthbook function

src/constants/featureFlags.ts
```diff
@@ -1,3 +1,4 @@
...
+  IS_GGS_WHITELISTED: "is_ggs_whitelisted",
```
harishv7, line 3: just to be clear are we doing a whitelist or blacklist?

src/constants/featureFlags.ts
```diff
@@ -1,3 +1,4 @@
 export const FEATURE_FLAGS = {
```
harishv7, line 1: can we add deploy notes to the PR mentioning to turn on this flag for prod during release

src/middleware/routeHandler.js
seaerchin: removed

src/constants/featureFlags.ts
seaerchin: changed naming to `IS_GGS_ENABLED` for clarity

src/constants/featureFlags.ts
seaerchin: done

src/services/db/RepoService.ts
```diff
@@ -38,27 +38,19 @@ export default class RepoService extends GitHubService {
...
+  isRepoWhitelisted(
+    repoName: string,
+    growthbook?: GrowthBook<FeatureFlags>
+  ): boolean {
     // TODO: Adding for initial debugging if required. Remove once stabilised
-    logger.info(
-      `Evaluating if ${repoName} is GGS whitelisted: ${ggsWhitelistedRepos.includes(
-        repoName
-      )}`
+    logger.info(`Evaluating if ${repoName} is GGS whitelisted.`)
```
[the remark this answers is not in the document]
seaerchin: function is removed

src/constants/featureFlags.ts
seaerchin: removed in code, added deploy note to remove in gb

harishv7 (approved):

PR #964, opened 2023-10-12
ref(media): paginate media endpoint

seaerchin:
## Problem
See [here](https://github.com/isomerpages/isomercms-frontend/pull/1578) for description

## Solution
- add `page` and `limit` to the media endpoint. we don't paginate directories (for now) due to reasons stated in the other PR. 
- also added 2 more endpoints for fetching files/directories separately. the reasons are given in the other PR. As github **does not** allow us ([no link header](https://docs.github.com/en/rest/guides/using-pagination-in-the-rest-api?apiVersion=2022-11-28#using-link-headers)) to paginate the repository content [endpoint](https://docs.github.com/en/rest/repos/contents?apiVersion=2022-11-28#get-repository-content), this pagination exists solely on our end, **not on github**. 
    - this means that we only save on 1 network call (from be -> fe) and we don't save on the network call from github -> be. this endpoint should have etags enabled so thankfully, only the first call out should be expensive.

## Note
tests coming in a follow up PR

### DIFF
--- a/src/services/db/GitHubService.js
+++ b/src/services/db/GitHubService.js
@@ -12,7 +12,7 @@ const {
 const { NotFoundError } = require("@errors/NotFoundError")
 const { UnprocessableError } = require("@errors/UnprocessableError")
 
-const logger = require("@root/logger/logger")
+const logger = require("@root/logger/logger").default
 
 const ReviewApi = require("./review")
 
--- a/src/services/db/GitHubService.js
+++ b/src/services/db/GitHubService.js
@@ -181,36 +181,6 @@ class GitHubService {
     return { content, sha }
   }
 
-  async readMedia(sessionData, { fileSha }) {
-    /**
-     * Files that are bigger than 1 MB needs to be retrieved
-     * via Github Blob API. The content can only be retrieved through
-     * the `sha` of the file.
-     */
-    const { accessToken } = sessionData
-    const { siteName } = sessionData
-    const params = {
-      ref: BRANCH_REF,
-    }
-
-    const blobEndpoint = this.getBlobPath({ siteName, fileSha })
-
-    const resp = await this.axiosInstance.get(blobEndpoint, {
-      validateStatus,
-      params,
-      headers: {
-        Authorization: `token ${accessToken}`,
-      },
-    })
-
-    if (resp.status === 404)
-      throw new NotFoundError("Media file does not exist")
-
-    const { content, sha } = resp.data
-
-    return { content, sha }
-  }
-
   async readDirectory(sessionData, { directoryName }) {
     const { accessToken } = sessionData
     const { siteName } = sessionData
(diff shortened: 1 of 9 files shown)


src/services/directoryServices/MediaDirectoryService.js
```diff
@@ -28,13 +28,16 @@ class MediaDirectoryService {
...
     if (!isMediaPathValid({ path: directoryName }))
       throw new BadRequestError("Invalid media folder name")
 
-    return await this.gitHubService.readMediaDirectory(
+    return this.gitHubService.readMediaDirectory(
       sessionData,
-      directoryName
+      directoryName,
+      // NOTE: Done here as frontend starts indexing from 1
+      page - 1,
```
dcshzj, line 39: Not a blocker but I feel that doing this might trip other engineers a little bit, because a "page 1" on the frontend is different from the backend. I have a couple of suggestions:
1. Switch to using offset instead (which seems to be recommended over using page, but I'm ok either way)
2. Bring this "minus 1" logic closer to the front, meaning the frontend's Paginator component should do this immediately on user change.
3. Bring this "minus 1" logic closer to the back, meaning this happens inside `getPaginatedDirectoryContents`.

I feel that having this in-between makes the definition of a page inconsistent and might add a little mental overhead.

src/services/db/GitHubService.js
```diff
@@ -12,7 +12,7 @@ const {
 const { NotFoundError } = require("@errors/NotFoundError")
 const { UnprocessableError } = require("@errors/UnprocessableError")
 
-const logger = require("@root/logger/logger")
+const logger = require("@root/logger/logger").default
```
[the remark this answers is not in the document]
seaerchin: yes - otherwise it won't work. you can check this for yourself @__@

src/services/directoryServices/MediaDirectoryService.js
seaerchin: could i get clarity on what you meant by using offset? 
 
we can't do #2 because the pagination components informs us of the selected page - and we maintain this using a `curPage` state variable, which feeds back into the component itself.

my main considerations here were that the router should extract the variables from the request and defer business logic to the service (the media service here), which is why i left it here. 

idt i'll change it to being in `getPaginatedDirectoryContents` but wdyt about having it on the router level then?

seaerchin: ok i derped on this - you're right wrt the subtraction! i'll put this in the FE, probably in a custom `usePaginate` hook for clarity (so it'll do the auto conversion).

seaerchin: wrt offset - agreed on more ideal. the reference i was looking at (can't remember offhand) used page + limit, which i duplicated here ._. but we can swap to a offset + limit approach (as outlined [here](https://specs.openstack.org/openstack/api-wg/guidelines/pagination_filter_sort.html)). 

i might not do this as idt it impacts clarity + involves both FE and BE changes which is annoying. lmk if you have concerns though (ie, 1 is more powerful than the other for eg) and we can change from there?

dcshzj (approved):
LGTM!

PR #965, opened 2023-10-12
feat(quickie): Create operation for commits

kishore03109:
## Problem

This this partially solves IS-644. This is not a complete feature, but the purposes of this PR is to obtain early feedback on the approach just to achieve the `Create` operation for both GGs and non GGs flow. This should not have any impact on any non-whitelisted amp reduction repos for both GGs AND non GGs flow. This would have to be tested manually for now.  

Tests are failing in this PR, will be fixed in upcoming PRs, and will hold off merging this into main line until upstream tests-fixing PR is approved. 

## Solution

Blank diagram - Page 1 (5)
(copy pasting from [RFC](https://www.notion.so/opengov/Reduce-Amplify-Build-Times-V-851b33e935d54aaf84eaf0742d02f96f?pvs=4))
This solution borrows a similar solution to what the repo service solved. We had to introduce a new drop in replacement class we didn’t wish to change most of the APIs. In addition about 

Currently, to cater for GGs and preserve APIs, there exists a `RepoService` that acts as a drop in replacement. However, `GitFileService` and `GithubService` are not polymorphic, since they fundamentally work slightly differently from one another. Because they work differently with differing APIs, having two `CommitService` per (GitFileService vs GithubService) is chosen. 

I did toy with the idea of alternative B, but this did not work as well as I thought it would due to the differing APIs, which would have led to quite a bit of `if-elses` which did not seem ideal. Opting to split up the classes instead to KISS. 

There is also a need to store the repos at two different places, see [here](https://www.notion.so/opengov/Reduce-Amplify-Build-Times-V-851b33e935d54aaf84eaf0742d02f96f?pvs=4#66eb7a0d98e247fc87df211054cf7e4e) for more info.

**Breaking Changes**

  - This is not intended to work for sites that are private, this is an intentional decision we made as a team to allow the public repos to gain some improvement in their build times. Concretely this would mean that private repos cannot be whitelisted in growthbook
- [] No - this PR is backwards compatible with ALL of the following feature flags in this [doc](https://www.notion.so/opengov/Existing-feature-flags-518ad2cdc325420893a105e88c432be5)

**Features**:

All `create` actions should be committed to staging branch.
All `create` actions on non-assets should be committed to `staging-lite` branch.

## Tests

Note: To the user, there should not be any difference between the GGs flow and the non-GGs flow. The key thing to note is that this particular Create operation on a non-asset change is lead to an update of both the `staging` and the `staging-lite` branch. 

### GGs flow, with quickie feature flagged
https://github.com/isomerpages/isomercms-backend/assets/42832651/86e15eaa-79a6-4f73-8d7e-843bbf8998a4

### Non-GGs flow, with quickie feature flagged

https://github.com/isomerpages/isomercms-backend/assets/42832651/d1955b6e-3a1e-401d-84bb-5e19405f1a23

[Note the two flows below are just to show that sites that are non feature flagged should be BAU, and not impact to their flow/staging branches]

### GGs flow, which quickie not feature flagged

https://github.com/isomerpages/isomercms-backend/assets/42832651/157e8c45-39a2-4ccf-8344-f536d90681be

### Non-GGs flow, with quike not feature flagged

https://github.com/isomerpages/isomercms-backend/assets/42832651/fa53cd8a-0e78-46a7-b664-1a9bd3e71924

## Manual tests in staging
1. Log into github: should be able to add a page to kishore-test-dev-gh and notice changes in staging and staging lite branch
2. log into email, should be able to add a page to `kishore-test` and notice changes in staging and staging lite.
3. All other repos, adding a page should have expected behaviour (adding an image should not throw errors + have a corresponding commit in github) 

- `env var` : env var details
#### modification  of env vars 
    - [ ] EFS_VOL_PATH : From `/efs/repos/` to `/efs/`


dcshzj:
I think the middle approach ("AFTER") is the most sane approach and the one that makes the most sense. However, I was thinking if the CommitService* should be merged with the service behind it, so we maintain the current structure but enhance both GitFileSystemService and GitHubService to have the ability to commit to 2 branches for assets.

Main reason is because this process of "committing to 2 different branches" is a logic that is determined by whether the commit is for an asset or for a page/settings/navbar. We can probably enhance the `commit()` function to handle specifying the repository folder to commit on, but the logic of committing to 2 different branches is isolated to asset-related functions (actually all the CRUD operations).

So in essence this is roughly the flow I imagine:
1. Determine if create/rename/delete operation is for an asset or for a page/settings/navbar.
    1. If it is a page/settings/navbar change, continue with current process.
    2. If it is an asset change, determine if the split branch strategy is adopted for the current repository.
        1. If split branch strategy is not adopted, continue with the current process.
        2. If split branch strategy is adopted, call `commit()` twice plus the other accompanying logic to make that happen.

For reads, we already have a split between reading from page/settings/navbar vs reading media, so we will do something similar like above to display the media file from the correct location.

Just my 2 cents with the opinion of not splitting up the operations further.

src/services/db/CommitServiceGitFile.ts
```diff
@@ -0,0 +1,91 @@
...
+        sessionData.siteName,
+        sessionData.isomerUserId,
+        content,
+        directoryName,
+        fileName,
+        isMedia ? "base64" : "utf-8",
+        STAGING_BRANCH
+      ),
+    ]
+    const shouldStagingLiteUpdate =
```
harishv7, line 54: nit: shouldUpdateStagingLite

src/services/db/CommitServiceGitFile.ts
```diff
@@ -0,0 +1,91 @@
...
+          sessionData.isomerUserId,
+          content,
+          directoryName,
+          fileName,
+          isMedia ? "base64" : "utf-8",
+          this.STAGING_LITE_BRANCH
+        )
+      )
+    }
+    const [resultToStaging, resultToStagingLite] = await Promise.all(
```
harishv7, line 72: nit: stagingCreateResult stagingLiteCreateResult

src/services/db/CommitServiceGitFile.ts
```diff
@@ -0,0 +1,91 @@
...
+      fileName,
+      directoryName,
+      isMedia = false,
+    }: {
+      content: string
+      fileName: string
+      directoryName: string
+      isMedia?: boolean
+    }
+  ) {
```
harishv7, line 42: might want to add return type

src/config/config.ts
```diff
@@ -187,12 +187,20 @@ const config = convict({
       },
     },
     efs: {
-      volPath: {
```
harishv7, line 190: To bounce off ideas, without the introduction of a new service, can we modify the existing flow/functions to support branch-level modifications?

So the conditional of whether it is a GGS flow or GitHub flow is already done at RepoService

Once the request reaches GitFileSystemService or GitHubService can these 2 services' functions be enhanced such that they are able to commit to staging-lite as well?

It could be as simple as introducing a new default parameter like `hasAssets` which is a flag that indicates the function should commit to both staging + staging-lite

Were the above considered for the approaches/design?

src/services/db/RepoService.ts
```diff
@@ -368,7 +384,7 @@ export default class RepoService extends GitHubService {
     }
   ): Promise<GitCommitResult> {
     if (
-      this.isRepoWhitelisted(
+      this.isRepoWhitelistedGgs(
```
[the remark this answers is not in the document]
kishore03109: else we have two flags -> one for whitelisted GGs, another for reduced build times. Renamed solely for clarity

src/config/config.ts
kishore03109: Yes, this could have been done without introducing any new services. However, as it is I felt that that repo service is already quite bulky, and handles quite a fair bit of github service level logic. I intended to 
1. lighten the responsibility of RepoSerice to only care about which flow (GGs or GH) to use
2. bring in all the logic of the commits over to the commit service
3. now that github service is in ts, bring more of the github level logic into github service so that commit service dont have to bloat -> bringing the API closer to one another but its done as incremental effort. 

I think the ideal to have here is 
reposervice descids what flow is it and calls the approporate API 
commit service calls what to commit to where 

github service + Git file service should move closer to having the same API, and this seem to move the needle closer to this ideal state.

src/config/config.ts
```diff
@@ -187,12 +187,20 @@ const config = convict({
...
-        doc: "Path to the EFS volume for storing the Git repositories",
-        env: "EFS_VOL_PATH",
+      volPathStaging: {
```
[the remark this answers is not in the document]
kishore03109: good idea 2c279f2f8d44f778a62963d8ec5cf6378aa02131!

src/services/db/CommitServiceGitFile.ts
kishore03109: 0f887f523e9e7f122b0b54f6d68989b5d311a4bc

src/services/db/CommitServiceGitFile.ts
kishore03109: 0f887f523e9e7f122b0b54f6d68989b5d311a4bc

src/services/db/CommitServiceGithub.ts
```diff
@@ -0,0 +1,80 @@
+import { GrowthBook } from "@growthbook/growthbook"
+import { AxiosCacheInstance } from "axios-cache-interceptor"
+
+import UserWithSiteSessionData from "@root/classes/UserWithSiteSessionData"
+import { FEATURE_FLAGS } from "@root/constants"
+import { FeatureFlags } from "@root/types/featureFlags"
+import { isReduceBuildTimesWhitelistedRepo } from "@root/utils/growthbook-utils"
```
seaerchin, line 7: casual reminder to remove unused vars

src/services/db/CommitServiceGithub.ts
```diff
@@ -0,0 +1,80 @@
...
+      directoryName,
+      isMedia = false,
+    }: {
+      content: string
+      fileName: string
+      directoryName: string
+      isMedia?: boolean
+    }
+  ): Promise<{
+    sha: any
```
seaerchin, line 47: could we type this as string?

seaerchin (commented):
given that the changes here have to be supported by both github/efs, there are alot of duplicated code. 

i think this is ok for now, given that we are planning to remove github dependencies in the future, as long as functionality is intact. 

i do have 1 question, though - given that the build times are reduced through comitting to `staging-lite` for files/images, why do we need to update `read` endpoints? wouldn't just reading from `staging` always be sufficient?

src/services/db/CommitServiceGithub.ts
kishore03109: 59732ce1515f2bbf561a3efdf4b9e6107e98e62f

src/services/db/RepoService.ts
```diff
@@ -516,7 +536,9 @@ export default class RepoService extends GitHubService {
         filePath,
         sha,
         sessionData.isomerUserId,
-        false
+        false,
+        //! TODO: this needs to be replaced with a call to commitService instead
+        STAGING_BRANCH
```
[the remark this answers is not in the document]
kishore03109: yes sir

src/constants/constants.ts
```diff
@@ -69,3 +71,8 @@ export const ISOMER_ADMIN_EMAIL = "admin@isomer.gov.sg"
 export const ISOMER_SUPPORT_EMAIL = "support@isomer.gov.sg"
 
 export const MAX_CONCURRENT_GIT_PROCESSES = 150
+
+export const EFS_VOL_PATH_STAGING = path.join(config.get("aws.efs.volPath"))
+export const EFS_VOL_PATH_STAGING_LITE = path.join(
+  config.get("aws.efs.volPath")
+)
```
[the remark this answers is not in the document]
kishore03109: 809069459077e4b58002e96277ce725c340defd4
opps, thanks for the catch

src/config/config.ts
```diff
@@ -191,7 +191,7 @@ const config = convict({
         doc: "Path to the EFS volume for storing the Git repositories",
         env: "EFS_VOL_PATH",
         format: "required-string",
-        default: "/efs/repos",
+        default: "/efs",
```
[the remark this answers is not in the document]
kishore03109: sure, we need two repo locations. see [here](https://www.notion.so/opengov/Reduce-Amplify-Build-Times-V-851b33e935d54aaf84eaf0742d02f96f?pvs=4#66eb7a0d98e247fc87df211054cf7e4e) to know more. 
Given that is the case, our current repo path is as such 
`efs/repos`

Additional repo path to add post addition of this PR: `efs/repos-lite`

Rather than adding another env var to store staging lite branch, we keep the root the same, ie `/efs` and store the paths for the staging + staging lite repos as constants in code.

src/services/db/CommitServiceGithub.ts
```diff
@@ -0,0 +1,77 @@
...
+import { isFileAsset } from "./CommitServiceGitFile"
+import GitFileSystemService from "./GitFileSystemService"
+import GitHubService, { STAGING_LITE_BRANCH } from "./GitHubService"
+
+/**
+ * Responsibilities of this class
+ * 1. Creates all commits to staging
+ * 2. Creates non-asset related commits to staging-lite
+ */
+export default class CommitServiceGitHub extends GitHubService {
```
harishv7, line 15: btw for naming prefer to have service, handler, router etc as the suffix
 
ie. GitHubCommitService

src/services/db/RepoService.ts
```diff
@@ -298,16 +310,18 @@ export default class RepoService extends GitHubService {
     sessionData: UserWithSiteSessionData,
     { directoryName }: { directoryName: string }
   ): Promise<GitDirectoryItem[]> {
+    const defaultBranch = STAGING_BRANCH
```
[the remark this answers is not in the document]
kishore03109: added this so i dont have to add it in as comment haha

src/services/db/CommitServiceGithub.ts
```diff
@@ -0,0 +1,77 @@
+import { AxiosCacheInstance } from "axios-cache-interceptor"
+
+import UserWithSiteSessionData from "@root/classes/UserWithSiteSessionData"
+import { isReduceBuildTimesWhitelistedRepo } from "@root/utils/growthbook-utils"
+
+import { isFileAsset } from "./CommitServiceGitFile"
```
harishv7, line 6: shld be in an utils instead of a service cross-dependency

src/config/config.ts
harishv7: Understand, but for now the CommitService's only job seems to be checking if the growthbook flag for the site is on and calling the underlying service twice, hence it feels like a passthrough service. In both CommitServices the same checks are being done to check (isFileAsset + growthbook)

And in future if we remove the feature flagging making this a perm solution, then will there be a need for this service?

kishore03109: Hmm no, it also decides whether or not a commit should be made to staging lite, not just the feature flag stuff. We dont intend to deprecate `gitfileCommitService` anytime soon. But, `githubCommitService` can be subsumed into `RepoService` when we completely move away from github + use other services for say our comments + prs

src/services/db/CommitServiceGithub.ts
kishore03109: i realised this after I pushed out the other prs :sadge: 
not sure how well git handles rebasing for renames, is it ok if I did this in the upsteam pr?

src/services/db/CommitServiceGithub.ts
kishore03109: fair 803e3699b8572a3b5dd7d83b05468d2cda20471b

src/config/config.ts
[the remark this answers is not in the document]
harishv7: This requires a change on SSM - pls include in PR description for on-call to know during release

src/constants/featureFlags.ts
```diff
@@ -1,3 +1,4 @@
 export const FEATURE_FLAGS = {
   GGS_WHITELISTED_REPOS: "ggs_whitelisted_repos",
+  REDUCE_BUILD_TIMES_WHITELISTED_REPOS: "reduce_build_times_whitelisted_repos",
```
harishv7, line 3: for feature flags going forward we prefer to use something like `isBuildTimeFlowWhitelisted` or `isQuickieWhitelisted`

src/services/db/CommitServiceGithub.ts
harishv7: sure! upstream pr will merge into this PR right?

src/services/db/CommitServiceGithub.ts
```diff
@@ -0,0 +1,77 @@
...
+          fileName,
+          isMedia,
+          branchName: STAGING_LITE_BRANCH,
+        })
+      )
+    }
+
+    // We still await and throw error if the commit to staging-lite fails,
+    // but the caller only gets the `resultToStaging` commit returned
+    const [resultToStaging, _] = await Promise.all(createPromises)
```
harishv7, line 73: just sanity check - promise.all will reject upon first promise failure. any possibility staging and staging-lite will diverge? I think probably not as we will rollback, but just re-checking

src/services/db/GitFileSystemService.ts
```diff
@@ -286,33 +338,107 @@ export default class GitFileSystemService {
...
+                  `An existing folder "${repoName}" exists in staging lite but is not the correct Git repo`
+                )
+              )
+            }
+            return okAsync(`${EFS_VOL_PATH_STAGING}/${repoName}`)
+          })
+      })
+  }
+
+  cloneStagingLite(repoName: string): ResultAsync<string, GitFileSystemError> {
```
harishv7, line 405: Do we really need 2 separate methods to clone for lite vs staging? seems like 80% logic will be the same right?

src/services/db/RepoService.ts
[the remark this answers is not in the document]
harishv7: okay!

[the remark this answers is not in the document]
harishv7: prefer name as isRepoGgsWhitelisted

src/services/db/CommitServiceGithub.ts
kishore03109: nope, am intending to merge this into mainline first then upstream pr will merge into main line to keep the lines per merge small

kishore03109: but ye this name change will go in by today

src/services/db/CommitServiceGithub.ts
kishore03109: No there is a case that might occur here. currently, the promise fails if ANY one of the promises fail, and eng would have to intervene to keep staging + staging lite the same (operation will still result in an error for the user). we could have a rollback handler here to keep staging + staging-lite in sync, but I do feel that it does not have to go in by today. 
I can add in a jira ticket and priortise accordingly if you feel that this is an important, thoughts?

src/config/config.ts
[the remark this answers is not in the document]
kishore03109: ps forgot to update, done!

src/services/db/RepoService.ts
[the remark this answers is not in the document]
kishore03109: sure befb48c4260a05c767e0404b320aedb7c2f87f18

src/services/db/CommitServiceGitFile.ts
```diff
@@ -0,0 +1,91 @@
+import { GrowthBook } from "@growthbook/growthbook"
+
+import UserWithSiteSessionData from "@root/classes/UserWithSiteSessionData"
+import { FEATURE_FLAGS } from "@root/constants"
+import { FeatureFlags } from "@root/types/featureFlags"
+import { isReduceBuildTimesWhitelistedRepo } from "@root/utils/growthbook-utils"
+
+import GitFileSystemService from "./GitFileSystemService"
+import GitHubService, { STAGING_BRANCH } from "./GitHubService"
```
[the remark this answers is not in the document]
kishore03109: updated!

src/constants/featureFlags.ts
kishore03109: 4be5c4826360d1592e35d47855f5c93ca0901053

src/services/db/GitFileSystemService.ts
kishore03109: derped on this, yup, better to combine b307b4b493b06e45da22d4137b8994865b9e99ed

src/services/db/GitFileSystemService.ts
```diff
@@ -215,13 +250,25 @@ export default class GitFileSystemService {
...
+   */
   getFilePathStats(
     repoName: string,
-    filePath: string
+    filePath: string,
+    isStaging: boolean
   ): ResultAsync<fs.Stats, NotFoundError | GitFileSystemError> {
+    const efsVolPath = isStaging
+      ? EFS_VOL_PATH_STAGING
+      : EFS_VOL_PATH_STAGING_LITE
```
dcshzj, line 269: Should we use `this.getEfsVolPath()` like the other functions do as well?

src/services/db/GitFileSystemService.ts
```diff
@@ -286,9 +338,26 @@ export default class GitFileSystemService {
...
+  }
+
+  cloneBranch(
+    repoName: string,
+    isStaging: boolean
+  ): ResultAsync<string, GitFileSystemError> {
     const originUrl = `git@github.com:${ISOMER_GITHUB_ORG_NAME}/${repoName}.git`
+    const efsVolPath = isStaging
+      ? EFS_VOL_PATH_STAGING
+      : EFS_VOL_PATH_STAGING_LITE
```
dcshzj, line 357: Same here for `this.getEfsVolPath()`.

src/services/db/GitFileSystemService.ts
```diff
@@ -450,19 +541,27 @@ export default class GitFileSystemService {
             }
           )
         )
-        .map(() => `${EFS_VOL_PATH}/${repoName}`)
+        .map(() => `${efsVolPath}/${repoName}`)
     })
   }
 
+  private getEfsVolPath(branchName: string) {
```
dcshzj, line 548: nit: this should be at the top of the file

src/services/db/CommitServiceGitFile.ts
```diff
@@ -0,0 +1,84 @@
...
+
+import GitFileSystemService from "./GitFileSystemService"
+import { STAGING_BRANCH } from "./GitHubService"
+
+/**
+ * Responsibilities of this class
+ * 1. Creates all commits to staging
+ * 2. Creates non-asset related commits to staging-lite
+ */
+export default class CommitServiceGitFile {
```
dcshzj, line 13: Should this `extends GitFileSystemService`?

src/services/db/CommitServiceGitFile.ts
```diff
@@ -0,0 +1,84 @@
...
+      fileName,
+      directoryName,
+      isMedia = false,
+    }: {
+      content: string
+      fileName: string
+      directoryName: string
+      isMedia?: boolean
+    }
+  ): Promise<{ sha: string }> {
```
dcshzj, line 35: Are we able to use `neverthrow` here?

src/services/db/CommitServiceGithub.ts
```diff
@@ -0,0 +1,77 @@
...
+ * Responsibilities of this class
+ * 1. Creates all commits to staging
+ * 2. Creates non-asset related commits to staging-lite
+ */
+export default class CommitServiceGitHub extends GitHubService {
+  private readonly STAGING_BRANCH = "staging"
+
+  private readonly STAGING_LITE_BRANCH = "staging-lite"
+
+  private readonly gitFileSystemService: GitFileSystemService
```
dcshzj, line 20: Should this be removed? This service is for GitHub not the EFS system.

src/services/db/CommitServiceGithub.ts
harishv7: But create failed operation does rollback right? src/services/db/GitFileSystemService.ts

but in any case, to answer the qn, addressing this in a subsequent PR is fine

seaerchin (approved):
i'm not too concerned with how the code looks here, more concerned with functionality + maintainability, 

beyond the feature flag, i think this is ok. as we are reviewing, we have to keep in mind that any improvement here is better than no improvement - as long as the code itself is suffficiently maintainable to be refactored in dec, this should roll out ASAP.

seaerchin (approved):
i'm not too concerned with how the code looks here, more concerned with functionality + maintainability, 

beyond the feature flag, i think this is ok. as we are reviewing, we have to keep in mind that any improvement here is better than no improvement - as long as the code itself is suffficiently maintainable to be refactored in dec, this should roll out ASAP.

src/services/db/CommitServiceGitFile.ts
kishore03109: yes but not now pls post release can refactor

src/services/db/CommitServiceGithub.ts
kishore03109: does not for github failures, agreed with chin here. addessing it post release

src/services/db/GitFileSystemService.ts
```diff
@@ -41,19 +45,29 @@ import type {
 import type { IsomerCommitMessage } from "@root/types/github"
 import { ALLOWED_FILE_EXTENSIONS } from "@root/utils/file-upload-utils"
 
-const EFS_VOL_PATH = config.get("aws.efs.volPath")
-const BRANCH_REF = config.get("github.branchRef")
-
+const STAGING_BRANCH = "staging"
+const STAGING_LITE_BRANCH = "staging-lite"
```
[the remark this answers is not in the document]
kishore03109: done upstream

src/services/db/CommitServiceGitFile.ts
kishore03109: nope, there is no need to here + aim is to not use this as subsitude!

src/constants/featureFlags.ts
```diff
@@ -1,3 +1,4 @@
 export const FEATURE_FLAGS = {
   GGS_WHITELISTED_REPOS: "ggs_whitelisted_repos",
+  IS_QUICKIE_WHITELISTED: "is_quickie_whitelisted",
```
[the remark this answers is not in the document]
kishore03109: this is fair. Ill change this in the leaf pr to this suggestion, and keep your suggestion cuz once its live hard to change the naming from growthbook

PR #966, opened 2023-10-13
fix(quickie): reduceAmpBuildTimesCreateTests

kishore03109:
[These tests will fail because dev tests are failing, interim solution in upstream pr #973]
## Problem

#965 has failing test cases. Had to convert js -> ts because I faced weird webpack issues that magically disappeared when shifting to TS. 

This does not add in new tests, but simply achieves parity with passing tests for #965. 

## Solution

Mostly changing APIs. 

## Tests
(same as #965) 
Manual tests in staging
Log into github: should be able to add a page to kishore-test-dev-gh and notice changes in staging and staging lite branch
log into email, should be able to add a page to kishore-test and notice changes in staging and staging lite.
All other repos, adding a page should have expected behaviour (adding an image should not throw errors + have a corresponding commit in github)


src/services/db/RepoService.ts
```diff
@@ -804,8 +804,6 @@ export default class RepoService extends GitHubService {
     sessionData: any,
     shouldMakePrivate: any
   ): Promise<any> {
-    return await super.changeRepoPrivacy(sessionData, {
-      shouldMakePrivate,
```
kishore03109, line 808: @alexanderleegs the API here seemed inconsistent, but I think I might be at the wrong here. happy to huddle this to shoulder check, but looking at the call sites should be able to just pass in boolean

src/services/db/GitHubService.ts
```diff
@@ -190,7 +239,13 @@ export default class GitHubService {
     return { content, sha }
   }
 
-  async readMedia(sessionData, { fileSha, branchName = STAGING_BRANCH }) {
+  async readMedia(
+    sessionData: { accessToken?: any; siteName?: any },
```
dcshzj, line 243: Should this be `UserWithSiteSessionData`?

src/services/db/GitHubService.ts
```diff
@@ -361,7 +451,7 @@ export default class GitHubService {
     return data
   }
 
-  async getRepoState(sessionData) {
+  async getRepoState(sessionData: { accessToken?: any; siteName?: any }) {
```
dcshzj, line 454: Same here regarding `UserWithSiteSessionData`.

src/services/db/GitHubService.ts
```diff
@@ -527,15 +641,17 @@ export default class GitHubService {
...
+      return okAsync(null)
     } catch (error) {
-      const { status } = error.response
-      // If user is unauthorized or site does not exist, show the same NotFoundError
-      if (status === 404 || status === 403) {
-        logger.error(
-          `User with id ${isomerUserId} attempted to change privacy of site ${siteName}`
-        )
-        return errAsync(new NotFoundError("Site does not exist"))
+      if (error instanceof AxiosError && error.response) {
```
dcshzj, line 646: Maybe `isAxiosError`?

src/services/db/GitHubService.ts
```diff
@@ -286,30 +357,46 @@ export default class GitHubService {
...
     } catch (err) {
       if (err instanceof NotFoundError) throw err
-      const { status } = err.response
-      if (status === 404) throw new NotFoundError("File does not exist")
-      if (status === 409)
-        throw new ConflictError(
-          "File has been changed recently, please try again"
-        )
-      throw err
+      if (axios.isAxiosError(err)) {
```
dcshzj, line 360: We have a `isAxiosError` util method, are we able to use that one?

harishv7 (approved):

src/services/db/__tests__/RepoService.spec.ts
```diff
@@ -198,10 +207,10 @@ describe("RepoService", () => {
...
 
       expect(actual).toEqual(expected)
-      expect(gitHubServiceCreate).toHaveBeenCalledWith(sessionData, {
-        content: mockContent,
-        fileName: mockFileName,
-        directoryName: mockDirectoryName,
+      expect(MockCommitServiceGitHub.create).toHaveBeenCalledWith(sessionData, {
+        content: "content",
+        fileName: "test.md",
+        directoryName: "",
```
dcshzj, line 213: We should use `mockContent`, `mockFileName` and `mockDirectoryName` above, the actual call should also be updated too.

kishore03109: b7fa30138bb1b5bacb5ff10b367e6c6d0d81473e

src/services/db/GitHubService.ts
kishore03109: b7fa30138bb1b5bacb5ff10b367e6c6d0d81473e

src/services/db/GitHubService.ts
kishore03109: b7fa30138bb1b5bacb5ff10b367e6c6d0d81473e

src/services/db/GitHubService.ts
kishore03109: b7fa30138bb1b5bacb5ff10b367e6c6d0d81473e

src/services/db/GitHubService.ts
kishore03109: b7fa30138bb1b5bacb5ff10b367e6c6d0d81473e

src/services/db/__tests__/RepoService.spec.ts
kishore03109: b7fa30138bb1b5bacb5ff10b367e6c6d0d81473e

PR #967, opened 2023-10-13
ref(media): defer image loading

seaerchin:
**NOTE: ** This PR requires the corresponding FE PR [here](https://github.com/isomerpages/isomercms-frontend/pull/1581)

## Problem
Right now, we return the image fully per call. This isn't optimal because we have to load the image fully even when we don't need it. As an example, consider the image move - we don't require the images (only titles), but we do a full load inclusive of the image data. 

## Solution
1. Defer fetching the image to the `readMediaFile` endpoint instead of returning wholesale from `listDirectoryContents`

**Tests coming in another PR**


alexanderleegs (approved):
lgtm

PR #970, opened 2023-10-16
feat(quickie): Github Commit Service Funcs

kishore03109:
## Problem

This PR introduces the complete CRUD functionality for Github Commit Service to function. 

## Solution

To be honest, there exists some functions that violate the single responsibility principle here, but am opting to keep refactor out side of this pr since we want to release this feature to our end users first. 

eg. all functions should look like `create` rather than what `renameSinglePath` looks like. 

**Breaking Changes**

## Tests
[for documentation only, will be more explicit about these tests in the leaf PR]

For all non gg whitelisted sites: 

1. For whitelisted quickie sites ->  all operations for CRUD on pages + assets should work as per normal, and commits should only appear in staging branch. All non assets should be committed to GitHub

2. For non-whitelisted quickie sites ->  all operations for CRUD on pages + assets should work as per normal, and commits should only appear in staging branch.

### DIFF
--- a/src/services/db/CommitServiceGithub.ts
+++ b/src/services/db/CommitServiceGithub.ts
@@ -74,13 +74,11 @@ export default class CommitServiceGitHub extends GitHubService {
       sha,
       fileName,
       directoryName,
-      branchName = STAGING_BRANCH,
     }: {
       fileContent: string
       sha: string
       fileName: string
       directoryName: string | undefined
-      branchName?: string
     }
   ): Promise<GitCommitResult> {
     const updatePromises = [
@@ -89,7 +87,7 @@ export default class CommitServiceGitHub extends GitHubService {
         sha,
         fileName,
         directoryName,
-        branchName,
+        branchName: STAGING_BRANCH,
       }),
     ]
     const shouldStagingLiteUpdate =
@@ -180,6 +178,7 @@ export default class CommitServiceGitHub extends GitHubService {
         })
       )
     }
+    await Promise.all(deletePromises)
   }
 
   async delete(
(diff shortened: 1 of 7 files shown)


src/services/db/GitHubService.ts
```diff
@@ -507,15 +508,16 @@ export default class GitHubService {
   async getTree(
     sessionData: UserWithSiteSessionData,
     githubSessionData: GithubSessionData,
-    { isRecursive }: { isRecursive: any }
-  ) {
+    { isRecursive }: any,
+    isStaging = true
```
dcshzj, line 512: Can we standardise to use `branchName`?

src/services/db/CommitServiceGithub.ts
```diff
@@ -1,7 +1,10 @@
 import { AxiosCacheInstance } from "axios-cache-interceptor"
 
+import GithubSessionData from "@root/classes/GithubSessionData"
 import UserWithSiteSessionData from "@root/classes/UserWithSiteSessionData"
-import { STAGING_LITE_BRANCH } from "@root/constants"
+import { STAGING_BRANCH, STAGING_LITE_BRANCH } from "@root/constants"
+import { ConflictError } from "@root/errors/ConflictError"
+import { GitCommitResult } from "@root/types/gitfilesystem"
 import { isReduceBuildTimesWhitelistedRepo } from "@root/utils/growthbook-utils"
 import isFileAsset from "@services/db/CommitServiceGitFile"
```
dcshzj, line 9: Should this be from `commit-utils.ts` instead?

src/services/db/CommitServiceGithub.ts
```diff
@@ -59,8 +62,330 @@ export default class CommitServiceGitHub extends GitHubService {
...
+    {
+      directoryName,
+      message,
+      githubSessionData,
+      isStaging,
+    }: {
+      directoryName: string
+      message: string
+      githubSessionData: GithubSessionData
+      isStaging?: boolean
```
dcshzj, line 129: Hmm why do we need to specify `isStaging`? Seems like it can be determined using `isFileAssets()`?

Similar comments to the other functions in this file.

src/services/db/GitHubService.ts
```diff
@@ -132,13 +133,13 @@ export default class GitHubService {
       fileName,
       directoryName,
       isMedia = false,
-      branchName = STAGING_BRANCH,
+      branchName,
```
seaerchin, line 136: this is abit weird - why is the default arg removed? we should just add this back to maintain behaviour. 

my concern here is that some caller doesn't pass an arg and relies on the default, which is no longer true here

src/services/db/GitHubService.ts
```diff
@@ -507,15 +508,16 @@ export default class GitHubService {
   async getTree(
     sessionData: UserWithSiteSessionData,
     githubSessionData: GithubSessionData,
-    { isRecursive }: { isRecursive: any }
-  ) {
+    { isRecursive }: any,
```
seaerchin, line 511: why is this typed as `any`? not a blocker as the prev revision already typed it as `any` too

src/services/db/RepoService.ts
```diff
@@ -590,64 +569,13 @@ export default class RepoService extends GitHubService {
...
-        // Delete old file
-        newGitTree.push({
-          ...item,
-          sha: null,
-        })
-      }
-    })
-
-    const newCommitSha = await super.updateTree(
+    return this.commitServiceGitHub.renameSinglePath(
```
seaerchin, line 572: didn't look into the method `renameSinglePath`, assuming is just a copy paste over

seaerchin (approved):
lgtm

src/services/db/GitHubService.ts
seaerchin: pls do this before merge - i scared of regression + it's a fast change

src/services/db/GitHubService.ts
kishore03109: solely cuz RepoService typed it as any, can become stricter in the future

src/services/db/GitHubService.ts
kishore03109: sure, ill add this in, but just to close the loop here this is not needed as 
1. only repoService calls githubService directly
2. repoService adds in this required param (both are in ts so is easier to guarentee)

kishore03109: ad0ee4c468d86c247401e2fc7fd49b350ad5d72f

src/services/db/RepoService.ts
kishore03109: this is the case

src/services/db/GitHubService.ts
kishore03109: this had quite a bit of fanout :sadge: deprio for this release lets revisit this since non fnctional?

PR #971, opened 2023-10-17
refactor(formsg-site-clone): remove and add to site creation

seaerchin:
## Problem
Previously we used a form to clone the site over to efs. Prod ops feedback that this should happen automatically on site creation, so this is done here

## Solution
Subsume cloning to be final step of site creation


src/routes/formsgSiteCreation.ts
```diff
@@ -82,20 +81,20 @@ export class FormsgRouter {
       if (!siteName) {
         const err = `A site name is required`
         await this.sendCreateError(requesterEmail, repoName, submissionId, err)
-        return res.sendStatus(200)
+        return cloneRes.sendStatus(200)
```
harishv7, line 84: where is cloneRes coming from?

harishv7: Where is the clone logic?

seaerchin: The rename in vscode renamed all instances instead of the variable...

I didn't catch this cos I didn't expect it to do a global rename of the word instead of the variable...

Will update?

src/routes/formsgSiteCreation.ts
```diff
@@ -119,6 +119,12 @@ export class FormsgRouter {
         repoName,
         isEmailLogin,
       })
+
+      // NOTE: Clone site to EFS
+      const cloneRes = await this.gitFileSystemService.clone(repoName)
```
seaerchin, line 124: @harishv7 clone logic is here

src/routes/formsgSiteCreation.ts
kishore03109, line 84: heads up that the types mismatch here !

kishore03109 (commented):
wah this is a more elegant solution, completely didnt think about this when reviewing pr :sadge:

src/routes/formsgSiteCreation.ts
seaerchin: yea due to vscode renaming the whole `res` stack instead of the var

src/routes/formsgSiteCreation.ts
```diff
@@ -119,6 +119,13 @@ export class FormsgRouter {
         repoName,
         isEmailLogin,
       })
+
+      if (isEmailLogin) {
+        // NOTE: Clone site to EFS
+        const cloneRes = await this.gitFileSystemService.clone(repoName)
+        if (cloneRes.isErr()) throw cloneRes.error
```
harishv7, line 126: should we add an alarm for this error? else we won't know right

seaerchin: outer loop has a `catch` that does `sendCreateError`

kishore03109 (approved):

PR #975, opened 2023-10-18
chore(commitService): proper naming

kishore03109:
## Problem

Changes the name to proper naming. This is also the leaf node that is intended to enter in production line by today.

## Tests

For all non gg + gg whitelisted sites:

For whitelisted quickie sites -> all operations for CRUD on pages + assets should work as per normal, and commits should only appear in staging branch. All non assets should be committed to GitHub

For non-whitelisted quickie sites -> all operations for CRUD on pages + assets should work as per normal, and commits should only appear in staging branch.

(Will update with videos after all reviews + rebasing are done for a final round of testing before going into main line) 

PS: will do this in a separate pr if bugs exist + during cutting of release


seaerchin (approved):

src/integration/Notifications.spec.ts
```diff
@@ -68,8 +68,8 @@ const MOCK_SITE_ID = "1"
 const MOCK_SITE_MEMBER_ID = "1"
 
 const gitFileSystemService = new GitFileSystemService(simpleGit())
-const commitServiceGitFile = new CommitServiceGitFile(gitFileSystemService)
-const commitServiceGitHub = new CommitServiceGitHub(isomerRepoAxiosInstance)
+const commitServiceGitFile = new GitFileCommitService(gitFileSystemService)
+const commitServiceGitHub = new GitHubCommitService(isomerRepoAxiosInstance)
```
dcshzj, line 72: We should rename the variable too, same for all other files.

harishv7 (approved):

PR #976, opened 2023-10-18
Feat/is 585 govt sgid login rollout

alexanderleegs:
## Problem

This PR adds support for govt officials to log in via singpass. To be reviewed in conjunction with PR [#1594](https://github.com/isomerpages/isomercms-frontend/pull/1594) on the isomercms-frontend repo. It changes the existing sgin-login endpoint to return an email instead of just the status, and adds an additional endpoint to allow for users with multiple emails to call a separate endpoint to login after the initial sgid login step. It also modifies the existing behaviour of the add collaborators endpoint so that new users can be added as a collaborator directly instead of requiring them to log into the cms first.

**Breaking Changes**

- [x]  Yes - this PR contains breaking changes
  - Minor breaking change - old sgid login will no longer work since the requested permissions no longer match. Acceptable behaviour because the existing sgid login is only usable by ogp officers currently

## Tests

- [ ] Run tests on associated frontend PR
- [ ] Check that a collaborator that does not previously exist can be added to a site


seaerchin:
Why would a user have separate emails?

src/errors/SgidErrors.ts
seaerchin, line 1: File name should be singular

src/routes/v2/sgidAuth.ts
```diff
@@ -86,25 +90,130 @@ export class SgidAuthRouter {
...
+    const userDataRes = await this.sgidAuthService
       .retrieveSgidAccessToken({ authCode, nonce, codeVerifier })
       .andThen(({ accessToken, sub }) =>
         // We can immediately process the access token - we only need to retrieve the email from sgid
-        this.sgidAuthService.retrieveSgidUserEmail(accessToken, sub)
+        this.sgidAuthService.retrieveSgidUserEmails(accessToken, sub)
       )
-      .andThen((email) => {
-        if (!email || !email.endsWith("@open.gov.sg")) {
+      .andThen((officerDetails) => {
```
seaerchin, line 100: Super nit: is it actually officer emails or officer details? If it's details then we should update the prior method call, otherwise this var name

src/routes/v2/sgidAuth.ts
```diff
@@ -86,25 +90,130 @@ export class SgidAuthRouter {
...
           )
         }
-        return okAsync(email)
+        return okAsync(officerDetails)
       })
-      .andThen((email) =>
-        // Login with email
+    if (userDataRes.isErr()) {
+      const { error } = userDataRes
+      if (
```
seaerchin, line 112: No error messages for this whole chunk - could we put in something meaningful to the user? 

Otherwise, they'd be redirected with no toast on why

src/routes/v2/sgidAuth.ts
```diff
@@ -86,25 +90,130 @@ export class SgidAuthRouter {
...
+            isomerUserId: user.id,
+            email: user.email,
+          }
+          Object.assign(req.session, { userInfo })
+          return res.status(200).send({ userData })
+        })
+        .mapErr((error) => res.status(500).send())
+    } else {
+      // User has multiple emails, defer to user to select email
+      const csrfTokenExpiry = new Date()
```
seaerchin, line 147: Why do we not require the same expiring and signing for single email users?

src/routes/v2/sgidAuth.ts
```diff
@@ -86,25 +90,130 @@ export class SgidAuthRouter {
...
+    never
+  > = async (req, res) => {
+    const { email } = req.body
+    const token = req.cookies[SGID_MULTIUSER_COOKIE_NAME]
+
+    const safeVerify = Result.fromThrowable(jwtUtils.verifyToken, (error) => {
+      logger.error(
+        `Error - invalid token for sgid multiuser login ${email}: ${error}`
+      )
+      return new SgidVerifyUserError()
```
seaerchin, line 175: Should we pass the message from sgid here?

seaerchin: Or at least provide some message for Fe to consume?

src/routes/v2/sgidAuth.ts
```diff
@@ -86,25 +90,130 @@ export class SgidAuthRouter {
...
+    })
+
+    await safeVerify(token)
+      .andThen((verifiedToken) => {
+        if (
```
seaerchin, line 180: If this is going to be repeated, we can extract as a function.

src/routes/v2/sgidAuth.ts
```diff
@@ -86,25 +90,130 @@ export class SgidAuthRouter {
...
+          `Error - token does not match expected format for sgid multiuser login ${email}`
+        )
+        return err(new SgidVerifyUserError())
+      })
+      .andThen((userData) => {
+        if (
+          Array.isArray(userData) &&
+          userData.every((item) => isPublicOfficerData(item))
+        )
+          return ok(userData as PublicOfficerData[])
```
seaerchin, line 197: nit: seems like these 2 should be packaged together

src/routes/v2/sgidAuth.ts
```diff
@@ -86,25 +90,130 @@ export class SgidAuthRouter {
...
+          userData.filter((data) => data.email === email).length > 0
+        if (!isValidUser) {
+          logger.error(
+            `Error - user with emails ${userData
+              .map((data) => data.email)
+              .join(", ")} attempting to login with unverified email ${email}`
+          )
+          return err(new SgidVerifyUserError())
+        }
+        return ok("")
```
seaerchin, line 214: supernit: can just return the data here or omit entirely so intention is clear

src/routes/v2/sgidAuth.ts
```diff
@@ -149,6 +250,10 @@ export class SgidAuthRouter {
       "/verify-login",
       attachReadRouteHandlerWrapper(this.handleSgidLogin)
     )
+    router.post(
+      "/verify-multiuser-login",
```
seaerchin, line 254: our path structure isn't ideal but this is something that can be fixed down the line. this should instead be `/verify/multi/login` or something

src/services/utilServices/SgidAuthService.ts
```diff
@@ -66,17 +81,28 @@ export default class SgidAuthService {
...
+          const employmentDetails: SgidPublicOfficerDetails[] = JSON.parse(
+            data[SGID_WORK_EMAIL_SCOPE]
+          )
+          const employmentResp = employmentDetails.map((detail) => ({
+            email: detail.work_email,
+            agencyName: detail.agency_name,
+            departmentName: detail.department_name,
+            employmentTitle: detail.employment_title,
+          }))
+          return employmentResp
```
seaerchin, line 104: supernit: can just combine (not imporant to fix)

src/types/sgid.ts
```diff
@@ -0,0 +1,19 @@
...
+  email: string
+  agencyName: string
+  departmentName: string
+  employmentTitle: string
+}
+
+export const isPublicOfficerData = (
+  item: unknown
+): item is PublicOfficerData => {
+  const officerData = item as PublicOfficerData
```
seaerchin, line 11: dangerous assertion - we first need an assertion of `!!item` as this doesn't guarantee existence of `item`

seaerchin (commented):
minor comments

src/routes/v2/sgidAuth.ts
alexanderleegs: The reason why no error messages are sent is because these endpoints only show up on the temporary redirect-in-progress page, which users should never see a toast/error message on. Our current behaviour is that the frontend handles the error messages according to status code (we only get 500/401 errors) upon the user getting redirected to the login page.

src/routes/v2/sgidAuth.ts
alexanderleegs: The expiring/signing is to ensure that a user attempting to use the multiuser login actually has access to that user - at this point we actually are able to log in the user as any of their approved emails already. But since we don't know which user they want to use, this requires additional input from the user, which requires us to ensure that an attacker cannot simply call this endpoint pretending to be the user.

src/routes/v2/sgidAuth.ts
alexanderleegs: Not repeated! single use, separated to make it easier to read

src/routes/v2/sgidAuth.ts
seaerchin: hmm ok - i was hoping for the frontend to extract the error message out from teh backend isntead of it deriving the message from status code. 

is this a fast change? if not, fine iwth leaving as is and doing lkater

src/services/utilServices/SgidAuthService.ts
alexanderleegs: Split because i needed to cast earlier!

src/services/identity/CollaboratorsService.ts
```diff
@@ -157,15 +157,8 @@ class CollaboratorsService {
...
-    const user = await this.usersService.findByEmail(email)
-    if (!user) {
-      // Error - user with a valid gov email does not exist
-      logger.error(`create collaborators error: user ${email} is not valid`)
-      return new NotFoundError(
-        `This user does not have an Isomer account. Ask them to log in to Isomer and try adding them again.`
-      )
-    }
+    // 3. Retrieve or create user if user doesn't already exist
+    const user = await this.usersService.findOrCreateByEmail(email)
```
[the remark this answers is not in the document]
seaerchin: won't happen - email now is from sgid

src/routes/v2/sgidAuth.ts
seaerchin: don't they hit this point **only after** they pass authorization checks from sgid? 

the multi-user email selection flow seems to be covered below

seaerchin: oh ok just double checking - this authorizes and retrieves data from sgid then passes back to FE

user selects email on FE and hits the below endpoint, **which verifies the security token** to ensure identity

if this is the flow, then please resolve

seaerchin (approved):
lgtm if my assumption on the login flow for multi user is correct. otherwise, **please dismiss**

src/types/sgid.ts
seaerchin: please do this before merge also

src/services/utilServices/SgidAuthService.ts
```diff
@@ -66,17 +81,28 @@ export default class SgidAuthService {
...
       this.sgidClient
         .userinfo({
           accessToken,
           sub,
         })
-        .then(({ data }) => data[SGID_WORK_EMAIL_SCOPE]),
+        .then(({ data }) => {
+          const employmentDetails: SgidPublicOfficerDetails[] = JSON.parse(
+            data[SGID_WORK_EMAIL_SCOPE]
+          )
```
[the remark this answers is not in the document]
alexanderleegs: Hmm this should be handled by the `ResultAsync` wrapper right?

src/services/identity/CollaboratorsService.ts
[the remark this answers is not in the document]
alexanderleegs: Yeah this is true - had a chat earlier and we decided that a quick fix for this would be to periodically go through the user table and remove users with no associated mobile numbers (since this means that have never made any cms edits before as you are blocked on doing anything until you fill in contact details)

src/errors/SgidErrors.ts
alexanderleegs: https://github.com/isomerpages/isomercms-backend/pull/976/commits/ee9bbe1313676bbf9580e93594a03fd6d302bc00

src/routes/v2/sgidAuth.ts
alexanderleegs: officer details! https://github.com/isomerpages/isomercms-backend/pull/976/commits/ee9bbe1313676bbf9580e93594a03fd6d302bc00

src/routes/v2/sgidAuth.ts
alexanderleegs: Let's do this in a separate PR?

src/types/sgid.ts
alexanderleegs: https://github.com/isomerpages/isomercms-backend/pull/976/commits/eacf951ad2a268bfa36dcf10b25b42a2ff1956db

src/routes/v2/sgidAuth.ts
alexanderleegs: 48a15e38

src/routes/v2/sgidAuth.ts
alexanderleegs: 48a15e38

src/routes/v2/sgidAuth.ts
```diff
@@ -86,25 +90,130 @@ export class SgidAuthRouter {
 
     const { nonce, codeVerifier } = cookieData
 
-    // Exchange the authorization code and code verifier for the access token, then use the access token to retrieve user's email
-    await this.sgidAuthService
+    // Exchange the authorization code and code verifier for the access token, then use the access token to retrieve user's data
```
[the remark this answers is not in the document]
alexanderleegs: b6434bda

PR #981, opened 2023-10-18
fix(githubservice): invalidate keys

seaerchin:
## Problem
Cache wasn't being invalidated so we might have used stale data 

## Solution
Invalidate the cache

### DIFF
--- a/src/services/db/GitHubService.ts
+++ b/src/services/db/GitHubService.ts
@@ -17,7 +17,7 @@ import { RawGitTreeEntry } from "@root/types/github"
 
 import * as ReviewApi from "./review"
 
-const CACHE_KEYS = {
+export const CACHE_KEYS = {
   read: {
     directory: "read-directory",
   },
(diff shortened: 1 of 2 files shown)


kishore03109 (approved):
I am quite confused on how this works, for we are introducing a cache key + then invalidating it, which seems to me would have the exact same behaviour if we did not have the cache key in the first place. I am approving this for now since this effort seems more of something to test a hypothesis that this does fix this flaky bug, + the time sensitive nature of this pr going into today's release, but I do request to 
1. follow up if this works
2. and if this doesnt, to remove this. This seems like the only place (correct me if I am wrong) where we are manually introducing this type of cache which might seem confusing + lead to behvaour that we dont typically expect.


Blocking comment: I am assuming the create + delete + update are the only CUD operations on images, and functions like `listMediaDirectory` + `getTree` operations use these under the hood. If this assumption is wrong, please dismiss this review

harishv7:
@seaerchin destination branch should be develop. will need to recut

harishv7:
@seaerchin also tests are failing

seaerchin:
> I am quite confused on how this works, for we are introducing a cache key + then invalidating it, which seems to me would have the exact same behaviour if we did not have the cache key in the first place. 

the behaviour is different because now the cache invalidation is controlled by the mutation also instead of purely on the etags. 

a good parallel is on our frontend, how we invalidate the read query key post mutation; this is a similar principle

PR #982, opened 2023-10-19
feat: allow Isomer admins to edit on email-login and adjust permissions

dcshzj:
## Problem

Closes IS-654.

## Solution

**Breaking Changes**

**Features**:

- A new collaborator role `ISOMERADMIN` has been added to identify users in the IsomerAdmins table. This additional role allows Isomer admins to have access to sites and act as if they are an admin of that particular site.

**Improvements**:

- The logic for the review request summary has been updated to reject GitHub login users if the site has already been migrated. This is to prevent GitHub users from editing their existing sites after we migrate them from GitHub to email login, and to encourage them to edit using their email-login account.

## Tests

- Unit tests
- Smoke tests (on https://github.com/isomerpages/isomercms-frontend/pull/1604)

## Deploy Notes

*None*

### DIFF
--- a/src/services/identity/__tests__/CollaboratorsService.spec.ts
+++ b/src/services/identity/__tests__/CollaboratorsService.spec.ts
@@ -286,7 +286,10 @@ describe("CollaboratorsService", () => {
       // Arrange
       collaboratorsService.deriveAllowedRoleFromEmail = (jest.fn(
         () => CollaboratorRoles.Admin
-      ) as unknown) as () => Promise<CollaboratorRoles | null>
+      ) as unknown) as () => Promise<Exclude<
+        CollaboratorRoles,
+        CollaboratorRoles.IsomerAdmin
+      > | null>
       mockSitesService.getBySiteName.mockReturnValue(
         okAsync({ id: mockSiteId })
       )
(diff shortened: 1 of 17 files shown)


dcshzj:
> @dcshzj hmm wait we already have a table for `isomer_admins`, why not just make anyone who is an isomer_admin admin for all sites ah? this way we dont have to maintain a map of site_memeber_admins -> sites bah

Hmm @kishore03109 are you able to clarify your comment? I'm using this table to determine who is an Isomer admin and enhancing existing endpoints to be able to retrieve this information from this isomer_admin table, so there is no adjustments made to the site_members table.

src/services/middlewareServices/AuthorizationMiddlewareService.ts
```diff
@@ -47,8 +47,10 @@ export default class AuthorizationMiddlewareService {
       userId
     )
     return collaboratorType === CollaboratorRoles.Admin
```
kishore03109, line 49: Semantically ok with this, but this took me a while to digest, nit would be to refactor this and use type guards. 
Assuming there will be a case where `collaboratorType.Admin` && `collaboratorRole === CollaboratorRoles.Contributor`, which is why this complexity is needed. Please change this logic if this assumption is wrong.

src/services/identity/__tests__/CollaboratorsService.spec.ts
```diff
@@ -286,7 +315,10 @@ describe("CollaboratorsService", () => {
       // Arrange
       collaboratorsService.deriveAllowedRoleFromEmail = (jest.fn(
         () => CollaboratorRoles.Admin
-      ) as unknown) as () => Promise<CollaboratorRoles | null>
+      ) as unknown) as () => Promise<Exclude<
```
kishore03109, line 318: shall we just create a nonIsomerAdmin role type since we are using this at multiple places?

kishore03109 (approved):
Arh i misunderstood!  this implementation is fine, approving with nits

kishore03109:
Dont merge in today ya, we upgrading version today

src/services/middlewareServices/AuthorizationMiddlewareService.ts
dcshzj: Hmm yeah I agree this should be refactored, because otherwise this would get very complex once we introduce more roles beyond Admin and Contributor. The assumption you have is correct.

src/services/identity/__tests__/CollaboratorsService.spec.ts
dcshzj: Adjusted in 771d9868!

PR #985, opened 2023-10-19
Feat/quickie/site-create-form

kishore03109:
## Problem

Moving forward to support quickie, all sites need to have 2 branches, `staging` + `staging-lite` AND 2 amplify apps. This modifies the site create form to allow this to happen.
Closes [insert issue #]

  - This does have implication with site privitisation. I have updated the [oncall runbook](https://www.notion.so/opengov/On-Call-Runbook-ab4bffa2472d47ef9b657242632da7f9?pvs=4#0ebf1ee6cf4c44a4a01352ce7fa79f20) on the additional step needed to make a site private AFTER this pr has been merged into main line. 

### DIFF
--- a/src/services/identity/DeploymentsService.ts
+++ b/src/services/identity/DeploymentsService.ts
@@ -48,12 +48,16 @@ class DeploymentsService {
       amplifyStagingLiteResult,
     ] = await this.createAmplifyAppsOnAws(repoName)
     if (amplifyStagingResult.isErr()) {
-      logger.error(`Amplify set up error: ${amplifyStagingResult.error}`)
+      logger.error(
+        `Amplify set up error for main app: ${amplifyStagingResult.error}`
+      )
       throw amplifyStagingResult.error
     }
 
     if (amplifyStagingLiteResult.isErr()) {
-      logger.error(`Amplify set up error: ${amplifyStagingLiteResult.error}`)
+      logger.error(
+        `Amplify set up error for staging-lite app: ${amplifyStagingLiteResult.error}`
+      )
       throw amplifyStagingLiteResult.error
     }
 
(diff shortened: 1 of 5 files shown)


src/services/identity/DeploymentsService.ts
```diff
@@ -43,32 +43,55 @@ class DeploymentsService {
...
+      amplifyStagingLiteResult,
+    ] = await this.createAmplifyAppsOnAws(repoName)
+    if (amplifyStagingResult.isErr()) {
+      logger.error(`Amplify set up error: ${amplifyStagingResult.error}`)
+      throw amplifyStagingResult.error
     }
-    const amplifyInfo = amplifyResult.value
+
+    if (amplifyStagingLiteResult.isErr()) {
+      logger.error(`Amplify set up error: ${amplifyStagingLiteResult.error}`)
```
dcshzj, line 56: Should we differentiate between this error message and the previous one? So that it is easier at a glance to know which app creation failed.

src/services/identity/ReposService.ts
```diff
@@ -206,45 +216,73 @@ export default class ReposService {
...
+      .clone(repoUrl, stgLiteDir)
+      .checkout("staging")
+      .rm(["-r", "images"])
+      .rm(["-r", "files"])
+
+    // Clear git
+    fs.rmSync(`${stgLiteDir}/.git`, { recursive: true, force: true })
+
+    // Prepare git repo
+
```
dcshzj, line 277: nit: spacing

dcshzj (approved):
Non-blocking comments

src/services/identity/ReposService.ts
kishore03109: 9fbbd471bd21ea9251aaffb7e7f4eabfa8794afe

src/services/identity/DeploymentsService.ts
kishore03109: fair point 9fbbd471bd21ea9251aaffb7e7f4eabfa8794afe

PR #987, opened 2023-10-20
Fix/quickieUpdatingHomepage

kishore03109:
## Problem

Top level updates to staging lite did not go through since the directoryName was undefined. 

## Solution

Rather than trying to compute this, am making isFileAsset smarter by making it compute the path so that a similar interface can be used by all of its callers.

Also I made a mistake yesterday during rebase, I think it accidentally committed the files. This is a safe delete, this is not referenced anywhere in the code base. 

## Tests

Changes made to homepage, navbar and contact us should all be reflected on staging lite for kishore-test in staging-cms.isomer.gov.sg

## Deploy Notes

Will be releasing it on Monday/Tuesday to gather early feedback before releasing it to all sites.


alexanderleegs (approved):
lgtm

PR #988, opened 2023-10-20
Update ci.yml

alexanderleegs:
Builds to new production environment and swap to use 18.x


alexanderleegs:
> just to double check, can we test the ci on staging by using the same configs for node_version too?

Yeah! we can push this to staging first to make sure it builds

kishore03109 (approved):
This seems like it needs to go in for release right?

harishv7:
@kishore03109 pls merge before release

PR #996, opened 2023-10-26
feat(GH<->GGs): Repair GGs 

kishore03109:
## Problem
Pains: 
1. Need for Ops to ask Eng to reset when there is a drift from GH and EFS
2. If quickie has bugs that cause drift between staging and staging-lite, Eng would have had to go into efs, rm the staging-lite repo, git clone and make the repo gitless and assetless before committing to staging-lite to make sure that the repos are in sync 
3. Need for Eng to clone repos into EFS during email migration. (Phase 2 shipposting)
4. Need for Eng to set up staging-lite for quickie (One time effort) 
This is also annoying since we need to go through a bunch of hops to enter into our efs in prod, which makes oncall a pain. 

Rather than having separate scripts for each of this pain point, this form intends to kill 4 birds in one stone. 

## Solution

This formsg script is a idempotent script to be used when 

1. When we want to clone a repo into GGs.
2. When we want to modify a repo directly in Github and we want to sync the repos.
3. Repair possible drifts between staging and staging lite branches.

addressing @alexanderleegs' comments at the PR desc level cuz its not v clear from just the code
>Hmm actually does this assume that a staging lite doesn't exist? it seems to attempt a clone + pull, but we also want this to be able to repair error right, is this the desired behaviour?

We always assume staging lite doesnt exist. the reason why we dont git pull is because once we git pull, .git reconciles and the folder bloats, and we lose the benefits of a small bundle size. so we always reset, and this is not too long (< 1min for nlb iirc. 

We also always clone and pull. the reason for this is because cloning takes time, with repos like nlb it can take a while, so we do a pull right before we reset staging lite. 

It is notable that during this period of time, any changes to the repo will screw this repair up, but actually this is ok because:
1. we can always re-run this script
2. we are not rate-limited in any way for the above  ^

 

**Breaking Changes**

  - flow has to change for phase 2 migration, updated in the [read me ](https://github.com/isomerpages/isomer-conversion-scripts/pull/75) post review

## Screenshots
Error state:
Screenshot 2023-10-27 at 12 00 59 PM

Success state:
Screenshot 2023-10-27 at 12 01 05 PM

## Tests
// staging ssm stuff is not working. 

## Deploy Notes

- `GGS_REPAIR_FORM_KEY` : Form secret key 
    - [X] added env var to 1PW + SSM script (`fetch_ssm_parameters.sh`)

**New scripts**:

- `script` : script details

**New dependencies**:

**New dev dependencies**:


src/routes/formsg/formsgSiteLaunch.ts
```diff
@@ -26,12 +26,12 @@ import { DigResponse, DigType } from "@root/types/dig"
 import UsersService from "@services/identity/UsersService"
 import InfraService from "@services/infra/InfraService"
 
-const { SITE_LAUNCH_FORM_KEY } = process.env
+const SITE_LAUNCH_FORM_KEY = config.get("formSg.siteLaunchFormKey")
```
kishore03109, line 29: not sure why this wasnt migrated.. considering that we dont use env vars anymore I am abit surprised how this works

alexanderleegs: Possibly we might not have had any site launches since the node 18 swapover? On the node 16 env we didn't delete our existing env vars, so it's possible that it was still reading from there

kishore03109: it is a safe change, since ssm populates our env. we are just changing the place where it reads here. resolving this ya

src/routes/formsg/formsgGGsRepair.ts
```diff
@@ -0,0 +1,264 @@
...
+                  return okAsync(true)
+                })
+            )
+          })
+          .orElse((error) => {
+            if (error === false) {
+              // Repo exists in EFS, no need to clone, but continue with syncing
+              return okAsync(false)
+            }
+            return errAsync<boolean, GitFileSystemError>(error)
```
alexanderleegs, line 147: Nit: isn't this guaranteed to be a `GitFileSystemError`?

src/routes/formsg/formsgGGsRepair.ts
```diff
@@ -0,0 +1,264 @@
...
+          .andThen(() =>
+            // repo exists in efs, but we need to pull for staging and reset staging lite
+            this.gitFileSystemService
+              .pull(repoName, "staging")
+              .andThen(() =>
+                fromPromise(
+                  this.reposService.setUpStagingLite(
+                    path.join(EFS_VOL_PATH_STAGING_LITE, repoName),
+                    repoUrl
+                  ),
```
alexanderleegs, line 158: Hmm actually does this assume that a staging lite doesn't exist? it seems to attempt a clone + pull, but we also want this to be able to repair error right, is this the desired behaviour?

src/services/middlewareServices/FormsProcessingService.ts
```diff
@@ -47,12 +38,12 @@ export default class FormsProcessingService {
...
 
-  decrypt = ({ formKey }: { formKey: string }) => (
+  decrypt = ({ formKey }: { formKey: string }) => async (
     req: Request,
     res: Response,
     next: NextFunction
-  ): void => {
-    const submission = this.formsg.crypto.decrypt(
+  ): Promise<void> => {
+    const submission = await this.formsg.crypto.decryptWithAttachments(
```
alexanderleegs, line 46: Do existing forms (without attachments) work with this? do we need a conditional on attachment existence?

alexanderleegs (commented):
Left comments!

Do also add the new env var to the deploy script to fetch ssm params

src/services/middlewareServices/FormsProcessingService.ts
kishore03109: yes, checked already

src/routes/formsg/formsgGGsRepair.ts
kishore03109: actually fair this not v clear, updated pr desc 
We always assume staging lite doesnt exist. the reason why we dont git pull is because once we git pull, .git reconciles and the folder bloats, and we lose the benefits of a small bundle size. so we always reset, and this is not too long (< 1min for nlb iirc.

We also always clone and pull. the reason for this is because cloning takes time, with repos like nlb it can take a while, so we do a pull right before we reset staging lite.

It is notable that during this period of time, any changes to the repo will screw this repair up, but actually this is ok because:

we can always re-run this script
we are not rate-limited in any way for the above ^

src/routes/formsg/formsgGGsRepair.ts
```diff
@@ -0,0 +1,264 @@
+/* eslint-disable import/prefer-default-export */ // todo remove this line post dev
```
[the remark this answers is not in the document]
kishore03109: uhm ill j leav eit here since other classes do the same + this eslint rule is tirival

src/routes/formsg/formsgGGsRepair.ts
kishore03109: 76b0c5c93475d96d36384aa1aa6c402527ab51e9

src/services/middlewareServices/FormsProcessingService.ts
kishore03109: Wait to be clear this was a breaking change to the api, so the other forms had to change and I checked that the other forms were able to parse the content (ie the individual fields)

kishore03109:
Do also add the new env var to the deploy script to fetch ssm params
@alexanderleegs , yup added

alexanderleegs (approved):
lgtm

PR #997, opened 2023-10-27
feat(media): include added time for every file

dcshzj:
## Problem

Part of 681 and 682.

## Solution

**Breaking Changes**

**Features**:

- The response when getting each media file now includes the timestamp that it was added.
    - For GGS repos, the added time is correct in the sense that it is the "birthtime" of the file.
    - For non-GGS repos, because Git has no concept of file creation timestamps, we have to get this information from the Git commit history. However, one caveat is that it is possible for the user to have deleted the file and created another with the same name, so we cannot rely on getting the earliest commit of that path. Since we generally only create commits when they upload, we can make a "best guess" using the latest commit.

**Improvements**:

- The `neverthrow` library has been upgraded to take advantage of [better type inferences](https://github.com/supermacro/neverthrow/releases/tag/v5.1.0). The breaking changes has been fixed in the codebase.

## Tests

- Unit tests.

## Deploy Notes

**UPGRADED dependencies**:

- `neverthrow` : for type inferences when using `combine`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,7 +73,7 @@
     "module-alias": "^2.2.2",
     "moment-timezone": "^0.5.35",
     "morgan": "~1.10.0",
-    "neverthrow": "^4.3.1",
+    "neverthrow": "^6.1.0",
     "nocache": "^3.0.4",
     "node-dig-dns": "^0.3.3",
     "otplib": "^12.0.1",
--- a/src/services/db/GitFileSystemService.ts
+++ b/src/services/db/GitFileSystemService.ts
@@ -792,7 +792,7 @@ export default class GitFileSystemService {
 
   getMimeType(fileExtension: string): Result<string, MediaTypeError> {
     if (!ALLOWED_FILE_EXTENSIONS.includes(fileExtension)) {
-      err(new MediaTypeError("Unsupported file extension found"))
+      return err(new MediaTypeError("Unsupported file extension found"))
     }
     switch (fileExtension) {
       case "svg":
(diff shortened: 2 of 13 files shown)


package.json
```diff
@@ -73,7 +73,7 @@
     "module-alias": "^2.2.2",
     "moment-timezone": "^0.5.35",
     "morgan": "~1.10.0",
-    "neverthrow": "^4.3.1",
+    "neverthrow": "^6.1.0",
```
kishore03109, line 76: wait the changes described in the pr, they are not breaking changes right? Ie the semantics of the code doesnt change with or without the declaration since now neverthrow does the inference?

src/services/db/GitFileSystemService.ts
```diff
@@ -800,7 +792,7 @@ export default class GitFileSystemService {
 
   getMimeType(fileExtension: string): Result<string, MediaTypeError> {
     if (!ALLOWED_FILE_EXTENSIONS.includes(fileExtension)) {
-      err(new MediaTypeError("Unsupported file extension found"))
+      return err(new MediaTypeError("Unsupported file extension found"))
```
kishore03109, line 795: previously hidden error? O.o
wah change in infering types by neverthrow seems timely

kishore03109 (approved):

package.json
dcshzj: Breaking changes because we no longer can import `combine` directly from `neverthrow`, so it needed to be updated when the library is updated.

src/services/db/GitFileSystemService.ts
dcshzj: Hmm sadly this was not because of `neverthrow` but rather due to the lack of unit tests. I noticed this when adding the unit tests for this function!

package.json
kishore03109: @dcshzj opps could i trouble you to rebase then + fix breaking changes then? i used combine in the latest pr that i merged

dcshzj: Updated in a594951d10e2bec46691890d820d8362f901e1b6!

PR #999, opened 2023-10-30
feat(quickie): correct staging url

kishore03109:
## Problem

Currently, the feature flag for quickie only allows/prevents commits to staging + staging-lite branches in Github. 
Enabling quickie for a site required a manual step to change the staging-urls directly in the db. Any rollback prior to this also required a manual intervention to go into the DB to mutate the staging-url manually. This is not scalable to both roll out + quickly rollback when a bug is found.

Closes [insert issue #]

## Solution

To keep this feature in line with what we typically expect to work with other feature flags, this PR 
1. produces the correct staging url when it has enough information to do so 
2. Updates our DB value with the correct value based on the dynamic boolean value of whether the feature flag is turned on for that repo

**Breaking Changes**

  - Details ...


dcshzj (approved):
LGTM!

QiluXie (approved):

PR #1003, opened 2023-11-01
fix(simple git): avoid chaining

kishore03109:
## Problem

We are facing issues with files being deleted for seemingly innocuous deletes from certain folders. simple git supports chaining, but note that each process is async, which they have have queue that is awaited
But hor, the var cwd is [shared](https://github.com/steveukx/git-js/blob/8a31c45a1359248cb38d768d22a5e29903ceef8a/simple-git/src/lib/runners/git-executor-chain.ts#L18), rather than attached to each process that is in the queue.

This means that at production level, we cannot afford to use their simple git’s chaining at all, and it will lead to all sorts of concurrency issues that we have no control over.
I think the reason why quickie affects this is because when creates are done, they are done almost simultaneously to two branches, so while this bug always existed in production, its hard to replicate this unless two git processes with this chaining occured at the exact same time, which is exactly kinda what happens with quickie

## Tests 
I am able to replicate this locally when running against ~30 repos. 

### Before 
Screenshot 2023-11-01 at 9 49 42 AM

### After 

Screenshot 2023-11-01 at 9 46 41 AM

### DIFF
--- a/src/services/identity/ReposService.ts
+++ b/src/services/identity/ReposService.ts
@@ -1,4 +1,3 @@
-import { exec } from "child_process"
 import fs from "fs"
 import path from "path"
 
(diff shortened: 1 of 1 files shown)


kishore03109:
closing this as better fix is #1004

PR #1004, opened 2023-11-01
fix: reposervice fix cwd

harishv7:
## Problem

By right we are supposed to use `root = false`, to quite the [doc](https://github.com/steveukx/git-js) 
```
.cwd({ path, root = false }) Sets the working directory for all future commands either in the current chain of commands (where root is omitted or set to false) or in the main instance (where root is true).
```

Ran this agains this against ~30 sites at one go, no errors like the ones we previously observed
Screenshot 2023-11-01 at 5 29 23 PM

### Tests 
- [ ] e2e tests for GGs 
- [ ] For a GGs site in staging, able to CRUD on pages + CRUD on images


kishore03109 (commented):
our git file system service also need changes

kishore03109:
lets branch off from last week's release + create 2 prs (one for dev, one to master) to do a hotfix today

also have we load tested this against chaining? (or you think no need)

kishore03109 (approved):

PR #1007, opened 2023-11-01
Fix/quickie/dbUpdate

kishore03109:
## Problem

Somehow the staging sites, while rendering correctly to the end user, did not mutate the DB as expected . 

Closes [insert issue #]

## Solution

Not too sure why the previous code didnt work, directly importing the deployments service to mutate this instead.

**Breaking Changes**

  - Details ...

## Tests

- [ ] in staging, log into kishore-test and view the staging url, since it is whitelisted it should be the staging-lite url 
- [ ] log into growthbook, go into the `is_show_staging_build_status_enabled` for staging and remove `kishore-test`
- [ ] refresh the cms 
- [ ] assert that the db value for `stagingUrl` have changed back to the `staging.amplifyapp.com`, the staging link should now be just the `staging.amplifyapp.com` (note that its NOT the staging-lite link anymore) 
- [ ] log into growthbook, go into the `is_show_staging_build_status_enabled` for staging and add `kishore-test` back in 
- [ ] refresh cms 
- [ ] assert that the db value for `stagingUrl` have changed back o `staging-lite.amplifyapp.com`
- [ ] the staging link should revert back to the `staging-lite.amplifyapp.com` (note that its NOT the staging link anymore)

### DIFF
--- a/src/fixtures/repoInfo.ts
+++ b/src/fixtures/repoInfo.ts
@@ -11,6 +11,9 @@ export const MOCK_STAGING_URL_CONFIGYML: NonNullable<
 export const MOCK_STAGING_URL_DB: NonNullable<
   ConfigYmlData["staging"]
 > = Brand.fromString("https://repo-staging-db.netlify.app")
+export const MOCK_STAGING_LITE_URL_DB: NonNullable<
+  ConfigYmlData["staging"]
+> = Brand.fromString("https://repo-staging-lite-db.netlify.app")
 
 export const MOCK_PRODUCTION_URL_GITHUB: NonNullable<
   ConfigYmlData["prod"]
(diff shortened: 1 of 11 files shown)


seaerchin (approved):
ok with this for now - just be aware that this fixes only if the 1-1 relation is the one causing the issue, otherwise it might still be present (just need to be aware). 

the tests don't seem to relate to the fix also - the fix is on `mutation` but for the manual test, we appear to be mutating directly?

kishore03109:
@seaerchin yea somehow that relation didnt work in production. 

wait ah why do you say that? the mutations are symmetrical at the repo level?  ie from `staging.amplifyapp.com` -> `staging-lite.amplifyapp.com`

let me change the test to actually directly update the growthbook for this to be clearer!

PR #1009, opened 2023-11-01
feat(mediafiles): add new mass creation endpoint

seaerchin:
## Problem
No mass creation endpoint - this is problematic because repeatedly calling the single media creation endpoint causes 429 errors (repo locked)

## Solution
duplicate most code from single endpoint

### DIFF
--- a/src/routes/v2/authenticatedSites/mediaFiles.js
+++ b/src/routes/v2/authenticatedSites/mediaFiles.js
@@ -13,6 +13,7 @@ const {
   CreateMediaFileRequestSchema,
   UpdateMediaFileRequestSchema,
   DeleteMediaFileRequestSchema,
+  CreateMediaFilesRequestSchema,
 } = require("@validators/RequestSchema")
 
 class MediaFilesRouter {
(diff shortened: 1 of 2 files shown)


src/routes/v2/authenticatedSites/mediaFiles.js
```diff
@@ -43,6 +44,35 @@ class MediaFilesRouter {
...
 
+  async createMediaFiles(req, res, next) {
+    const { userWithSiteSessionData } = res.locals
+
+    const { directoryName } = req.params
+    const { error } = CreateMediaFilesRequestSchema.validate(req.body)
+    if (error) throw new BadRequestError(error.message)
+
+    const { files } = req.body
+    const resolvedFiles = await Promise.allSettled(
```
dcshzj, line 55: I think this might not work for GitHub, because they might throw an error about some hash mismatch if the requests are sent too quickly.

For local Git, we commit (and push) on every create, which means we might encounter the Git lock when multiple requests are committing at the same time.

We might need a new method that either has the ability to retry for x number of times, or to sequentially make these create requests.

seaerchin: Dang ok it's not a big deal to do this sequentially; this actually worked locally but it did get a 429

kishore03109 (approved):
Please only merge this after the sequential calls

seaerchin:
might close, doing sequential on FE

PR #1015, opened 2023-11-03
fix(showBuildStatus): add logging

kishore03109:
## Problem
1. This API design throws 4xx errors in a large proportion. While this has no user impact, eng is getting alarms.
2. Am facing some difficulities with the stagin build status badges showing up in production, and I have no idea why. Works in staging + local but not in prod, adding logging to figure out why (prob a permissioning issue <- hard to test since prems need a redploy, and I dont want to redploy when our instances are unhealthy due to large proportion of 4xx errors) 

TLDR: returning a 2xx success no matter what


src/routes/v2/authenticated/sites.ts
```diff
@@ -184,7 +185,8 @@ export class SitesRouter {
     if (result.isOk()) {
       return res.status(200).json(result.value)
     }
-    return res.status(404).json({ message: "Unable to get staging status" })
+    logger.error({ message: result.error.message })
+    return res.status(200).json(undefined)
```
seaerchin, line 189: does our frontend actually make use of this? (we went from returning `{message: string}` to `undefined`, which might have unintended consequences if FE always expects `message` to be present

seaerchin (approved):

kishore03109:
closing this for better api design of not pinging this endpoing at all if not defined

PR #1018, opened 2023-11-06
refactor(v1): completely nuke v1 endpoints and associated classes

dcshzj:
**Note: This change has medium risks and requires extensive testing to ensure no breakages**

## Problem

The v1 API endpoints are no longer used.

## Solution

**Breaking Changes**

**Improvements**:

- All v1 endpoints and their associated classes have been removed, marking the end of an era.

## Tests

Unit tests, e2e tests on frontend.

## Deploy Notes

*None*


kishore03109 (approved):
Wah this is skimmed because I assume this is removal of all v1 points that was already remvoed.

HIGHLY worried about bugs, I think this makes e2e a must + potential QA, thoughts?

PR #1019, opened 2023-11-07
fix(media): change media sorting to addedTime descending

dcshzj:
## Problem

The ordering of the media files should be by the addedTime in descending order.

## Solution

**Breaking Changes**

**Improvements**:

- The media files are now sorted by the addedTime in descending order for users on GGS (i.e. email-login users).
    - Unfortunately, it is very expensive to get the addedTime for each file from the GitHub API, as there is no concept of creation timestamp (since it is entirely based on commits). We currently do get this information when individual files are requested, but it would get very expensive if we do this for all files at once.
    - Hence, we are retaining the existing behaviour of sorting by the file name for GitHub-login users, but we will let email-login users have the sorting by addedTime.

## Tests

Unit tests, smoke tests on the frontend.

- [ ] Navigate to any site and visit the images workspace as an email-login user
- [ ] Verify that the images are sorted by their added time in descending order
- [ ] Navigate to any site and visit the images workspace as a GitHub-login user
- [ ] Verify that the images are still sorted by their file names in ascending order

## Deploy Notes

*None*

### DIFF
--- a/src/services/db/GitFileSystemService.ts
+++ b/src/services/db/GitFileSystemService.ts
@@ -927,6 +927,7 @@ export default class GitFileSystemService {
                 sha,
                 path,
                 size: type === "dir" ? 0 : stats.size,
+                addedTime: stats.birthtimeMs,
               }
 
               return okAsync(result)
--- a/src/services/db/RepoService.ts
+++ b/src/services/db/RepoService.ts
@@ -1,4 +1,3 @@
-import { GrowthBook } from "@growthbook/growthbook"
 import { AxiosCacheInstance } from "axios-cache-interceptor"
 import _ from "lodash"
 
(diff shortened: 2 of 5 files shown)


src/services/db/GitFileSystemService.ts
```diff
@@ -927,6 +927,7 @@ export default class GitFileSystemService {
                 sha,
                 path,
                 size: type === "dir" ? 0 : stats.size,
+                addedTime: stats.birthtimeMs,
```
seaerchin, line 930: do we not need this for github?

seaerchin (commented):
double checking - we only sort by recency for **email** users? for github users still sort by name

seaerchin (approved):
saw write up

src/services/db/GitFileSystemService.ts
seaerchin: also, why not use `stats.mtimeMs` (last modified time - see [here](https://nodejs.org/docs/v9.9.0/api/fs.html#fs_stats_birthtimems))

PR #1027, opened 2023-11-10
fix(cm): extra timeout

kishore03109:
## Problem

cm giving us issues. Increase time out, else eng should be able to turn it off to reduce biz impact

### DIFF
--- a/src/types/featureFlags.ts
+++ b/src/types/featureFlags.ts
@@ -2,7 +2,7 @@
 // Add BE feature flags here to mirror that on GrowthBook
 // Note: key should mirror GrowthBook exactly as it is
 
-export interface cmConfigType {
+export interface CloudmersiveConfigType {
   is_enabled: boolean
   timeout: number
 }
@@ -11,7 +11,7 @@ export interface FeatureFlags {
   is_build_times_reduction_enabled: boolean
   is_ggs_enabled: boolean
   is_show_staging_build_status_enabled: boolean
-  is_cloudmersive_enabled: cmConfigType
+  is_cloudmersive_enabled: CloudmersiveConfigType
 }
 
 // List of attributes we set in GrowthBook Instance in auth middleware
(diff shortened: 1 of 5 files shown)


src/utils/growthbook-utils.ts
```diff
@@ -45,3 +45,18 @@ export const isShowStagingBuildStatusWhitelistedRepo = (
     false
   )
 }
+
+export const isCloudmersiveEnabled = (
+  growthbook: GrowthBook<FeatureFlags> | undefined
+): cmConfigType => {
+  const defaultConfig = {
+    is_enabled: false,
```
dcshzj, line 53: nit: should we default to true instead to be safe if growthbook is down?

src/types/featureFlags.ts
```diff
@@ -1,10 +1,17 @@
 // Use for type safety with GrowthBook
 // Add BE feature flags here to mirror that on GrowthBook
 // Note: key should mirror GrowthBook exactly as it is
+
+export interface cmConfigType {
```
dcshzj, line 5: nit: the interface name should start with caps, and might be better to spell Cloudmersive out in full because it might not be clear what it stands for when we look at this in the future.

dcshzj (approved):
Non-blocking comments but best addressed before merging.

src/utils/growthbook-utils.ts
kishore03109: @dcshzj wah bro that was my initial approach. but i was worried that eng in the future might toggle gb to be OFF -> but then CM is still enabled cuz if the feature toggle off -> CM still on cuz null defaults to ON 
I ok to j leave a comment in gb about this unintuitive behaviour, i do have a preference for it being dafult on. Thoughts

dcshzj: Ah I see okay yeah that makes sense, I think probably won't need a comment here but maybe can put inside growthbook instead?

PR #1036, opened 2023-11-15
fix(siteCreate): add redirect rules

kishore03109:
## Problem

Stg lite never configure properly, the url not correct + the redirect rules were not present + for reasons I do not know why there were blips when some of the sites had their jobs queued + it did not start even after a while. As such am pinging Amplify to release the jobs. Not sure if this completely fixes it, not sure why we have this issue now, so intent here is to add the release job + monitor for tickets. 

## Tests

running this in server.js 
```

deploymentsService.createAmplifyAppOnAws(
  "kishore-test-dev-gh",
  "kishore-test-again-delete-me",
  false
)

deploymentsService.createAmplifyAppOnAws(
  "kishore-test-dev-gh",
  "kishore-test-again-delete-me-staging-lite",
  true
)

```

note that there will be two apps being created in amplify. Staging lite will have the redirect rules for images + files, while the main app will have just the main 404 redirect. 

Stg lite
Screenshot 2023-11-16 at 9 37 51 AM
Screenshot 2023-11-16 at 9 37 45 AM

Stg
Screenshot 2023-11-16 at 9 37 31 AM
Screenshot 2023-11-16 at 9 37 26 AM

when visiting stg lite, we should see the images as such : https://staging-lite.da33yn9c7fu8b.amplifyapp.com/images/favicon-isomer.ico

### DIFF
--- a/src/services/identity/DeploymentClient.ts
+++ b/src/services/identity/DeploymentClient.ts
@@ -79,17 +79,40 @@ class DeploymentClient {
 
   generateCreateAppInput = (
     repoName: string,
-    repoUrl: string
-  ): CreateAppCommandInput => ({
-    name: repoName,
-    accessToken: SYSTEM_GITHUB_TOKEN,
-    repository: repoUrl,
-    buildSpec: AMPLIFY_BUILD_SPEC,
-    environmentVariables: {
-      JEKYLL_ENV: "development",
-    },
-    customRules: [{ source: "/<*>", target: "/404.html", status: "404" }],
-  })
+    repoUrl: string,
+    isStagingLite: boolean
+  ): CreateAppCommandInput => {
+    const stgLiteRedirectRules = [
+      {
+        source: "/files/<*>",
+        target: `https://raw.githubusercontent.com/isomerpages/${repoName}/staging/files/<*>`,
+        status: "200",
+      },
+      {
+        source: "/images/<*>",
+        target: `https://raw.githubusercontent.com/isomerpages/${repoName}/staging/images/<*>`,
+        status: "200",
+      },
+    ]
+    const defaultRedirectRules = [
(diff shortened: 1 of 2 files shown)


src/services/identity/DeploymentClient.ts
```diff
@@ -13,6 +13,9 @@ import {
   ListJobsCommand,
```
harishv7, line 13: does this mean we need to re-run this on all the quickiefied apps so far? they don't have redirects present?

src/services/identity/DeploymentsService.ts
```diff
@@ -152,17 +154,38 @@ class DeploymentsService {
...
+              branchName: "master",
+              jobType: "RELEASE",
+            })
+          )
           .map(() => amplifyInfo)
           .andThen(() =>
             this.deploymentClient
               .sendCreateBranch(createStagingBranchInput)
+              .andThen(() =>
+                this.deploymentClient.sendStartJobCommand({
```
harishv7, line 183: how is it working currently without this btw?

kishore03109: I am not sure why this doesnt work all the time

src/services/identity/DeploymentClient.ts
kishore03109: yes this would be the case, will be looking at all the apps for the past mth to change redirect rules

harishv7 (approved):

harishv7:
Verified as per above steps, cc  @kishore03109

harishv7:
Verified that the apps are created, imgs loading (e.g. https://staging-lite.da2l10qaui7tv.amplifyapp.com/images/favicon-isomer.ico) 
cc @kishore03109

PR #1043, opened 2023-12-03
chore(quickie): delete quickie for gh

kishore03109:
## Problem

Actual removal of the no longer needed `GithubCommitService` file to reduce complexity + fix test cases.


src/services/db/GitHubService.ts
```diff
@@ -241,10 +235,7 @@ export default class GitHubService {
 
   async readMedia(
     sessionData: UserWithSiteSessionData,
-    {
-      fileSha,
-      branchName = STAGING_BRANCH,
-    }: { fileSha: any; branchName?: string }
+    { fileSha }: { fileSha: any; branchName?: string }
```
seaerchin, line 238: nit: wrong typing - remove `branchName`! could we also stricten `fileSha` to `string`?

src/services/db/GitHubService.ts
```diff
@@ -209,18 +207,14 @@ export default class GitHubService {
 
   async read(
     sessionData: UserWithSiteSessionData,
-    {
-      fileName,
-      directoryName,
-      branchName = STAGING_BRANCH,
-    }: { fileName: any; directoryName: any; branchName?: string }
+    { fileName, directoryName }: { fileName: any; directoryName: any }
```
seaerchin, line 210: nit: could we stricten to `string`? i'm ok w leaving as-is and having another PR tidying types (similarly for type related nits gg forward in this PR)

src/services/db/GitHubService.ts
```diff
@@ -379,24 +370,21 @@ export default class GitHubService {
       sha,
       fileName,
       directoryName,
-      branchName = STAGING_BRANCH,
     }: {
```
seaerchin, line 373: just a sanity check (feel free to mark as resolved) - all call sites either ignore (default to `STAGING_BRANCH`) or pass in `STAGING_BRANCH` explicitly right? otherwise there'll be a regression

src/services/db/GitHubService.ts
```diff
@@ -767,11 +755,13 @@ export default class GitHubService {
...
     sessionData: UserWithSiteSessionData,
-    { commitSha }: { commitSha: any }
+    { commitSha, branchName }: { commitSha: any; branchName?: string }
   ) {
     const { accessToken } = sessionData
     const { siteName } = sessionData
-    const refEndpoint = `${siteName}/git/refs/heads/${STAGING_BRANCH}`
+    const refEndpoint = `${siteName}/git/refs/heads/${
+      branchName || STAGING_BRANCH
+    }`
```
seaerchin, line 764: i might be lacking context here but it seems abit weird to me to see `branchName` allowed as an argument here when we're explicitly removing it elsewhere (cos we know the branch should be `STAGING_BRANCH`) - is this because there are some call sites that need the branch to be different? (a merge to `master` perhaps)

seaerchin (commented):
only looked at code changes and skimmed tests - not approving first because of sanity checks. code wise lgtm

src/services/db/GitHubService.ts
kishore03109: Yes, this is the case! Specifically we have a `RepoManagementService` that calls on this branch name, and thus this API should be preserved

src/services/db/GitHubService.ts
kishore03109: good catch, removed

src/services/db/GitHubService.ts
kishore03109: I change those that can easily change ah (ie change + ts dont complain) will deprior those that require a change in control flow for now

kishore03109:
@seaerchin 
> just a sanity check (feel free to mark as resolved) - all call sites either ignore (default to STAGING_BRANCH) or pass in STAGING_BRANCH explicitly right? otherwise there'll be a regression

Agreed, looked through all the call sites and found one pending `getTree` that still maintains the branchName as a arg but actually we dont need it already. 

For the rest, the way i check is if the fun exists in `RepoService`, then make sure the call are compatible. 
There exists some calls of func A in `RepoService` that calls func B in `GithubService` by using super (eg. `readMediaFile`. For those i check if the call sites are compatible. 
^ reason for verbalising this is cuz yes quite a bit of code change, but i think this is sufficient to prevent regression, thoughts?

src/services/db/GitHubService.ts
seaerchin: abit confused here - you mean that this method **itself** has a call site that does not ignore it? if this is the case, we can't remove the `branchName` argument here right

kishore03109: arh i misunderstood this. but yes, from scanning all call sites dont pass in the staging arg, so we default to staging

src/services/db/GitHubService.ts
seaerchin: thanks for checking - seems like we (isomer admins) use it to reset branches so it should still be `staging`. that being said, it's ok to leave as-is cos this is hard to tell for sure + the added flexibility of resetting a diff branch might be useful for `staging/staging-lite/master`

seaerchin (approved):
lgtm

kishore03109:
### Merge activity

* **Dec 6, 5:51 AM**: @kishore03109 started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/isomerpages/isomercms-backend/1043).

PR #1044, opened 2023-12-03
fix(rollback handler): convert to ts for safety

kishore03109:
Now that there exists commits to 2 different branches, there needs to be a rollback handler to cater for the case of network failures that exist for a commit to 1 branch. else, we might have a scary situation whereby staging-lite updates, but staging doesnt, and this leads to a wrongly output production site. this effectively means that there should be any use of the `writeroutehandler` since any write can lead to deviation of staging and staging lite

Note that we dont need to do any thing with regards checking if a site is whitelisted for quickie since all sites have the staging lite branch infra already set up. the whitelisting only dictates whether or not the staging lite file gets updated or not.

Previously, we were not using the rollback properly, this pr converts the file to ts to allow for easier type checking. 
there were also some weird errors that occured that could be avoided with the retry. however, staging lite needs to be pushed with a `.push(gitOptions)`, which the retry mechanism does not have.  as a bandage, retry twice with the options, check in with @harishv7 on why there exists a retry in `GitFileSystemService` in the first place.

## Manual tests for rollback handler 
[Testing for quickie whitelisted site that is GGs] 
 - [ ] use a email login site, ensure that that it is whitelisted for build time reduction in growthbook
 - [ ] create a file, ensure that no error are obsevered
 - [ ] go into the `create` function of `GitFileSystemService` and short circuit the function by adding a early `// return errAsync(new ConflictError("this is a test failure"))` eg.
```
create(
    repoName: string,
    userId: string,
    content: string,
    directoryName: string,
    fileName: string,
    encoding: "utf-8" | "base64" = "utf-8",
    branchName: string
  ): ResultAsync<
    GitCommitResult,
    ConflictError | GitFileSystemError | NotFoundError
  > {
    // short circuit for testing
    return errAsync(new ConflictError("this is a test failure"))
... // rest of code
}
```
- [ ] re-run the create operation, notice that all rollbacks occur for both staging + staging-lite.

[Testing for non-quickie whitelisted site that is GGs] 
- [ ] do the same as above but not quickie should not be whitelisted 

[Testing for GH sites for rollback handler] (all gh sites are NOT whitelisted for quickie) 
- [ ] throw error in the create function for `GithubService` 
- [ ] notice that rollback works as intended -> it goes back to the last known state sucessfully.

[Testing normal crud operations]
## Tests
In staging environment,
`kishore-test` is qucikie + `harish-test` is NON-quickie + `kishore-test-dev-gh` on github (non qucikie) 
- [ ] rename of folders with pages inside 
- [ ] changes made to edit a page


src/middleware/routeHandler.ts
```diff
@@ -0,0 +1,250 @@
...
+      originalStagingCommitSha = currentStgCommitSha
+      originalStagingLiteCommitSha = currentStgLiteCommitSha
+    } catch (err) {
+      await unlock(siteName)
+      next(err)
+      return
+    }
+  }
+
+  await routeHandler(req, res, next).catch(async (err: any) => {
```
harishv7, line 181: we converted to a ts file but I still see widespread use of any in this file. Was the type check meant specifically for some parts only?

src/middleware/routeHandler.ts
```diff
@@ -0,0 +1,250 @@
...
+    try {
+      if (shouldUseGitFileSystem) {
+        await backOff(() =>
+          convertNeverThrowToPromise(
```
harishv7, line 185: why must convert to promise?

src/middleware/routeHandler.ts
```diff
@@ -0,0 +1,250 @@
...
+            gitFileSystemService.rollback(
+              siteName,
+              originalStagingCommitSha,
+              STAGING_BRANCH
+            )
+          )
+        )
+
+        await backOff(() =>
```
harishv7, line 194: how many times does this retry?

harishv7, line 194: how does this know to retry? based on if the wrapped function throws or not?

src/middleware/routeHandler.ts
```diff
@@ -0,0 +1,250 @@
+import { backOff } from "exponential-backoff"
```
harishv7, line 1: regarding this:

> there needs to be a rollback handler to cater for the case of network failures that exist for a commit to 1 branch. else, we might have a scary situation whereby staging-lite updates, but staging doesnt, and this leads to a wrongly output production site

Correct me if otherwise, but we alr have a rollback mechanism right? the focus of thePR is to implement retry on in case rollback fails right?

src/middleware/routeHandler.ts
```diff
@@ -0,0 +1,250 @@
...
+import simpleGit from "simple-git"
```
harishv7, line 2: > we were not using the rollback properly,

What does "properly" mean?

src/routes/v2/authenticatedSites/contactUs.js
```diff
@@ -53,7 +53,7 @@ class ContactUsRouter {
     const router = express.Router({ mergeParams: true })
 
     router.get("/", attachReadRouteHandlerWrapper(this.readContactUs))
-    router.post("/", attachWriteRouteHandlerWrapper(this.updateContactUs))
+    router.post("/", attachRollbackRouteHandlerWrapper(this.updateContactUs))
```
harishv7, line 56: how do we choose bet using `attachWriteRouteHandlerWrapper` and `attachRollbackRouteHandlerWrapper`?

src/middleware/routeHandler.ts
```diff
@@ -0,0 +1,250 @@
...
+
```
harishv7, line 3: > check in with @harishv7 on why there exists a retry in GitFileSystemService in the first place.

which line is this?

src/middleware/routeHandler.ts
kishore03109: yes the point of this is not to make everything type safe. there were some obvious errors (esp with the use of neverthrow) which lead the logic not being correct. 
tldr; the intent for this pr is a scoped marginal improvement

src/middleware/routeHandler.ts
kishore03109: because the library that we are using (backoff) expects a promise 
this is impt becuase the existence of the throw -> allows for the retry.
never throw never throws -> no retry occurs

src/middleware/routeHandler.ts
kishore03109: max default 10

kishore03109: yes

src/middleware/routeHandler.ts
kishore03109: > but we alr have a rollback mechanism right

this is the rollback mechanism, but it wasnt working as intended for some call sites since never throw never throws the error -> leading to the the rollback not occuring as intended.

src/middleware/routeHandler.ts
kishore03109: simply put, we were not giving the correct types that `backoff` expected

src/routes/v2/authenticatedSites/contactUs.js
kishore03109: attachWriteRouteHandlerWrapper -> single write only operation 

attachRollbackRouteHandlerWrapper -> multiple write operations 

reason for this change is becuase now iwth quickie, every "write" operation is two fold, and as such it follows the latter definition rather than the former

src/services/db/GitFileSystemService.ts
```diff
@@ -539,6 +539,29 @@ export default class GitFileSystemService {
         )
         .orElse(() =>
           // Retry push once
+          ResultAsync.fromPromise(
```
kishore03109, line 542: @harishv7 ps should have been clearer, specificly here, why do we have a retry here ah? + why is it on the second retry we pass in diferent args?

harishv7: Oh this is just cos in case the first attempt fails, we have a backup to retry once more.

regarding the options, I think there may be a bug - we should be passing in the same options both for first & 2nd attempts

kishore03109: do you know why it (i think consistently) fails on file rename?

src/middleware/routeHandler.ts
harishv7: 1st attempt
200ms
2nd
400ms
3rd
800ms
4th
1600ms 
5th
3200ms
7th
6400ms

and so on right? is the above backoff sequence correct?

was just thinking in a worst case scenario the request will be open for quite long? will this timeout on client?

kishore03109: would need to dive into source code here to get the exact algo (https://github.com/coveooss/exponential-backoff), but based on the doc i would assume that would be the case

wait ah this backoff occurs when the call has an error bah, so the FE would have already gotten an error code that we try to fix via exponential backoff

src/middleware/routeHandler.ts
```diff
@@ -0,0 +1,250 @@
...
+})
+const gitFileSystemService = new GitFileSystemService(simpleGitInstance)
+
+const handleGitFileLock = async (
+  repoName: string,
+  next: (arg0: any) => void
+) => {
+  const result = await gitFileSystemService.hasGitFileLock(repoName, true)
+  if (result.isErr()) {
+    next(result.error)
```
seaerchin, line 35: that bugfix :monkas:

src/middleware/routeHandler.ts
harishv7: hmm this is a handler middleware right? so it will completed through the execution of all handlers before request completes right?

src/services/db/GitFileSystemService.ts
harishv7: > do you know why it (i think consistently) fails on file rename?

hmm, consistent failure on 1st try is not an expected behaviour

src/middleware/routeHandler.ts
kishore03109: for clarity ya, this is in the .catch() control flow, would this not mean that this occurs after a certain error short circuits the entire chain of executions?

src/middleware/routeHandler.ts
```diff
@@ -0,0 +1,250 @@
...
+  STAGING_LITE_BRANCH,
+} from "@constants/constants"
+
+import { FEATURE_FLAGS } from "@root/constants/featureFlags"
+import LockedError from "@root/errors/LockedError"
+import logger from "@root/logger/logger"
+import convertNeverThrowToPromise from "@root/utils/neverthrow"
+import GitFileSystemService from "@services/db/GitFileSystemService"
+
+const BRANCH_REF = config.get("github.branchRef")
```
seaerchin, line 23: nit: this appears to be unused - perhaps we should remove this?

src/middleware/routeHandler.ts
```diff
@@ -0,0 +1,250 @@
...
+  let originalStagingLiteCommitSha: any
+
+  if (shouldUseGitFileSystem) {
+    const results = await Promise.all([
+      gitFileSystemService.getLatestCommitOfBranch(siteName, STAGING_BRANCH),
+      gitFileSystemService.getLatestCommitOfBranch(
+        siteName,
+        STAGING_LITE_BRANCH
+      ),
+    ])
```
seaerchin, line 138: because this returns result async, which wraps the promise underneath, i'm not entirely sure how `Promise.all` (a native js construct) interacts with this.

`neverthrow` exports `ResultAsync.combine` which does the same thing, but for `ResultAsync` - perhaps we should use that instead?

src/middleware/routeHandler.ts
```diff
@@ -0,0 +1,250 @@
...
+      await unlock(siteName)
+      if (stagingResult.isErr()) next(stagingResult.error)
+      if (stagingLiteResult.isErr()) next(stagingLiteResult.error)
+      return
+    }
+    originalStagingCommitSha = stagingResult.value.sha
+    originalStagingLiteCommitSha = stagingLiteResult.value.sha
+    if (!originalStagingCommitSha || !originalStagingLiteCommitSha) {
+      await unlock(siteName)
+      if (stagingResult.isErr()) next(stagingResult.error)
```
seaerchin, line 151: this should never hit because we will do an early return previously. maybe we can pass a string or something of the same shape if `next` does work on the `error` itself.

src/middleware/routeHandler.ts
kishore03109: this is copy-pasted code :sadge:

src/middleware/routeHandler.ts
```diff
@@ -0,0 +1,250 @@
...
+  } else {
+    try {
+      const {
+        currentCommitSha: currentStgCommitSha,
+        treeSha: stgTreeSha,
+      } = await getCommitAndTreeSha(siteName, accessToken, STAGING_BRANCH)
+
+      const {
+        currentCommitSha: currentStgLiteCommitSha,
+      } = await getCommitAndTreeSha(siteName, accessToken, STAGING_LITE_BRANCH)
```
seaerchin, line 168: would we ever need this as this implies they're not on efs (they're not using git file system)?

src/middleware/routeHandler.ts
```diff
@@ -0,0 +1,250 @@
...
+          let pushRes = gitFileSystemService.push(
+            siteName,
+            STAGING_BRANCH,
+            true
+          )
+          if (originalStagingLiteCommitSha) {
+            pushRes = pushRes.andThen(() =>
+              gitFileSystemService.push(siteName, STAGING_LITE_BRANCH, true)
+            )
+          }
```
seaerchin: given that this is taken mostly as-is, it's ok to keep with the existing structure. just thinking out loud here, no real action expected on your end - why wouldn't we chain the results together then do a single `backOff` at the end?

seaerchin: there might be semantic differences and i'm not entirely sure the existing impl is correct here but given that the structure is kept, i'm not too keen on rewriting this now.

src/middleware/routeHandler.ts
harishv7: follow up suggestion, option 1 + alarm if it fails for 5 times as failing 5 times signifies some deeper issue not resolvable with retry

src/middleware/routeHandler.ts
```diff
@@ -0,0 +1,262 @@
...
+              originalStagingLiteCommitSha,
+              siteName,
+              accessToken,
+              STAGING_LITE_BRANCH
+            ),
+          backoffOptions
+        )
+      }
+    } catch (retryErr) {
+      await unlock(siteName)
```
seaerchin, line 250: tbvh this whole chunk is very jank and confusing but this PR doesn't focus on code clarity but migration from js -> ts so ok with it. we should, however, look into probably refactoring this into a sensible form.

src/middleware/routeHandler.ts
kishore03109: sure option 1 [implemented](https://github.com/isomerpages/isomercms-backend/pull/1044/commits/f626ef7dd6ad56130ff3dfe6191878f2538bf72a)

src/routes/v2/authenticatedSites/settings.js
```diff
@@ -136,7 +135,7 @@ class SettingsRouter {
     router.post(
       "/repo-password",
       this.authorizationMiddleware.verifyIsEmailUser,
-      attachWriteRouteHandlerWrapper(this.updateRepoPassword)
+      attachReadRouteHandlerWrapper(this.updateRepoPassword)
```
seaerchin, line 138: shouldn't updating a repo's password be a write op?

src/services/db/GitFileSystemService.ts
```diff
@@ -542,10 +542,33 @@ export default class GitFileSystemService {
             isForce
               ? this.git
                   .cwd({ path: `${efsVolPath}/${repoName}`, root: false })
-                  .push(["--force"])
+                  .push([...gitOptions, "--force"])
```
seaerchin, line 545: why the change here?

harishv7 (approved):
@kishore03109 can add a follow up minor todo - add alarm if 5 retries all fail

src/services/db/GitFileSystemService.ts
```diff
@@ -542,10 +542,33 @@ export default class GitFileSystemService {
...
+                return new GitFileSystemError(
+                  "Unable to push latest changes of repo"
+                )
+              }
+
+              return new GitFileSystemError("An unknown error occurred")
+            }
+          )
+        )
+        .orElse(() =>
```
seaerchin, line 562: in order for this to work, should we check that the previous failure was over the network (ie, github doesn't see the commit)? otherwise, we might have a failure from a non-github issue -> push -> dup commits or fail, isn't it?

src/utils/neverthrow.ts
```diff
@@ -0,0 +1,16 @@
+import { ResultAsync } from "neverthrow"
+
+/**
+ * This function is only used when integrating with third party libraries that
+ * expect a .catch() method on the returned promise. This should not be used in most
+ * control flows as it removes the benefits that neverthrow provides.
+ */
+const convertNeverThrowToPromise = <T, E>(x: ResultAsync<T, E>): Promise<T> =>
```
seaerchin, line 8: does `x._unsafeUnwrap` work here?

seaerchin (commented):
focused mostly on ensuring no regression; can't be fully sure but kinda sure. 

we should aim to fix low-hanging fruits here + checking the scope to see that the changes made on API are safe

src/middleware/routeHandler.ts
kishore03109: hmm sure about alarm? worried about numberous false positives tho
What I can do is add logging for this, can review later if this occurs too freq with false positives. will add todo regardless in jira

src/utils/neverthrow.ts
seaerchin: just tried, the below snippet works!

```ts
  const res = await x
  return res._unsafeUnwrap()

```

src/middleware/routeHandler.ts
harishv7: okay

harishv7: @kishore03109 pls test once on staging/local

src/utils/neverthrow.ts
kishore03109: done [097f1f4](https://github.com/isomerpages/isomercms-backend/pull/1044/commits/097f1f47d81f1c92cd5b915262eea832d0c26e7f)

src/services/db/GitFileSystemService.ts
kishore03109: after disc resolving this as an edge case

src/services/db/GitFileSystemService.ts
kishore03109: pushes are supposed to have the required options

src/routes/v2/authenticatedSites/settings.js
kishore03109: [79cdf6a](https://github.com/isomerpages/isomercms-backend/pull/1044/commits/79cdf6a57e0888d1011884f852ef834e562605ca)

src/middleware/routeHandler.ts
kishore03109: [79cdf6a](https://github.com/isomerpages/isomercms-backend/pull/1044/commits/79cdf6a57e0888d1011884f852ef834e562605ca)

src/middleware/routeHandler.ts
```diff
@@ -0,0 +1,250 @@
...
+
+  const { accessToken } = userSessionData
+  const { growthbook } = req
+
+  const shouldUseGitFileSystem = !!growthbook?.getFeatureValue(
+    FEATURE_FLAGS.IS_GGS_ENABLED,
+    false
+  )
+
+  const isGitAvailable = await handleGitFileLock(siteName, next)
```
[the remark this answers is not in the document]
kishore03109: [79cdf6a](https://github.com/isomerpages/isomercms-backend/pull/1044/commits/79cdf6a57e0888d1011884f852ef834e562605ca)

src/middleware/routeHandler.ts
```diff
@@ -0,0 +1,250 @@
...
+  if (!isGitAvailable) return
+  try {
+    await lock(siteName)
+  } catch (err) {
+    next(err)
+    return
+  }
+
+  let originalStagingCommitSha: any
+  let originalStagingLiteCommitSha: any
```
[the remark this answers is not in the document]
kishore03109: [79cdf6a](https://github.com/isomerpages/isomercms-backend/pull/1044/commits/79cdf6a57e0888d1011884f852ef834e562605ca)

src/middleware/routeHandler.ts
```diff
@@ -0,0 +1,250 @@
...
+const simpleGitInstance = simpleGit({
+  maxConcurrentProcesses: MAX_CONCURRENT_GIT_PROCESSES,
+})
+const gitFileSystemService = new GitFileSystemService(simpleGitInstance)
+
+const handleGitFileLock = async (
+  repoName: string,
+  next: (arg0: any) => void
+) => {
+  const result = await gitFileSystemService.hasGitFileLock(repoName, true)
```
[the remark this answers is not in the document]
kishore03109: done

kishore03109:
### Merge activity

* **Dec 6, 5:51 AM**: @kishore03109 started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/isomerpages/isomercms-backend/1044).

PR #1045, opened 2023-12-03
test(githubService): add tests

kishore03109:
adding test cases for the two new functions `renameSinglePath` and `deleteDirectory` introduced into `GithubService` in #1042. Note that these functions already existed, but this pr adds the test cases to the right file.

PR #1046, opened 2023-12-03
feat(gitFileSystem): safer api

kishore03109:
## Problem

1. the presence of default arguments lead to sneaky bugs introduced due to caller unknowingly calling the wrong branch. 
2. there is a scary instance whereby the lack of update to staging-lite could occur, but the update to staging does not, leading to issues downstream. 

Closes [insert issue #]

## Solution
1. to keep things simple, we make all the args compulsory.
2. make staging update first before updating staging lite.

### DIFF
--- a/src/services/db/GitFileSystemService.ts
+++ b/src/services/db/GitFileSystemService.ts
@@ -250,7 +250,6 @@ export default class GitFileSystemService {
         logger.error(
           `Error when getting Git blob hash: ${error} when trying to access ${efsVolPath}/${repoName}`
         )
-        console.log(`revparse: HEAD:${filePath} for ${efsVolPath}/${repoName}`)
 
         if (error instanceof GitError) {
           return new GitFileSystemError("Unable to determine Git blob hash")


kishore03109:
* develop
  * **PR #1042** <a href="https://app.graphite.dev/github/pr/isomerpages/isomercms-backend/1042?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #1043** <a href="https://app.graphite.dev/github/pr/isomerpages/isomercms-backend/1043?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #1044** <a href="https://app.graphite.dev/github/pr/isomerpages/isomercms-backend/1044?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
        * **PR #1045** <a href="https://app.graphite.dev/github/pr/isomerpages/isomercms-backend/1045?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
          * **PR #1046** <a href="https://app.graphite.dev/github/pr/isomerpages/isomercms-backend/1046?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/isomerpages/isomercms-backend/1046?utm_source=stack-comment).

src/services/db/GitFileCommitService.ts
```diff
@@ -36,51 +56,43 @@ export default class GitFileCommitService {
...
     }
 
-    this.gitFileSystemService.push(sessionData.siteName, STAGING_BRANCH)
-    if (shouldUpdateStagingLite) {
-      this.gitFileSystemService.push(
-        sessionData.siteName,
-        this.STAGING_LITE_BRANCH
-      )
-    }
+    this.pushToGithub(sessionData, shouldUpdateStagingLite)
```
seaerchin, line 95: this can fail silently - we should probably just chain off this. alternatively, what we can do is just chain off the results sequentially and throw an error at the end (the behaviour now), which makes things clearer to read. 

also eliminates the issue of `staging-lite` updating but not `staging` cos if the initial call to `staging` fails, `staging-lite` won't even run

src/services/db/GitFileCommitService.ts
```diff
@@ -101,51 +113,43 @@ export default class GitFileCommitService {
...
     }
 
-    this.gitFileSystemService.push(sessionData.siteName, defaultBranch)
-    if (shouldUpdateStagingLite) {
-      this.gitFileSystemService.push(
-        sessionData.siteName,
-        this.STAGING_LITE_BRANCH
-      )
-    }
+    this.pushToGithub(sessionData, shouldUpdateStagingLite)
```
seaerchin, line 152: same issue as above; there's also a repeated pattern - we could probably just separate out the steps as functions and chain them together.

src/services/db/GitFileCommitService.ts
seaerchin, line 1: this file has alot of the same pattern of do some change -> check the result of that change -> update staging lite -> update github -> throw error if any surface.

we could probably separate those out and make this whole file clearer. that being said, we don't have to do it _now_. 

however, there is an issue where updates are allowed to silently fail at the push step. because our call returns a `result`, it won't throw by design and we need to check if it's recoverable or we should just fail fast (throw error).

src/services/db/GitFileSystemService.ts
```diff
@@ -248,6 +250,7 @@ export default class GitFileSystemService {
         logger.error(
           `Error when getting Git blob hash: ${error} when trying to access ${efsVolPath}/${repoName}`
         )
+        console.log(`revparse: HEAD:${filePath} for ${efsVolPath}/${repoName}`)
```
seaerchin, line 253: nit: could we remove this `console.log`?

seaerchin (commented):
the changes are fine, albeit repetitive. similar work should probably be abstracted out as a function to avoid slight behavioural drift over time that creates bugs. this is a maintainability issue but it's not too bad for now.

otoh, we have a possibility of a silent failure, where we commit to EFS but fail the push to github. this could potentially be very confusing and we could fail fast here or attempt to handle the error properly. no pref, as long as the error itself is handled.

src/services/db/GitFileCommitService.ts
kishore03109: Hey this was an intentional trade off we made as a team not to await for calls for they are expensive. But I do air your concerns regarding silent failures, this is the cause of the divergence that we are seeing. Dont intend to tackle it here!

For info only + outside of scope of pr: I tried adding an alarm for this, but it was TOO noisy, there is prob a way to make this such that we only throw alarms for this silent pushes, but havnt had time to invetigate

kishore03109:
@seaerchin

> the changes are fine, albeit repetitive. similar work should probably be abstracted out as a function to avoid slight behavioural drift over time that creates bugs. this is a maintainability issue but it's not too bad for now.
 
I am quite curious how you would have structured this instead (but not today kek release first), maybe i book time w u after i come back to learn from you

>otoh, we have a possibility of a silent failure, where we commit to EFS but fail the push to github. this could potentially be very confusing and we could fail fast here or attempt to handle the error properly. no pref, as long as the error itself is handled.

I do think this is outside of the scope of this pr, since this is parity to what currently exists in production! thoughts?

src/services/db/GitFileCommitService.ts
```diff
@@ -14,14 +14,34 @@ import GitFileSystemService from "./GitFileSystemService"
...
+
+    if (shouldUpdateStagingLite) {
+      res.andThen(() =>
+        this.gitFileSystemService.push(
+          sessionData.siteName,
+          STAGING_LITE_BRANCH
+        )
+      )
+    }
+    await res
```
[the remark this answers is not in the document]
kishore03109: can

src/services/db/GitFileCommitService.ts
seaerchin: we have a few different ways to tackle this tbh, just spitballing ideas - not an ask to implement. 

1. separate out the push to github and let the router handle this. the router can always send response early then fail later (no cost to user) if the push to github fails. 
2. because `neverthrow` is safe, we could just write a wrapper for retrying `Errors<T>`
3. just send an email on failure 

we probably want to decide what categories of errors we care about - transient errors vs persistent errors; in terms of our usage, transient is probably outages <10 mins in length (we don't have that much pushes tbvh) and persistent >=10. 

if transient, retry w/ exponential backoff will work well; if persistent, we probably want a notif + manual fix. to be honest, the divergence should arise only if we edit on github directly whilst there's a push failure so should be minimal but the problem is we got high barrier of entry to edit on efs (hence why we edit on github directly).

seaerchin:
> I am quite curious how you would have structured this instead (but not today kek release first), maybe i book time w u after i come back to learn from you

ye sure

> I do think this is outside of the scope of this pr, since this is parity to what currently exists in production! thoughts?
if there's no easy win here then sure, we can just go with this

src/services/db/GitFileSystemService.ts
seaerchin: @kishore03109 pls remove thx

seaerchin (approved):
lgtm if no quick way of feeding back failures to us. 

do remove extra console log before merge.

kishore03109:
### Merge activity

* **Dec 6, 5:51 AM**: @kishore03109 started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/isomerpages/isomercms-backend/1046).

PR #1048, opened 2023-12-04
hotfix(repair-form): set remote url correctly

kishore03109:
## Problem

The remore was not set properly. Reason why this error was not caught eariler was beacuse git itself is tolerant to this error, but our BE logic is not as we check a strict string check. 
Closes [insert issue #]

## Solution

add .git at the end


seaerchin (commented):
lgtm - technically this should work (and i checked against the git ui), but just to be safe: 

1. how does this error manifest?
2. how did we fix the error? (assumption is just go on efs and fix error manually?)

kishore03109:
@seaerchin yes gh handles this gracefully. In our rollback handlers however, we do a strict string check in the function call `isRemoteOriginCorrect`, which was how i discovered that error

seaerchin:
eh dk why i didn't approve cos i clicked approve sia cb

seaerchin (approved):

PR #1052, opened 2023-12-12
IS-802 bugs with rollback

harishv7:
Fix multiple issues with rollback, locking, logging and other errors


seaerchin (approved):
Lgtm, as follow-up, let's look into cleaning up the various git services for consistency.

PR #1058, opened 2023-12-14
ref(media): upgrade media router to typescript

dcshzj:
## Problem

The media router is in JavaScript.

## Solution

**Breaking Changes**

**Improvements**:

- Upgraded the two media routers to TypeScript.
- Also combined MediaFilesRouter and MediaCategoriesRouter into a single MediaRouter. Having such a distinction can be confusing at times, especially when dealing with bulk actions.

**No logic changes were made, only conversion into TypeScript.**

## Tests

Unit tests.

## Deploy Notes

*None*


src/routes/v2/authenticatedSites/media.ts
```diff
@@ -0,0 +1,419 @@
...
+    const createResp = await this.mediaDirectoryService.createMediaDirectory(
+      userWithSiteSessionData,
+      githubSessionData,
+      {
+        directoryName: newDirectoryName,
+        objArray: items,
+      }
+    )
+
+    return res.status(200).json(createResp)
```
seaerchin, line 160: why did we omit the `next()` call here?

src/routes/v2/authenticatedSites/media.ts
```diff
@@ -0,0 +1,419 @@
...
+    await this.mediaDirectoryService.renameMediaDirectory(
+      userWithSiteSessionData,
+      githubSessionData,
+      {
+        directoryName,
+        newDirectoryName,
+      }
+    )
+
+    return res.status(200).send("OK")
```
seaerchin, line 190: same here on omission of `next`

seaerchin (approved):
have a few questions: 
1. noticed the omission of `next` broadly throughout the file - i assuem this is cos the route handler is the final function in the stack so the `next` call is useless? (from what i can tell at the `getRouter` function
2. how did we determine the route params/return type? it seems fine but because the router uses the `service` under teh hood (which is still in `js`), we run the risk of our types being wrong without any signal from ts

i suspect the most value-add might come from shifting `mediaDirectoryService` and `mediaFileService` to ts to properly typecheck but for now this is ok

src/routes/v2/authenticatedSites/media.ts
dcshzj: The `next()` call as I understand it is to let express know to continue the execution by matching other routes, which is mainly used for middleware. In this case, we don't actually need this `next()` call since this is the final route and we want the execution to end here.

Not too sure why it was added in the past, probably a copy-paste from somewhere, but verified that this is not necessary (especially since we already send a status 200).

dcshzj:
Route params and return types were inferred from their usages, along with the schema validations that were already present in the JS files. I agree tho, this would be more complete when the underlying services are migrated to TS.

PR #1060, opened 2023-12-15
fix(media): filter first before getting git blob hash

dcshzj:
## Problem

Sites with a lot of images cannot load their images on the CMS.

## Solution

**Breaking Changes**

**Improvements**:

- Perform a filter first before getting the Git blob hash, which was found to be quite expensive for a large number of files.

**Bug Fixes**:

- The addedTime attribute was accidentally reverted earlier, fixed that so that the added times are now correct on EFS.
- Skip checking the `.git` folder, so we should have less error messages in the logs.

## Tests

Smoke test:
- [ ] Navigate to `moe-ast-moehc` with the repo on GGS (i.e. email-login)
- [ ] Attempt to load the images page
- [ ] The page should load fairly quickly

## Deploy Notes

*None*

### DIFF
--- a/src/services/db/GitHubService.ts
+++ b/src/services/db/GitHubService.ts
@@ -14,6 +14,7 @@ import { STAGING_BRANCH } from "@root/constants"
 import logger from "@root/logger/logger"
 import { GitCommitResult } from "@root/types/gitfilesystem"
 import { RawGitTreeEntry } from "@root/types/github"
+import { getPaginatedDirectoryContents } from "@root/utils/files"
 
 import * as ReviewApi from "./review"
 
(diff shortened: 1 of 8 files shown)


src/services/db/GitHubService.ts
```diff
@@ -14,6 +14,7 @@ import { STAGING_BRANCH } from "@root/constants"
 import logger from "@root/logger/logger"
 import { GitCommitResult } from "@root/types/gitfilesystem"
 import { RawGitTreeEntry } from "@root/types/github"
+import { getPaginatedDirectoryContents } from "@root/utils/files"
```
seaerchin, line 17: nit: can remove

seaerchin (commented):
lgtm. was wondering why we still force-fit the pagination stuff into `listDirectoryContents` given that they have differing goals (and resulted in `listDirectoryContent` having a huge API but default arguments help).

if we foresee the pagination stuff to **always** occur with listing the contents then ok, otherwise more in favour of breaking this up

dcshzj:
> if we foresee the pagination stuff to always occur with listing the contents then ok, otherwise more in favour of breaking this up

I don't have a good answer to this, so I have split the function up. Reading a media directory is the only time that needs pagination right now I believe, and it is also the only time that requires the Git blob hash.

seaerchin (approved):

PR #1063, opened 2023-12-18
IS-835 Removed all instances of any from GitHubService.ts

QiluXie:
## Problem

Closes [IS-835](https://isomeropengov.atlassian.net/browse/IS-835?atlOrigin=eyJpIjoiNGJiMzk4NTM4NmRlNGYxYmEzODU5MjY5MzY1YWQwOGYiLCJwIjoiaiJ9)

There are 13 instances of `any` in the file `GitHubService.ts`. The usage of `any` should be avoid if possible to maintain the type safety of Typescript. 

## Solution

Defined explicit types for all instances of `any` in `GitHubService.ts` 

**Breaking Changes**

  - Details ...

**Improvements**:

- Ensure type safety by replacing all instances of `any` with explicit types in `GitHubService.ts` file 

## Tests

Clone the branch and run backend locally to confirm there is no error from type checking

[IS-835]: https://isomeropengov.atlassian.net/browse/IS-835?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/services/db/GitHubService.ts
+++ b/src/services/db/GitHubService.ts
@@ -207,7 +207,10 @@ export default class GitHubService {
 
   async read(
     sessionData: UserWithSiteSessionData,
-    { fileName, directoryName }: { fileName: any; directoryName: any }
+    {
+      fileName,
+      directoryName,
+    }: { fileName: string; directoryName: string | undefined }
   ) {
     const { accessToken } = sessionData
     const { siteName } = sessionData
@@ -514,7 +517,7 @@ export default class GitHubService {
   async getTree(
     sessionData: UserWithSiteSessionData,
     githubSessionData: GithubSessionData,
-    { isRecursive }: any
+    { isRecursive }: { isRecursive: boolean }
   ): Promise<RawGitTreeEntry[]> {
     const { accessToken } = sessionData
     const { siteName } = sessionData
@@ -541,7 +544,7 @@ export default class GitHubService {
   async updateTree(
     sessionData: UserWithSiteSessionData,
     githubSessionData: GithubSessionData,
-    { gitTree, message }: { gitTree: any; message: any }
+    { gitTree, message }: { gitTree: unknown; message: unknown }
   ) {
     const { accessToken, siteName, isomerUserId: userId } = sessionData
     const { treeSha, currentCommitSha } = githubSessionData.getGithubState()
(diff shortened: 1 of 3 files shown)


src/services/db/GitHubService.ts
```diff
@@ -541,7 +544,7 @@ export default class GitHubService {
   async updateTree(
     sessionData: UserWithSiteSessionData,
     githubSessionData: GithubSessionData,
-    { gitTree, message }: { gitTree: any; message: any }
+    { gitTree, message }: { gitTree: unknown; message: unknown }
```
harishv7, line 547: are we able to get more specific than unknown here?

src/types/github.ts
```diff
@@ -97,7 +97,7 @@ export type RawGitTreeEntry = {
   path: string
   mode: string
   type: "tree" | "file"
-  sha: string
+  sha: string | null
```
harishv7, line 100: so sha is optional? would `sha?` work?

QiluXie: No, `sha?` doesn't work because there are places where `sha` is assigned specifically to `null`. 
```
 newGitTree.push({
            ...item,
            sha: null,
          })
```
If we do `sha? string` that's equivalent to `sha: string | undefined` which cannot be assigned to `null`.

src/services/db/GitHubService.ts
QiluXie: Good catch! Modified it in 8f15d84de12b5550a58fba2e0046a7a628c28d53

seaerchin:
fyi, i think there's a test spec being skipped due ot this typing change. could i trouble you to update the variable used in the test?

src/types/github.ts
harishv7: got it

harishv7 (approved):
lgtm, except the test changes @seaerchin has commented

QiluXie:
> fyi, i think there's a test spec being skipped due ot this typing change. could i trouble you to update the variable used in the test?
> 
> Screenshot 2023-12-19 at 1 26 24 PM

Thanks for the catch @seaerchin ! Removed unnecessary string declaration of `gitTree` since it's already defined in the upper scope. Done in 49eeebece8a352c996088cee64d025019bbc2e92
Can you take another look and approve before I merge it?

seaerchin:
> > fyi, i think there's a test spec being skipped due ot this typing change. could i trouble you to update the variable used in the test?
> > Screenshot 2023-12-19 at 1 26 24 PM
> 
> Thanks for the catch @seaerchin ! Removed unnecessary string declaration of `gitTree` since it's already defined in the upper scope. Done in [49eeebe](https://github.com/isomerpages/isomercms-backend/commit/49eeebece8a352c996088cee64d025019bbc2e92) Can you take another look and approve before I merge it?

ci is green, should be fine

PR #1065, opened 2023-12-19
fix(media): introduce delete multiple files endpoint

dcshzj:
> [!NOTE]
> Large PR due to unit tests, feel free to browse by commits

## Problem

The current delete file endpoint often encounters the repo lock, prevent users from deleting multiple files at once.

## Solution

**Breaking Changes**

**Improvements**:

- A new endpoint has been introduced to allow for deleting of multiple files in a single repo lock transaction.

## Tests

- Unit tests.
- Smoke tests (written on https://github.com/isomerpages/isomercms-frontend/pull/1753)

## Deploy Notes

*None*

### DIFF
--- a/src/services/db/GitFileCommitService.ts
+++ b/src/services/db/GitFileCommitService.ts
@@ -258,6 +258,49 @@ export default class GitFileCommitService {
     this.pushToGithub(sessionData, shouldUpdateStagingLite)
   }
 
+  async deleteMultipleFiles(
+    sessionData: UserWithSiteSessionData,
+    githubSessionData: GithubSessionData,
+    { items }: { items: Array<{ filePath: string; sha: string }> }
+  ): Promise<void> {
+    logger.info(
+      `Deleting multiple files in local Git file system for repo: ${sessionData.siteName}`
+    )
+
+    const stagingDeleteResult = await this.gitFileSystemService.deleteMultipleFiles(
+      sessionData.siteName,
+      items,
+      sessionData.isomerUserId,
+      STAGING_BRANCH
+    )
+
+    const shouldUpdateStagingLite =
+      isReduceBuildTimesWhitelistedRepo(sessionData.growthbook) &&
+      !items.some(({ filePath }) => isFileAsset({ directoryName: filePath }))
+
+    let stagingLiteDeleteResult
+    if (shouldUpdateStagingLite) {
+      stagingLiteDeleteResult = await this.gitFileSystemService.deleteMultipleFiles(
+        sessionData.siteName,
+        items,
+        sessionData.isomerUserId,
+        STAGING_LITE_BRANCH
+      )
+    }
+
+    if (stagingDeleteResult.isErr()) {
+      throw stagingDeleteResult.error
+    } else if (
+      shouldUpdateStagingLite &&
+      stagingLiteDeleteResult &&
+      stagingLiteDeleteResult.isErr()
+    ) {
+      throw stagingLiteDeleteResult.error
+    }
+
+    this.pushToGithub(sessionData, shouldUpdateStagingLite)
+  }
+
   async renameSinglePath(
     sessionData: UserWithSiteSessionData,
     githubSessionData: GithubSessionData,
--- a/src/services/db/GitFileSystemService.ts
+++ b/src/services/db/GitFileSystemService.ts
@@ -620,8 +620,6 @@ export default class GitFileSystemService {
         )
       }
 
-      // Note: We only accept commits that change 1 file at once (pathSpec.length == 1)
-      // Or commits that move/rename files (pathSpec.length == 2)
       if (pathSpec.length < 1) {
         return errAsync(
           new GitFileSystemError(
@@ -1289,7 +1287,7 @@ export default class GitFileSystemService {
             )
           )
         }
-        oldStateSha = latestCommit.sha as string
+        oldStateSha = latestCommit.sha
         return okAsync(true)
       })
       .andThen(() =>
--- a/src/services/db/GitFileSystemService.ts
+++ b/src/services/db/GitFileSystemService.ts
@@ -9,8 +9,6 @@ import {
   LogResult,
 } from "simple-git"
 
-import { config } from "@config/config"
-
 import logger from "@logger/logger"
 
 import { BadRequestError } from "@errors/BadRequestError"
@@ -624,10 +622,10 @@ export default class GitFileSystemService {
 
       // Note: We only accept commits that change 1 file at once (pathSpec.length == 1)
       // Or commits that move/rename files (pathSpec.length == 2)
-      if (pathSpec.length < 1 || pathSpec.length > 2) {
+      if (pathSpec.length < 1) {
         return errAsync(
           new GitFileSystemError(
-            `Invalid pathSpec length: ${pathSpec.length}. Expected 1 or 2`
+            `Invalid pathSpec length: ${pathSpec.length}. Expected 1 or more`
           )
         )
       }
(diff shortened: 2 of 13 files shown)


src/services/db/GitFileSystemService.ts
```diff
@@ -651,25 +649,32 @@ export default class GitFileSystemService {
...
 
-          return ResultAsync.fromPromise(
-            this.git
-              .cwd({ path: `${efsVolPath}/${repoName}`, root: false })
-              .add(pathSpec),
-            (error) => {
-              logger.error(
-                `Error when Git adding files to ${repoName}: ${error}`
-              )
+          // Note: We need to add files sequentially due to the Git lock
```
harishv7, line 652: why can't we do a `git add .`?

src/services/db/GitFileSystemService.ts
```diff
@@ -1236,6 +1241,131 @@ export default class GitFileSystemService {
...
+            new ConflictError(
+              "File has been changed recently, please try again"
+            )
+          )
+        }
+
+        return okAsync(itemsWithStats)
+      })
+      .andThen((itemsWithStats) =>
+        // Note: All deletions must be successful, otherwise we rollback
```
harishv7, line 1314: would this mean a single failure in a batch of 20 deletions will rollback all deletions? we don't support partial success of the request?

src/services/db/GitFileCommitService.ts
```diff
@@ -258,6 +258,49 @@ export default class GitFileCommitService {
...
+    )
+
+    const stagingDeleteResult = await this.gitFileSystemService.deleteMultipleFiles(
+      sessionData.siteName,
+      items,
+      sessionData.isomerUserId,
+      STAGING_BRANCH
+    )
+
+    const shouldUpdateStagingLite =
```
harishv7, line 277: staging lite doesn't contain images or files right? in that case why do we need to update stg lite branch?

src/services/db/GitFileSystemService.ts
dcshzj: We don't have sufficient guarantees that doing a `git add .` only adds the files that were indeed changed by the user. There may be other changes that were made on the repo (which may be there for unknown reasons) but not yet committed, we don't want to accidentally add them here.

src/services/db/GitFileSystemService.ts
dcshzj: This is more for simplicity. There can be many reasons why failure happens, so I decided to treat it as an atomic transaction instead of having some success some failure. On the frontend, we don't also have an easy way to show the user which files failed to delete, so I decided to go for getting the user to retry again.

src/services/db/GitFileCommitService.ts
dcshzj: We don't have that guarantee at this stage, because we only know the `filePath` and `sha`. Doing this is safer in case we somehow reuse this function to bulk delete pages in the future.

src/services/db/RepoService.ts
```diff
@@ -242,6 +243,10 @@ export default class RepoService extends GitHubService {
     )
     const { private: isPrivate } = await super.getRepoInfo(sessionData)
 
+    if (targetFile === undefined) {
```
kishore03109, line 246: perf: we can reduce a redundant `super.getRepoInfo(sessionData)` in this control flow by adding this line before `const { private: isPrivate } = await super.getRepoInfo(sessionData)`

src/services/fileServices/MdPageServices/__tests__/MediaFileService.spec.js
```diff
@@ -236,4 +237,40 @@ describe("Media File Service", () => {
       )
     })
   })
+
+  describe("DeleteMultipleFiles", () => {
+    it("rejects page names with special characters", async () => {
+      await expect(
+        service.deleteMultipleFiles(sessionData, mockGithubSessionData, {
+          items: [
+            { filePath: "file/file%%%name.pdf", sha },
```
kishore03109, line 246: was the point of this to check for encoding issues hahah

src/services/db/GitFileSystemService.ts
```diff
@@ -1236,6 +1241,131 @@ export default class GitFileSystemService {
...
+    return this.getLatestCommitOfBranch(repoName, branchName)
+      .andThen((latestCommit) => {
+        if (!latestCommit.sha) {
+          return errAsync(
+            new GitFileSystemError(
+              `Unable to find latest commit of repo: ${repoName} on branch "${branchName}"`
+            )
+          )
+        }
+        oldStateSha = latestCommit.sha as string
```
kishore03109, line 1264: nit: no need casing already since we have already declared in the func args

kishore03109 (approved):
mostly nits, please dismiss my review if my comment regarding the line `return result.isOk() && result.value` has a wrong assumption

src/services/db/GitFileSystemService.ts
```diff
@@ -1236,6 +1241,131 @@ export default class GitFileSystemService {
...
+            isStaging
+          ).andThen((sha) => {
+            if (sha !== itemStats.sha) {
+              return okAsync(false)
+            }
+
+            return okAsync(true)
+          })
+
+          return result.isOk() && result.value
```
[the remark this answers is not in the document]
dcshzj: Hmm yes, the `result.value` is the one that does the check for the unwrapped `ok`. Effectively there are 3 cases:
1. If an error occurred, `result.isOk()` is false, so the entire return is false.
2. If no error occurred but there is a sha mismatch, then result.value is false, so true && false becomes false.
3. If no error occurred and everything is good, then result.value is true, so true && true becomes true, which is what we want.

This is only specific for a single item, so it will loop through all the items before we determine the value of `isFileConflict`, which will then determine if we continue the function execution or not.

src/services/db/GitFileSystemService.ts
```diff
@@ -624,10 +622,10 @@ export default class GitFileSystemService {
 
       // Note: We only accept commits that change 1 file at once (pathSpec.length == 1)
       // Or commits that move/rename files (pathSpec.length == 2)
```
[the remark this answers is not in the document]
dcshzj: Oops thanks! Removed in ad2521fd.

src/services/db/RepoService.ts
dcshzj: Oh good idea, updated in ad2521fd.

src/services/db/GitFileSystemService.ts
dcshzj: Updated in ad2521fd!

src/services/fileServices/MdPageServices/__tests__/MediaFileService.spec.js
dcshzj: Oops not really haha, it was just a randomly chosen special character

src/services/db/GitFileSystemService.ts
```diff
@@ -1236,6 +1241,131 @@ export default class GitFileSystemService {
...
+            const deletePromise = itemStats.stats.isDirectory()
+              ? fs.promises.rm(
+                  `${efsVolPath}/${repoName}/${itemStats.filePath}`,
+                  {
+                    recursive: true,
+                    force: true,
+                  }
+                )
+              : fs.promises.rm(
+                  `${efsVolPath}/${repoName}/${itemStats.filePath}`
```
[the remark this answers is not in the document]
dcshzj: Hmm no good arguments for this other than just being cautious about it, basically not doing a force delete unless absolutely necessary (for directories), so we avoid masking potential errors.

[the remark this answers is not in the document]
kishore03109: I think the [-f flag ignores non-existent files and doesn't ask to prompt](https://man7.org/linux/man-pages/man1/rm.1.html), seems like safer here?

but is a nit so i dont feel strong about this too, let you make the final decision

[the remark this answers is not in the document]
dcshzj: I was thinking it ignores non-existent files but if at this stage we are deleting a non-existent file then we might have encountered some issue where someone else deleted the file, which might be grounds for a conflict :(

PR #1066, opened 2023-12-19
ref(ts migration): replaced all instances of any with explicit types in repoService.ts

QiluXie:
## Problem

Closes [IS-836](https://isomeropengov.atlassian.net/browse/IS-836?atlOrigin=eyJpIjoiM2JkYjcxYjAwZjkwNDdkNWI3NmQ5ZGE2NmY1ZTdkN2IiLCJwIjoiaiJ9)

There are a lot of instances of type `any` in the file `repoService.ts`, and we need to replace them with explicit types to ensure type safety. 

## Solution

Replaced all instances of type `any` in `repoService.ts`

**Breaking Changes**

  - Details ...

**Improvements**:

- Replaced all instances of type `any` in `repoService.ts`

## Tests

CI test to confirm no error in type checking 

[IS-836]: https://isomeropengov.atlassian.net/browse/IS-836?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/services/db/GitHubService.ts
+++ b/src/services/db/GitHubService.ts
@@ -1,6 +1,6 @@
 import { AxiosCacheInstance } from "axios-cache-interceptor"
 import { Base64 } from "js-base64"
-import { okAsync, errAsync } from "neverthrow"
+import { okAsync, errAsync, ResultAsync } from "neverthrow"
 
 import { ConflictError, inputNameConflictErrorMsg } from "@errors/ConflictError"
 import { NotFoundError } from "@errors/NotFoundError"
@@ -11,9 +11,10 @@ import { isAxiosError, validateStatus } from "@utils/axios-utils"
 import GithubSessionData from "@root/classes/GithubSessionData"
 import UserWithSiteSessionData from "@root/classes/UserWithSiteSessionData"
 import { STAGING_BRANCH } from "@root/constants"
+import GitHubApiError from "@root/errors/GitHubApiError"
 import logger from "@root/logger/logger"
 import { GitCommitResult } from "@root/types/gitfilesystem"
-import { RawGitTreeEntry } from "@root/types/github"
+import { GitHubRepoInfo, RawGitTreeEntry, RepoState } from "@root/types/github"
 
 import * as ReviewApi from "./review"
 
@@ -418,7 +419,9 @@ export default class GitHubService {
     }
   }
 
-  async getRepoInfo(sessionData: UserWithSiteSessionData) {
+  async getRepoInfo(
+    sessionData: UserWithSiteSessionData
+  ): Promise<GitHubRepoInfo> {
     const { siteName } = sessionData
     const { accessToken } = sessionData
     const endpoint = `${siteName}`
@@ -437,7 +440,7 @@ export default class GitHubService {
     return data
   }
 
-  async getRepoState(sessionData: UserWithSiteSessionData) {
+  async getRepoState(sessionData: UserWithSiteSessionData): Promise<RepoState> {
     const { accessToken } = sessionData
     const { siteName } = sessionData
     const endpoint = `${siteName}/commits`
@@ -791,7 +794,7 @@ export default class GitHubService {
   async changeRepoPrivacy(
     sessionData: { siteName: string; isomerUserId: string },
     shouldMakePrivate: boolean
-  ) {
+  ): Promise<ResultAsync<null, NotFoundError | GitHubApiError>> {
     const { siteName, isomerUserId } = sessionData
     const endpoint = `${siteName}`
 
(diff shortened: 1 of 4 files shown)


src/services/db/GitHubService.ts
```diff
@@ -791,7 +794,7 @@ export default class GitHubService {
   async changeRepoPrivacy(
     sessionData: { siteName: string; isomerUserId: string },
     shouldMakePrivate: boolean
-  ) {
+  ): Promise<ResultAsync<null, NotFoundError | GitHubApiError>> {
```
dcshzj, line 797: nit: `ResultAsync` itself is a promise, so we don't need to wrap in a `Promise`

src/services/db/RepoService.ts
```diff
@@ -1,5 +1,7 @@
+import { User } from "aws-sdk/clients/budgets"
```
dcshzj, line 1: nit: I think this is not needed

src/services/db/RepoService.ts
```diff
@@ -112,19 +116,39 @@ export default class RepoService extends GitHubService {
...
   }
 
-  getBlobPath({ siteName, fileSha }: any): any {
+  getBlobPath({
+    siteName,
+    fileSha,
+  }: {
+    siteName: string
+    fileSha: string
+  }): string {
```
dcshzj, line 141: nit: This `string` return type can be inferred from the parent so we don't need to redefine here

src/services/db/RepoService.ts
```diff
@@ -592,14 +618,14 @@ export default class RepoService extends GitHubService {
...
     return super.checkHasAccess(sessionData)
   }
 
   async changeRepoPrivacy(
-    sessionData: any,
-    shouldMakePrivate: any
-  ): Promise<any> {
+    sessionData: UserWithSiteSessionData,
+    shouldMakePrivate: boolean
+  ): Promise<ResultAsync<null, NotFoundError | GitHubApiError>> {
```
dcshzj, line 628: nit: same here regarding the wrapping of `ResultAsync`

src/services/db/GitHubService.ts
QiluXie: In principle you are right but removing it here would cause error in typescript. See the attached screenshot. 
Because the ResultAsync type from the `neverthrow` library is a wrapper around a `Promise` that provides additional utility methods and properties. However, TypeScript doesn't recognize it as a global `Promise` because it's not an instance of the global `Promise` type.
Screenshot 2023-12-20 at 12 07 15 PM

Removing the `async` keyword would solve this error but we are using `await` inside the function, so it's not a viable option. 

Therefore, we should wrap `ResultAsync` inside the global `Promise` here for `Typescript` to correctly recognise it.

src/services/db/RepoService.ts
QiluXie: Same as the[ comment here](https://github.com/isomerpages/isomercms-backend/pull/1066#discussion_r1432225854)

src/services/db/RepoService.ts
QiluXie: Thanks for the catch! Removed in ce3615eb571009f5f1bf3074191def1847dfa87b

src/services/db/RepoService.ts
QiluXie: Sure, thanks for the catch. Removed in ce3615eb571009f5f1bf3074191def1847dfa87b

src/services/db/GitHubService.ts
dcshzj: Hmm I think it is a symptom of a bigger problem in the function where we are mixing `neverthrow` with async/await. Ideally, we should be wrapping the `this.axiosInstance.patch` inside a `ResultAsync.fromPromise`, then doing the error handling inside. That way, we can remove the async/await from this function and the return type is preserved as `ResultAsync`.

I'm ok with fixing this in a separate PR if you prefer.

QiluXie: Yeah I think you are right. Refactored it in 1cb950bcd9aa3b41314912c8e30bb51cfc6e9c52
Wrapped the patch call inside `ResultAsync.fromPromise` as you suggested. 

Upon a quick search, there are other instances of similar mixed use of `async/await` and `ResultAsync`, will address them in separate PR.

src/services/db/RepoService.ts
QiluXie: Refactored in 1cb950bcd9aa3b41314912c8e30bb51cfc6e9c52 along with [the comment here](https://github.com/isomerpages/isomercms-backend/pull/1066#discussion_r1432250037)

dcshzj (approved):

QiluXie:
CI passed, all green. Merging now.

PR #1068, opened 2023-12-19
fix(db): make use of the reader endpoint for read queries

dcshzj:
> [!WARNING]
> **High risk change**: Need PR author to shoulder check!
> The env vars have been added to SSM but **NOT** on 1pw

## Problem

We have always been using a single DB instance for all queries, which makes the reader instance pointless/wasteful.

## Solution

**Breaking Changes**

**Bug Fixes**:

- Make use of [Sequelize's built-in functionality](https://sequelize.org/docs/v6/other-topics/read-replication/) to use the reader endpoint when performing read queries.

## Tests

- Need to randomly test read/write actions, might need to set up staging to have a reader instance temporarily

## Deploy Notes

- `DB_READ_URI` : The connection string for the reader endpoint
- `DB_WRITE_URI`: The connection string for the writer endpoint
    - [x] added env var to 1PW + SSM script (`fetch_ssm_parameters.sh`)

**Deploy steps**
- [ ] ~Set up reader instance for staging environment~ RDS already comes with a read/write endpoint, using that instead
- [x] Set up `DB_READ_URI` and `DB_WRITE_URI` env vars on SSM for both staging and prod
- [x] Deploy on staging and verify DB-related actions work (test with review requests)
- [ ] Update 1pw with `DB_READ_URI` and `DB_WRITE_URI` env vars
- [ ] Deploy to prod


src/database/config.js
```diff
@@ -7,22 +7,37 @@ const { parse } = require("pg-connection-string")
...
 
 module.exports = {
   // Connection settings
-  database: parsed.database || "isomercms_dev",
-  host: parsed.host || "localhost",
-  username: parsed.user,
-  password: parsed.password,
-  port,
+  replication: {
+    read: [
```
harishv7, line 24: we have added the endpoints into the configs. but , how are we conditionally using the read and write endpoints

dcshzj: This is handled automatically by Sequelize. For example, if we use a `findOne` function, it is known by Sequelize to be a read-only query, so that query will be directed to the read endpoint. Same goes for update functions.

harishv7: sweet, this is what were talking about on Mon!

harishv7 (approved):
lgtm, let's be sure to test well on staging before merging. also be careful abt the env var changes

timotheeg:
As discussed earlier,

TODOs:
1.  Set a read replica on staging to be able to test this "properly"
2.  Identify if we have any case of sync reads (i.e. when a read is done directly following a write, with the expectation that it is consistent). This can of consistency cannot be guaranteed with a read replica (which has lag), and typically the ORM must be instructed to read from master specifically instead. In sequelize, that is done by specifying a `useMaster:true` in the queries ([documentation here](https://sequelize.org/api/v6/class/src/sequelize.js~sequelize#instance-method-query))

To do 2) typically we perform static analysis on the codebase to find the synchronous reads, and we can look at the list of all sql queries made by the app to find if we do have such cases ([list in DD here](https://ogp.datadoghq.com/apm/traces?query=env%3Aproduction%20service%3Aisomer-postgres&agg_q=resource_name&agg_q_source=base&cols=core_service%2Ccore_resource_name%2Clog_duration%2Clog_http.method%2Clog_http.status_code&fromUser=true&historicalData=true&messageDisplay=inline&query_translation_version=v0&sort=desc&sort_m=&sort_m_source=&sort_t=&spanType=all&top_n=100&top_o=top&traceQuery=&view=spans&viz=toplist&x_missing=true&start=1711958049037&end=1712562849037&paused=false))

timotheeg:
I rebased this from develop

dcshzj:
Closing this, too much effort.

PR #1072, opened 2023-12-20
fix: add conditional sha calc

harishv7:
fix for renaming subfolders

## context

We made a change to fast load images as huge repos with a lot of images had very slow load times.

We used to do: get all files -> calculate hash -> filter/paginate
Now, we get all files -> filter/paginate -> calculate hash

For the get all files to be fast, we removed the sha calculation in the first step of get all files. However, this affected other ops which use list directory contents

Specifically, when we try to rename a subfolder, we need to delete the current directory and re-create a new one
When deleting the current directory, it uses list directory method, but the sha is not present. So when it does a verification of oldSha === newSha, it fails as oldSha is undefined. This causes BE to think that the file was recently changed, so it blocks the delete operation.

Hence, the rename op fails, and we do a rollback. We return 409 saying someone edited the file.

## tests

- Ensure ci/test passes
- On staging, test renaming sub-folders on email login
- Sites with > 15 images should be paginated and switching across pages should be smooth
- Verify same for GitHub login

### DIFF
--- a/src/routes/v2/authenticatedSites/media.ts
+++ b/src/routes/v2/authenticatedSites/media.ts
@@ -108,13 +108,11 @@ export class MediaRouter {
     { directoryName: string },
     { directories: MediaDirOutput[] },
     never,
-    { page: number },
+    never,
     { userWithSiteSessionData: UserWithSiteSessionData }
   > = async (req, res) => {
     const { userWithSiteSessionData } = res.locals
-
     const { directoryName } = req.params
-    const { page } = req.query
 
     const {
       directories,
@@ -122,8 +120,8 @@ export class MediaRouter {
       userWithSiteSessionData,
       {
         directoryName,
-        page,
-        limit: 0,
+        page: 0,
+        limit: 1,
         search: "",
       }
     )
--- a/src/services/db/GitFileSystemService.ts
+++ b/src/services/db/GitFileSystemService.ts
@@ -967,20 +967,16 @@ export default class GitFileSystemService {
               )
               .andThen((shaAndStats) => {
                 const [sha, stats] = shaAndStats
-                return this.getFilePathStats(repoName, path, isStaging).andThen(
-                  (stats) => {
-                    const result: GitDirectoryItem = {
-                      name,
-                      type,
-                      sha,
-                      path,
-                      size: type === "dir" ? 0 : stats.size,
-                      addedTime: stats.ctimeMs,
-                    }
+                const result: GitDirectoryItem = {
+                  name,
+                  type,
+                  sha,
+                  path,
+                  size: type === "dir" ? 0 : stats.size,
+                  addedTime: stats.ctimeMs,
+                }
 
-                    return okAsync(result)
-                  }
-                )
+                return okAsync(result)
               })
           }
           return this.getFilePathStats(repoName, path, isStaging).andThen(
@@ -1022,7 +1018,6 @@ export default class GitFileSystemService {
           getPaginatedDirectoryContents(directoryContents, page, limit, search)
         )
       )
-      .andThen((dirContents) => okAsync(dirContents))
       .andThen((paginatedDirectoryContents) => {
         const directories = paginatedDirectoryContents.directories.map(
           (directory) =>
--- a/src/services/db/GitFileSystemService.ts
+++ b/src/services/db/GitFileSystemService.ts
@@ -1109,7 +1109,6 @@ export default class GitFileSystemService {
       .andThen(() =>
         this.getGitBlobHash(repoName, filePath, isStaging).andThen((sha) => {
           if (sha !== oldSha) {
-            logger.debug(`Old sha: ${oldSha}, new: ${sha}`)
             return errAsync(
               new ConflictError(
                 "File has been changed recently, please try again"
@@ -1216,7 +1215,6 @@ export default class GitFileSystemService {
           return okAsync(true) // If it's a directory, skip the blob hash verification
         }
         return this.getGitBlobHash(repoName, path, isStaging).andThen((sha) => {
-          logger.debug(`Delete: Old sha: ${oldSha}, new: ${sha}`)
           if (sha !== oldSha) {
             return errAsync(
               new ConflictError(
--- a/src/services/db/GitFileSystemService.ts
+++ b/src/services/db/GitFileSystemService.ts
@@ -912,7 +912,8 @@ export default class GitFileSystemService {
   listDirectoryContents(
     repoName: string,
     directoryPath: string,
-    branchName: string
+    branchName: string,
+    includeSha = true
   ): ResultAsync<GitDirectoryItem[], GitFileSystemError | NotFoundError> {
     const efsVolPath = this.getEfsVolPathFromBranch(branchName)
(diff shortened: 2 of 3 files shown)


src/routes/v2/authenticatedSites/media.ts
```diff
@@ -108,22 +108,20 @@ export class MediaRouter {
     { directoryName: string },
     { directories: MediaDirOutput[] },
     never,
-    { page: number },
+    never,
```
kishore03109, line 111: hey was this intentional, why we making this never ah?

kishore03109, line 111: hey was this intentional, why we making this never ah?

kishore03109, line 111: hey was this intentional, why we making this never ah?

src/routes/v2/authenticatedSites/media.ts
```diff
@@ -108,22 +108,20 @@ export class MediaRouter {
...
 
     const {
       directories,
     } = await this.mediaDirectoryService.listMediaDirectoryContent(
       userWithSiteSessionData,
       {
         directoryName,
-        page,
-        limit: 0,
+        page: 0,
```
kishore03109, line 123: hey am confused here, arent we supposed to use the page number from the fe, rather than defaulting to a value?

src/routes/v2/authenticatedSites/media.ts
```diff
@@ -108,22 +108,20 @@ export class MediaRouter {
...
+        limit: 1,
```
kishore03109, line 124: May I get context behind this change?

src/services/db/GitFileSystemService.ts
```diff
@@ -984,12 +1011,18 @@ export default class GitFileSystemService {
...
+      directoryPath,
+      branchName,
+      false
+    )
       .andThen((directoryContents) =>
         okAsync(
           getPaginatedDirectoryContents(directoryContents, page, limit, search)
         )
       )
+      .andThen((dirContents) => okAsync(dirContents))
```
kishore03109, line 1025: hey i think this was accidentally added!

src/services/db/GitFileSystemService.ts
```diff
@@ -912,7 +912,8 @@ export default class GitFileSystemService {
   listDirectoryContents(
     repoName: string,
     directoryPath: string,
-    branchName: string
+    branchName: string,
+    includeSha = true
```
kishore03109, line 916: may i get the context for why we are adding this param?

src/services/db/GitFileSystemService.ts
```diff
@@ -955,6 +956,33 @@ export default class GitFileSystemService {
...
+              .orElse(() => okAsync(""))
+              .andThen((sha) =>
+                ResultAsync.combine([
+                  okAsync(sha),
+                  this.getFilePathStats(repoName, path, isStaging),
+                ])
+              )
+              .andThen((shaAndStats) => {
+                const [sha, stats] = shaAndStats
+                return this.getFilePathStats(repoName, path, isStaging).andThen(
```
seaerchin, line 970: can probably remove the call to `this.getFilePathStats` cos we alr have it from prev step

seaerchin (commented):
so on delete/rename, we check for `oldSha` and `newSha`? and because now we set `includeSha` to false, this allows us to list paginated directories without issue, thereby allowing us to delete?

not super sure on how this change fixes the issue, would be good if we verified this on staging against 
1. base directories
2. subdirectories
3. recursively nested directories w/ depth > 2 (albums/files folder)

src/routes/v2/authenticatedSites/media.ts
seaerchin: from huddle: we need to set this because otherwise it invokes `.value`, which will cause it to get all the files

src/services/db/GitFileSystemService.ts
seaerchin: this param is required to skip/include the sha; if we skip teh sha, then we can delete safely without encountering a `409` error

src/routes/v2/authenticatedSites/media.ts
seaerchin: this is to list the subdirectories, so we pass in sane defaults to avoid listing the actual pages/images

seaerchin (approved):
lgtm for now cos to resolve prod issue

src/routes/v2/authenticatedSites/media.ts
harishv7: we do not do any pagination/slicing here

src/routes/v2/authenticatedSites/media.ts
harishv7: we are not performing pagination here

src/routes/v2/authenticatedSites/media.ts
harishv7: limit 0 loads everything as you can see in src/utils/files.ts

```
 const paginatedFiles =
    limit === 0
      ? sortedFiles.value()
      : sortedFiles
          .drop(page * limit)
          .take(limit)
          .value()
```

src/services/db/GitFileSystemService.ts
harishv7: ZJ made a change to fast load images as huge repos with a lot of images had very slow load times. 

We used to do: get all files -> calculate hash -> filter/paginate
Now, we get all files -> filter/paginate -> calculate hash

For the get all files to be fast, we removed the sha calculation in the first step of get all files. However, this affected other ops which use list directory contents

Specifically, when we try to rename a subfolder, we need to delete the current directory and re-create a new one
When deleting the current directory, it uses list directory method, but the sha is not present. So when it does a verification of oldSha === newSha, it fails as oldSha is undefined. This causes BE to think that the file was recently changed, so it blocks the delete operation.

Hence, the rename op fails, and we do a rollback. We return 409 saying someone edited the file.

src/services/db/GitFileSystemService.ts
harishv7: will remove

src/services/db/GitFileSystemService.ts
kishore03109: ok understand this is to fix prod, so dont have to fix in this pr but i do have concerns of this API that it is confusing. would rather have some "force=true" than have `includeSha = true` but am not sure how that

src/services/db/GitFileSystemService.ts
harishv7: fixed

src/services/db/GitFileSystemService.ts
harishv7: knowing the sha will allow oldSha === newSha be true for .keep and all files under the subdirectory, which allows the delete op to go through

harishv7: includeSha is more specific than "force=true". Includesha flag tells the method to compute sha for all the files it is fetching in listDirectoryContents

seaerchin:
@harishv7 got 1 failing test btw maybe need fix

PR #1075, opened 2023-12-21
fix(betterUptime): make better uptime on success

kishore03109:
## Problem
Currently our better uptime is done on create, but it should be on success instead. 

Closes [insert issue #]

## Solution

shifting the creation of this to post success, not during the launch process. 

**Breaking Changes**

  - Details ...

Tests

- [ ] create site alunch in staging, notice that better uptime monitor was [created](https://opengovproducts.slack.com/archives/C04RDJ504DC/p1704255875013479)

### DIFF
--- a/src/routes/formsg/formsgSiteLaunch.ts
+++ b/src/routes/formsg/formsgSiteLaunch.ts
@@ -308,7 +308,6 @@ export class FormsgSiteLaunchRouter {
             )
           }
           // Create better uptime monitor
-          await this.createMonitor(launchResult.value.primaryDomainSource)
           successResults.push(successResult)
         }
       }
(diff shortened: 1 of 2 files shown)


src/routes/formsg/formsgSiteLaunch.ts
alexanderleegs, line 1: We're still attempting to create the monitor in lines 310-311
(```
// Create better uptime monitor
await this.createMonitor(launchResult.value.primaryDomainSource)```)
,

this should be removed right!

kishore03109: 59ab717d8de7899bb18feae6465f414a801cf0bd

kishore03109: ps forgot to remove

alexanderleegs (approved):
lgtm

PR #1078, opened 2023-12-27
fix(migrations): set db uri correctly

seaerchin:
## Problem
When we swap to docker, we had to swap the `DB_URI` out, which meant that migrations ran using the command weren't working properly.

## Solution
inject env var at command. didn't create separate env file for this cos didn't wanna overload the `.env` files we have alr.


harishv7 (approved):

seaerchin:
### Merge activity

* **Dec 27, 12:54 AM**: @seaerchin started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/isomerpages/isomercms-backend/1078).

PR #1084, opened 2023-12-28
fix(ci): reverts ci changes to allow staging updates

seaerchin:
## Problem
we previously deployed to ecs on push to the `staging` branch. this PR swaps it back as efs has issues mounting onto a separate vpc. 

## Solution
swap out `ci.yml` changes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,7 +102,8 @@ jobs:
           import os
           ref = os.environ['GITHUB_REF']
           prod = os.environ['PRODUCTION_BRANCH']
-          if ref == prod:
+          staging = os.environ['STAGING_BRANCH']
+          if ref == prod or ref == staging:
             print('::set-output name=proceed::true')
           else:
             print('::set-output name=proceed::false')
--- a/.github/workflows/deploy_staging.yml
+++ b/.github/workflows/deploy_staging.yml
@@ -7,7 +7,8 @@ concurrency:
 on:
   push:
     branches:
-      - staging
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -102,7 +102,8 @@ jobs:
           import os
           ref = os.environ['GITHUB_REF']
           prod = os.environ['PRODUCTION_BRANCH']
-          if ref == prod:
+          staging = os.environ['STAGING_BRANCH']
+          if ref == prod or ref == staging:
```
harishv7, line 106: @seaerchin is this a temp change or perm? Also what does "efs has issues mounting" mean, is it an unexpected error? could you provide a little more context?

seaerchin: This is temp for now, will try to debug efs mounting when I'm back 

The problem is that ecs itself resides in a different vpc than efs. I've created a peering connection from the eb private vpc to the compute vpc (together with a ingress rule that allows NFS traffic from ecs subnet to eb private subnet). Using aws reachability analyzed shows that nfs traffic over the peering connection reaches our efs.

However, at deployment time, what happens is that the filesystem id of our efs is unable to be resolved. We already created a mount target for our efs inside the existing private subnet, so we can't create another one in the same AZ. 

I'm not entirely sure what's happening here (deployment logs on ecs just shows that the mount failed), maybe we just need to allow egress traffic but (when I'm back), gonna try a few things and if fail, just create our ecs inside the existing vpc...

harishv7 (approved):

PR #1088, opened 2024-01-02
fix(amplify): use rewrite instead of redirect for 404 error

dcshzj:
## Problem

We redirect when a page is not found instead of rewriting (to preserve the URL). This is semantically incorrect and causes SEO issues.

## Solution

**Breaking Changes**

**Features**:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- Set the custom rule to rewrite instead of redirect. The `404-200` comes from [their API reference](https://docs.aws.amazon.com/amplify/latest/APIReference/API_CustomRule.html).

## Deploy Notes

*None*

### DIFF
--- a/src/services/identity/DeploymentClient.ts
+++ b/src/services/identity/DeploymentClient.ts
@@ -110,7 +110,7 @@ class DeploymentClient {
       },
     ]
     const defaultRedirectRules = [
-      { source: "/<*>", target: "/404.html", status: "404" },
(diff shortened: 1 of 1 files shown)


kishore03109 (approved):

dcshzj:
Nope, we were wrong. It doesn't work if the site has been launched, weird.

PR #1090, opened 2024-01-02
feat(logging): remove logging

kishore03109:
## Problem
logging damn noisy, and the logging doesnt show the actual arguments anyways. maybe a better design is to allow this for prod, but I have never used these statements for production logging, and to be honest, most of the time it has hindered my ability to debug becasue this has been pure noise. 

## Solution
I can actually see my console.logs now without scrolling infinitely

### DIFF
--- a/src/database/config.js
+++ b/src/database/config.js
@@ -43,4 +43,5 @@ module.exports = {
     max: DB_MAX_POOL,
     acquire: DB_ACQUIRE,
   },
+  logging: false,
 }


src/database/config.js
```diff
@@ -43,4 +43,5 @@ module.exports = {
     max: DB_MAX_POOL,
     acquire: DB_ACQUIRE,
   },
+  logging: false,
```
harishv7, line 46: This also disables logs on prod?

harishv7 (approved):

src/database/config.js
kishore03109: yes, i think its useless on prod since we dont actually log the params. 
i can see why this would be useful but it cannot be output to raw logs if not hard to debug uh. 
@dcshzj mentioned of making our own by formatted sql statements which to be honest sounds like the most ideal solution, but that can be done as a separate pr uh

PR #1092, opened 2024-01-03
IS-653: add functionality for RR comments in DB

harishv7:
# NOTE: This PR requires DB migration to add new table

## Problem

Task: We rely on GitHub to manage comments. Right now, comments are not linked to a specific file nor line on a changed file. The comments apply to the entire RR. This is easy to replicate within CMS BE. 

## Solution

What we need is:

A new table to store comments (as strings). The rough fields are RR_ID, User_ID, Site_ID, Comment (String)

Impact: Eliminates dependency on GitHub API Requests, Pull Requests, and consequently GitHub Token usage

Note: For now, the logic will fetch comments from both GitHub + DB till we completely phase out GitHub usage, as there may be some already in-progress RRs which have comments stored on GitHub

**Breaking Changes**

  - Details ...
- [] No - this PR is backwards compatible with ALL of the following feature flags in this [doc](https://www.notion.so/opengov/Existing-feature-flags-518ad2cdc325420893a105e88c432be5)

## Tests

- [ ] Ensure all tests (unit + integration) passes
- [ ] Create a new RR
- [ ] Add some comments to the RR, this will be piped to GitHub before the migration
- [ ] Check out from this branch
- [ ] Run the DB migration locally / on staging
- [ ] Create new RR
- [ ] Add new comments, this will and should be populated in the DB. Use TablePlus to verify
- [ ] Visit the RR on CMS, you should be able to see both the comments from GitHub + DB

## Deploy Notes

**Note: A DB migration is required for this PR.**


src/services/review/ReviewCommentService.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  private readonly reviewers: ModelStatic<Reviewer>
+
+  private readonly sequelize: Sequelize
+
+  constructor(
+    users: ModelStatic<User>,
+    repository: ModelStatic<ReviewComment>,
+    reviewers: ModelStatic<Reviewer>,
+    sequelize: Sequelize
```
alexanderleegs, line 23: `users`, `reviewers` and `sequelize` seem to be unused here!

src/services/review/ReviewCommentService.ts
```diff
@@ -0,0 +1,49 @@
+import { ModelStatic } from "sequelize"
+import { Sequelize } from "sequelize-typescript"
+
+import { ReviewComment } from "@database/models/ReviewComment"
+import { Reviewer } from "@database/models/Reviewers"
+import { ReviewMeta } from "@database/models/ReviewMeta"
+import { ReviewRequest } from "@database/models/ReviewRequest"
```
alexanderleegs, line 7: Nit: `ReviewMeta` and `ReviewRequest` are unused

alexanderleegs (approved):
lgtm, minor nits!

Could I also clarify what the rollout plan of this is? Are we planning to do a swapover for only comments at a later date (once we ensure that no remaining open PRs on github are open with comments), or are we planning to phase out the open PRs on github as well? Currently we still make use of the pull request system on Github even though we also maintain our own internal record of PRs

src/database/models/ReviewComment.ts
```diff
@@ -0,0 +1,48 @@
...
+  @ForeignKey(() => ReviewRequest)
+  @Column
+  reviewId!: string
+
+  @Column
+  comment!: string
+
+  @BelongsTo(() => User, {
+    onUpdate: "CASCADE",
+    onDelete: "CASCADE",
```
kishore03109, line 39: are we doing soft deletes? We shouldnt be deleting the comments when a user is deleted, since we dont have that audit trail anymore

harishv7:
> lgtm, minor nits!
> 
> Could I also clarify what the rollout plan of this is? Are we planning to do a swapover for only comments at a later date (once we ensure that no remaining open PRs on github are open with comments), or are we planning to phase out the open PRs on github as well? Currently we still make use of the pull request system on Github even though we also maintain our own internal record of PRs
> 
> `users`, `reviewers` and `sequelize` seem to be unused here!
> 
> Nit: `ReviewMeta` and `ReviewRequest` are unused

@alexanderleegs The changes are compatible with GitHub comments. So it pulls from both GitHub and DB before returning to FE.

However, new comments are only created in our DB. So in the future, say 1-2 months when we are relatively confident most of the open RRs (before the date of merge) are closed, we can remove the logic to fetch from GH - saves 1 network call as well

src/database/models/ReviewComment.ts
harishv7: @kishore03109  good point, will remove the cascade to keep the trail. But even so, for audit trail that means a User should never be deleted right?

By that logic, then cascading delete will make sense, because even if we retain this when a user is deleted, the relation to the table is gone. So how are we going to identify the user?

alexanderleegs: I think regardless we shouldn't be hard deleting! for users there's a bunch of related information so we'll need to do a soft delete because of audit purposes

harishv7: @alexanderleegs Yes that's exactly my point - we shouldn't hard delete for User. Hence, this cascade delete of comments will never happen.

But to be sure, I will remove the cascade delete

kishore03109:
> @alexanderleegs The changes are compatible with GitHub comments. So it pulls from both GitHub and DB before returning to FE.

However, new comments are only created in our DB. So in the future, say 1-2 months when we are relatively confident most of the open RRs (before the date of merge) are closed, we can remove the logic to fetch from GH - saves 1 network call as well

This rollout plan is sane, just confirming here, this would not require any manual remediation right? since there will be no deuplciates since a review has its comments either stored in gh OR db, never gh AND db, right?

seaerchin:
just double checking here - the intent of this PR is to complete phase out github?

PR #1094, opened 2024-01-04
Fix/privatisation quickie interaction

alexanderleegs:
This PR fixes an issue with the integration of privatisation with quickie. To be reviewed in conjunction with PR[#1763](https://github.com/isomerpages/isomercms-frontend/pull/1763) on the frontend repo. There are two existing issues being tackled by this PR:

- Repos which are being made private/unprivate should also affect their respective staging-lite sites, otherwise a site might still contain a publicly accessible version in staging-lite
- Our DB automatically self corrects to switch any site which has been placed on quickie via growthbook to the `staging-lite.` variant, but this should not be true for private sites

Note that this PR does not handle the issue of inconsistent states fully right now - it is possible that a user who attempts to privatise/unprivatise a site will have the request to the second amplify site fail, leaving us in an inconsistent state - this needs to be manually resolved for now, as the resolution is complicated. We can look into making this more robust if we find that this is a consistent issue, but I am opting to leave this as is for now and pipe alarms that notify us instead. The issue is also fixable on the user's end if they try again - the inconsistent state will not block cms operations.

TODOs:

- [x] ~Add cloudwatch alarm for inconsistent state resolution~
- [x] ~Add entry in runbook for handling inconsistent state~

Tests

- [ ] Privatise a site
  - [ ] Ensure that the staging and staging lite sites on amplify get updated with the password
- [ ] Deprivatise a site
  - [ ] Ensure that the staging and staging-lite sites on amplify have removed password

### DIFF
--- a/src/services/identity/DeploymentsService.ts
+++ b/src/services/identity/DeploymentsService.ts
@@ -202,9 +202,14 @@ class DeploymentsService {
     return okAsync(deploymentInfo)
   }
 
-  deletePassword = async (appId: string, deploymentId: number) => {
+  deletePassword = async (
+    appId: string,
+    deploymentId: number,
+    isStagingLite: boolean
+  ) => {
     const updateAppInput = this.deploymentClient.generateDeletePasswordInput(
-      appId
+      appId,
+      isStagingLite ? "staging-lite" : "staging"
     )
     const updateResp = await this.deploymentClient.sendUpdateApp(updateAppInput)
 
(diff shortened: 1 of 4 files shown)


src/services/identity/DeploymentsService.ts
```diff
@@ -249,9 +254,22 @@ class DeploymentsService {
...
 
-    if (!enablePassword) return this.deletePassword(appId, id)
+    if (!enablePassword) {
+      const stagingRes = await this.deletePassword(appId, id, false)
+      if (!stagingLiteId || stagingRes.isErr()) return stagingRes
+      const stagingLiteRes = await this.deletePassword(stagingLiteId, id, true)
+      if (stagingLiteRes.isErr())
+        logger.error(
+          `Privatisation adjustment failed for ${repoName} - requires manual fixing of inconsistent state`
+        )
```
seaerchin, line 270: do we have an alarm for this? otherwise, we won't know when this occurs until someone feedsback to us, rendering this log moot

seaerchin (approved):
lgtm with one question

src/services/identity/DeploymentsService.ts
alexanderleegs: Hey yup the alarm's already been set up!
https://ap-southeast-1.console.aws.amazon.com/cloudwatch/home?region=ap-southeast-1#alarmsV2:alarm/cms-backend-prod-node18-privatisation-inconsistency-error?~(search~'privatisation)

PR #1100, opened 2024-01-08
chore: upgrade axios

alexanderleegs:


package.json
```diff
@@ -37,7 +37,7 @@
     "auto-bind": "^4.0.0",
     "aws-lambda": "^1.0.7",
     "aws-sdk": "^2.1428.0",
-    "axios": "~0.27.2",
+    "axios": "^1.6.5",
```
harishv7, line 40: looks like a major version change, have we verified if any breaking changes?

alexanderleegs: tests run - https://cloud.cypress.io/projects/nxbty1/runs/477/specs

Seems to be mostly similar errors as our normal tests

harishv7 (approved):

PR #1102, opened 2024-01-10
fix(rr): skip checking the existence of review request for blob

dcshzj:
## Problem

When a user is creating a new review request, we also try to obtain the diff between the staging and master branches. However, the logic requires a valid review request to be present, which may not be the case when the user is in the midst of creating one.

## Solution

**Breaking Changes**

**Bug Fixes**:

- For the time being, there is no reason for us to know if the review request is present, as we are only comparing between the staging and master branches. Hence, we can skip the check for the existence of the review request.
    - This is important when creating a new review request, as the review request ID is not present.

## Tests

- [ ] Navigate to any site and click on create a review request.
- [ ] Click on the Edited items tab (also ensure that you have a difference between staging and master).
- [ ] Click on any of the item's eye icon. You should be able to view the diff.
- [ ] Actually create the review request.
- [ ] Verify that you can view the same diff in the created review request.

## Deploy Notes

*None*

### DIFF
--- a/src/routes/v2/authenticated/review.ts
+++ b/src/routes/v2/authenticated/review.ts
@@ -1274,7 +1274,9 @@ export class ReviewsRouter {
     { userWithSiteSessionData: UserWithSiteSessionData }
   > = async (req, res) => {
     // Step 1: Check that the site exists
-    const { siteName, requestId } = req.params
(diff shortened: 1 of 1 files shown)


alexanderleegs (approved):
lgtm

PR #1108, opened 2024-01-11
chore: update formsg sdk

alexanderleegs:


seaerchin (commented):
codewise ok but ci seems to be failing - could you take a look?

alexanderleegs:
> codewise ok but ci seems to be failing - could you take a look?

hmm seems to work fine locally, going to just rerun

seaerchin (approved):

PR #1109, opened 2024-01-14
fix: remove unnecessary push logs

harishv7:
## Problem

We get a lot of false positives on git push failures which is affecting on-call productivity. 

Closes IS-847

## Solution

we should log only when git push is a complete failure (i.e. after 3 tries)

**Breaking Changes**

  - Details ...

## Tests

- On first 2 push failures, it should log differently from the third try
- To test, we might need to simulate a divergence in EFS vs GH states
- Perform edits on GH for a repo that is currently on EFS
- Go to CMS now and perform some edits. These edits cannot be push due to the divergence
- Check the logs to ensure that only the third try and error case prints out the actual Git error (i.e. should read like `error: failed to push some refs .......` )

### DIFF
--- a/src/services/db/GitFileSystemService.ts
+++ b/src/services/db/GitFileSystemService.ts
@@ -539,6 +539,10 @@ export default class GitFileSystemService {
                   .cwd({ path: `${efsVolPath}/${repoName}`, root: false })
                   .push(gitOptions),
             (error) => {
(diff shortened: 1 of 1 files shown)


alexanderleegs (approved):
lgtm!

Have we also modified our alarms to account for the new error message to search for?

PR #1114, opened 2024-01-25
fix(ecs): misc fixes for deployment workflow for staging

seaerchin:
## Problem
There were issues previously which would lead to a failed deployment of our code to ecs. These are fixed here and will be explained below

## Solution
1. add `appspec.json` - we require this for aws codedeploy, which is used under the hood to do a blue green deployment of our app 
2. update task definitions for `staging` + `prod` 
    - env vars can only be string so non-string values were coerced to string 
    - add missing env vars (i think max limit was hit previously? unsure because i added this using a script) 
    - update log groups for `staging` (was piping to `prod` :monkas:) 
    - update task + exec roles 
3. update deployment workflow for staging **(NOTE: this file does not exist for prod)** to use codedeploy (ecs component in our ogp lib uses it under the hood) 
4. add command to list safe directories in dockerfile, required so that `git` won't complain
5. add double quotes for ssh key fetching; will fail otherwise as hte newline will be stripped.


.github/workflows/ci.yml
```diff
@@ -162,16 +161,11 @@ jobs:
...
             branch = os.environ['GITHUB_REF']
             production = os.environ['PRODUCTION_BRANCH']
-            staging = os.environ['STAGING_BRANCH']
             eb_app = os.environ['EB_APP']
             eb_env_production = os.environ['EB_ENV_PRODUCTION']
-            eb_env_staging = os.environ['EB_ENV_STAGING']
             if branch == production:
               print('::set-output name=eb_app::' + eb_app)
               print('::set-output name=eb_env::' + eb_env_production)
-            elif branch == staging:
```
kishore03109, line 172: hey sense check here, is this intentional

kishore03109 (approved):

.github/workflows/ci.yml
seaerchin: yeap, instead of using this to deploy to staging, we're using a new workflow `deploy_staging.yml`!

don't have the braincells to combine them so leaving them apart

PR #1119, opened 2024-01-31
core parser logic

kishore03109:
## Problem
this pr adds the core parsing and reporting logic for the site link checker. 

Closes [insert issue #]

## Solution
high level algo 
1. grabs all images and files path 
2. parses all `.md` files and grabs a list of permalinks
3. if any duplicate permalinks are found, output to logs
4. parses all `.md` files and takes the content and convert everything to html.
5. parse the html code to get instances of `img` and `a` elements, and checks if the referenced path are all ok. 
6. outputs the list of incorrect referenced path to logs

**Breaking Changes**

- [ ] Yes - this PR contains breaking changes
  - Details ...
- [X] No - this PR is backwards compatible with ALL of the following feature flags in this [doc](https://www.notion.so/opengov/Existing-feature-flags-518ad2cdc325420893a105e88c432be5)

## Tests
add this in server.js to see the link checker run in the repo 
```
console.log("repoCheckerService")
const repoCheckerService = new RepoCheckerService()
const result = repoCheckerService.checkRepo("<name-of-repo>")
console.log("result")
console.info(result)
```

PR #1118, opened 2024-01-31
build(deps): add parser as dep

kishore03109:
## Problem

adding relevant deps for upcomming prs regarding site link checker 

note: change in dockerfile is needed here cuz i couldnt get the new package to download otherwise. we also should not be using node modules for consistency (notice node modules in the docker ignore)

this package was chosen in particular as it faster and produced better parsing than (markdown-it)[https://github.com/markdown-it/markdown-it], (showdown)[https://github.com/showdownjs/showdown]

PR #1121, opened 2024-02-06
integration with form

kishore03109:
## Problem

Allows us to run the script and update the isomer link checker repo directly.  NOTE: this is not just for observability, the function calls in `RepoCheckerService` will be used in the subsequent PRs as well! 

## Solution

Hook to formsg

**Breaking Changes**

  - Details ...

**Features**:

// todo run this in staging env and notice that in staging, the things have been running as expected 

**Improvements**:

- Details ...

**Bug Fixes**:

- Details ...

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Deploy Notes

- `SITE_CHECKER_FORM_KEY` : form key bah 

**New scripts**:

- `script` : script details

**New dependencies**:

- `papaparse` : easily parse the errors data objects into csv  with ease 

**New dev dependencies**:


kishore03109:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/isomerpages/isomercms-backend/1121?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* `develop`
  * **PR #1118** <a href="https://app.graphite.dev/github/pr/isomerpages/isomercms-backend/1118?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #1119** <a href="https://app.graphite.dev/github/pr/isomerpages/isomercms-backend/1119?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #1121** <a href="https://app.graphite.dev/github/pr/isomerpages/isomercms-backend/1121?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/isomerpages/isomercms-backend/1121?utm_source=stack-comment).

src/database/models/Repo.ts
seaerchin, line 1: Imports here don't seem linked to code?

src/database/models/SiteMember.ts
seaerchin, line 1: Imports here don't seem linked to code?

src/services/identity/SitesService.ts
seaerchin, line 1: Nit: can we remove the commented out code 

Also, do check that downstream callers of this method don't consume the repo as this method omits it now

seaerchin (approved):
This method requires a manual trigger on our end and doesn't enforce the link fix.

Should we instead just make the checking a GitHub action and add it to our repos + enforce that merges to production have 0 failures?

GitHub exposes an actions API so this should be doable! We can expose the output as a txt artifact and show it to admins too.

src/services/identity/SitesService.ts
kishore03109: eh omg this was not suppsed to go in. wrongly did a `git add .`

kishore03109:
>This method requires a manual trigger on our end and doesn't enforce the link fix.
Should we instead just make the checking a GitHub action and add it to our repos + enforce that merges to production have 0 failures?
GitHub exposes an actions API so this should be doable! We can expose the output as a txt artifact and show it to admins too.

Hey @seaerchin PR is only scoped to get observability for now to justify to ourselves that this is a worthy problem to be solved. This is not yet close to the final output of what the users will see via the CMS with regards to broken links!

seaerchin:
> > This method requires a manual trigger on our end and doesn't enforce the link fix.
> > Should we instead just make the checking a GitHub action and add it to our repos + enforce that merges to production have 0 failures?
> > GitHub exposes an actions API so this should be doable! We can expose the output as a txt artifact and show it to admins too.
> 
> Hey @seaerchin PR is only scoped to get observability for now to justify to ourselves that this is a worthy problem to be solved. This is not yet close to the final output of what the users will see via the CMS with regards to broken links!

ok so let me clarify - if the medium that we're choosing now (presenting a leaderboard) is a temporary measure to get feedback from users, this is fine!

my main worry is that there's not much incentive for users to go fix their links and hence, i'd be more in favour of just piggybacking off gha, which solves the problem of forcing users to fix their broken links cleanly w/ no need for manual triggers. 

if this is just for proto, lmk and i'll take a look/most probably approve

kishore03109:
@seaerchin 
>ok so let me clarify - if the medium that we're choosing now (presenting a leaderboard) is a temporary measure to get feedback from users, this is fine!
my main worry is that there's not much incentive for users to go fix their links and hence, i'd be more in favour of just piggybacking off gha, which solves the problem of forcing users to fix their broken links cleanly w/ no need for manual triggers.
if this is just for proto, lmk and i'll take a look/most probably approve

Hey its not for external consumption! will be covering soon on how we will present it to users ya. But yes the mental model here is a proto, since we are not aware how users feel about their sites/are deligent about say older pages with broken links. 
This exists as a sot to track this features success. if our actions dont shift the needle, this is the yardstick we will use to shift and change tactics

src/services/review/RepoCheckerService.ts
```diff
@@ -4,110 +4,358 @@ import path from "path"
...
+              return
             }
+            const html = await marked.parse(fileContent)
+            const dom = new JSDOM(html)
+            const anchorTags = dom.window.document.querySelectorAll("a")
+            forEach(anchorTags, (tag) => {
+              const href = tag.getAttribute("href")
+              const text = tag.textContent || ""
+              if (href) {
+                const encodedHref = this.mediaPathDecoder(href)
```
seaerchin, line 257: should this be `decodedHref`?

src/services/review/RepoCheckerService.ts
```diff
@@ -4,110 +4,358 @@ import path from "path"
...
+    const remover = () => {
+      if (!repoExists) {
+        return this.gitFileSystemService
+          .removeRepo(repo, STAGING_BRANCH)
+          .mapErr((error) => new SiteCheckerError(`${error}`))
+      }
+      return okAsync(undefined)
+    }
+
+    const checker = () =>
```
seaerchin, line 212: i think for each submethod here, we probably want to separate them out into their own method so the dependencies are clear.

src/services/review/RepoCheckerService.ts
```diff
@@ -4,110 +4,358 @@ import path from "path"
...
+                    linkToAsset: href,
+
+                    viewablePageInCms: this.getPathInCms(fileName),
+                    viewablePageInStaging: path.join(
+                      stagingLink,
+                      filePermalink
+                    ),
+                    linkedText: text,
+                  }
+                  errors.push(error)
```
seaerchin, line 275: since this is a proto, ok w/ this for now but this is a code smell. the error here is a derived state and we can jsut return it directly

src/services/review/RepoCheckerService.ts
```diff
@@ -4,110 +4,358 @@ import path from "path"
...
+              }
+            })
+            const imgTags = dom.window.document.querySelectorAll("img")
+            forEach(imgTags, (tag) => {
+              const src = tag.getAttribute("src")
+              if (src) {
+                const encodedSrc = this.mediaPathDecoder(src)
+                if (
+                  !this.isExternalLinkOrPageRef(encodedSrc) &&
+                  !setOfAllMediaAndPagesPath.has(encodedSrc)
```
seaerchin, line 286: this might potentially be buggy - our upstream has a list of [media](https://github.com/isomerpages/isomerpages-template/tree/staging/assets/img) that won't be present here, but will exist in the output. 

we might wanna add those to an excludes list or run this on the final built site, when all assets are pulled

src/services/review/RepoCheckerService.ts
```diff
@@ -4,110 +4,358 @@ import path from "path"
...
+                    linkedText: text,
+                  }
+                  errors.push(error)
+                }
+              }
+            })
+            const imgTags = dom.window.document.querySelectorAll("img")
+            forEach(imgTags, (tag) => {
+              const src = tag.getAttribute("src")
+              if (src) {
```
seaerchin, line 282: if there's no `src`, the image will fail to load - we should also report that as an error

src/services/review/RepoCheckerService.ts
```diff
@@ -4,110 +4,358 @@ import path from "path"
...
+            if (!fileContent) {
+              return
             }
+            const html = await marked.parse(fileContent)
+            const dom = new JSDOM(html)
+            const anchorTags = dom.window.document.querySelectorAll("a")
+            forEach(anchorTags, (tag) => {
+              const href = tag.getAttribute("href")
+              const text = tag.textContent || ""
+              if (href) {
```
seaerchin, line 256: if there's no href, we might wanna report an error

src/services/review/RepoCheckerService.ts
```diff
@@ -4,110 +4,358 @@ import path from "path"
...
+                const encodedHref = this.mediaPathDecoder(href)
+                if (
+                  !this.isExternalLinkOrPageRef(encodedHref) &&
+                  !setOfAllMediaAndPagesPath.has(encodedHref) &&
+                  // Amplify supports adding of trailing slash
+                  !setOfAllMediaAndPagesPath.has(`${encodedHref}/`)
```
seaerchin, line 262: i think this condition checks for internal links but the check itself (`isExternalLink`) is buggy - we'd need to check for scheme, **not** for substring. see [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#href) for more details. as is, we will allow `tel + sms` links and we omit a check for buggy mailtos, which exist in PA-corp.

seaerchin:
if this proto is jsut to get feedback, could we:
1. rename vars from `a, b` to somethign more descriptive
2. add error for empty href/img src
3. separate sub methods out 
4. fix the `async + forEach` 

rest can ignore

src/services/review/RepoCheckerService.ts
```diff
@@ -4,110 +4,358 @@ import path from "path"
...
+      })
+  }
+
+  /**
+   * Run the site checker for all repos sequentially
+   * Not using neverthrow here as we want to run all the checkers
+   * in sequence without short circuiting, and neverthrow does not
+   * support this.
+   */
+  async runCheckerForReposInSeq(repos: string[]): Promise<void> {
```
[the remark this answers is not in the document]
seaerchin: the comment seems to suggest that we want this to be the case?

src/services/review/RepoCheckerService.ts
```diff
@@ -4,110 +4,358 @@ import path from "path"
...
+      }
+      return okAsync(undefined)
+    }
+
+    const checker = () =>
+      this.getListOfMarkdownFiles(repoPath)
+        .andThen((files) => {
+          files.forEach((file) => {
+            mdFiles.add(file)
+          })
```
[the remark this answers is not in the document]
seaerchin: fwiw - the var is mutable as we can call `mdFiles.add` + scoping wise, the current one is less strict as it's in a bigger scope and locality of information is lost.

this is not a big deal here cos htis is proto but just smth to consider

src/services/review/RepoCheckerService.ts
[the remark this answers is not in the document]
kishore03109: @seaerchin ps typo, this ran concurrently** while we want sequentially

src/services/review/RepoCheckerService.ts
```diff
@@ -4,110 +4,358 @@ import path from "path"
...
+          return this.getAllMediaPath(repoPath)
+        })
+        .andThen((setOfAllMediaPath) =>
+          ResultAsync.fromPromise(
+            this.getAllPermalinks(repoPath, mdFiles),
```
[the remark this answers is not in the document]
seaerchin: hmm, it feels like they should be 2 different methods for keeping the methods focused and easier to use in the future.

we can do this in follow up

src/services/review/RepoCheckerService.ts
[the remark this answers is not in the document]
seaerchin: this is not important now, but why prefer sequential > concurrent if it doesn't short circuit? 

this would only be the case if we hold some lock or our current operation depends on a prior operation, neither of which seems to be the case

src/services/review/RepoCheckerService.ts
seaerchin: hmm, no not exactly - the `error` here is a derived state of our repos.

we shouldn't need to declare it at hte top and then do a `.push` - this loses locality and makes us prone to errors (scope is within lifetime of function).

we might, instead, want to have smaller methods that just return each category of error and we do the combine ourselves:

```ts
.andThen(prevErrors => { 
  const brokenLinks = checkForBrokenLinks(file)
  return ok([...prevErrors, ...brokenLinks])
})
```

src/services/review/RepoCheckerService.ts
```diff
@@ -177,6 +493,60 @@ export default class RepoCheckerService {
...
+      path.join(SITE_CHECKER_REPO_PATH, repoName, "log.csv"),
+      "utf8"
+    )
+
+    const report: string[] = []
+
+    Papa.parse(csvData, {
+      header: true,
+      complete(results) {
+        results.data.forEach((row: any) => {
```
seaerchin, line 507: please add a type here - we can just cast but having `any` is prone to error and doesn't preserve information

seaerchin (approved):
didn't see if the `a/b` variable name was fixed but i'm assuming it is.

do fix the `async forEach` + `any` prior to merge so that they don't stick around for long in our codebase

src/services/review/RepoCheckerService.ts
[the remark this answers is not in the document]
kishore03109: hey i dont have a strong argument apart from fear of using compute resources. ie i dont want to accidentally ddos us by running a program to parse 600+ folders at once. 

but actually this seems like a pretty trivial worry since node should handle this for us + io ops have pretty generous limits. 

i am prioritising pushing the current working prototype to prod first + and then rethink this/see if can use your gh actions idea

kishore03109:
> hmm, no not exactly - the error here is a derived state of our repos.
we shouldn't need to declare it at hte top and then do a .push - this loses locality and makes us prone to errors (scope is within lifetime of function).
we might, instead, want to have smaller methods that just return each category of error and we do the combine ourselves:

hmm wait arh while this is more declarative it is a lot less performant leh, and will scale up to (c*n) time complexity, where c is the number of errors that we acknowledge to exist! maybe a bit of context here is that in our user interviews, they did mention time taken to run the scan is a impt factor to whether they will use this feature, and actually after looking at the errors we are noticing multiple types of errors, which we might need to further separate. while we havent made a final decision yet, i do want to maintain this pattern for performance reasons, lmk if you have concerns against this

src/services/review/RepoCheckerService.ts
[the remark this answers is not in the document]
seaerchin: I think we would add a gh action for every repo to check on push to staging and pr to masterz so this wouldn't be an issue.

There wouldn't be a ddos concern because this should be a script or docker container that we can just pull and use as is off of github's resources

seaerchin:
> hmm wait arh while this is more declarative it is a lot less performant leh, and will scale up to (c*n) time complexity, where c is the number of errors that we acknowledge to exist which is a growing number as we are doing more user interviews! maybe a bit of context here is that in our user interviews, they did mention time taken to run the scan is a impt factor to whether they will use this feature, and actually after looking at the errors we are noticing multiple types of errors, which we might need to further separate. while we havent made a final decision on the type of errors we want, i do want to maintain this pattern for performance reasons, lmk if you have concerns against this

I think splitting the errors up into smaller methods and calling them shouldn't incur extra runtime - the check now is just consolidated into 1 method, which is significantly harder to reason about and extend, and each individual check still takes time.

Under our current model, we do image etc + a href check (2) in 1 pass against all internal links (n).

If we separate out the check so that we call the check by each internal link and not by the list of internal links, it should still be the same run time (was just using an eg previously)

src/services/review/RepoCheckerService.ts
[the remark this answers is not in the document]
kishore03109: i do intend to try this but gh actions are known to take a while and we have a realtime requirement of 1 minute to meet, can try later and see if this works as intended 

also, there is a push to reduce dep on gh, so i am thinking of timeboxing this effort too, thoughts?

kishore03109:
### Merge activity

* **Feb 22, 3:02 AM EST**: @kishore03109 started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/isomerpages/isomercms-backend/1121).

PR #1122, opened 2024-02-07
allow apng ext

harishv7:
## Problem

team wants to upload "APNG" format, the extension is still in ".png", but when we validate the sanitised file format it is interpreted as APNG and is blocked from uploading. Suggested using GIF but their reason was the following:

But because GIF has limited colour palette and low resolution, the usability CoP advised us to change from gif to apng for better quality

Which makes sense actually. I saw their sample file and the resolution is much better

## Solution

Whitelist apng


seaerchin:
great picture

seaerchin (approved):
lgtm - we might wanna check if jekyll can build this asset! jekyll's [site](https://jekyllrb.com/docs/assets/) doesn't say sadly enuf so we have to try ourselves ._.

PR #1128, opened 2024-02-14
Fix/validate site name

alexanderleegs:
Adds a middleware to validate proper site name for associated routes. This is a quick fix solution - see https://opengovproducts.slack.com/archives/C06FXS46AFP/p1707916279240099 for details.


seaerchin (approved):
Lgtm, are we not able to attach the middleware to the router as a whole?

alexanderleegs:
> Lgtm, are we not able to attach the middleware to the router as a whole?

Hmm in order to retrieve `siteName` from params the middleware must be used in a route has already defined a `:siteName`  in the route! So I put them in only after they were referenced

PR #1148, opened 2024-02-20
feat(routes): add link checker routes

kishore03109:
## Problem

adding backend routes to allow for site checker to be done via the fe. corresponding fe pr will be up soon
Closes [insert issue #]

## Solution
hooking up existing func call done for form sg into a route to be called form fe.  gb will be introduced in the fe so as to not introduce 4xx errors. this is not private info anyways, dont really care if an agency is able to view their own site's broken links just by querying this path. 

a locking pattern is added to prevent multiple runs of the same repo state that might use unnecessary compute. 

there are two main api calls, a checker and a result getter. the front end will be call the checker through a manual button for now, (this is user triggered). the getter will be called once every 10 seconds. (or window refetch i dont have a strong opinion here). 

### Newer Changes

After offline disc with @harishv7 
1. Error states will persist when there is an error during a trigger, by the existence of an `error.log` file
2. We want to enable users to be self-reliant, so we allow them to retry during error states. By extension, `error.log` will be deleted + the function will be retried during this session. (DANGER, but I cannot seem to think of a case where this would lead to recursive calls, lmk if this is ok)
3. The two main function calls in `RepoCheckerService` are `readRepoErrors` and `runBrokenLinkChecker`. For observability, am wrapping an overall `mapErr` to log with the specific string that starts with `SiteCheckerError: `. This will be useful for devs to monitor error states from this newly rolled out feature.
4. only push to github is only done as i am returning the result back to the user. this allows for prod + design to look easily while not affecting user flow.

**Breaking Changes**

  - Details ...

### DIFF
--- a/.aws/deploy/backend-task-definition.prod.json
+++ b/.aws/deploy/backend-task-definition.prod.json
@@ -181,6 +181,10 @@
         {
           "name": "UPTIME_ROBOT_API_KEY",
           "valueFrom": "PROD_UPTIME_ROBOT_API_KEY"
+        },
+        {
+          "name": "SITE_CHECKER_FORM_KEY",
+          "valueFrom": "PROD_SITE_CHECKER_FORM_KEY"
         }
       ],
       "logConfiguration": {
--- a/src/routes/v2/authenticated/sites.ts
+++ b/src/routes/v2/authenticated/sites.ts
@@ -295,7 +295,7 @@ export class SitesRouter {
       "/:siteName/checkLinks",
       routeCheckerMiddleware.verifySiteName,
       attachSiteHandler,
-      this.authorizationMiddleware.verifySiteAdmin,
+      this.authorizationMiddleware.verifySiteMember,
       attachReadRouteHandlerWrapper(this.checkLinks)
     )
 
--- a/src/routes/v2/authenticated/sites.ts
+++ b/src/routes/v2/authenticated/sites.ts
@@ -210,7 +210,7 @@ export class SitesRouter {
     if (result.isOk()) {
       return res.status(200).json(result.value)
     }
-    return res.status(404).json({ status: "error" })
+    return res.status(400).json({ status: "error" })
   }
 
   getRouter() {
--- a/src/routes/v2/authenticated/sites.ts
+++ b/src/routes/v2/authenticated/sites.ts
@@ -50,7 +50,6 @@ export class SitesRouter {
     infraService,
     repoCheckerService,
   }: SitesRouterProps) {
-    console.log({ repoCheckerService }, "in sites router")
     this.sitesService = sitesService
     this.authorizationMiddleware = authorizationMiddleware
     this.statsMiddleware = statsMiddleware
--- a/src/routes/v2/authenticated/sites.ts
+++ b/src/routes/v2/authenticated/sites.ts
@@ -212,7 +212,7 @@ export class SitesRouter {
     return res.status(404).json({ status: "error" })
   }
 
-  checkLinks: RequestHandler<
+  triggerCheckLinks: RequestHandler<
     { siteName: string },
     BrokenLinkErrorDto | ResponseErrorBody,
     never,
@@ -296,7 +296,7 @@ export class SitesRouter {
       routeCheckerMiddleware.verifySiteName,
       attachSiteHandler,
       this.authorizationMiddleware.verifySiteMember,
-      attachReadRouteHandlerWrapper(this.checkLinks)
+      attachReadRouteHandlerWrapper(this.triggerCheckLinks)
     )
 
     // The /sites/preview is a POST endpoint as the frontend sends
--- a/src/routes/v2/authenticated/sites.ts
+++ b/src/routes/v2/authenticated/sites.ts
@@ -18,7 +18,7 @@ import { ResponseErrorBody } from "@root/types/dto/error"
 import { ProdPermalink, StagingPermalink } from "@root/types/pages"
 import { PreviewInfo } from "@root/types/previewInfo"
 import { RepositoryData } from "@root/types/repoInfo"
-import { RepoErrorDto } from "@root/types/siteChecker"
+import { BrokenLinkErrorDto } from "@root/types/siteChecker"
 import { SiteInfo, SiteLaunchDto } from "@root/types/siteInfo"
 import { StagingBuildStatus } from "@root/types/stagingBuildStatus"
 import type SitesService from "@services/identity/SitesService"
@@ -197,7 +197,7 @@ export class SitesRouter {
 
   getLinkCheckerStatus: RequestHandler<
     { siteName: string },
-    RepoErrorDto | ResponseErrorBody,
+    BrokenLinkErrorDto | ResponseErrorBody,
     never,
     never,
     { userWithSiteSessionData: UserWithSiteSessionData }
--- a/src/routes/v2/authenticated/sites.ts
+++ b/src/routes/v2/authenticated/sites.ts
@@ -209,7 +209,7 @@ export class SitesRouter {
     if (result.isOk()) {
       return res.status(200).json(result.value)
     }
-    return res.status(400).json({ status: "error" })
+    return res.status(404).json({ status: "error" })
   }
 
   checkLinks: RequestHandler<
--- a/src/routes/v2/authenticated/sites.ts
+++ b/src/routes/v2/authenticated/sites.ts
@@ -203,6 +203,23 @@ export class SitesRouter {
     { userWithSiteSessionData: UserWithSiteSessionData }
   > = async (req, res) => {
     const { userWithSiteSessionData } = res.locals
+    const result = await this.repoCheckerService.readRepoErrors(
+      userWithSiteSessionData.siteName
+    )
+    if (result.isOk()) {
+      return res.status(200).json(result.value)
+    }
+    return res.status(400).json({ status: "error" })
+  }
+
+  checkLinks: RequestHandler<
+    { siteName: string },
+    BrokenLinkErrorDto | ResponseErrorBody,
+    never,
+    never,
+    { userWithSiteSessionData: UserWithSiteSessionData }
+  > = async (req, res) => {
+    const { userWithSiteSessionData } = res.locals
     const result = await this.repoCheckerService.checkRepo(
       userWithSiteSessionData.siteName
     )
@@ -273,6 +290,13 @@ export class SitesRouter {
       attachSiteHandler,
       this.authorizationMiddleware.verifySiteMember,
       attachReadRouteHandlerWrapper(this.getLinkCheckerStatus)
+    )
+    router.post(
+      "/:siteName/checkLinks",
+      routeCheckerMiddleware.verifySiteName,
+      attachSiteHandler,
+      this.authorizationMiddleware.verifySiteAdmin,
+      attachReadRouteHandlerWrapper(this.checkLinks)
     )
 
     // The /sites/preview is a POST endpoint as the frontend sends
--- a/src/routes/v2/authenticated/sites.ts
+++ b/src/routes/v2/authenticated/sites.ts
@@ -12,11 +12,13 @@ import { attachSiteHandler } from "@root/middleware"
 import { RouteCheckerMiddleware } from "@root/middleware/routeChecker"
 import { StatsMiddleware } from "@root/middleware/stats"
 import InfraService from "@root/services/infra/InfraService"
+import RepoCheckerService from "@root/services/review/RepoCheckerService"
 import type { RequestHandler } from "@root/types"
 import { ResponseErrorBody } from "@root/types/dto/error"
 import { ProdPermalink, StagingPermalink } from "@root/types/pages"
 import { PreviewInfo } from "@root/types/previewInfo"
 import { RepositoryData } from "@root/types/repoInfo"
+import { RepoErrorDto } from "@root/types/siteChecker"
 import { SiteInfo, SiteLaunchDto } from "@root/types/siteInfo"
 import { StagingBuildStatus } from "@root/types/stagingBuildStatus"
 import type SitesService from "@services/identity/SitesService"
@@ -26,6 +28,7 @@ type SitesRouterProps = {
   infraService: InfraService
   authorizationMiddleware: AuthorizationMiddleware
   statsMiddleware: StatsMiddleware
+  repoCheckerService: RepoCheckerService
 }
 
 // eslint-disable-next-line import/prefer-default-export
@@ -38,16 +41,21 @@ export class SitesRouter {
 
   private readonly infraService
 
+  private readonly repoCheckerService
+
   constructor({
     sitesService,
     authorizationMiddleware,
     statsMiddleware,
     infraService,
+    repoCheckerService,
   }: SitesRouterProps) {
+    console.log({ repoCheckerService }, "in sites router")
     this.sitesService = sitesService
     this.authorizationMiddleware = authorizationMiddleware
     this.statsMiddleware = statsMiddleware
     this.infraService = infraService
+    this.repoCheckerService = repoCheckerService
     // We need to bind all methods because we don't invoke them from the class directly
     autoBind(this)
   }
@@ -188,6 +196,23 @@ export class SitesRouter {
     return res.status(404).json({ message: "Unable to get staging status" })
   }
 
+  getLinkCheckerStatus: RequestHandler<
+    { siteName: string },
+    RepoErrorDto | ResponseErrorBody,
+    never,
+    never,
+    { userWithSiteSessionData: UserWithSiteSessionData }
+  > = async (req, res) => {
+    const { userWithSiteSessionData } = res.locals
+    const result = await this.repoCheckerService.checkRepo(
+      userWithSiteSessionData.siteName
+    )
+    if (result.isOk()) {
+      return res.status(200).json(result.value)
+    }
+    return res.status(404).json({ status: "error" })
+  }
+
   getRouter() {
     const router = express.Router({ mergeParams: true })
     const routeCheckerMiddleware = new RouteCheckerMiddleware()
@@ -242,6 +267,13 @@ export class SitesRouter {
       attachSiteHandler,
       this.authorizationMiddleware.verifySiteMember,
       attachReadRouteHandlerWrapper(this.getUserStagingSiteBuildStatus)
+    )
+    router.get(
+      "/:siteName/getLinkCheckerStatus",
+      routeCheckerMiddleware.verifySiteName,
+      attachSiteHandler,
+      this.authorizationMiddleware.verifySiteMember,
+      attachReadRouteHandlerWrapper(this.getLinkCheckerStatus)
     )
 
     // The /sites/preview is a POST endpoint as the frontend sends
--- a/src/routes/v2/authenticated/sites.ts
+++ b/src/routes/v2/authenticated/sites.ts
@@ -220,7 +220,7 @@ export class SitesRouter {
     { userWithSiteSessionData: UserWithSiteSessionData }
   > = async (req, res) => {
     const { userWithSiteSessionData } = res.locals
-    const result = await this.repoCheckerService.checkRepo(
+    const result = await this.repoCheckerService.runBrokenLinkChecker(
       userWithSiteSessionData.siteName
     )
     if (result.isOk()) {
--- a/src/services/review/RepoCheckerService.ts
+++ b/src/services/review/RepoCheckerService.ts
@@ -9,7 +9,7 @@ import { marked } from "marked"
 import { ResultAsync, errAsync, ok, okAsync } from "neverthrow"
 import Papa from "papaparse"
 import { ModelStatic } from "sequelize"
-import { PushResult, SimpleGit } from "simple-git"
+import { PullResult, PushResult, SimpleGit } from "simple-git"
 
 import config from "@root/config/config"
 import { EFS_VOL_PATH_STAGING, STAGING_BRANCH } from "@root/constants"
@@ -43,6 +43,9 @@ export const SITE_CHECKER_REPO_PATH = path.join(
 )
 
 const LOCK_CONTENT = "checker.lock"
+const REPO_ERROR_LOG = "error.log"
+const REPO_LOG = "log.csv"
+
 export default class RepoCheckerService {
   private readonly siteMemberRepository: RepoCheckerServiceProps["siteMemberRepository"]
 
@@ -83,8 +86,12 @@ export default class RepoCheckerService {
   }
 
   createLock(repo: string): ResultAsync<PushResult, SiteCheckerError> {
+    const folderPath = path.join(SITE_CHECKER_REPO_PATH, repo)
+    if (!fs.existsSync(folderPath)) {
+      fs.mkdirSync(folderPath, { recursive: true })
+    }
     // create a checker.lock file
-    const lockFilePath = path.join(SITE_CHECKER_REPO_PATH, repo, LOCK_CONTENT)
+    const lockFilePath = path.join(folderPath, LOCK_CONTENT)
     return ResultAsync.fromPromise(
       fs.promises.writeFile(lockFilePath, "locked"),
       (error) => new SiteCheckerError(`${error}`)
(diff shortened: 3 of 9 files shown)


kishore03109:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/isomerpages/isomercms-backend/1148?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* `develop`
  * **PR #1118** <a href="https://app.graphite.dev/github/pr/isomerpages/isomercms-backend/1118?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #1119** <a href="https://app.graphite.dev/github/pr/isomerpages/isomercms-backend/1119?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #1121** <a href="https://app.graphite.dev/github/pr/isomerpages/isomercms-backend/1121?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
        * **PR #1148** <a href="https://app.graphite.dev/github/pr/isomerpages/isomercms-backend/1148?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/isomerpages/isomercms-backend/1148?utm_source=stack-comment).

src/services/review/RepoCheckerService.ts
```diff
@@ -64,6 +69,65 @@ export default class RepoCheckerService {
...
+
+    // check if checker.lock exists
+    const lockFilePath = path.join(logsFilePath, "checker.lock")
+    return ResultAsync.fromPromise(
+      fs.promises.access(lockFilePath, fs.constants.F_OK),
+      () => okAsync(false)
+    )
+      .map(() => true)
+      .orElse(() => ok(false))
+  }
```
seaerchin, line 88: we could make this better by returning an `error` that describes the failure mode instead of just `false`

src/services/review/RepoCheckerService.ts
```diff
@@ -64,6 +69,65 @@ export default class RepoCheckerService {
...
+        ResultAsync.fromPromise(
+          this.git
+            .cwd({ path: SITE_CHECKER_REPO_PATH, root: false })
+            .add([`${path.join(repo, "checker.lock")}`])
+            .commit(`Create lock for ${repo}`)
+            .push(),
+          (error) => new SiteCheckerError(`${error}`)
+        )
+      )
+      .map(() => undefined)
```
seaerchin, line 107: do we need this? i feel that returning the result of the operation might be more useful?

src/services/review/RepoCheckerService.ts
```diff
@@ -64,6 +69,65 @@ export default class RepoCheckerService {
...
+  }
+
+  deleteLock(repo: string): ResultAsync<undefined, SiteCheckerError> {
+    // delete the checker.lock file
+    const lockFilePath = path.join(SITE_CHECKER_REPO_PATH, repo, "checker.lock")
+    return ResultAsync.fromPromise(
+      fs.promises.unlink(lockFilePath),
+      (error) => new SiteCheckerError(`${error}`)
+    )
+      .map(() => undefined)
```
seaerchin, line 117: i think we can omit this `map`?

src/services/review/RepoCheckerService.ts
```diff
@@ -64,6 +69,65 @@ export default class RepoCheckerService {
...
+            .push(),
+          (error) => new SiteCheckerError(`${error}`)
+        )
+      )
+      .map(() => undefined)
+  }
+
+  deleteLock(repo: string): ResultAsync<undefined, SiteCheckerError> {
+    // delete the checker.lock file
+    const lockFilePath = path.join(SITE_CHECKER_REPO_PATH, repo, "checker.lock")
```
seaerchin, line 112: should `checker.lock` just be a constnat?

src/services/review/RepoCheckerService.ts
```diff
@@ -64,6 +69,65 @@ export default class RepoCheckerService {
...
+        ResultAsync.fromPromise(
+          this.git
+            .cwd({ path: SITE_CHECKER_REPO_PATH, root: false })
+            .add([`${path.join(repo, "checker.lock")}`])
+            .commit(`Delete lock for ${repo}`)
+            .push(),
+          (error) => new SiteCheckerError(`${error}`)
+        )
+      )
+      .map(() => undefined)
```
seaerchin, line 128: same here

src/services/review/RepoCheckerService.ts
```diff
@@ -190,6 +254,11 @@ export default class RepoCheckerService {
   }
 
   remover(repo: string, repoExists: boolean) {
+    const NODE_ENV = config.get("env")
+    if (NODE_ENV === "prod") {
+      // by right this should not happen, but just in case
+      return okAsync(undefined)
+    }
```
seaerchin, line 261: sorry, would it be possible to get a more complete explanation here?

src/services/review/RepoCheckerService.ts
```diff
@@ -346,24 +418,56 @@ export default class RepoCheckerService {
         .commit("site checker logs")
         .push(),
       (error) => new SiteCheckerError(`${error}`)
-    )
+    ).map(() => errors)
   }
 
-  checkRepo(repo: string): ResultAsync<undefined, SiteCheckerError> {
+  checkRepo(repo: string): ResultAsync<RepoErrorDto, RepoErrorDto> {
```
seaerchin, line 424: hmm, this looks abit weird - should we return the same type of error if it succeeds?

src/services/review/RepoCheckerService.ts
```diff
@@ -346,24 +418,56 @@ export default class RepoCheckerService {
...
+
+          return this.checker(repo)
+            .andThen((errors) => this.reporter(repo, errors))
+            .map<RepoErrorDto>((errors) => ({
+              status: "success",
+              errors,
+            }))
+
+            .andThen((errors) =>
+              this.remover(repo, repoExists).andThen(() => ok(errors))
```
seaerchin, line 452: double checking - `remover` returns a `resultAsync`? if so, can return directly

src/services/review/RepoCheckerService.ts
```diff
@@ -346,24 +418,56 @@ export default class RepoCheckerService {
...
+            )
+        })
+      )
+      .orElse((error) => {
+        if (error.status === "loading") {
+          return okAsync<RepoErrorDto>({
+            status: "loading",
+          })
+        }
+        this.deleteLock(repo)
```
seaerchin, line 465: in this case, the lock is preserved only if the `status === "loading"` which is fine irght?

seaerchin (commented):
minor nits on code; apart from the nits, the mental model here seems to be a github action, no?

seaerchin:
could we add the required deps to the tests?

src/services/review/RepoCheckerService.ts
kishore03109: but it has no use for the variable errors, so am passing this through to make the dependencies explicit!

src/routes/v2/authenticated/sites.ts
```diff
@@ -188,6 +196,23 @@ export class SitesRouter {
...
+    { userWithSiteSessionData: UserWithSiteSessionData }
+  > = async (req, res) => {
+    const { userWithSiteSessionData } = res.locals
+    const result = await this.repoCheckerService.checkRepo(
+      userWithSiteSessionData.siteName
+    )
+    if (result.isOk()) {
+      return res.status(200).json(result.value)
+    }
+    return res.status(400).json({ status: "error" })
```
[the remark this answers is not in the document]
kishore03109: like the error toast ah?

src/routes/v2/authenticated/sites.ts
```diff
@@ -38,16 +41,21 @@ export class SitesRouter {
...
+  private readonly repoCheckerService
+
   constructor({
     sitesService,
     authorizationMiddleware,
     statsMiddleware,
     infraService,
+    repoCheckerService,
   }: SitesRouterProps) {
+    console.log({ repoCheckerService }, "in sites router")
```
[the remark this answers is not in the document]
kishore03109: opps removed

src/services/review/RepoCheckerService.ts
kishore03109: correct, since some other process is currently running this check!

src/services/review/RepoCheckerService.ts
kishore03109: thanks for the callout, lmk if still not clear

src/services/review/RepoCheckerService.ts
kishore03109: i think is this weird quirk of neverthrow, types it as void and then void and undefined conflict

src/services/review/RepoCheckerService.ts
kishore03109: yup, rmed

src/services/review/RepoCheckerService.ts
kishore03109: uhm wait ah if not success we error out no? the info is not lost, it returns an error?

src/services/review/RepoCheckerService.ts
kishore03109: wah ok i clearly need to rename this then. The error here was trying to refer to the errors found in the repo which we are going to report to the user, not an error in the cms. will rename it to BrokenLinkErrorDto instead

src/services/review/RepoCheckerService.ts
kishore03109: Hmm may i clarify if this is a nit? 

from the pov of the caller, feels like `ResultAsync<boolean, never>` is better contract than `ResultAsync<true, false>`, since the latter needs to do a  `.orElse(ok(false))`. 
ie an error thrown here is not a sign of something went wrong in trying to get the result of "is this repo locked" (which signals intent of some error recovery is needed), but rather just the repo not locked at the moment

src/routes/v2/authenticated/sites.ts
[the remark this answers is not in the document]
seaerchin: err not the error toast per se, but more the error message! (which the toast uses to display)

it's ok for now because we're testing this with users

src/services/review/RepoCheckerService.ts
seaerchin: this is a nit in the context of this PR (get this out to proto); `ResultAsync` contains 2 types in itself - we want to return `ResultAsync<true, someError>` because the error itself is more descriptive that `false`. taken in context, this means that: 

1. if i return `ResultAsync<success, false>` - i know only about the presence of an error, but idk _what_ error occurred. 
2. if i return `ResultAsync<boolean, never>` - `neverthrow` isn't adding much value here. the type signature indicates that this doesn't throw so why not just return `boolean`? (but then same point as above) 

we don't necessarily have to fix this _now_ cos this is a proto

src/services/review/RepoCheckerService.ts
seaerchin: `map` only ever operates on success values! you might be looking for `mapErr` here

src/services/review/RepoCheckerService.ts
seaerchin: not entirely sure - i think we can fix this just by hinting the return type as `void` above adn omitting this step. i think this is a quick fix that eliminates uncertainty in our codebase, which is why i pointed it out!

src/services/review/RepoCheckerService.ts
seaerchin: ah ok sgtm

src/services/review/RepoCheckerService.ts
```diff
@@ -439,28 +437,32 @@ export default class RepoCheckerService {
...
         if (error.status === "loading") {
-          return okAsync<RepoErrorDto>({
+          return okAsync<BrokenLinkErrorDto>({
             status: "loading",
           })
         }
         this.deleteLock(repo)
         logger.error(`Error running site checker for repo ${repo}: ${error}`)
-        return errAsync<RepoErrorDto, RepoErrorDto>({
+        return errAsync<BrokenLinkErrorDto, BrokenLinkErrorDto>({
```
seaerchin: honestly, this is my only gripe atm - returning an error **in both success and error** states is very confusing.

there should eb some differences between them, no?

seaerchin (commented):
1 fix

src/services/review/RepoCheckerService.ts
kishore03109: good point. am changing this to return the push result instead, rather than void

src/services/review/RepoCheckerService.ts
kishore03109: changed api for this

src/services/review/RepoCheckerService.ts
kishore03109: calling this directly at the router level, and fe expect error dto mah. 

how would you structure this ah? cuz its either this or at the router call site, we need to define the shape which feels like a weird place to construct the dto

kishore03109:
> the mental model here seems to be a github action

@seaerchin no leh i only exploring that if we cannot sustain the compute load, for now this one is just run in the be

src/routes/v2/authenticated/sites.ts
```diff
@@ -188,6 +195,40 @@ export class SitesRouter {
...
+    { userWithSiteSessionData: UserWithSiteSessionData }
+  > = async (req, res) => {
+    const { userWithSiteSessionData } = res.locals
+    const result = await this.repoCheckerService.readRepoErrors(
+      userWithSiteSessionData.siteName
+    )
+    if (result.isOk()) {
+      return res.status(200).json(result.value)
+    }
+    return res.status(400).json({ status: "error" })
```
harishv7, line 212: Is this really a 400 status? This seems to be a catch all instead of error handling we can send to the FE

harishv7: What are the possible failure modes in this?

src/routes/v2/authenticated/sites.ts
```diff
@@ -188,6 +195,40 @@ export class SitesRouter {
...
+
+  checkLinks: RequestHandler<
+    { siteName: string },
+    BrokenLinkErrorDto | ResponseErrorBody,
+    never,
+    never,
+    { userWithSiteSessionData: UserWithSiteSessionData }
+  > = async (req, res) => {
+    const { userWithSiteSessionData } = res.locals
+    const result = await this.repoCheckerService.checkRepo(
```
harishv7, line 223: I was thinking this will be async? But looks like a sync response here to the user

src/services/review/RepoCheckerService.ts
```diff
@@ -53,17 +55,69 @@ export default class RepoCheckerService {
...
   }: RepoCheckerServiceProps) {
     this.siteMemberRepository = siteMemberRepository
-    this.sitesService = sitesService
+
     this.gitFileSystemService = gitFileSystemService
     this.git = git
     this.repoRepository = repoRepository
   }
 
+  isCurrentlyLocked(repo: string): ResultAsync<true, SiteCheckerError> {
```
harishv7, line 68: We seem to be having multiple locks. There is a DDB, then we introduced a GIT efs lock on repo level and this is a new lock. Is there a way to simplify this?

src/services/review/RepoCheckerService.ts
```diff
@@ -189,8 +243,12 @@ export default class RepoCheckerService {
       .mapErr((error) => new SiteCheckerError(`${error}`))
   }
 
-  remover(repo: string, repoExists: boolean) {
-    if (repoExists) {
+  // This function removes the recently cloned repo. This function should only be called in non-prod environments
+  remover(repo: string) {
```
harishv7, line 247: When's this function used?

src/services/review/RepoCheckerService.ts
```diff
@@ -346,24 +407,99 @@ export default class RepoCheckerService {
         .commit("site checker logs")
         .push(),
       (error) => new SiteCheckerError(`${error}`)
-    )
+    ).map(() => errors)
+  }
+
+  // adapted from https://stackoverflow.com/questions/56427009/how-to-return-papa-parsed-csv-via-promise-async-await
+  csvToRepoError = (repo: string): Promise<RepoError[]> => {
```
harishv7, line 414: The name was confusing at first maybe can use something like generateRepoErrorsFromCsv

src/services/review/RepoCheckerService.ts
kishore03109: anywhere but prod. 
enables running of this in dev mode + testing by the team in staging.

src/services/review/RepoCheckerService.ts
kishore03109: you bring a good point, but the three locks here solve varying levels of locking.
1. git locks is for not each git operation
2. ddb lock is at a repo level (eg it is possible for a save to end up committing twice)
3. repo checking lock, which is at this time, not related to any repo changes. While that could be linked in the furture, i rather keep it simple for to get the initial traction

src/routes/v2/authenticated/sites.ts
kishore03109: good point, seems like 404 is more semantically meaninful here. ie the list of possible errors that is supposed to exist is not found. 

failure modes -> multiple, but the as of the current state, there is only one recovery path (eg, by re-reunning the checker again)

kishore03109: [d1d1814](https://github.com/isomerpages/isomercms-backend/pull/1148/commits/d1d181420865d3535d44d952a88589ce12cb9047) changed here

src/routes/v2/authenticated/sites.ts
kishore03109: i dont understand this comment, will chat offline

src/services/review/RepoCheckerService.ts
harishv7: okay fair enough

src/routes/v2/authenticated/sites.ts
harishv7: understand we are returning a loading state if the lock is present, resolving this comment

harishv7 (dismissed):

src/services/review/RepoCheckerService.ts
```diff
@@ -346,24 +407,103 @@ export default class RepoCheckerService {
...
      * To allow for an easy running of the script, we can temporarily clone
      * the repo to the EFS
      */
-
     let repoExists = true
+    return this.isCurrentlyLocked(repo)
+      .andThen(() => okAsync<BrokenLinkErrorDto>({ status: "loading" }))
+      .orElse(() => {
+        // this process can run async, so we can just return the loading state early
+        this.createLock(repo)
```
harishv7, line 473: we are not awaiting here because we want to do this async right?

harishv7: So if there are errors during this async run, it will just fail?

kishore03109: actually not wrong uh in this case will fail silently

kishore03109: but users are able to retry, which is i think more impt here than showing error state

harishv7: > but users are able to retry, which is i think more impt here than showing error state
So whats' the user flow? 
They start the run but if it fails, they just see a empty state? Or do they know if there is an error and they need to retry?

kishore03109: on error we are redirecting over to the dashboard

.aws/deploy/backend-task-definition.prod.json
```diff
@@ -181,6 +181,10 @@
         {
           "name": "UPTIME_ROBOT_API_KEY",
           "valueFrom": "PROD_UPTIME_ROBOT_API_KEY"
+        },
+        {
+          "name": "SITE_CHECKER_FORM_KEY",
```
harishv7, line 186: is this a new env var? note that this needs to be applied for both ECS and EB for time being. Has it been added to SSM + updated in the fetch ssm params scripts?

src/routes/v2/authenticated/sites.ts
```diff
@@ -188,6 +195,40 @@ export class SitesRouter {
...
+    { userWithSiteSessionData: UserWithSiteSessionData }
+  > = async (req, res) => {
+    const { userWithSiteSessionData } = res.locals
+    const result = await this.repoCheckerService.readRepoErrors(
+      userWithSiteSessionData.siteName
+    )
+    if (result.isOk()) {
+      return res.status(200).json(result.value)
+    }
+    return res.status(404).json({ status: "error" })
```
harishv7, line 212: status could be better phrased e.g. error: no log files found

src/routes/v2/authenticated/sites.ts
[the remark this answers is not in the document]
harishv7: @kishore03109 are we updating the status code/error msg here?

src/services/review/RepoCheckerService.ts
```diff
@@ -53,17 +58,102 @@ export default class RepoCheckerService {
...
+    const errorLogFilePath = path.join(
+      SITE_CHECKER_REPO_PATH,
+      repo,
+      REPO_ERROR_LOG
+    )
+    return ResultAsync.fromPromise(
+      fs.promises.access(errorLogFilePath, fs.constants.F_OK),
+      (err) => new SiteCheckerError(`${err}`)
+    )
+      .map(() => true)
```
harishv7, line 136: should this be okAsync(true)?

src/services/review/RepoCheckerService.ts
```diff
@@ -53,17 +58,102 @@ export default class RepoCheckerService {
...
+
+  isCurrentlyErrored(repo: string): ResultAsync<boolean, SiteCheckerError> {
+    const errorLogFilePath = path.join(
+      SITE_CHECKER_REPO_PATH,
+      repo,
+      REPO_ERROR_LOG
+    )
+    return ResultAsync.fromPromise(
+      fs.promises.access(errorLogFilePath, fs.constants.F_OK),
+      (err) => new SiteCheckerError(`${err}`)
```
harishv7, line 134: just sanity check - all errors when returned to FE are remapped right? so that we don't show weird errors like filesystem errors on FE

src/services/review/RepoCheckerService.ts
```diff
@@ -316,54 +410,170 @@ export default class RepoCheckerService {
...
+            okAsync<BrokenLinkErrorDto>({
+              errors,
+              status: "success",
+            })
+          )
+        )
+    })
+  }
+
+  checkRepo(repo: string): ResultAsync<BrokenLinkErrorDto, SiteCheckerError> {
```
harishv7, line 509: checkRepo is mapped to the POST call made by the user clicking on the run site check?

src/services/review/RepoCheckerService.ts
```diff
@@ -316,54 +410,170 @@ export default class RepoCheckerService {
...
+
+    // time the process taken to check the repo
+    const start = process.hrtime()
+    return this.isCurrentlyLocked(repo)
+      .andThen(() => okAsync<BrokenLinkErrorDto>({ status: "loading" }))
+      .orElse(() =>
+        this.isCurrentlyErrored(repo).andThen((isError) => {
+          if (isError) {
+            // delete the error.log file + retry, safe operation since this is only user triggered and not automated
+            return this.deleteErrorLog(repo).andThen(() => this.checkRepo(repo))
```
harishv7, line 525: at this point why must recurse again? just delete the error log and continue as per normal?

.aws/deploy/backend-task-definition.prod.json
kishore03109: yup this has been added [here](https://github.com/isomerpages/isomercms-backend/pull/1169/files#diff-6c4a4e2ae0fb03622002fcaf7aea00bf45c7bd2d584314ff1f60eeeef8ce3894R86).
this just was added after the rebase.

the ssm already contains this as well 
Screenshot 2024-02-28 at 9 25 44 AM

src/routes/v2/authenticated/sites.ts
kishore03109: hmm wait this the shape of the error dto. for now our fe only has one error state, which the frontend takes care of, so am inclined to keep it simple for the initial rollout.

additionally, this could have multiple failure modes, eg file could not be parsed
but it is no meaningful to return these varying error states since from the user pov, there is only one way to recover, which is to retry

src/services/review/RepoCheckerService.ts
kishore03109: good qn, this is done at the route level. if the result is of an `err` variant, will be returning 
`return res.status(404).json({ status: "error" })`

src/services/review/RepoCheckerService.ts
kishore03109: are we confusing `map` with `andThen`? 

this would be similar to 
```
ResultAsync.fromPromise(
      fs.promises.access(errorLogFilePath, fs.constants.F_OK),
      (err) => new SiteCheckerError(`${err}`)
    )
      .andThen(() => okAsync(true))
      .orElse(() => okAsync(false))
```

preference for map here ya, less verbose

src/services/review/RepoCheckerService.ts
kishore03109: yes, this is the call that updates our internal state

src/services/review/RepoCheckerService.ts
kishore03109: fair, have modified code [here](https://github.com/isomerpages/isomercms-backend/pull/1148/commits/621e0ea37b6815c81c9ca37ada6d2057d79a620e)

.aws/deploy/backend-task-definition.prod.json
harishv7: @kishore03109 i don't see any script files being modified in the above PR you shared

src/services/review/RepoCheckerService.ts
```diff
@@ -316,54 +410,130 @@ export default class RepoCheckerService {
...
     if (!fs.existsSync(filePath)) {
       fs.appendFileSync(filePath, stringifiedErrors)
     } else {
       fs.writeFileSync(filePath, stringifiedErrors)
     }
-
+    // pushing since there is no real time requirement for user
     return ResultAsync.fromPromise(
       this.git
         .cwd({ path: SITE_CHECKER_REPO_PATH, root: false })
```
harishv7, line 441: Should you assert what branch you are on first?

src/services/review/RepoCheckerService.ts
```diff
@@ -316,54 +410,130 @@ export default class RepoCheckerService {
...
+        .fetch()
+        .merge(["origin/main"])
```
harishv7, line 443: What's this merge for? Will there be conflicts?

kishore03109: yes, conflicts due to cron job that runs + commits for every push. (https://github.com/isomerpages/isomer-site-checker/)

src/services/review/RepoCheckerService.ts
kishore03109: sure [b43b410](https://github.com/isomerpages/isomercms-backend/pull/1148/commits/b43b410281fb3a00b8ac5118e0094424e3f20bd0)

src/services/review/RepoCheckerService.ts
harishv7: @kishore03109 so how are conflicts resolved? what's the merge strategy? do we need to specify that?

harishv7 (approved):

PR #1149, opened 2024-02-20
Chore/lock repos when repairing

alexanderleegs:
This PR locks the repo during the time site repair is being done, to prevent editing by users during this period.

### DIFF
--- a/src/utils/mutex-utils.js
+++ b/src/utils/mutex-utils.js
@@ -37,7 +37,7 @@ const mockUnlock = (siteName) => {
 
 const lock = async (siteName, lockLengthSeconds = 60) => {
   try {
-    const ONE_MIN_FROM_CURR_DATE_IN_SECONDS_FROM_EPOCH_TIME =
+    const expiryTime =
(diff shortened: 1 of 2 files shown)


src/utils/mutex-utils.js
```diff
@@ -35,10 +35,10 @@ const mockUnlock = (siteName) => {
   mockMutexObj[siteName] = false
 }
 
-const lock = async (siteName) => {
+const lock = async (siteName, lockLengthSeconds = 60) => {
   try {
     const ONE_MIN_FROM_CURR_DATE_IN_SECONDS_FROM_EPOCH_TIME =
```
dcshzj, line 40: nit: the "one min" isn't true anymore, might want to rename the variable to avoid confusion?

dcshzj (approved):

src/utils/mutex-utils.js
alexanderleegs: Fixed in https://github.com/isomerpages/isomercms-backend/pull/1149/commits/0cc6344c3cb654f3f7fd5dc3b009b514ccf9aa23!

PR #1156, opened 2024-02-21
fix(dompurify): further limit src

kishore03109:
copy paste to backend for fe code [here](https://github.com/isomerpages/isomercms-frontend/pull/1809)

### DIFF
--- a/src/services/utilServices/Sanitizer.ts
+++ b/src/services/utilServices/Sanitizer.ts
@@ -1,5 +1,19 @@
 import DOMPurify from "isomorphic-dompurify"
 
+/**
+ * While we do have a CSP in place, we want to further restrict the content that
+ * that the user can input.
+ * NOTE: Any changes to this list should also be updated in frontend code
+ */
+const ALLOWED_SRC = [
+  "//www.instagram.com/embed.js",
+  "https://www.evvochannel.tv/jwplayer7/jwplayer.js",
+  "/jquery/resize-tables.js",
+  "/jquery/jquery.min.js",
+  "/jquery/bp-menu-new-tab.js",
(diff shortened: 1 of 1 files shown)


src/services/utilServices/Sanitizer.ts
```diff
@@ -1,5 +1,19 @@
...
+ * NOTE: Any changes to this list should also be updated in frontend code
+ */
+const ALLOWED_SRC = [
+  "//www.instagram.com/embed.js",
+  "https://www.evvochannel.tv/jwplayer7/jwplayer.js",
+  "/jquery/resize-tables.js",
+  "/jquery/jquery.min.js",
+  "/jquery/bp-menu-new-tab.js",
+  "//s3.amazonaws.com/downloads.mailchimp.com/js/mc-validate.js",
+]
```
dcshzj, line 15: question (blocking): where did you get `evvochannel.tv` and `s3.amazonaws.com` ones from? Those domains are not whitelisted inside our CSP currently.

src/services/utilServices/Sanitizer.ts
```diff
@@ -18,9 +32,14 @@ DOMPurify.setConfig({
 DOMPurify.addHook("uponSanitizeElement", (node, data) => {
   // Allow script tags if it has a src attribute
   // Script sources are handled by our CSP sanitiser
```
dcshzj, line 34: nit: can remove this comment now since we no longer rely on the CSP.

src/services/utilServices/Sanitizer.ts
kishore03109: wah interesting, i did a grep search on existing repos to maintain parity, which is where i got this. Will remove these two + let ops know that this functionality for these sites are broken

dcshzj (approved):

PR #1166, opened 2024-02-23
02 01 build deps add parser as dep

kishore03109:
## Problem

duplicate of #1118. had to rollback during release due to package lock issues that i faced and could not debug immediately and did not want to block release.


timotheeg (approved):
lgtm

PR #1168, opened 2024-02-25
02 01 core parser logic

kishore03109:
## Problem

dup of already approved #1119. this was cut from release due to package lock issues that occurred during release


src/services/review/RepoCheckerService.ts
```diff
@@ -0,0 +1,223 @@
...
+            return
+          }
+          const html = await marked.parse(fileContent)
+          const dom = new JSDOM(html)
+          const anchorTags = dom.window.document.querySelectorAll("a")
+          forEach(anchorTags, (tag) => {
+            const href = tag.getAttribute("href")
+            const text = tag.textContent
+            if (href) {
+              const encodedHref = this.documentPathDecoder(href)
```
timotheeg, line 65: Hm 🤔 , the var name is `encodedHref`, but conditionally `documentPathDecoder()` calls `decodeURIComponent()`, so the result is a decoded href?. In fact based on this code, if `href` is a full url, `documentPathDecoder()` should use `decodeUri()`, rather than `decodeURIComponent()`, no? 🤔

package.json
```diff
@@ -118,7 +118,7 @@
     "@types/jsdom": "^21.1.1",
     "@types/lodash": "^4.14.186",
     "@types/mock-fs": "^4.13.1",
-    "@types/node": "^17.0.21",
+    "@types/node": "^18.19.18",
```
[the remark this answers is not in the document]
kishore03109: Yup, this pr introduces code that uses node fs, which has some type conflicts between versions 17 and 18!

src/services/review/RepoCheckerService.ts
```diff
@@ -0,0 +1,223 @@
...
+  }
+
+  documentPathDecoder(documentPath: string) {
+    for (let i = 0; i < ALLOWED_FILE_EXTENSIONS.length; i += 1) {
+      if (documentPath.endsWith(ALLOWED_FILE_EXTENSIONS[i])) {
+        return decodeURIComponent(documentPath)
+      }
+    }
+    return documentPath
+  }
```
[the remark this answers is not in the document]
kishore03109: hey these are fair points!

here are some context:

1. there are other files (eg, md files) that fe we dont need to decode the content
2. in our codebase, we restrict the types of files being uploaded into the cms, and part of this check ensures that the file extension is no encoded. ie we are guarenteed by the cms that the repo should not have the example in this comment.


happy to huddle if this does not make sense!

src/services/review/RepoCheckerService.ts
kishore03109: 213feca5939bbbcc3ff0453d7496f22c347b5ed0
fair point

src/services/review/RepoCheckerService.ts
```diff
@@ -0,0 +1,223 @@
...
+                `Duplicate permalink ${permalink} found in ${file}`
+              ),
+            ])
+          } else {
+            setOfAllPermalinks.add(permalink)
+          }
+        })
+      })
+
+    await Promise.all(promises)
```
harishv7, line 218: Just wanted to check - intentional use of `all` over `allSettled`? All settled will ensure all promises are fulfilled even if some are rejected

kishore03109: yup, this is the case!

kishore03109: No point short circuiting here, we want other possible successful runs to occur regardless of some error states since they are independent events!

harishv7: @kishore03109 wait shouldn't we be using allSettled per ur description?

> Promise.all() will reject immediately upon any of the input promises rejecting

> Promise.allSettled() will wait for all input promises to complete, regardless of whether or not one rejects.

kishore03109: arh apologies thought this was referring to a different code. yes, this one has to be `.all()` since we need to parse all files to get a good checker result

kishore03109: to be clearer: I thought the code `promise.all()` here was in the function call `checkAllRepos()` (this is introduced in a later pr)

here, for a sane check of the repo,  I would have to parse the entire repo because it would not be possible check if an internal link is valid without the enumeration through all the permalinks in a repo

PR #1172, opened 2024-02-26
feat: get diff from git instead of GitHub

jacobkwan:
## Problem
This PR is part of a wider effort to reduce dependency on GitHub for functionality that GGS/EFS is capable of handling.
Specifically, this PR allows for getting diffs via EFS instead of GitHub.

## Solution
Main changes are made in `ReviewRequestService.compareDiff`, also tried to refactor the existing code for clarity

Highlighting some of the important helper methods added to `GitFileSystemService`:
1. `getFilesChanged`: returns list of filenames representing files that have been changed between staging and master
2. `getLatestCommitOfPath`: getting latest commit on staging for a certain full file path (should work for deleted files as well, unlike current buggy response from GitHub API)
3. `fastForwardMaster`: self-explanatory; called during the merge review request flow since we need to keep local master up to date, now that we are getting diff from EFS

Not sure if I have to update the repair form to handle master branch now too(?) @kishore03109 

## Doubts
Feel free to comment on anything; I have some specific things that I wanted to get feedback on
1. if I'm writing tests wrongly
2. if I'm writing non-idiomatic `neverthrow` code
3. if I missed anything regarding possible ways that master can diverge/failure modes/edge cases

**Breaking Changes**
I believe this PR should be backward compatible.
Since our repos on EFS are not currently set up with `master`, `getGitDiff` also handles creation of the local tracking branch if it doesn't exist yet.

  - Details ...

## Deploy Notes
- [x] Add monitors for errors during `compareDiffLocal` and `fastForwardMaster`
https://ogp.datadoghq.com/monitors/140997853?view=spans
https://ogp.datadoghq.com/monitors/140998024?view=spans

## Testing
Tested in staging
- [ ] Toggle feature flag `is_local_diff_enabled` for account being tested
- [ ] Create a page
- [ ] Edit a page
- [ ] Delete a page
- [ ] Move a page
- [ ] Try to create an RR and check that the diff shown is correct for all files (this tests that /compare works)
- [ ] After creating the RR, navigate to the RR page and check that the diff shown is correct (this tests that /listFullReviewRequest works)
- [ ] Approve and publish the RR
- [ ] Check that diff is empty now (tests that master has been fastforwarded)
- [ ] Finally, last sanity check that the diff still works after making another edit


src/services/review/ReviewRequestService.ts
```diff
@@ -840,6 +836,7 @@ export default class ReviewRequestService {
     )
 
     await this.apiService.mergePullRequest(repoNameInGithub, pullRequestNumber)
+    await this.apiService.mergeStagingToMaster(repoNameInGithub)
```
dcshzj, line 839: question(blocking): correct me if I'm wrong here, but would this create two separate merge commits, one on GitHub and one locally, resulting in divergent master branches?

src/services/db/GitFileSystemService.ts
```diff
@@ -306,6 +307,92 @@ export default class GitFileSystemService {
     )
   }
 
+  /**
+   * Wrapper over `git diff --name-only` that also creates a local tracking branch for `base` and/or `head` if they do not exist.
+   * Note that `base` and `head` should be branch names, not commit hashes or other usually valid arguments for `git diff`.
+   */
+  getGitDiff(
```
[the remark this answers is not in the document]
jacobkwan: https://github.com/isomerpages/isomercms-backend/pull/1172/commits/11ad1275a4f0b4e8feb9624d2524596a94e716a6

src/services/db/GitFileSystemService.ts
```diff
@@ -306,6 +307,92 @@ export default class GitFileSystemService {
...
+  ): ResultAsync<string[], GitFileSystemError> {
+    if (base === head) {
+      return okAsync([])
+    }
+    return this.createLocalTrackingBranchIfNotExists(repoName, base)
+      .andThen(() => this.createLocalTrackingBranchIfNotExists(repoName, head))
+      .andThen(() => this._getGitDiff(repoName, base, head))
+  }
+
+  private _getGitDiff(
```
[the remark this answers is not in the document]
jacobkwan: https://github.com/isomerpages/isomercms-backend/pull/1172/commits/11ad1275a4f0b4e8feb9624d2524596a94e716a6

src/services/review/ReviewRequestService.ts
```diff
@@ -128,96 +129,111 @@ export default class ReviewRequestService {
...
+        ).map((author) => {
+          const lastChangedTime = new Date(latestLog.date).getTime()
           return {
-            ...item,
-            lastEditedBy: mappings[hash]?.author || "Unknown user",
-            lastEditedTime: mappings[hash]?.unixTime || 0,
+            lastEditedBy: author?.email || latestLog.author_email,
+            lastEditedTime: lastChangedTime,
           }
         })
```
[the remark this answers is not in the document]
jacobkwan: https://github.com/isomerpages/isomercms-backend/pull/1172/commits/7d84489dc11cbc11f4c219dec141748578d9d5ef

src/services/review/ReviewRequestService.ts
jacobkwan: yea actually you're right, changed it to pull from master instead, can remove pull request from the whole flow in a separate PR

https://github.com/isomerpages/isomercms-backend/pull/1172/commits/33e665e414587f170036520bd69ac4c0bb619087

src/services/db/GitFileSystemService.ts
```diff
@@ -1686,4 +1816,56 @@ export default class GitFileSystemService {
...
+  createLocalTrackingBranchIfNotExists(repoName: string, branchName: string) {
+    return this.doesLocalBranchExist(repoName, branchName).andThen((exists) => {
+      if (exists) {
+        return okAsync(true)
+      }
+      const efsVolPath = this.getEfsVolPathFromBranch(branchName)
+      return ResultAsync.fromPromise(
+        this.git
+          .cwd({ path: `${efsVolPath}/${repoName}`, root: false })
+          .branch(["--track", branchName, `origin/${branchName}`]),
```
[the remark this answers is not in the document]
jacobkwan: I don't think `git branch --track master origin/master` will do any checking out so this is safe; tested it myself also to be sure

src/services/db/GitFileSystemService.ts
```diff
@@ -306,6 +307,92 @@ export default class GitFileSystemService {
     )
   }
 
+  /**
+   * Wrapper over `git diff --name-only` that also creates a local tracking branch for `base` and/or `head` if they do not exist.
+   * Note that `base` and `head` should be branch names, not commit hashes or other usually valid arguments for `git diff`.
```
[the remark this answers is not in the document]
jacobkwan: oh its just because I do the defensive branch creation inside this method for convenience

src/services/db/GitFileSystemService.ts
```diff
@@ -706,6 +793,49 @@ export default class GitFileSystemService {
...
+        )
+      }
+
+      return (
+        this.ensureCorrectBranch(repoName, "master")
+          .andThen(() =>
+            ResultAsync.fromPromise(
+              this.git
+                .cwd({ path: `${efsVolPath}/${repoName}`, root: false })
+                .merge(["staging", "--no-ff"]), // ensure there is a merge commit
```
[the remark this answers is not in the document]
jacobkwan: no longer using git merge

src/services/db/GitFileSystemService.ts
```diff
@@ -706,6 +793,49 @@ export default class GitFileSystemService {
...
+              (error) => {
+                logger.error(
+                  `Error when merging staging to master for ${repoName}: ${error}`
+                )
```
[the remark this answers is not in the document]
jacobkwan: now that I've changed to pull master, should we have an alarm on `gitFileSystemService.pull` ?
or whats the correct way to monitor this

jacobkwan:
question regarding the implementation of `GitFileSystemService.pull`:
https://github.com/isomerpages/isomercms-backend/blob/d996b4bda16950e047f080f2a939edab80cd2955/src/services/db/GitFileSystemService.ts#L500-L532

not sure why we can ignore those specific error messages? for this feature should I move the `logger.error` to the top so all errors are logged? or is there a better way for monitoring

src/services/db/GitFileSystemService.ts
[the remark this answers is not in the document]
dcshzj: Ah okay, thanks for clarifying! I initially thought that it would switch to the master branch but looks like it doesn't

src/services/db/GitFileSystemService.ts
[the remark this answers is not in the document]
dcshzj: Yeah I think an alarm to monitor for any errors that happen during the pull operation would be good.

dcshzj:
> not sure why we can ignore those specific error messages? for this feature should I move the `logger.error` to the top so all errors are logged? or is there a better way for monitoring

I think this is quite a while back when we were doing a pull operation on every update operation (because it was planned to  have GGS support read operations only). At that time I think we could ignore these errors because multiple pulls were happening concurrently, so some of these pulls would throw an error but the overall operation is successful (since only 1 pull needed to be successful).

This may not be so relevant now (and we only pull when doing GGS repair), so it is worth checking again if these error messages can still be safely ignored.

The `neverthrow` library forces you to handle the error cases, so just make sure to log when there are such errors and create the relevant monitors for these error logs.

jacobkwan:
> > not sure why we can ignore those specific error messages? for this feature should I move the `logger.error` to the top so all errors are logged? or is there a better way for monitoring
> 
> I think this is quite a while back when we were doing a pull operation on every update operation (because it was planned to have GGS support read operations only). At that time I think we could ignore these errors because multiple pulls were happening concurrently, so some of these pulls would throw an error but the overall operation is successful (since only 1 pull needed to be successful).
> 
> This may not be so relevant now (and we only pull when doing GGS repair), so it is worth checking again if these error messages can still be safely ignored.
> 
> The `neverthrow` library forces you to handle the error cases, so just make sure to log when there are such errors and create the relevant monitors for these error logs.

got it, I went to read about those error messages, I think we will want to be alerted if those happen during git pull especially now that its part of the critical path

src/fixtures/review.ts
```diff
@@ -20,6 +20,10 @@ import {
   MOCK_GITHUB_COMMIT_AUTHOR_TWO,
```
harishv7, line 20: What are the failure modes for this feature? Are we considering feature-flags?

You mentioned this is backward-compatible? So if we have a flag to enable this and disable the flag, the site shouldn't be affected right?

src/services/db/GitFileSystemService.ts
```diff
@@ -337,6 +338,86 @@ export default class GitFileSystemService {
     )
   }
 
+  /**
+   * Wrapper over `git diff --name-only` that also creates a local tracking branch for `base` and/or `head` if they do not exist.
```
harishv7, line 342: local tracking branch means? if for example, master does not exist, it creates a branch called master?

btw base/head are symbolic references than branch names right

src/services/db/GitFileSystemService.ts
```diff
@@ -337,6 +338,86 @@ export default class GitFileSystemService {
...
+    repoName: string,
+    base = "master",
+    head = "staging"
+  ): ResultAsync<string[], GitFileSystemError> {
+    if (base === head) {
+      return okAsync([])
+    }
+    return this.createLocalTrackingBranchIfNotExists(repoName, base)
+      .andThen(() => this.createLocalTrackingBranchIfNotExists(repoName, head))
+      .andThen(() => {
```
harishv7, line 355: do we need an orElse here?

src/services/db/GitFileSystemService.ts
```diff
@@ -337,6 +338,86 @@ export default class GitFileSystemService {
...
+              )
+            }
+
+            return new GitFileSystemError("An unknown error occurred")
+          }
+        ).map((files) =>
+          files
+            .trim()
+            .split("\n")
+            .filter((file) => file !== "")
```
harishv7, line 378: some of the files maybe `.keep`?

src/services/db/GitFileSystemService.ts
```diff
@@ -337,6 +338,86 @@ export default class GitFileSystemService {
...
+        )
+      })
+  }
+
+  /**
+   * Get latest commit for a file path on a branch (including deleted files)
+   */
+  getLatestCommitOfPath(
+    repoName: string,
+    path: string,
```
harishv7, line 388: where is the path coming from? Is any part of this coming from a user input?

for e.g. if the path was "../../../<some other repo>/<path>" it should not be giving other repo's info

src/services/db/GitFileSystemService.ts
```diff
@@ -337,6 +338,86 @@ export default class GitFileSystemService {
...
+    branch = "staging"
+  ): ResultAsync<DefaultLogFields & ListLogLine, GitFileSystemError> {
+    const efsVolPath = this.getEfsVolPathFromBranch(branch)
+    return ResultAsync.fromPromise(
+      this.git
+        .cwd({ path: `${efsVolPath}/${repoName}`, root: false })
+        .log(["-1", branch, "--", path]), // adding -- allows us to get logs even for deleted files
```
harishv7, line 395: what's -1?

src/services/db/GitFileSystemService.ts
```diff
@@ -737,6 +792,49 @@ export default class GitFileSystemService {
...
+        )
+      }
+
+      return (
+        this.ensureCorrectBranch(repoName, "master")
+          .andThen(() =>
+            ResultAsync.fromPromise(
+              this.git
+                .cwd({ path: `${efsVolPath}/${repoName}`, root: false })
+                .merge(["staging", "--no-ff"]), // ensure there is a merge commit
```
harishv7, line 814: could you add comments on what -no-ff is doing?

src/services/db/GitFileSystemService.ts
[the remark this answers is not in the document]
harishv7: yea let's setup sufficient monitoring for the possible failure modes here - currently we are using a string search, so can follow the same on DD monitors

src/services/db/GitFileSystemService.ts
```diff
@@ -1724,6 +1822,33 @@ export default class GitFileSystemService {
     })
   }
 
+  /**
+   * Does not create a new branch if it already exists
```
harishv7, line 1826: should the comment be "creates a new branch iff it doesn't exist"

src/services/review/ReviewRequestService.ts
```diff
@@ -134,96 +114,108 @@ export default class ReviewRequestService {
...
+        return ResultAsync.fromPromise(
+          this.users.findByPk(userId),
+          () => new DatabaseError()
+        ).map((author) => ({
+          lastEditedBy: author?.email || latestLog.author_email,
+          lastEditedTime: new Date(latestLog.date).getTime(),
+        }))
+      })
+      .orElse(() =>
+        ok({
```
harishv7, line 173: is this expected ?

harishv7: should we log a warn msg here?

src/services/db/GitFileSystemService.ts
```diff
@@ -1724,6 +1822,33 @@ export default class GitFileSystemService {
...
+   */
+  createLocalTrackingBranchIfNotExists(repoName: string, branchName: string) {
```
harishv7, line 1828: can we add return types for the various functions? can see the other funcs for example

e.g. `: ResultAsync<string, GitFileSystemError | GitFileSystemNeedsRollbackError>`

src/services/db/GitFileSystemService.ts
jacobkwan: I don't think so, this error isn't recoverable (in the context of /compare)

src/services/db/GitFileSystemService.ts
jacobkwan: it returns only the latest commit
`https://git-scm.com/docs/git-log#Documentation/git-log.txt--ltnumbergt`

technically not necessary since I'm only returning `logs.latest` below which is functionally the same, but just thought it's clearer

src/services/review/ReviewRequestService.ts
jacobkwan: yea this actually follows current behaviour, but yes can definitely add logs here

src/fixtures/review.ts
jacobkwan: hmm, I didn't consider feature flagging it
however, if we were to have a feature flag that is on, and then toggle it off, wouldn't there be divergence between `origin/master` and `master` if user's merge any RRs after that?

or are you suggesting to only feature flag the /compare
but not the `git pull` from `origin/master` during RR merge, so that master is always in sync?

src/services/db/GitFileSystemService.ts
jacobkwan: hmm, I wanted to keep this method as generic as possible and let the caller handle specific edge cases like .keep. The `.keep` files are indeed filtered out in `ReviewRequestService.compareDiff` by checking for the `type !== "placeholder" (`type` is populated when constructing the `EditedItemInfo` for each file)

This was also the existing behaviour previously

src/services/db/GitFileSystemService.ts
jacobkwan: if you're referring to user input via an endpoint then no it is not
the filenames from `getFilesChanged` are passed into this method
so it is not possible to receive a malicious path here
furthermore the git log is only done in their repo as seen by `git.cwd({path: efsvolPath.....`

src/services/db/GitFileSystemService.ts
jacobkwan: yes this is a certified derp moment

kishore03109 (changes requested):
Hey Jacob! Congrats on your first big pr! (I think HAHAH) 
I love that you are quite clear getting feedback on your code style.

That being said, I do have strong concerns against the implementation of this PR.

1. Simple Git Commands are async in nature. This makes checking out to master branch scary. 
Consider this flow:
1. user A merges a pr for the first time.
2. user B makes a change at the exact time.

i am not sure of when exactly we checkout master, my assumption here is that this is only done when there is a merge of a pr.

in the worse case, the git commands executed would be
```
git checkout master 
<some files changed by user B>
git push (while still in master branch!) 
 ```
have we thought about this edge case? 



I am still a bit confused on how does this play with closing of GitHub prs? Techincally a merge locally + push to master = a pr merge. If we merge locally and then attempt to merge the pr upstream in GitHub we might get inconsistent states. I dont quite understand the need for `GitFileSystemService.mergeStagingToMaster`

Feedback regarding PR desc:
Understand that you mentioned that you have tested in staging, but this would be clearer with screenshots + explain which control flows were actually tested in staging!

File: src/services/db/GitFileSystemService.ts (hunk 7 above)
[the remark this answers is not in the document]
Comment (jacobkwan) [reply src/services/db/GitFileSystemService.ts]: my bad, I meant that this method isn't being used anymore as I am pulling from origin/master instead of merging in staging locally

just to avoid confusion, I've removed this function from the PR
[27d6435](https://github.com/isomerpages/isomercms-backend/pull/1172/commits/27d643548cec3739e018faa88a04e97a76e6e23a)

File: src/services/db/GitFileSystemService.ts (hunk 15 above)
Comment (jacobkwan) [reply src/services/db/GitFileSystemService.ts]: added the comment, but removing this function from this PR for now as its unused and was confusing to others
[27d6435](https://github.com/isomerpages/isomercms-backend/pull/1172/commits/27d643548cec3739e018faa88a04e97a76e6e23a)

File: src/services/db/GitFileSystemService.ts (hunk 16 above)
Comment (jacobkwan) [reply src/services/db/GitFileSystemService.ts]: [99bcdd0](https://github.com/isomerpages/isomercms-backend/pull/1172/commits/99bcdd04c4f7725a1ef9a1e04efab3fbc190442a)

File: src/services/review/ReviewRequestService.ts (hunk 17 above)
Comment (jacobkwan) [reply src/services/review/ReviewRequestService.ts]: [cd07757](https://github.com/isomerpages/isomercms-backend/pull/1172/commits/cd077572994c84265dfb274d2504dedafe546f65)

File: src/services/db/GitFileSystemService.ts (hunk 18 above)
Comment (jacobkwan) [reply src/services/db/GitFileSystemService.ts]: [dc591e7](https://github.com/isomerpages/isomercms-backend/pull/1172/commits/dc591e71d2efaf59bb0735973a4c1a484aea57f0)

File: src/services/review/ReviewRequestService.ts [hunk 19]
```diff
@@ -134,96 +114,108 @@ export default class ReviewRequestService {
...
-          // This is because the sha present on the file is
-          // a checksum of the files contents.
-          // And the actual commit sha is given by the ref param
-          const hash = items[items.length - 1]
-          return {
-            ...item,
-            lastEditedBy: mappings[hash]?.author || "Unknown user",
-            lastEditedTime: mappings[hash]?.unixTime || 0,
-          }
+  extractEditMeta = (siteName: string, filename: string) =>
```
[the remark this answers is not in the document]
jacobkwan: [dc591e7](https://github.com/isomerpages/isomercms-backend/pull/1172/commits/dc591e71d2efaf59bb0735973a4c1a484aea57f0)

src/services/db/GitFileSystemService.ts
```diff
@@ -337,6 +338,86 @@ export default class GitFileSystemService {
     )
   }
 
+  /**
+   * Wrapper over `git diff --name-only` that also creates a local tracking branch for `base` and/or `head` if they do not exist.
+   * Note that `base` and `head` should be branch names, not commit hashes or other usually valid arguments for `git diff`.
+   */
+  getFilesChanged(
+    repoName: string,
+    base = "master",
```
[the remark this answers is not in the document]
jacobkwan: great point! fixed in [91bdcac](https://github.com/isomerpages/isomercms-backend/pull/1172/commits/91bdcac5c30872334176e79501b38bcbfd78254b)

src/services/db/GitFileSystemService.ts
jacobkwan: > hmm irrc we need the .keep, because in git it will not commit a folder if it is not empty folders!

not sure what you mean by this, but this method does not impact git commit activity, its only used when **reading** the files changed for the `/compare` and `/:requestId` endpoints

jacobkwan:
> Hey Jacob! Congrats on your first big pr! (I think HAHAH) I love that you are getting feedback on your code style.
> 
> That being said, I do have strong concerns against the implementation of this PR.
> 
> 1. Simple Git Commands are async in nature. This makes checking out to master branch scary.
>    Consider this flow:
> 2. user A merges a pr for the first time.
> 3. user B makes a change at the exact time.
> 
> i am not sure of when exactly we checkout master, my assumption here is that this is only done when there is a merge of a pr.
> 
> in the worse case, the git commands executed would be
> 
> ```
> git checkout master 
> <some files changed by user B>
> git push (while still in master branch!) 
> ```
> 
> have we thought about this edge case?
> 
> I am still a bit confused on how does this play with closing of GitHub prs? Techincally a merge locally + push to master = a pr merge. If we merge locally and then attempt to merge the pr upstream in GitHub we might get inconsistent states. I dont quite understand the need for `GitFileSystemService.mergeStagingToMaster`
> 
> Feedback regarding PR description: Understand that you mentioned that you have tested in staging, but this would be clearer with screenshots + explain which control flows were actually tested in staging!

we're no longer merging staging to master but going with pulling origin/master into master instead, but you still raised a good point about the possible issues with concurrent git actions.
However, looking at `GitFileSystemService.commit` it seems as that we are protected from this as we gate the commit with a call to `ensureCorrectBranch` first, which would ensure we are on staging 

Noted on the feedback, will add in some screenshots!

src/fixtures/review.ts
harishv7: I thought to feature flag for a few sites then progressively expand this to 100% rollout. On growthbook,  you can whitelist for specific sites. For starters, this will be done for test sites, then 1-2 agency sites and so on

src/services/db/GitFileSystemService.ts
harishv7: can, just add as comments for other engs

src/fixtures/review.ts
```diff
@@ -20,6 +20,10 @@ import {
...
   MOCK_GITHUB_DATE_ONE,
   MOCK_COMMIT_FILEPATH_PLACEHOLDER,
+  MOCK_COMMIT_FILENAME_ONE,
+  MOCK_COMMIT_FILEPATH_ONE,
+  MOCK_COMMIT_FILENAME_TWO,
+  MOCK_GITHUB_DATE_TWO,
 } from "./identity"
```
harishv7, line 27: The tests in your description should allow the on-call to test step by step before deployment. Sample here https://github.com/isomerpages/isomercms-backend/pull/1173

kishore03109:
> we're no longer merging staging to master but going with pulling origin/master into master instead, but you still raised a good point about the possible issues with concurrent git actions.
However, looking at GitFileSystemService.commit it seems as that we are protected from this as we gate the commit with a call to ensureCorrectBranch first, which would ensure we are on staging

hmm no it is not about check per se, but the async nature of the calls. as long each chain in the simple git command is an independent promise, we cannot guarantee that user A and user B's actions are _**isolated**_ from one another. all we know is from the individual pov of user A and user B, their promises are fulfilled in order.

what we want here is to avoid this complexity entirely. if we are not going to merge locally, we can indeed update our master branch within efs using a simple [fetch](https://stackoverflow.com/questions/3216360/merge-update-and-pull-git-branches-without-using-checkouts) rather than a checkout.

src/services/db/GitFileSystemService.ts
kishore03109: oh no i agree with you @jacobkwan, we shouldnt filter out keep

src/fixtures/review.ts
jacobkwan: good point! I've added back the old `compareDiff` and feature flagged it as `is_local_diff_enabled`
[9920250](https://github.com/isomerpages/isomercms-backend/pull/1172/commits/99202504538a0ac2fd9aeb7ee856f4f8bc8c4aa8)

https://app.growthbook.io/features/is_local_diff_enabled

src/services/db/GitFileSystemService.ts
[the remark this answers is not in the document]
jacobkwan: added monitors here!
https://ogp.datadoghq.com/monitors/140997853?view=spans
https://ogp.datadoghq.com/monitors/140998024?view=spans

src/services/db/GitFileSystemService.ts
```diff
@@ -476,12 +551,46 @@ export default class GitFileSystemService {
...
+          new GitFileSystemError(
+            `Folder "${repoName}" for EFS vol path: "${efsVolPath}" is not a valid Git repo`
+          )
+        )
+      }
+      return this.createLocalTrackingBranchIfNotExists(repoName, "master")
+        .andThen(() =>
+          ResultAsync.fromPromise(
+            this.git
+              .cwd({ path: `${efsVolPath}/${repoName}`, root: false })
```
kishore03109, line 570: lets use fs.path for this rather than constructing this by ourselves

src/routes/v2/authenticated/review.ts
```diff
@@ -125,12 +129,23 @@ export class ReviewsRouter {
...
+        ) {
+          return this.reviewRequestService.compareDiffLocal(
+            userWithSiteSessionData,
+            stagingLink
+          )
+        }
+        return this.reviewRequestService.compareDiff(
           userWithSiteSessionData,
           stagingLink
         )
```
kishore03109, line 147: hmm this logic seems like it should live in the `service` level rather than at the router level. 

could we refactor this such that this lives in `reviewRequestService` instead?

src/services/db/GitFileSystemService.ts
```diff
@@ -476,12 +551,46 @@ export default class GitFileSystemService {
       })
   }
 
+  fastForwardMaster(
+    repoName: string
+  ): ResultAsync<boolean, GitFileSystemError> {
+    const efsVolPath = this.getEfsVolPathFromBranch("master")
+    return this.isValidGitRepo(repoName, "master").andThen((isValid) => {
```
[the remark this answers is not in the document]
jacobkwan: can I refactor this in a separate PR? tbh I think it's a lot cleaner if all the methods (`isValidGitRepo`, `getFilePathStats`, `isGitInitialized` etc) just take in the correct `efsVolPath`, instead of each one taking in a `isStaging` boolean and then deciding which EFS vol path to use

meaning we just need to `getEfsVolFromBranch` once at the top, and then just pass that to all the random util methods; at the end of the day its just the efs vol that matters

src/constants/featureFlags.ts
```diff
@@ -3,4 +3,5 @@ export const FEATURE_FLAGS = {
   IS_GGS_ENABLED: "is_ggs_enabled",
   IS_SHOW_STAGING_BUILD_STATUS_ENABLED: "is_show_staging_build_status_enabled",
   IS_CLOUDMERSIVE_ENABLED: "is_cloudmersive_enabled",
+  IS_LOCAL_DIFF_ENABLED: "is_local_diff_enabled",
```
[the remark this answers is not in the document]
jacobkwan: didn't want to use `git_diff` to avoid confusion as its not exactly a `git diff` operation

src/services/db/GitFileSystemService.ts
jacobkwan: [6931527](https://github.com/isomerpages/isomercms-backend/pull/1172/commits/6931527c71f805ec5b316922a88d17ec02ab81e4)

src/routes/v2/authenticated/review.ts
jacobkwan: [00fa655](https://github.com/isomerpages/isomercms-backend/pull/1172/commits/00fa6554e11a223043fcadb139e25e1154cdaa1f)

src/constants/featureFlags.ts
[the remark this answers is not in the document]
kishore03109: wait now i am confused haha under the hood is this not `git diff`? is a nit anyways

src/services/db/GitFileSystemService.ts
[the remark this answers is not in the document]
kishore03109: sure

kishore03109 (approved):

harishv7 (approved):

PR #1173, opened 2024-02-26
fix: determine if the branch exists first

dcshzj:
## Problem

We get a lot of such errors when users access their site dashboard when retrieving the last saved and last published information:

> Error when getting latest commit of "master" branch: Error: fatal: ambiguous argument 'master': unknown revision or path not in the working tree.

This happens because we try to get the Git log for the `master` branch first, before attempting on `origin/master`. The first one will fail as the `master` branch does not exist on EFS, but `origin/master` does.

## Solution

**Breaking Changes**

**Bug Fixes**:

- Determine if the branch exists first before continuing to get the Git log, so that we avoid encountering the error unnecessarily.

## Tests

- [ ] Navigate to any site dashboard.
- [ ] Verify from the logs that there are no errors shown that matches the error log above.
- [ ] Verify that you can see `Unable to ascertain that "master" branch...` log line as an info log.

## Deploy Notes

*None*

### DIFF
--- a/src/services/db/GitFileSystemService.ts
+++ b/src/services/db/GitFileSystemService.ts
@@ -203,6 +203,28 @@ export default class GitFileSystemService {
       })
   }
 
+  // Determine if the given branch is present in the repo
+  isBranchPresent(
+    repoName: string,
+    branchName: string
+  ): ResultAsync<boolean, never> {
+    const efsVolPath = this.getEfsVolPathFromBranch(branchName)
+    return ResultAsync.fromPromise(
+      this.git
+        .cwd({ path: `${efsVolPath}/${repoName}`, root: false })
+        .revparse(["--verify", branchName]),
+      (error) => {
+        logger.info(
+          `Unable to ascertain that "${branchName}" branch is in ${efsVolPath}/${repoName}: ${error}`
+        )
+
+        return false
+      }
+    )
+      .andThen(() => okAsync(true))
+      .orElse(() => okAsync(false))
+  }
+
   // Ensure that the repository is in the specified branch
   ensureCorrectBranch(
     repoName: string,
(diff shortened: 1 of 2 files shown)


jacobkwan (approved):

src/services/db/GitFileSystemService.ts
```diff
@@ -203,6 +203,28 @@ export default class GitFileSystemService {
...
+  // Determine if the given branch is present in the repo
+  isBranchPresent(
+    repoName: string,
+    branchName: string
+  ): ResultAsync<boolean, never> {
+    const efsVolPath = this.getEfsVolPathFromBranch(branchName)
+    return ResultAsync.fromPromise(
+      this.git
+        .cwd({ path: `${efsVolPath}/${repoName}`, root: false })
+        .revparse(["--verify", branchName]),
```
jacobkwan, line 215: can consider using `.branchLocal` instead

jacobkwan: also regarding what you mentioned just now about making it not return an error, I think as a common util method, this should still return an error. The caller should be the one to decide if they want to ignore the error i.e `orElse(()=> false)` or something along those lines

src/services/db/GitFileSystemService.ts
```diff
@@ -203,6 +203,34 @@ export default class GitFileSystemService {
...
+
+        if (error instanceof GitError) {
+          return new GitFileSystemError(
+            "Unable to get the list of local branches"
+          )
+        }
+
+        return new GitFileSystemError("An unknown error occurred")
+      }
+    ).andThen((result) => okAsync(result.all.includes(branchName)))
```
jacobkwan, line 231: nit: can just use `map` here instead of `andThen`

dcshzj: Updated in eb91c65d3632fed7a53946180a9a650e7870dc08.

jacobkwan (approved):

PR #1180, opened 2024-03-04
feat(mail): support sending emails with attachments

dcshzj:
## Problem

The current mail sending functionality does not support adding of attachments to the email.

## Solution

**Breaking Changes**

- [ ] Yes - this PR contains breaking changes
- [x] No - this PR is backwards compatible with ALL of the following feature flags in this [doc](https://www.notion.so/opengov/Existing-feature-flags-518ad2cdc325420893a105e88c432be5)

**Features**:

- Use `form-data` to craft the message object and also allow for inserting of files as attachments.
- Adjust email from address to `donotreply@isomer.gov.sg`, which is required for sending emails with attachments.

## Tests

- [ ] Update the `POSTMAN_API_KEY` to the one that starts with `donotreply_`
- [ ] Attempt to login with email OTP
- [ ] Verify that the OTP email is sent successfully from `donotreply@isomer.gov.sg`

## Deploy Notes

**New dependencies**:

- `form-data` : Package for handling multipart form data requests

**Other deploy steps**
- [ ] Create the new `/efs/audit-logs` folder on EFS before deployment.
- [ ] Update `PROD_POSTMAN_API_KEY` on SSM

PR #1181, opened 2024-03-04
feat: add site audit logs generator

dcshzj:
## Problem

Site audit logs are manually created using some script that we have, which brings a lot of overhead and is not an ideal long-term solution.

## Solution

**Breaking Changes**

**Features**:

- Allow users to self-service request for their site audit logs.

## Tests

- [ ] Fill in the form, ensuring that you are either an Isomer Admin or a collaborator of the repo(s) that you are requesting site audit logs for.
- [ ] Verify that you receive an email with the site audit logs for the repos that you have requested for.
- [ ] Try again with an account that is not an Isomer Admin account and the account does not have collaborator access to the repo(s) that you are requesting site audit logs for.
- [ ] Verify that you still receive an email, but with an error message saying that you are not a collaborator of the repo(s).

## Deploy Notes

- `SITE_AUDIT_LOGS_FORM_KEY` : Secret key for the site audit logs form
    - [x] added env var to 1PW + SSM script (`fetch_ssm_parameters.sh`)


dcshzj:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/isomerpages/isomercms-backend/1181?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1181** <a href="https://app.graphite.dev/github/pr/isomerpages/isomercms-backend/1181?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1180** <a href="https://app.graphite.dev/github/pr/isomerpages/isomercms-backend/1180?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `develop`

<h2></h2>

 Join @dcshzj and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

src/routes/formsg/formsgSiteAuditLogs.ts
```diff
@@ -0,0 +1,128 @@
...
+        .getMonth()
+        .toString()
+        .padStart(2, "0")}-${endDateObject.getDate()}`
+    } else {
+      const startDateField = getField(responses, LOGS_TIMEFRAME_START_FIELD)
+      const endDateField = getField(responses, LOGS_TIMEFRAME_END_FIELD)
+      if (startDateField && endDateField) {
+        startDate = startDateField
+        endDate = endDateField
+      }
```
kishore03109, line 98: hmm missing a control flow here right? if the user never put the start date and end date correctly?

src/services/admin/AuditLogsService.ts
```diff
@@ -0,0 +1,500 @@
...
+      `Received request to get audit logs, submission ID: ${formSubmissionId}`
+    )
+
+    // Step 1: Check if the user exists
+    return ResultAsync.fromPromise(this.usersService.findByEmail(email), () => {
+      logger.error(
+        `Error occurred while retrieving user data from the database for email ${email}`
+      )
+
+      return new DatabaseError(
```
kishore03109, line 330: are we emailing this error to anyone ah? seems like silent failure here :(

src/services/admin/AuditLogsService.ts
```diff
@@ -0,0 +1,500 @@
...
+            )
+          ),
+        ])
+      )
+      .andThen(([user, ...roles]) => {
+        const isUserNotAdminCollaborator = roles.some(
+          (role) =>
+            !role ||
+            (role !== CollaboratorRoles.Admin &&
+              role !== CollaboratorRoles.IsomerAdmin)
```
kishore03109, line 365: hmm i thought we are gating this to devs only? shouldnt this just be isomerAdmin?

src/services/admin/AuditLogsService.ts
```diff
@@ -0,0 +1,500 @@
...
+          okAsync(user),
+          tokenServiceInstance.getAccessToken(),
+        ])
+      )
+      .andThen(([user, accessToken]) =>
+        // Step 4: Construct a fake user session data object and get audit logs
+        ResultAsync.combine(
+          repoNames.map((repoName) => {
+            const userSessionData = new UserWithSiteSessionData({
+              githubId: user.githubId,
```
kishore03109, line 396: what if this doesnt exist? ie an admin that uses this form that only has email sites?

src/services/admin/AuditLogsService.ts
```diff
@@ -0,0 +1,500 @@
...
+                  .getUTCSeconds()
+                  .toString()
+                  .padStart(2, "0")
+                const recordTime = `${recordHour}:${recordMinute}:${recordSecond} (UTC)`
+
+                return `${csv}${recordDate},${recordTime},${activity},"${actor}","${page}","${remarks}"\n`
+              },
+              auditLogHeader
+            )
+            const csvFileDate = snapshotTime.toISOString().split("T")[0]
```
kishore03109, line 440: nit: use papaparse?

src/services/admin/AuditLogsService.ts
```diff
@@ -0,0 +1,500 @@
...
+          mailer.sendMail(email, emailSubject, emailBody, csvFilePaths),
+          () =>
+            new AuditLogsError(
+              `Error occurred while sending audit log CSV files to ${email} (submission ID: ${formSubmissionId}). Requested repos: ${repoNames.join(
+                ", "
+              )}`
+            )
+        )
+      })
+      .orElse((error) => {
```
kishore03109, line 478: error here is typed to a `unknown`, i would have expected a obj/string... thoughts?

kishore03109: also we are using result.combine which short circuits! this means that we will either send the success state if all the repos pass, or fail for when only one fails. this means that repos that pass, will sill be reported as failed here due to this control flow. is this intended?

src/types/auditLog.ts
```diff
@@ -0,0 +1,17 @@
+export enum AuditableActivity {
```
kishore03109, line 1: aye shall we not use enums hahah

https://dev.to/ivanzm123/dont-use-enums-in-typescript-they-are-very-dangerous-57bh#:~:text=Excessive%20use%20of%20regular%20ENUMs,new%20values%20can%20be%20added.

src/services/admin/AuditLogsService.ts
```diff
@@ -0,0 +1,500 @@
...
+                    return okAsync(
+                      commit.commit.author?.name || "Unknown author"
+                    )
+                  }
+
+                  return errAsync(error)
+                })
+                .andThen((actorEmail) =>
+                  okAsync({
+                    timestamp: new Date(commit.commit.author?.date || ""),
```
kishore03109, line 233: hmm this makes it to today's timestamp, rather than idk the timestamp... is this intended?

src/routes/formsg/formsgSiteAuditLogs.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    const { responses } = res.locals.submission.content
+
+    const requesterEmail = getField(responses, REQUESTER_EMAIL_FIELD)
+
+    if (!requesterEmail) {
+      logger.error(
+        "No requester email was provided in site audit logs form submission"
+      )
+      return
```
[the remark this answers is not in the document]
dcshzj: I think maybe we can test with other internal forms first instead? This form is more intended for agencies so they might get duplicate emails if retries are enabled.

src/routes/formsg/formsgSiteAuditLogs.ts
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    repoNamesFromTable.forEach((repoName) => {
+      if (typeof repoName === "string") {
+        // actually wont happen based on our formsg form, but this code
+        // is added defensively
+        repoNames.add(repoName)
+      } else {
+        repoNames.add(repoName[0])
```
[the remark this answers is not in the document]
dcshzj: I took this from the GGS repair form but my understanding is that `repoNamesFromTable` is a table, so we want the values of the first column here.

src/routes/formsg/formsgSiteAuditLogs.ts
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  getAuditLogsHandler: RequestHandler<
+    never,
+    Record<string, never>,
+    { data: { submissionId: string } },
+    never,
+    { submission: DecryptedContentAndAttachments }
+  > = async (req, res) => {
+    let startDate = "1970-01-01"
```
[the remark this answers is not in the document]
dcshzj: Hmm I'm not sure if 6 years ago would be arbitrary here as compared to using epoch, which would be clearer in the intent?

src/routes/formsg/formsgSiteAuditLogs.ts
dcshzj: In that case then it would be kept to the defaults of epoch and now.

src/services/admin/AuditLogsService.ts
dcshzj: This is handled in the final `orElse` flow, which is then sent to the requestor.

src/services/admin/AuditLogsService.ts
dcshzj: Oops I meant this form is going to be used by agencies themselves as it is supposed to be self-service. For a start, we will be the ones using it (since they don't know the existence of the form) but the main intent is for them to use.

src/services/admin/AuditLogsService.ts
dcshzj: It would be set to undefined, but that is not used anyway since we rely on the accessToken from TokenService.

src/services/admin/AuditLogsService.ts
dcshzj: No strong opinion on this, opted to use this way originally since the amount of logs is not very large. I think we can optimise this later if we find that this is running into performance issues?

src/services/admin/AuditLogsService.ts
dcshzj: Not sure why the typings died halfway through, adjusted in 6a46e291197690df3fc3825f46e0efc311bab2be.

The short circuiting is intended! I'm erring on the side of caution here to fail often and only provide the logs if everything is successful.

src/types/auditLog.ts
dcshzj: Adjusted in 06dbe93cdd89d76d2ed8be5280d966bbdf2a47db.

src/services/admin/AuditLogsService.ts
dcshzj: This is more to satisfy the types, because every commit will have a timestamp.

src/routes/formsg/formsgSiteAuditLogs.ts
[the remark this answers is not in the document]
kishore03109: sure, but it's still good practice to return once the form is processed to keep our api latency low ya
concretely, am thinking of 

```
res(400) 
return
```

thoughts?

src/routes/formsg/formsgSiteAuditLogs.ts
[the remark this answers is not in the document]
kishore03109: oh right i remember this 2d array thing zzz

src/routes/formsg/formsgSiteAuditLogs.ts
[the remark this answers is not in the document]
kishore03109: oh its not an arbitrary number, its when isomer was born haha

src/services/admin/AuditLogsService.ts
kishore03109: ill let you make the final call since non functional, but i feel that its not just pref issues (not really a concern since no real time req).

what i valued more was their out of the box handling for nity things like new lines. so in the future when you want to make this into a full fleded feature, you can easily parse and unparse heh

src/services/admin/AuditLogsService.ts
kishore03109: sure
noticed that we are returning the error directly to the user. have we ensured that we are not returning cryptic/sensitive error info? eg. some "sequalize error string expected" shouldnt reach our user
also, have we set up monitoring for this feature? will us devs be aware if suddenly all of our forms are failing?

src/services/admin/AuditLogsService.ts
kishore03109: this is confusing to me :(
if it is not needed here, why declare it? doesnt seem like a necessary key too

src/services/admin/AuditLogsService.ts
kishore03109: arh ok

kishore03109 (commented):
only real blocking is the error messages returned to agency users

src/routes/formsg/formsgSiteAuditLogs.ts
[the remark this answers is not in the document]
dcshzj: Fixed in c9f56b7caad419be32ca53b53bacea6db4d09908.

src/services/admin/AuditLogsService.ts
dcshzj: My bad, it is defined and is needed because of this control flow:

https://github.com/isomerpages/isomercms-backend/blob/24658ec51af72594c7da2c6368ebc0f7be5d5c52/src/classes/UserSessionData.ts#L32-L35

Otherwise, it would not be possible for us to set the accessToken for the request.

src/services/admin/AuditLogsService.ts
dcshzj: Adjusted in 85d9b84f4f7a3953776fec96f23fba80f6ef790f.

src/services/admin/AuditLogsService.ts
dcshzj: Decided to add a common prefix for error logs here so we can just monitor for this prefix to see if any errors occur: 24658ec51af72594c7da2c6368ebc0f7be5d5c52.

We are not returning sensitive/cryptic information back to the user, because the error that is returned only contains text that we define and not the actual error message that comes from say sequelize.

src/routes/formsg/formsgSiteAuditLogs.ts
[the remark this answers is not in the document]
dcshzj: This one I might want to push back. I don't have an example but setting it to say 2018-01-01 means we may miss potential commits that occur before that date, which can happen if for some reason users "backdate" their commits. I try to avoid making this decision for users just in case, especially this is audit logs that needs as much information captured as possible.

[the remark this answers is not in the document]
kishore03109: sure, lgtm

src/services/admin/AuditLogsService.ts
kishore03109: wait this is an issue right? not all users have a github id, wouldnt this then fail for those users?

kishore03109 (commented):
only blocking is in verifying that non gh users admin users should be able to use this form to get their site's audit logs

src/services/admin/AuditLogsService.ts
dcshzj: Adjusted in ae8410a1d315b913370a8333c2427fd174c9c067.

kishore03109 (approved):

PR #1183, opened 2024-03-05
fix(rateLimiter): correct rate limits

kishore03109:
## Problem

Our rate limiter is implemented wrongly. Couple of problems exist:
1. careless conversion of units. the unit of env var was already ms, but the code assumes it is in s instead. for clarity, renaming the env var from `tokenExpiry` to `tokenExpiryInMs`
2.  We are currently trusting proxy, where we trust the proxy to expose the client ip. The [documentation](https://expressjs.com/en/guide/behind-proxies.html) states that 

> When setting to true, it is important to ensure that the last reverse proxy trusted is removing/overwriting all of the following HTTP headers: X-Forwarded-For, X-Forwarded-Host, and X-Forwarded-Proto otherwise it may be possible for the client to provide any value.

This is not entirely true from reading the Cloudflare's [documentation](https://developers.cloudflare.com/fundamentals/reference/http-request-headers/#x-forwarded-for).  

> If, on the other hand, an X-Forwarded-For header was already present in the request to Cloudflare, Cloudflare will append the IP address of the HTTP proxy connecting to Cloudflare to the header. 

This PR uses the recommended approach of using the `CF-Connecting-IP` that cloudflare [provides](https://developers.cloudflare.com/fundamentals/reference/http-request-headers/#cf-connecting-ip) to assert the ip of the client instead.

Moving forward, we never trust the proxy. When using Cloudflare in production env, we should use the `CF-Connecting-IP` instead to verify the cilent ip instead. We continue to use `req.ip` for dev environments

I have verified that the `cf-incoming-ip` exists in staging env by logging it
Screenshot 2024-03-05 at 12 14 04 PM

Closes GTA-24-011 WP3.

**Breaking Changes**

  - Details ...

## Tests
- [ ] create a file called ddos.js
```
const stg = "https://staging-cms-api.isomer.gov.sg/v2/auth/verify"
async function send() {
  try {
    const resp = await fetch(stg, {
      method: "POST",
      body: JSON.stringify({
        email: "test@test.com",
        otp: "111111",
      }),
      headers: {
        "Content-Type": "application/json",
        "X-Forwarded-For": generateRandomIp(),
      },
    })
    const text = await resp.text()
    console.log(text)
    console.log({
      Limit: resp.headers.get("Ratelimit-Limit"),
      Remaining: resp.headers.get("Ratelimit-Remaining"),
      Reset: resp.headers.get("Ratelimit-Reset"),
    })
  } catch (err) {
    console.log(err.message)
  }
}
for (let i = 1; i <= 25; i++) {
  send()
}

```

- [ ] run `node ddos.js` 

assert that the reset time is around 84400 (units are in seconds)

Reset value before:
Screenshot 2024-03-05 at 12 44 20 PM

Reset value after:

Screenshot 2024-03-05 at 12 45 08 PM

### DIFF
--- a/src/services/utilServices/RateLimiter.ts
+++ b/src/services/utilServices/RateLimiter.ts
@@ -1,7 +1,5 @@
 import rateLimit from "express-rate-limit"
 
-import { config } from "@config/config"
-
 import { BaseIsomerError } from "@root/errors/BaseError"
 import { isSecure } from "@root/utils/auth-utils"
 
@@ -16,8 +14,8 @@ const DEFAULT_AUTH_TOKEN_EXPIRY_MILLISECONDS = 900000
 // but not on the other, leading to inconsistent behaviour.
 // eslint-disable-next-line import/prefer-default-export
 export const rateLimiter = rateLimit({
-  windowMs: config.get("auth.tokenExpiryInMs"),
-  max: 100, // Limit each IP to 100 requests per `window` (here, per 1 hour)
+  windowMs: DEFAULT_AUTH_TOKEN_EXPIRY_MILLISECONDS,
+  max: 100, // Limit each IP to 100 requests per `window` (here, per 15 minutes)
   standardHeaders: true, // Return rate limit info in the `RateLimit-*` headers
   legacyHeaders: false, // Disable the `X-RateLimit-*` headers
   // We know that this key exists in a secure env (Cloudflare)
--- a/src/services/utilServices/RateLimiter.ts
+++ b/src/services/utilServices/RateLimiter.ts
@@ -2,6 +2,9 @@ import rateLimit from "express-rate-limit"
 
 import { config } from "@config/config"
 
+import { BaseIsomerError } from "@root/errors/BaseError"
+import { isSecure } from "@root/utils/auth-utils"
+
 const DEFAULT_AUTH_TOKEN_EXPIRY_MILLISECONDS = 900000
 
 // NOTE: Refer here for more information regarding the implementation:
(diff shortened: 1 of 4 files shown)


kishore03109:
* **#1183** <a href="https://app.graphite.dev/github/pr/isomerpages/isomercms-backend/1183?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop`

<h2></h2>

src/services/utilServices/RateLimiter.ts
```diff
@@ -13,8 +16,21 @@ const DEFAULT_AUTH_TOKEN_EXPIRY_MILLISECONDS = 900000
 // but not on the other, leading to inconsistent behaviour.
 // eslint-disable-next-line import/prefer-default-export
 export const rateLimiter = rateLimit({
-  windowMs: config.get("auth.tokenExpiry") / (1000 * 60),
-  max: 100, // Limit each IP to 100 requests per `window` (here, per 15 minutes)
+  windowMs: config.get("auth.tokenExpiryInMs"),
```
seaerchin, line 19: i think you might have made a mistake here - going off the prod env var, this is `86400000 ms`, which is a day.

i think this token expiry is different from the rate limits - we want to limit incoming auth requests to a max of 100 requests per window, this has nothing to do with the token expiry, which is how long a user can stay signed in for before we invalidate their session.

could we instead just use `DEFAULT_AUTH_TOKEN_EXPIRY_MILLISECONDS` defined above? (an hour or 15 minutes is both fine. i think 15 might be more permissive but this isn't likely to be ddos or abused given the low rate)

seaerchin (changes requested):

src/services/utilServices/RateLimiter.ts
kishore03109: sure modified to 15 mins [here](https://github.com/isomerpages/isomercms-backend/pull/1183/commits/168074ec0971e3083649f157e7bf6d06f5c81a31)

seaerchin (approved):

PR #1186, opened 2024-03-05
fix(otp): increment instead of update for concurrency

dcshzj:
## Problem

The update operation on the number of attempts for each OTP is subject to race conditions, which causes the value to be overwritten when many requests are made simultaneously.

## Solution

**Breaking Changes**

**Bug Fixes**:

- Use [sequelize's increment operation](https://sequelize.org/docs/v6/core-concepts/model-instances/#incrementing-and-decrementing-integer-values) instead of update, which is done on the database itself and does not run into concurrency issues.
- Use transactions to lock the row from further updates.

## Tests

- [ ] Use the script provided in the VAPT report on page 17 and 18
- [ ] Adjust the URL to point to your test instance
- [ ] Adjust the email address to be one that is valid (i.e. your own account) and attempt to log in (without keying in the correct OTP)
- [ ] Run the script and verify that you hit the max attempts after 5 tries

## Deploy Notes

*None*

### DIFF
--- a/src/services/identity/UsersService.ts
+++ b/src/services/identity/UsersService.ts
@@ -1,4 +1,5 @@
-import { Op, ModelStatic } from "sequelize"
+import { ResultAsync, errAsync, okAsync } from "neverthrow"
+import { Op, ModelStatic, Transaction } from "sequelize"
 import { Sequelize } from "sequelize-typescript"
 import { RequireAtLeastOne } from "type-fest"
 
@@ -6,6 +7,8 @@ import { config } from "@config/config"
 
 import { Otp, Repo, Site, User, Whitelist, SiteMember } from "@database/models"
 import { BadRequestError } from "@root/errors/BadRequestError"
+import DatabaseError from "@root/errors/DatabaseError"
+import logger from "@root/logger/logger"
 import { milliSecondsToMinutes } from "@root/utils/time-utils"
 import SmsClient from "@services/identity/SmsClient"
 import MailClient from "@services/utilServices/MailClient"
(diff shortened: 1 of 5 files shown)


src/services/identity/UsersService.ts
```diff
@@ -230,56 +233,201 @@ class UsersService {
...
+          .map(() => true)
+      )
+      .orElse((error) =>
+        ResultAsync.fromPromise(transaction.commit(), (txError) => {
+          logger.error(
+            `Error committing transaction: ${JSON.stringify(txError)}`
+          )
+          return new DatabaseError("Error committing transaction")
+        }).andThen(() => errAsync(error))
+      )
```
seaerchin, line 356: there's honestly alot of repeated `fromPromise` -> error handling code here. 

this is fine, but i suspect the error handling code is too verbose and takes away from the actual business logic of what we're trying to do here. 

maybe a better way to do it would be to write a wrapper for `fromPromise` that is localised to this module and does a simple log of the error + returns a `DatabaseError` (cos everything here seems to return that) 

don't have to do it now, just food for thought

src/services/utilServices/AuthService.js
```diff
@@ -124,18 +126,28 @@ class AuthService {
...
+          }
+        )
+      )
+      .map((user) => ({
+        isomerUserId: user.id,
+        email: user.email,
+      }))
+      .mapErr((error) => {
+        throw error
+      })
```
seaerchin, line 150: this is kinda-self defeating but given that this is a `.js` module, `neverthrow` has limited value here anyways

seaerchin (approved):

src/services/identity/UsersService.ts
```diff
@@ -230,56 +233,201 @@ class UsersService {
...
-      throw new BadRequestError("Max number of attempts reached")
-    }
-
-    if (!otpEntry?.hashedOtp) {
-      await otpEntry.destroy()
-      throw new BadRequestError("Hashed OTP not found")
-    }
+    return okAsync(otp)
+      .andThen((otp) => {
+        if (!otp || otp === "") {
```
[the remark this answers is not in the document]
dcshzj: Adjusted in 5a7e90a8ca8dcb1bdee2b236d1f728f82779636e.

src/services/identity/UsersService.ts
```diff
@@ -230,56 +233,201 @@ class UsersService {
...
+
+            return new DatabaseError("Error incrementing OTP attempts")
+          }
+        ).map(() => otpDbEntry)
+      )
+      .andThen((otpDbEntry) =>
+        ResultAsync.combine([
+          okAsync(otpDbEntry),
+          ResultAsync.fromPromise(
+            this.otpService.verifyOtp(otp, otpDbEntry.hashedOtp),
```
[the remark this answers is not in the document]
dcshzj: The verification on the number of attempts is only done here, and we increment beforehand because the number of attempts starts from 0. This means that the first attempt is when attempts = 1, then the fifth attempt is when attempts = 5. Thereafter, they can no longer attempt (so we limit users to 5 OTP attempts).

src/services/utilServices/AuthService.js
dcshzj: Yeah kept this mainly for backward compat

PR #1187, opened 2024-03-05
fix(linkChecker): bug fixes

kishore03109:
## Problem

Users have given feedback regarding link checker. 

1. Decoding issues 
The repo had the file 
`/images/Life @ JPJC/2023 Events/LTC2023/ltc photo 1.JPG`

User had sourced this in one of their pages using the path `/images/Life%20%40%20JPJC/2023%20Events/LTC2023/ltc%20photo%201.JPG`

In the live site, this is a vaild link.

When running 
```
const uri = '/images/Life%20%40%20JPJC/2023%20Events/LTC2023/ltc%20photo%201.JPG';
  console.log(decodeURIComponent(uri));
  console.log(decodeURI(uri));
```

we get the result 
```
> "/images/Life @ JPJC/2023 Events/LTC2023/ltc photo 1.JPG"
> "/images/Life %40 JPJC/2023 Events/LTC2023/ltc photo 1.JPG"
```

which is expected as the [documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURI#decodeuri_vs._decodeuricomponent) states 
> decodeURI() assumes the input is a full URI, so it does not decode characters that are part of the URI syntax.

since we are dealing with user supplied inputs, we have to make the least guarantees, and assume that the user supplied value might contain characters that are part of the URI syntax. 

2. Anchor links
We have an agency that used pattern of 
`<a id="blah"></a>` 
for their anchor links. To them, this is not a broken link and a false positive, which is understandable. Since this is something that the end citizen will not notice, I am opting to make a call to not report any anchor tags without any text. 

3. Image links
There exist dangling image tags in .md files, eg
`` 
and this has confused the user. Similar to the reasoning above, removing reporting of empty img srcs. 

4. Trivial permalink normalisation issues. eg leading quotation marks 
5. any resources _post page in cms was not rendered properly
6. Remove query params when checking for local references, amplify allows it and returns a 200

Other optimisations
Allow faster dev iteration by removing the push to the remote repository isomer-link-checker

this pr should make link checker stable for the upcoming 8 sites to be whitelisted after release

**Breaking Changes**

  - Details ...

## Tests

nil

### DIFF
--- a/src/services/review/RepoCheckerService.ts
+++ b/src/services/review/RepoCheckerService.ts
@@ -9,7 +9,7 @@ import { marked } from "marked"
 import { ResultAsync, errAsync, ok, okAsync } from "neverthrow"
 import Papa from "papaparse"
 import { ModelStatic } from "sequelize"
-import { SimpleGit } from "simple-git"
+import { CommitResult, PushResult, SimpleGit, Response } from "simple-git"
 
 import config from "@root/config/config"
 import { EFS_VOL_PATH_STAGING, STAGING_BRANCH } from "@root/constants"
@@ -293,6 +293,16 @@ export default class RepoCheckerService {
       .mapErr((error) => new SiteCheckerError(`${error}`))
   }
 
+  getResourceRoomPath(repoPath: string) {
+    const configPath = path.join(repoPath, "_config.yml")
+    const fileContent = fs.readFileSync(configPath, "utf8")
+    const resourceRoom = fileContent
+      .split("resources_name: ")[1]
+      .split("\n")[0]
+      .trim()
+    return resourceRoom
+  }
+
   checker(repo: string): ResultAsync<RepoError[], SiteCheckerError> {
     const repoPath = path.join(EFS_VOL_PATH_STAGING, repo)
     const mdFiles = new Set<string>()
@@ -330,6 +340,7 @@ export default class RepoCheckerService {
           if (!filePermalink) {
             continue
           }
+          const normalisedPermalink = this.permalinkNormaliser(filePermalink)
 
           // we are only parsing out the front matter
           const fileContent = file.split("---")?.slice(2).join("---")
@@ -345,62 +356,59 @@ export default class RepoCheckerService {
               const anchorTags = dom.window.document.querySelectorAll("a")
               forEach(anchorTags, (tag) => {
                 const href = tag.getAttribute("href")
-                const text = tag.textContent || ""
-                if (!href) {
+                const text = tag.textContent
+
+                if (!text) {
+                  // while rendered in the dom, unlikely to be
+                  // noticed by end citizen.
+                  return
+                }
+                if (!tag.hasAttribute("href")) {
                   const error: RepoError = {
                     type: RepoErrorTypes.BROKEN_LINK,
                     linkToAsset: "",
                     viewablePageInCms: this.getPathInCms(fileName, repo),
-                    viewablePageInStaging: path.join(filePermalink),
+                    viewablePageInStaging: path.join(normalisedPermalink),
                     linkedText: text,
                   }
                   errors.push(error)
-                } else {
-                  const decodedHref = this.mediaPathDecoder(href)
-                  if (
-                    !this.isExternalLinkOrPageRef(decodedHref) &&
-                    !setOfAllMediaAndPagesPath.has(decodedHref) &&
-                    // Amplify supports adding of trailing slash
-                    !setOfAllMediaAndPagesPath.has(`${decodedHref}/`)
-                  ) {
-                    const error: RepoError = {
-                      type: RepoErrorTypes.BROKEN_LINK,
-                      linkToAsset: href,
+                }
+                if (!href) {
+                  // could be intentional linking to self
+                  // eg. <a href>Back to top</a>
+                  return
+                }
 
-                      viewablePageInCms: this.getPathInCms(fileName, repo),
-                      viewablePageInStaging: path.join(filePermalink),
-                      linkedText: text,
-                    }
-                    errors.push(error)
+                if (!this.hasValidReference(setOfAllMediaAndPagesPath, href)) {
+                  const error: RepoError = {
+                    type: RepoErrorTypes.BROKEN_LINK,
+                    linkToAsset: href,
+
+                    viewablePageInCms: this.getPathInCms(fileName, repo),
+                    viewablePageInStaging: path.join(normalisedPermalink),
+                    linkedText: text,
                   }
+                  errors.push(error)
                 }
               })
               const imgTags = dom.window.document.querySelectorAll("img")
               forEach(imgTags, (tag) => {
                 const src = tag.getAttribute("src")
                 if (!src) {
+                  // while rendered in the dom, unlikely to be
+                  // noticed by end citizen.
+                  return
+                }
+
+                //! todo: include check for post bundled files https://github.com/isomerpages/isomerpages-template/tree/staging/assets/img
+                if (!this.hasValidReference(setOfAllMediaAndPagesPath, src)) {
                   const error: RepoError = {
                     type: RepoErrorTypes.BROKEN_IMAGE,
-                    linkToAsset: "",
+                    linkToAsset: src,
                     viewablePageInCms: this.getPathInCms(fileName, repo),
-                    viewablePageInStaging: path.join(filePermalink),
+                    viewablePageInStaging: path.join(normalisedPermalink),
                   }
                   errors.push(error)
-                } else {
-                  const decodedSrc = this.mediaPathDecoder(src)
-                  //! todo: include check for post bundled files https://github.com/isomerpages/isomerpages-template/tree/staging/assets/img
-                  if (
-                    !this.isExternalLinkOrPageRef(decodedSrc) &&
-                    !setOfAllMediaAndPagesPath.has(decodedSrc)
-                  ) {
-                    const error: RepoError = {
-                      type: RepoErrorTypes.BROKEN_IMAGE,
-                      linkToAsset: src,
-                      viewablePageInCms: this.getPathInCms(fileName, repo),
-                      viewablePageInStaging: path.join(filePermalink),
-                    }
-                    errors.push(error)
-                  }
                 }
               })
               return ok(undefined)
@@ -420,6 +428,24 @@ export default class RepoCheckerService {
       })
   }
 
+  hasValidReference(setOfAllMediaAndPagesPath: Set<string>, src: string) {
+    if (this.isExternalLinkOrPageRef(src)) {
+      return true
+    }
+
+    const decodedSrc = this.mediaPathDecoder(src)
+
+    if (setOfAllMediaAndPagesPath.has(decodedSrc)) {
+      return true
+    }
+
+    // Amplify allows for trailing slash
+    if (setOfAllMediaAndPagesPath.has(`${decodedSrc}/`)) {
+      return true
+    }
+    return false
+  }
+
   reporter(repo: string, errors: RepoError[]): ResultAsync<RepoError[], never> {
     const folderPath = path.join(SITE_CHECKER_REPO_PATH, repo)
     if (!fs.existsSync(folderPath)) {
@@ -433,16 +459,24 @@ export default class RepoCheckerService {
       fs.writeFileSync(filePath, stringifiedErrors)
     }
 
+    let gitOperations:
+      | Response<CommitResult>
+      | Response<PushResult> = this.git
+      .cwd({ path: SITE_CHECKER_REPO_PATH, root: false })
+      .checkout("main")
+      .fetch()
+      .merge(["origin/main"])
+      .add(["."])
+      .commit(`Site checker logs added for ${repo}`)
+
+    const isProd = config.get("env") === "prod"
+    if (isProd) {
+      gitOperations = gitOperations.push()
+    }
+
     // pushing since there is no real time requirement for user
-    return ResultAsync.fromPromise(
-      this.git
-        .cwd({ path: SITE_CHECKER_REPO_PATH, root: false })
-        .checkout("main")
-        .fetch()
-        .merge(["origin/main"])
-        .add(["."])
-        .commit(`Site checker logs added for ${repo}`)
-        .push(),
+    return ResultAsync.fromPromise<CommitResult | PushResult, SiteCheckerError>(
+      gitOperations,
       (error) => new SiteCheckerError(`${error}`)
     )
       .map(() => errors)
@@ -568,7 +602,7 @@ export default class RepoCheckerService {
     const paths = permalink.split("/")
     const initialPath = `/sites/${repoName}`
 
-    const isUngroupedPages = paths.length === 2
+    const isUngroupedPages = paths.length === 2 && paths[0] === "pages"
     if (isUngroupedPages) {
       const fileName = paths[1]
       return `${initialPath}/editPage/${fileName}`
@@ -578,7 +612,7 @@ export default class RepoCheckerService {
     if (isResource) {
       const resourceRoomName = paths[0]
       const resourceCategoryName = paths[1]
-      const resourceName = paths[2]
+      const resourceName = paths[3]
       return `${initialPath}/resourceRoom/${resourceRoomName}/resourceCategory/${resourceCategoryName}/editPage/${resourceName}`
     }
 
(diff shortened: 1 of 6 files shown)


kishore03109:
* **#1187** <a href="https://app.graphite.dev/github/pr/isomerpages/isomercms-backend/1187?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop`

<h2></h2>

src/services/review/RepoCheckerService.ts
```diff
@@ -293,6 +293,16 @@ export default class RepoCheckerService {
...
+  getResourceRoomPath(repoPath: string) {
+    const configPath = path.join(repoPath, "_config.yml")
+    const fileContent = fs.readFileSync(configPath, "utf8")
+    const resourceRoom = fileContent
+      .split("resources_name: ")[1]
+      .split("\n")[0]
+      .trim()
+    return resourceRoom
+  }
+
```
seaerchin, line 305: tbh i think `PageService` has this already but it's fine ba

src/services/review/RepoCheckerService.ts
```diff
@@ -345,62 +356,59 @@ export default class RepoCheckerService {
...
                 const href = tag.getAttribute("href")
-                const text = tag.textContent || ""
-                if (!href) {
+                const text = tag.textContent
+
+                if (!text) {
+                  // while rendered in the dom, unlikely to be
+                  // noticed by end citizen.
+                  return
+                }
```
seaerchin, line 365: honestly i think we should still report this, because it's more likely that this is an error than not. it might also be tab targetted/picked up by screen readers because those rely on the dom structure.

src/services/review/RepoCheckerService.ts
```diff
@@ -345,62 +356,59 @@ export default class RepoCheckerService {
...
-                      viewablePageInStaging: path.join(filePermalink),
-                      linkedText: text,
-                    }
-                    errors.push(error)
+                }
+                if (!href) {
+                  // could be intentional linking to self
+                  // eg. <a href>Back to top</a>
+                  return
+                }
```
seaerchin, line 380: i saw that the users are using empty hrefs - just wanted to check if [this](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#onclick_events) is what they intended? 

if so, we should report an error (not necessarily a broken link error, but something else certainly) so that they can use a button or something more suited for the job.

wdyt?

src/services/review/RepoCheckerService.ts
```diff
@@ -345,62 +356,59 @@ export default class RepoCheckerService {
...
+                  errors.push(error)
                 }
               })
               const imgTags = dom.window.document.querySelectorAll("img")
               forEach(imgTags, (tag) => {
                 const src = tag.getAttribute("src")
                 if (!src) {
+                  // while rendered in the dom, unlikely to be
+                  // noticed by end citizen.
+                  return
```
seaerchin, line 400: similar comment to above

src/services/review/RepoCheckerService.ts
```diff
@@ -433,16 +459,24 @@ export default class RepoCheckerService {
...
-      this.git
-        .cwd({ path: SITE_CHECKER_REPO_PATH, root: false })
-        .checkout("main")
-        .fetch()
-        .merge(["origin/main"])
-        .add(["."])
-        .commit(`Site checker logs added for ${repo}`)
-        .push(),
+    return ResultAsync.fromPromise<CommitResult | PushResult, SiteCheckerError>(
+      gitOperations,
```
seaerchin, line 479: can we put the whole operation into the `if isProd` branch

src/services/review/RepoCheckerService.ts
```diff
@@ -578,7 +612,7 @@ export default class RepoCheckerService {
     if (isResource) {
       const resourceRoomName = paths[0]
       const resourceCategoryName = paths[1]
-      const resourceName = paths[2]
+      const resourceName = paths[3]
```
seaerchin, line 615: this check might be wrong - `isResource` only checks that the permalink doesn't start with an underscore, which doesn't seem strong enough to distinguish between collection/resource pages. 

fwiw, `PageService` has this exact method

src/services/review/RepoCheckerService.ts
```diff
@@ -705,8 +739,16 @@ export default class RepoCheckerService {
 
   mediaPathDecoder = (mediaPath: string) => {
     for (let i = 0; i < ALLOWED_FILE_EXTENSIONS.length; i += 1) {
-      if (mediaPath.toLowerCase().endsWith(ALLOWED_FILE_EXTENSIONS[i])) {
-        return decodeURI(mediaPath)
+      let pathWithoutQueryParams = mediaPath
+      if (mediaPath.includes("?")) {
```
seaerchin, line 734: you might need to handle url encoded strings here too

src/services/review/RepoCheckerService.ts
```diff
@@ -705,8 +739,16 @@ export default class RepoCheckerService {
...
+        pathWithoutQueryParams = mediaPath.split("?").slice(0, -1).join("")
```
seaerchin, line 735: to be safe here, i think we should take **only the first portion** - if there's 2 `?`, this might give the wrong result (only ignores ending `?`)

src/services/review/RepoCheckerService.ts
```diff
@@ -705,8 +739,16 @@ export default class RepoCheckerService {
...
+      }
+      if (
+        pathWithoutQueryParams
+          .toLowerCase()
+          .endsWith(ALLOWED_FILE_EXTENSIONS[i])
```
seaerchin, line 740: idt this is strong enough - `ALLOWED_FILE_EXTENSIONS` only has the ending file ext **without** the dot. this wouldn't be able to distinguish between `png` and `apng` for example

src/services/review/RepoCheckerService.ts
```diff
@@ -819,6 +861,29 @@ export default class RepoCheckerService {
     })
   }
 
+  permalinkNormaliser(permalink: string) {
```
seaerchin, line 855: nit: could we call this `normalisePermalink`

src/services/review/RepoCheckerService.ts
```diff
@@ -819,6 +852,29 @@ export default class RepoCheckerService {
...
+    if (finalPermalink.startsWith(`"`)) {
+      finalPermalink = finalPermalink.slice(1)
+    }
+    if (finalPermalink.endsWith(`"`)) {
+      finalPermalink = finalPermalink.slice(0, -1)
+    }
+
+    if (!finalPermalink.startsWith("/")) {
+      finalPermalink = `/${finalPermalink}`
+    }
```
timotheeg, line 873: This code looks weird 🤔 It looks like you expect to see values wrapped in both double quotes and backticks. Is that  normal?

src/services/review/RepoCheckerService.ts
```diff
@@ -819,6 +852,29 @@ export default class RepoCheckerService {
...
+    if (finalPermalink.endsWith(`'`)) {
+      finalPermalink = finalPermalink.slice(0, -1)
+    }
+    if (finalPermalink.startsWith(`"`)) {
+      finalPermalink = finalPermalink.slice(1)
+    }
+    if (finalPermalink.endsWith(`"`)) {
+      finalPermalink = finalPermalink.slice(0, -1)
+    }
```
seaerchin, line 869: just thinking out loud here - would it be possible for a permalink to have a starting `'` but not a terminating `'`? (also for double quotes) 

this, for example, would also sanitise `'"some_permalink"'` to `some_permalink` but not `"'some_permalink'"` (only removes outer `"`) - again, this is contrived but i'm not sure what behaviour we're aiming for here

seaerchin:
again, this feels like maintenance debt **which we can accomplish just by using an external package**. i'll leave the value judgement to you but from an outsider perspective, checking for "broken" links (direct to 404 on live site) seems better done with an ext package adn we might want to focus our efforts on checking for eg: staging links, which are a 200 but are wrong in the isomer context.

src/services/review/RepoCheckerService.ts
```diff
@@ -705,8 +730,16 @@ export default class RepoCheckerService {
 
   mediaPathDecoder = (mediaPath: string) => {
     for (let i = 0; i < ALLOWED_FILE_EXTENSIONS.length; i += 1) {
-      if (mediaPath.toLowerCase().endsWith(ALLOWED_FILE_EXTENSIONS[i])) {
-        return decodeURI(mediaPath)
+      let pathWithoutQueryParams = mediaPath
```
kishore03109, line 733: create a new uri + get the base url?

src/services/review/RepoCheckerService.ts
kishore03109: hey thanks for bringing this up! abit of early feedback that we have received is that the errors are not true errors, and this has led to agency not fixing the issues.  
currently the product direction for this feature is that we want to make it as easy and as least overwhelming as possible to the users, which is why we are deprio any non-obvious errors to users first, and think we can revisit this again maybe thr design (maybe nice to haves vs actual errors)

src/services/review/RepoCheckerService.ts
kishore03109: same reasoning as above

src/services/review/RepoCheckerService.ts
kishore03109: hey may i clarify if this is a nit ya? this would be useful for testing on staging branch, just without pushing to the remote repository so that only prod uses it. happy to huddle to explain this concept over ya

seaerchin: hmm, for me, i think the problem here is that we `commit` on non-prod environments and don't do anything with it. in that case, why bother committing in the first place? 

ie, we do a `commit + push` only `if isProd`, which makes it clearer

src/services/review/RepoCheckerService.ts
seaerchin: yeap, i think what's important here is _who_ does the determination of errors in ambiguous cases like these. 

as you've rightfully pointed out, some agencies **don't** think that this is an error and rightfully so! but in the case of agencies that just forgot to put a link in, these may actually be errors. 

in that case, rather than a blanket judgement on our end to deem these as non-errors, perhaps what would be more useful here is a categorisation of the severities of the broken links that the checker flags out? 

eg: if your `img` or `a` tag is missing a link entirely, it's flagged as informational, but if those tags have a link pointing to a missing item (eg: `href="some_missing_item"`), we flag it as error. 

that seems to fit what the agency wants, as well as still helping out potential broken links use case

src/services/review/RepoCheckerService.ts
kishore03109: changed to regex pattern [6a6fc29](https://github.com/isomerpages/isomercms-backend/pull/1187/commits/6a6fc296ed62a9292c47c50ad3dc0917908c8d85)

src/services/review/RepoCheckerService.ts
kishore03109: [6a6fc29](https://github.com/isomerpages/isomercms-backend/pull/1187/commits/6a6fc296ed62a9292c47c50ad3dc0917908c8d85) the code to use page service instead

src/services/review/RepoCheckerService.ts
kishore03109: while I am aligned with you that users should be aware of the less severe errors in the long run, I might have to push back on this for this pr. looking at the datadog, we saw instances of repo owners looking at the broken links and then proceeding to approve and merge changes to production 

as of right now, the users of this feature in production have yet to fix the higher priority bugs. the discussion of how to tackle low prior bugs shd occur after we see evidence of agency users tackling high prior bugs. we can then have a discussion with design regarding how to display the not so severe errors, what are your thoughts?

src/services/review/RepoCheckerService.ts
kishore03109: modified in [6a6fc29](https://github.com/isomerpages/isomercms-backend/pull/1187/commits/6a6fc296ed62a9292c47c50ad3dc0917908c8d85) to use page service instead

src/services/review/RepoCheckerService.ts
kishore03109: hey have modified this in this commit [6a6fc29](https://github.com/isomerpages/isomercms-backend/pull/1187/commits/6a6fc296ed62a9292c47c50ad3dc0917908c8d85)

src/services/review/RepoCheckerService.ts
kishore03109: modified [6a6fc29](https://github.com/isomerpages/isomercms-backend/pull/1187/commits/6a6fc296ed62a9292c47c50ad3dc0917908c8d85)

src/services/review/RepoCheckerService.ts
kishore03109: [6a6fc29](https://github.com/isomerpages/isomercms-backend/pull/1187/commits/6a6fc296ed62a9292c47c50ad3dc0917908c8d85)

src/services/review/RepoCheckerService.ts
kishore03109: [6a6fc29](https://github.com/isomerpages/isomercms-backend/pull/1187/commits/6a6fc296ed62a9292c47c50ad3dc0917908c8d85)

src/services/review/RepoCheckerService.ts
kishore03109: using regex now [6a6fc29](https://github.com/isomerpages/isomercms-backend/pull/1187/commits/6a6fc296ed62a9292c47c50ad3dc0917908c8d85)

kishore03109:
> again, this feels like maintenance debt which we can accomplish just by using an external package. i'll leave the value judgement to you but from an outsider perspective, checking for "broken" links (direct to 404 on live site) seems better done with an ext package adn we might want to focus our efforts on checking for eg: staging links, which are a 200 but are wrong in the isomer context.

hey i hear you here, i intend to do a deeper eng dive on the approach next week! this approach isnt set in stone, but what we want to still see is 
1. are our initial users of this feature fixing their broken links 
2. how can the ui for this be improved to help ^ 1

I am ok to treat this as throwaway code after looking into html proofer + other checkers next week, but we still need something in production to get data on 1 and 2!

src/services/review/RepoCheckerService.ts
seaerchin: > the discussion of how to tackle low prior bugs shd occur after we see evidence of agency users tackling high prior bugs

yeap, that's fine but the approach taken here is to declassify them as bugs, which seems to run counter to what you're saying ._.

we can just leave these checks in first then instead of pre-emptively removing them?

src/services/review/RepoCheckerService.ts
kishore03109: sure [5cb14f5](https://github.com/isomerpages/isomercms-backend/pull/1187/commits/5cb14f5edd3233d524e4014177680b4c9957a33f)

src/services/review/RepoCheckerService.ts
kishore03109: resolving this to talk about this in a single thread

src/services/review/RepoCheckerService.ts
kishore03109: discussed offline, filing ticket for this!

seaerchin: copy pasting from slack discussion: 
1. we shouldn't enable user laziness - regardless of whether it is a bug or not, the user should always be the one making the judgement rather than us determining that it isn't and aiding them to avoid that judgement
2. this shouldn't be a permanent measure - just a temporary one, with clear hypothesis of what this should encourage/discourage and follow-ups from there. 

i'm ok with this **as a temporary measure** - and we need to be clear about this. i've filed two tickets tracking this and we should have a clear yes/no decision following on

src/services/review/RepoCheckerService.ts
seaerchin: can't see the commit but marking as resolved

kishore03109: disappeared after rebase [8bb11a1](https://github.com/isomerpages/isomercms-backend/pull/1187/commits/8bb11a12e16b22fac09cfe5aaf1b4e79f919d0b6)

src/services/review/RepoCheckerService.ts
kishore03109: commit changed after rebase [8bb11a1](https://github.com/isomerpages/isomercms-backend/pull/1187/commits/8bb11a12e16b22fac09cfe5aaf1b4e79f919d0b6)

seaerchin (approved):
lgtm - assuming that we'll relook at presentation + bug fixes are in

PR #1189, opened 2024-03-07
feat(dev productivity): faster hot reloads

kishore03109:
I am not sure about the rest but using docker during local dev has been a pain due to the following reasons:

1. reloads take ard 34 seconds compared to 3 seconds that running directly took

using docker:
Screenshot 2024-03-13 at 11 46 53 AM

using local:
Screenshot 2024-03-13 at 11 56 08 AM

2. sometimes the docker hangs + need to restart which takes ard 1 minute 
Screenshot 2024-03-13 at 11 49 57 AM

4. (cannot replicate this) other eng have raised that everytime got new dep they need to delete and rebuild the image
5. given enough reloads, i keep getting pipe errors, which is only fixed by a restart
Screenshot 2024-03-13 at 11 47 43 AM

I have been using this to speed up my development process, codifying this in case this is useful to other devs. 

not advocating for the default to be run locally, code that revolves around logic that affects ecs should still be tested via the docker. but for dev that is purely app level, the above limitations were annoying enough :(


kishore03109:
* **#1189** <a href="https://app.graphite.dev/github/pr/isomerpages/isomercms-backend/1189?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop`

<h2></h2>

seaerchin:
> the hot reloads take a while, and I keep running to weird PIPE related errors.

could you describe what pipe errors you got

seaerchin (commented):
actually 1 thing about using docker here is that the execution context is exactly the same as that of prod/staging, so that errors can be flagged out if they exist.

i think the solution here should be to fix hte errors w/ docker rather than shifting the service out of docker itself cos then things like ssh keys injection/setting of known hosts get omitted

kishore03109:
@seaerchin added more info to the pr desc, ya stance is should not be default, am not updating readme for this, just a local optimisation only

kishore03109 (commented):
no need already, updating local docker works already

PR #1193, opened 2024-03-07
fix: send updated data back to client, from all update pages endpoints

jacobkwan:
## Problem
https://opengovproducts.slack.com/archives/CK68JNFHR/p1709007514442959?thread_ts=1708332491.637599&cid=CK68JNFHR

Fixes issue where users see warning modal even after saving changes

Closes [insert issue #]

## Solution
1. updating/saving/syncing state logic is honestly quite convoluted on the FE
2. However, frontend is indeed updating editor content **after** the POST request returns (see `useUpdatePageHook` in `EditPageLayout`)
3. issue is that `/update` endpoints are not returning the updated `frontmatter` and `pageContent` to the client

This seems like the simplest solution to fixing this bug, i.e returning updated `pageContent` to the client

**Breaking Changes**

  - Details ...

**Features**:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- Details ...

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Tests
1. Go to Prod CMS, edit a page to leave trailing whitespace then save changes
2. Click Back on the navbar -> should see the unsaved changes modal
3. Go to Staging CMS, edit a page to leave trailing whitespace then save changes
4. After saving, should see that trailing whitespace is removed from the editor (to match backend)
5. Click Back on the navbar -> should not see the modal

## Deploy Notes

- `env var` : env var details

**New scripts**:

- `script` : script details

**New dependencies**:

**New dev dependencies**:

### DIFF
--- a/src/services/fileServices/MdPageServices/CollectionPageService.js
+++ b/src/services/fileServices/MdPageServices/CollectionPageService.js
@@ -70,6 +70,10 @@ class CollectionPageService {
   ) {
     const parsedCollectionName = `_${collectionName}`
     const newContent = convertDataToMarkdown(frontMatter, content)
+    const {
+      frontMatter: newFrontMatter,
+      pageContent: newPageContent,
+    } = retrieveDataFromMarkdown(newContent)
     const { newSha } = await this.gitHubService.update(sessionData, {
       fileContent: newContent,
       sha,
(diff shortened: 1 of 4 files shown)


kishore03109 (commented):
hmm i dont understand this bug.
is this deterministic? 
> issue is that /update endpoints are not returning the updated frontmatter and pageContent to the client

hmm what are the changes between frontMatter and new frontmatter? 

the ticket is for nlb, which means that it is not required to actually do a post call to gh, so am unclear what the diff here would be?

kishore03109 (approved):
discussed offline, understand this now

PR #1198, opened 2024-03-08
chore: add cloudmersive details

alexanderleegs:
Adds additional details to our cloudmersive API call in case of failure.

### DIFF
--- a/src/utils/file-upload-utils.js
+++ b/src/utils/file-upload-utils.js
@@ -47,7 +47,7 @@ const scanFileForVirus = (fileBuffer, timeout) => {
     apiInstance.scanFile(fileBuffer, (error, data) => {
       if (error) {
         logger.error(
-          `Error when calling Cloudmersive Virus Scan API: ${error.message}`
+          `Error when calling Cloudmersive Virus Scan API for the following file content: ${fileBuffer}: ${error.message}`
         )
         failure(error)
       } else {


kishore03109 (approved):

src/utils/file-upload-utils.js
```diff
@@ -47,7 +47,7 @@ const scanFileForVirus = (fileBuffer, timeout) => {
     apiInstance.scanFile(fileBuffer, (error, data) => {
       if (error) {
         logger.error(
-          `Error when calling Cloudmersive Virus Scan API: ${error.message}`
+          `Error when calling Cloudmersive Virus Scan API for the following file content: ${fileBuffer}: ${error.message}`
```
harishv7, line 50: does this show the file extension?

fileBuffer is not the binary right? are we printing out the entire media inside the buffer?

alexanderleegs: I was thinking we needed the actual file content to debug issues with cloudmersive - what we're logging here is exactly what we submit to cloudmersive so that we can debug if necessary

harishv7: Hmm would that be helpful? a file buffer is going to be huge, what we probably need is the actual file extension and the mime type - https://www.npmjs.com/package/mime-types?

alexanderleegs:
Closing - as discussed offline, logging the file binary doesn't help us to debug much and clogs up our logs - what might be more important here is that the oncall engineer knows how to debug this kind of scenario. I'll add an additional entry to our runbook on debugging this situation instead

PR #1200, opened 2024-03-11
feat(auditlogs): add relevant GitHub link for each audit log entry

dcshzj:
## Problem

The CSV file is very limiting and is very difficult for users to tell exactly what changed from the commit message.

## Solution

**Breaking Changes**

**Features**:

- Add a link to the relevant GitHub item (either commit or PR) for users to find out more information about the individual audit log items.
- Also added some basic unit tests for some of the supporting functionality. The core functionality remains untested as it is still subject to change.

## Tests

- [ ] Fill in the form for staging to get the audit logs for a valid test site
- [ ] Verify that the link columns show up links to the respective GitHub commit or PR

## Deploy Notes

_None_

### DIFF
--- a/src/types/auditLog.ts
+++ b/src/types/auditLog.ts
@@ -16,4 +16,5 @@ export type AuditLog = {
   actor: string
   page: string
   remarks: string
+  link?: string
 }
--- a/src/services/admin/AuditLogsService.ts
+++ b/src/services/admin/AuditLogsService.ts
@@ -215,12 +215,13 @@ class AuditLogsService {
                       return errAsync(error)
                     })
                     .andThen((actorEmail) =>
-                      okAsync({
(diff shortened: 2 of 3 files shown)


src/types/auditLog.ts
```diff
@@ -16,4 +16,5 @@ export type AuditLog = {
   actor: string
   page: string
   remarks: string
+  link?: string
```
kishore03109, line 19: why is this optional ah? isnt this available for all commits?

dcshzj: Hmm this one is for forward compatibility, because not every audit log item will have a relevant link. For instance, when we support adding/removing collaborators, we won't be able to provide a link for this, so I opted to keep it optional for now.

kishore03109 (approved):

PR #1206, opened 2024-03-12
fix(jwt-utils): remove deprecated lib

seaerchin:
## Problem
`crypto-js` is outdated and lacks support. instead of using `crypto-js`, we should use node's inbuilt `crypto` library 

## Solution
- swapover to using `crypto` rather than `crypto-js` for encryption. we are **not** using `gcm` or using an `iv` because this incurs downtime on our app (the case where you have old version of BE required to decrypt smth encrypted w/ iv or gcm will never work hence requiring downtime) 
- this trade-off is ok and has been communicated to @jandonyada; we're only losing security between db/app and what we're guarding against here is losing tokens in plaintext hence giving attackers a vector of abuse for said tokens

### DIFF
--- a/src/middleware/authentication.ts
+++ b/src/middleware/authentication.ts
@@ -9,7 +9,7 @@ import { GrowthBookAttributes } from "@root/types/featureFlags"
 import { RequestWithGrowthBook } from "@root/types/request"
 
 interface RequestWithSession extends RequestWithGrowthBook {
-  session: Session & SessionData & { algo: "aes-256-gcm" }
+  session: Session & SessionData
 }
 
 // eslint-disable-next-line import/prefer-default-export
--- a/src/middleware/authentication.ts
+++ b/src/middleware/authentication.ts
@@ -42,7 +42,6 @@ export class AuthenticationMiddleware {
         cookies,
         url,
         userInfo: session.userInfo,
-        algo: session.algo,
       })
       const userSessionData = new UserSessionData({
         isomerUserId,
--- a/src/middleware/authentication.ts
+++ b/src/middleware/authentication.ts
@@ -9,7 +9,7 @@ import { GrowthBookAttributes } from "@root/types/featureFlags"
 import { RequestWithGrowthBook } from "@root/types/request"
 
 interface RequestWithSession extends RequestWithGrowthBook {
-  session: Session & SessionData
+  session: Session & SessionData & { algo: "aes-256-gcm" }
 }
 
 // eslint-disable-next-line import/prefer-default-export
@@ -42,6 +42,7 @@ export class AuthenticationMiddleware {
         cookies,
         url,
         userInfo: session.userInfo,
+        algo: session.algo,
       })
       const userSessionData = new UserSessionData({
         isomerUserId,
--- a/src/utils/jwt-utils.js
+++ b/src/utils/jwt-utils.js
@@ -6,8 +6,12 @@ const _ = require("lodash")
 const { config } = require("@config/config")
 
 const JWT_SECRET = config.get("auth.jwtSecret")
-const ENCRYPTION_SECRET = config.get("auth.encryptionSecret")
-const AUTH_TOKEN_EXPIRY_MS = config.get("auth.tokenExpiryInMs").toString()
+// const ENCRYPTION_SECRET = config.get("auth.encryptionSecret")
+const ENCRYPTION_SECRET = Buffer.from(crypto.randomBytes(16)).toString("hex")
+const AUTH_TOKEN_EXPIRY_MS = config.get("auth.tokenExpiry").toString()
+
+const IV_LENGTH = 16
+const AUTH_TAG_LENGTH = 32
 
 const jwtUtil = {
   decodeToken: _.wrap(jwt.decode, (decode, token) => decode(token)),
@@ -25,20 +29,28 @@ const jwtUtil = {
     // This is because we generate 8 random bytes (1 byte = 8 bits)
     // but hex is 4 bits per character..
     const iv = crypto.randomBytes(8).toString("hex")
-    const cipher = crypto.createCipheriv("aes-256-cbc", ENCRYPTION_SECRET, iv)
+    const cipher = crypto.createCipheriv("aes-256-gcm", ENCRYPTION_SECRET, iv)
     let encrypted = cipher.update(token, "utf8", "base64")
     encrypted += cipher.final("base64")
 
-    return `${iv}${encrypted}`
+    // NOTE: Auth tag is 32 chars long
+    const authTag = cipher.getAuthTag().toString("hex")
+
+    return `${iv}${authTag}${encrypted}`
   },
   decryptToken: (encrypted) => {
-    const iv = encrypted.slice(0, 16)
-    const encryptedText = encrypted.slice(16)
+    const iv = encrypted.slice(0, IV_LENGTH)
+    const authTag = Buffer.from(
+      encrypted.slice(IV_LENGTH, IV_LENGTH + AUTH_TAG_LENGTH),
+      "hex"
+    )
+    const encryptedText = encrypted.slice(32 + 16)
     const decipher = crypto.createDecipheriv(
-      "aes-256-cbc",
+      "aes-256-gcm",
       ENCRYPTION_SECRET,
       iv
     )
+    decipher.setAuthTag(authTag)
     let decrypted = decipher.update(encryptedText, "base64", "utf8")
     decrypted += decipher.final("utf8")
     return decrypted
--- a/src/utils/jwt-utils.js
+++ b/src/utils/jwt-utils.js
@@ -1,5 +1,5 @@
-const CryptoJS = require("crypto-js")
-const AES = require("crypto-js/aes")
+const crypto = require("crypto")
+
 const jwt = require("jsonwebtoken")
 const _ = require("lodash")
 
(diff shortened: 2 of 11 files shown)


src/utils/jwt-utils.js
```diff
@@ -20,12 +20,29 @@ const jwtUtil = {
...
-    encrypt(token, ENCRYPTION_SECRET).toString()
-  ),
-  decryptToken: _.wrap(AES.decrypt, (decrypt, token) =>
-    decrypt(token, ENCRYPTION_SECRET).toString(CryptoJS.enc.Utf8)
-  ),
+  encryptToken: (token) => {
+    // NOTE: The iv here is 16 characters long.
+    // This is because we generate 8 random bytes (1 byte = 8 bits)
+    // but hex is 4 bits per character..
+    const iv = crypto.randomBytes(8).toString("hex")
```
kishore03109, line 27: hmm i think this wont work, we need the same iv over all the instances!
what we want here is an env var rather than creating this randomly at runtime!

kishore03109 (changes requested):
I think since this flow affects logins, could i trouble you to ensure the main flows here are tested on staging? mainly sgid + github app auth flow?

src/utils/jwt-utils.js
seaerchin: we don't want an env var - the iv needs to be unpredictable and unique (best is cryptographically random, see [here](https://en.wikipedia.org/wiki/Initialization_vector))

this works by appending the iv itself to the encrypted token and using it at the decrypt stage to check.

Screenshot 2024-03-12 at 6 50 17 PM

seaerchin:
> I think since this flow affects logins, could i trouble you to ensure the main flows here are tested on staging? mainly sgid + github app auth flow?

it should only affect github logins but yea will test. tbh, integration/unit tests here is sufficient, because we should decrypt to the same value but doesn't hurt to test

src/utils/jwt-utils.js
kishore03109: hmm but there is a chance that the encrypt and decrypt steps occur in different different instances? with the different values over say two instances, how will this work ah?

kishore03109 (commented):
not sure if i have a misconception here, but if since the decrypt step and encrypt step can occur at different times over different instances, and we need the same iv to encrypt and decrypt a token, not sure how this would work when we have multiple instances with different versions of iv. if this is a misunderstanding, can revoke my request for changes 

also this pr is not backward compatible right? during that 10 mins of rolling deploy, some jwts will be signed with an iv while others will not, have we gotten clearance with the team regarding this deployment?

seaerchin:
> not sure if i have a misconception here, but if since the decrypt step and encrypt step can occur at different times over different instances, and we need the same iv to encrypt and decrypt a token, not sure how this would work when we have multiple instances with different versions of iv. if this is a misunderstanding, can revoke my request for changes

i'll try to map out the login flow for github (where this is used) as best as i can here, if you still have queries, we can huddle.

1. user clicks on github login on FE 
2. this brings them to `/v2/auth/github-redirect` 
3. we redirect them to github's oauth flow with a **signed** state token 
4. user signs in on github, github redirects them to `/v2/auth/`
5. verify cookie state, get access token + encrypt, store stuff into `req.session` (connected to db) and redirect users to `/sites` on frontend.

when user comes in, the session id from cookie is then used to fetch from `req.session`.

src/utils/jwt-utils.js
spaceraccoon: Initialisation vectors **must be randomized** otherwise it breaks the security of encryption. The initialisation vector is **prepended** to the cipher text, which the decryption then uses to initialise itself as well.
https://knowledge-base.secureflag.com/vulnerabilities/broken_cryptography/reused_iv_key_pair_vulnerability.html:

> The consequences of reusing the same IV-Key pair depend on the actual cipher and mode used. In the case of Cipher Block Chaining (CBC) and Cipher Feedback (CFB) modes, the leaks of information from the first block and commonalities are enough to lead an empowered attacker to fully compromising the encryption layer of protection in the given system.

spaceraccoon: To emphasise: the initialisation vector is randomised but not encrypted, it's just prepended before the actual encrypted cipher text. It's used to initialise the cipher randomly otherwise it's vulnerable to typical bit flipping and known planet attacks. So there's no problem across different instances, cos the decrypting cipher only needs to take it from the prepended IV in the cipher text.

src/utils/jwt-utils.js
```diff
@@ -20,12 +20,29 @@ const jwtUtil = {
...
+    const cipher = crypto.createCipheriv("aes-256-cbc", ENCRYPTION_SECRET, iv)
```
spaceraccoon, line 28:
```suggestion
    const cipher = crypto.createCipheriv("aes-256-gcm", ENCRYPTION_SECRET, iv)
```

Btw, why not AES-GCM? Not super necessary given the context, but has the advantage of preventing forged cipher texts as long as your IV is randomised.

spaceraccoon (commented):
LGTM - there's some discussion whether to use `aes-256-gcm` instead

src/utils/jwt-utils.js
seaerchin: actually i didn't wanna do it cos i'm not super familiar with crypto (or security stuff) in general. 

i changed it here cos it was simple enuf tho pog

src/middleware/authentication.ts
```diff
@@ -42,6 +42,7 @@ export class AuthenticationMiddleware {
         cookies,
         url,
         userInfo: session.userInfo,
+        algo: session.algo,
```
kishore03109: sir type error here

src/utils/jwt-utils.ts
```diff
@@ -0,0 +1,52 @@
...
+  signToken: _.wrap(jwt.sign, (sign, content: string) =>
+    sign(content, JWT_SECRET, {
+      algorithm: "HS256",
+      expiresIn: AUTH_TOKEN_EXPIRY_MS,
+    })
+  ),
+  encryptToken: (token: string, secret?: string) => {
+    const encryptionSecret = secret ?? ENCRYPTION_SECRET
+
+    const cipher = crypto.createCipher(
```
kishore03109, line 25: `crypto.createCipher` is deperciated. lets use `createCipheriv()` instead ya

src/utils/jwt-utils.ts
```diff
@@ -0,0 +1,52 @@
...
+    )
+    let encrypted = cipher.update(token, "utf8", "base64")
+    encrypted += cipher.final("base64")
+
+    return encrypted
+  },
+  decryptToken: (encrypted: string, secret?: string) => {
+    const encryptionSecret = secret ?? ENCRYPTION_SECRET
+
+    const decipher = crypto.createDecipher(
```
kishore03109, line 40: similar comment as above, use `createDecipheriv()` instead ya

kishore03109 (commented):
I am approving this pr with the following assumptions:

`algo: session.algo,` in `authentication.ts` is was added by accident and to be removed
Jan is aware that this means that we are accepting the risk in the vapt report
We modify the `crypto` calls to use the non-depreciated variants

if these assumptions do not hold, please revoke this review

kishore03109 (dismissed):

src/utils/jwt-utils.ts
seaerchin: we can't, see PR description

> we are not using gcm or using an iv because this incurs downtime on our app (the case where you have old version of BE required to decrypt smth encrypted w/ iv or gcm will never work hence requiring downtime)

i've double-checked w @spaceraccoon on this and we're aligned that this is ok - the threat vector we want to mitigate is leaking plaintext creds, which this does help with. 

sadly using `iv` and/or `gcm` will always incur downtime so this was a conscious choice.

src/middleware/authentication.ts
seaerchin: fixed

seaerchin:
> I am approving this pr with the following assumptions:
> 
> 1. `algo: session.algo,` in `authentication.ts` is was added by accident and to be removed
> 2. Jan is aware that this means that we are accepting the risk in the vapt report
> 3. We modify the `crypto` calls to use the non-depreciated variants
> 
> if these assumptions do not hold, please revoke this review

re: 1 - was added to try and do a zero-downtime deploy but it's not possible.
re: 2 - yes. the [risk](https://opengovproducts.slack.com/archives/C06FXS46AFP/p1710397654845389) is really only between db/app.
re: 3 - no, won't happen for reasons given in PR desc. revokign review cos of this

src/utils/jwt-utils.ts
kishore03109: @seaerchin 
no bah this just changing the function, iv can be null when calling `createCipheriv()` according to the [docs](https://nodejs.org/docs/latest-v18.x/api/crypto.html#cryptocreatecipherivalgorithm-key-iv-options) 

> If the cipher does not need an initialization vector, iv may be null.

kishore03109 (approved):
approving again since i think got misconception here haha

ok with not using iv, just need to use the non-depreciated func call with an null iv can alr

seaerchin:
> approving again since i think got misconception here haha
> 
> ok with not using iv, just need to use the non-depreciated func call with an null iv can alr

can you try using `null`? the tests don't pass if i call using the `iv` function with `iv=null`

PR #1207, opened 2024-03-13
feat(pino): add logger

seaerchin:
## Problem
this pr adds a structured logging library (`pino`) to our codebase. it also supersedes `winston` (which hasn't been working anyway).

Closes [ISOM-772]

## Solution
1. add both `pino` and `pino-pretty` (will be described below) 
2. replaced `consoleLogger` implementation with a call to `pino` with transport as `pino-pretty`. `pino` pipes logs **by default** to `stdout`. this means that on local, the logs get transformed by `pino-pretty` before going to `stdout`, so it'll be seen as unstructured text instead of json. 
3. for production loggers, we **also pipe to `stdout`** - this gets picked up by aws and forwarded to the correct log group. **this is actually how logging works now** - our `winston` logger has been missing a `transport` and broken since forever. this was also chosen **over** using `pino-cloudwatch` because the latter is **legacy**.

Screenshot 2024-03-13 at 2 56 56 PM

see examples of the logger at work [https://ogp.datadoghq.com/logs?query=service:isomer-infra%20env:%22staging%22%20sourc[…]=stream&from_ts=1710241200000&to_ts=1710244740000&live=false](https://ogp.datadoghq.com/logs?query=service:isomer-infra%20env:%22staging%22%20source:%22elasticbeanstalk/cms-backend-staging-node18/var/log/web.stdout.log%22%20-@aws.awslogs.logStream:i-*%20&cols=host,service&fromUser=true&index=*&messageDisplay=inline&sort=time&storage=hot&stream_sort=time,asc&view=spans&viz=stream&from_ts=1710241200000&to_ts=1710244740000&live=false)

### DIFF
--- a/src/logger/constants.ts
+++ b/src/logger/constants.ts
@@ -1,0 +1,14 @@
+import pino from "pino"
+
+// eslint-disable-next-line import/prefer-default-export
+export const PINO_OPTIONS = {
+  // NOTE: DO NOT format time in-process
+  // as this will affect perf of the logger.
+  // See here: https://getpino.io/#/docs/api?id=timestamp-boolean-function
+  timestamp: pino.stdTimeFunctions.isoTime,
+  errorKey: "error",
+  messageKey: "message",
+  formatters: {
+    level: (label: string) => ({ level: label.toUpperCase() }),
+  },
+}
--- a/src/logger/logger.ts
+++ b/src/logger/logger.ts
@@ -76,12 +76,10 @@ export class IsomerLogger implements ExtendedLogger {
 
 const logger = new IsomerLogger()
 
-if (!USE_CONSOLE_LOGGER) {
-  logger.use(pinoLogger)
-}
-
 if (USE_CONSOLE_LOGGER) {
   logger.use(consoleLogger)
+} else {
+  logger.use(pinoLogger)
 }
 
 export default logger
--- a/src/logger/logger.ts
+++ b/src/logger/logger.ts
@@ -1,8 +1,5 @@
-import moment from "moment-timezone"
-
 import { config } from "@root/config/config"
 
-import CloudWatchLogger from "./cloudwatch.logger"
 import { consoleLogger } from "./console.logger"
 import {
   Formatter,
@@ -14,14 +11,11 @@ import {
   LoggerVariants,
   WithFatal,
 } from "./logger.types"
+import { pinoLogger } from "./pino.logger"
 
 const NODE_ENV = config.get("env")
-const useCloudwatchLogger =
-  NODE_ENV === "prod" || NODE_ENV === "vapt" || NODE_ENV === "staging"
-const useConsoleLogger = !(NODE_ENV === "test")
 
-const timestampGenerator = () =>
-  moment().tz("Asia/Singapore").format("YYYY-MM-DD HH:mm:ss")
+const USE_CONSOLE_LOGGER = NODE_ENV === "dev" || NODE_ENV === "test"
 
 const hasDebug = (logger: LoggerVariants): logger is WithDebug<Logger> =>
   (logger as WithDebug<Logger>).debug !== undefined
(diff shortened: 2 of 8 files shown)


seaerchin:
auth spec is somehow failing but this doesn't touch it

dcshzj:
@seaerchin the auth spec fixed alr, can rebase from develop

src/logger/logger.ts
```diff
@@ -81,8 +75,13 @@ export class IsomerLogger implements ExtendedLogger {
 }
 
 const logger = new IsomerLogger()
-if (useConsoleLogger) logger.use(consoleLogger)
-if (useCloudwatchLogger) logger.use(new CloudWatchLogger())
-logger.useFormatter((message) => `${timestampGenerator()}: ${message}`)
+
+if (!USE_CONSOLE_LOGGER) {
```
harishv7, line 79: nit: which is there 2 if statements, can just combine to 1 if else?

harishv7 (commented):
lgtm 1 minor comment

harishv7:
@seaerchin will adding of context attributes in our log statements be done in future PRs?

harishv7 (approved):

src/logger/logger.ts
seaerchin: done

src/logger/constants.ts
```diff
@@ -0,0 +1,14 @@
+import pino from "pino"
+
+// eslint-disable-next-line import/prefer-default-export
+export const PINO_OPTIONS = {
+  // NOTE: DO NOT format time in-process
+  // as this will affect perf of the logger.
+  // See here: https://getpino.io/#/docs/api?id=timestamp-boolean-function
+  timestamp: pino.stdTimeFunctions.isoTime,
+  errorKey: "error",
```
harishv7, line 9: we are not yet injecting the error / message in json for our log statements right?

dcshzj (approved):

seaerchin:
> @seaerchin will adding of context attributes in our log statements be done in future PRs?

this is already done at present unless you mean **per module** attributes. if that's the case, we can use pino child [loggers](https://getpino.io/#/docs/child-loggers) on a per module basis

src/logger/constants.ts
seaerchin: we are!

timotheeg:
Right, I didn't think we'd merge that as-is but it's fine, it should work and we can iterate to make things better:

Known issue so far:
- Logging from axios instance are not connected to APM traces. That needs to be understood and remedied

Also for reference, the condition selection of loggers from env vars is a bit iffy, Armoury has recently adopted a different system which is simpler and more reliable and we can consider adopting. Reference PR [here](https://github.com/opengovsg/armoury/pull/485).

seaerchin:
> Known issue so far:
> 
> * Logging from axios instance are not connected to APM traces. That needs to be understood and remedied

re: axios logging not connected - my hunch is that it's not being connected **because we start a new trace**. due to the call of `tracer.use`, we create a new trace and the logs are not connected to the original trace - we _should_ be able to verify this post rollout tomorrow. 

> Also for reference, the condition selection of loggers from env vars is a bit iffy, Armoury has recently adopted a different system which is simpler and more reliable and we can consider adopting. Reference PR [here](https://github.com/opengovsg/armoury/pull/485).

actually this seems smarter hahahahahahah will change

src/logger/constants.ts
```diff
@@ -0,0 +1,14 @@
+import pino from "pino"
+
+// eslint-disable-next-line import/prefer-default-export
+export const PINO_OPTIONS = {
+  // NOTE: DO NOT format time in-process
+  // as this will affect perf of the logger.
+  // See here: https://getpino.io/#/docs/api?id=timestamp-boolean-function
+  timestamp: pino.stdTimeFunctions.isoTime,
```
timotheeg, line 8: Just a note on that: using `pino.stdTimeFunctions.isoTime` **is** formatting time in process 😑😅  . The "cheap" option is to leave the default `timestamp:true` to get the plain integer timestamp and "out of process" means you transform to human readable string when you use/process logs.

[Implementation in pino for reference](https://github.com/pinojs/pino/blob/dbb394fc8d4bafad26814e39ee9fbfab1f59fe70/lib/time.js#L9)

seaerchin: wow i got scammed, i thought their docs on the `timestamp` there would be out of process. tbh, pino uses a worker process to pipe logs over to cw, so it depends on what their "process" here refers to (main thread or worker thread). 

in any case, no harm just removing this - if we're relying on log integration, the trace itself should have a timestamp and there's already 2 hops for our log (console -> log group, log group -> dd) so i think less latency the better

PR #1208, opened 2024-03-13
fix(tracer): fix tag setting on span

seaerchin:
## Problem
we were setting the tags for the trace as a **bootstrap** process. this means that the `tracer.init` call does not bootstrap properly (as `tracer.use` is meant to be called at start-up, **prior** to actual tracing). 

this PR changes it so that the tags are set on the span.

## Solution
1. instead of calling `tracer.use`, which is a bootstrap call, change to get the active span and calling `setTag` on that span.

## Tests

- [ ] Find traces in isomer's APM like [this one](https://ogp.datadoghq.com/apm/trace/65e9892a000000002705741a8c4efda0?colorBy=service&env=production&graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=7086274656857169657&spanViewType=metadata&timeHint=1709803818752.001&view=spans), and verify that the tags are present

### DIFF
--- a/src/services/api/AxiosInstance.ts
+++ b/src/services/api/AxiosInstance.ts
@@ -46,17 +46,12 @@ const requestFormatter = async (axiosConfig: AxiosRequestConfig) => {
   // NOTE: This also implies that the user has not provided
   // their own github token and hence, are email login users.
   const isEmailLoginUser = getIsEmailUserFromAuthMessage(authMessage)
+  const span = tracer.scope().active()
 
   if (isEmailLoginUser) {
     const accessToken = await tokenServiceInstance.getAccessToken()
     if (accessToken.isOk()) {
-      tracer.use("http", {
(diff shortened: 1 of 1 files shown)


timotheeg (commented):
I suppose we can merge this first since it's meant to be equivalent as what was there before, just cleaner.

But it's still weird to add tags to the scope's span, when we are in the context of a specific axios call.

Since a request handled by express can make multiple axios calls, typically we want to:
1. track the request user itself once at the level of the request's scope, and
2. if we need to track information at the level of the individual axios calls, then they should have their own spans, which would carry their own tags. That, or have log entries with enough structured data.

For 1. there's even a [tracer.setUser()](https://datadoghq.dev/dd-trace-js/index.html#set-user) utility for that

timotheeg (approved):

PR #1212, opened 2024-03-14
feat(pino): removes extraneous stuff

seaerchin:
## Problem
we don't need to differentiate between prod/console loggers due to how pino implements it as a transport (thx pino) so we removed the whole logger folder.

additionally, we shifted the prettifying to be on the cli side so that there's no logic at the code level.

## Solution
1. shift conditional to just invoking via cli 
2. set `logInjection` to be dynamic based on the env as otherwise the dd object would be appended to **every** single pino log to the console


src/utils/tracer.ts
```diff
@@ -1,7 +1,13 @@
...
 
+import { config } from "../config/config"
+
+const env = config.get("dataDog.env")
+
 tracer.init({
-  logInjection: true, // https://docs.datadoghq.com/tracing/other_telemetry/connect_logs_and_traces/nodejs/
+  // NOTE: We require that the `logInjection` be dynamically set
+  // as otherwise it'll inject the dd object on local
+  logInjection: env !== "local", // https://docs.datadoghq.com/tracing/other_telemetry/connect_logs_and_traces/nodejs/
```
kishore03109, line 10: Hmm, why do we have two separate envs for dd ah? like why cant we just use node_env here?

kishore03109 (approved):
no functional blockers, but feels like we can remove an env var here

src/utils/tracer.ts
seaerchin: we need 3 env vars for datadog [unified service tagging](https://docs.datadoghq.com/getting_started/tagging/unified_service_tagging/). since this is a datadog specific config for env, i opted to use this rather than node env.

timotheeg (approved):
lgtm, great to see a PR full of red for deletions!

src/middleware/stats.ts
```diff
@@ -14,7 +14,7 @@ const wrapAsRequestHandler = (sideEffect: SideEffect): RequestHandler => (
   res,
   next
 ) => {
-  sideEffect().catch(logger.info)
+  sideEffect().catch((err) => logger.error(err))
```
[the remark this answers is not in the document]
seaerchin: from their [docs](https://getpino.io/#/docs/api?id=mergingobject-object), yes!

PR #1220, opened 2024-03-15
perf(I/O): rm blocking fs calls

kishore03109:
## Problem

We have quite a number of blocking I/O calls that has lead to unwanted CPU spikes in the event pool thread. 

This PR scrapes any existence of I/O calls and ensures that there is no lingering sync fs calls made from node. 

## Solution
use `fs.promises` instead

Looking at the spike after running the site checker
### NOTE: initial spike in event loop delay is due to the startup push to staging environment, this is not due to the site checker calls that were made
Screenshot 2024-03-19 at 2 30 02 PM

Screenshot 2024-03-19 at 1 30 03 PM

**Breaking Changes**

  - Details ...

**Features**:

when running the site repair form:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- Details ...

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Tests

- [ ] Run the following command from the command line: 
```
grep -rE '(accessSync|appendFileSync|chmodSync|chownSync|closeSync|copyFileSync|cpSync|existsSync|fchmodSync|fchownSync|fdatasyncSync|fstatSync|fsyncSync|ftruncateSync|futimesSync|lchmodSync|lchownSync|lutimesSync|linkSync|lstatSync|mkdirSync|mkdtempSync|opendirSync|openSync|readdirSync|readFileSync|readlinkSync|readSync|readvSync|realpathSync|renameSync|rmdirSync|rmSync|statSync|statfsSync|symlinkSync|truncateSync|unlinkSync|utimesSync|writeFileSync|writeSync|writevSync)\b' src 
```

the only results should be from the `GitFileSystemService.spec.ts` which is fine since this test file runs locally and not in prod line

- [ ] Submit a form [here](https://form.gov.sg/65e4a3d2c25a061b046f3f01) for a repo in staging efs, and assert that the attachments are sent properly

- [ ] submit the site create form 

## Deploy Notes

- `env var` : env var details

**New scripts**:

- `script` : script details

**New dependencies**:

**New dev dependencies**:

### DIFF
--- a/src/services/identity/ReposService.ts
+++ b/src/services/identity/ReposService.ts
@@ -399,21 +399,29 @@ export const createRecords = (zoneId: string): Record[] => {
       .cwd({ path: stgLiteDir, root: false })
       .checkout("staging")
 
-    const doesImagesFolderExist = await doesDirectoryExist(
+    const doesImagesFolderExistResult = await doesDirectoryExist(
       path.join(`${stgLiteDir}`, `images`)
     )
 
-    if (doesImagesFolderExist) {
+    if (doesImagesFolderExistResult.isErr()) {
+      throw doesImagesFolderExistResult.error
+    }
+
+    if (doesImagesFolderExistResult.value) {
       await this.simpleGit
         .cwd({ path: stgLiteDir, root: false })
         .rm(["-r", "images"])
     }
 
-    const doesFilesFolderExist = await doesDirectoryExist(
+    const doesFilesFolderExistResult = await doesDirectoryExist(
       path.join(`${stgLiteDir}`, `files`)
     )
 
-    if (doesFilesFolderExist) {
+    if (doesFilesFolderExistResult.isErr()) {
+      throw doesFilesFolderExistResult.error
+    }
+
+    if (doesFilesFolderExistResult.value) {
       await this.simpleGit
         .cwd({ path: stgLiteDir, root: false })
         .rm(["-r", "files"])
--- a/src/services/identity/ReposService.ts
+++ b/src/services/identity/ReposService.ts
@@ -1,4 +1,3 @@
-import { exec } from "child_process"
 import fs from "fs"
 import path from "path"
 
@@ -22,6 +21,7 @@ import {
 } from "@root/constants"
 import GitHubApiError from "@root/errors/GitHubApiError"
 import logger from "@root/logger/logger"
+import { doesDirectoryExist } from "@root/utils/fs-utils"
 
 const SYSTEM_GITHUB_TOKEN = config.get("github.systemToken")
 const octokit = new Octokit({ auth: SYSTEM_GITHUB_TOKEN })
@@ -120,7 +120,7 @@ export default class ReposService {
     const dir = this.getLocalStagingRepoPath(repoName)
 
     // 1. Set URLs in local _config.yml
-    this.setUrlsInLocalConfig(dir, repoName, stagingUrl, productionUrl)
+    await this.setUrlsInLocalConfig(dir, repoName, stagingUrl, productionUrl)
 
     // 2. Commit changes in local repo
     await this.simpleGit
@@ -151,14 +151,21 @@ export default class ReposService {
     await this.simpleGit.cwd({ path: dir, root: false }).checkout("staging")
   }
 
-  private setUrlsInLocalConfig(
+  private async setUrlsInLocalConfig(
     dir: string,
     repoName: string,
     stagingUrl: string,
     productionUrl: string
   ) {
     const configPath = `${dir}/_config.yml`
-    const configFile = fs.readFileSync(configPath, "utf-8")
+    let configFile: string
+    try {
+      configFile = await fs.promises.readFile(configPath, "utf-8")
+    } catch (error) {
+      throw new UnprocessableError(
+        `Error reading _config.yml for '${repoName}': ${error}`
+      )
+    }
     const lines = configFile.split("\n")
     const stagingIdx = lines.findIndex((line) => line.startsWith("staging:"))
     const prodIdx = lines.findIndex((line) => line.startsWith("prod:"))
@@ -169,7 +176,7 @@ export default class ReposService {
     }
     lines[stagingIdx] = `staging: ${stagingUrl}`
     lines[prodIdx] = `prod: ${productionUrl}`
-    fs.writeFileSync(configPath, lines.join("\n"))
+    await fs.promises.writeFile(configPath, lines.join("\n"))
   }
 
   createRepoOnGithub = (
@@ -227,16 +234,16 @@ export default class ReposService {
     const stgLiteDir = this.getLocalStagingLiteRepoPath(repoName)
 
     // Make sure the local path is empty, just in case dir was used on a previous attempt.
-    fs.rmSync(`${stgDir}`, { recursive: true, force: true })
+    await fs.promises.rm(`${stgDir}`, { recursive: true, force: true })
 
     // Clone base repo locally
-    fs.mkdirSync(stgDir)
+    await fs.promises.mkdir(stgDir)
     await this.simpleGit
       .cwd({ path: stgDir, root: false })
       .clone(SITE_CREATION_BASE_REPO_URL, stgDir, ["-b", "staging"])
 
     // Clear git
-    fs.rmSync(`${stgDir}/.git`, { recursive: true, force: true })
+    await fs.promises.rm(`${stgDir}/.git`, { recursive: true, force: true })
 
     // Prepare git repo
     await this.simpleGit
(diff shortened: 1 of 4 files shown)


kishore03109:
* **#1220** <a href="https://app.graphite.dev/github/pr/isomerpages/isomercms-backend/1220?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop`

<h2></h2>

src/services/identity/ReposService.ts
```diff
@@ -151,14 +151,21 @@ export default class ReposService {
...
     const configPath = `${dir}/_config.yml`
-    const configFile = fs.readFileSync(configPath, "utf-8")
+    let configFile: string
+    try {
+      configFile = await fs.promises.readFile(configPath, "utf-8")
+    } catch (error) {
+      throw new UnprocessableError(
+        `Error reading _config.yml for '${repoName}': ${error}`
+      )
+    }
```
seaerchin, line 168: ok with the `throw` but how do we return something useful back to our user?

src/services/review/RepoCheckerService.ts
```diff
@@ -77,33 +78,57 @@ export default class RepoCheckerService {
...
+          fs.promises.mkdir(logsFilePath, { recursive: true }),
+          (err) => {
+            logger.error(`SiteCheckerError: Unable to create logs dir: ${err}`)
+            return new SiteCheckerError(
+              `SiteCheckerError: Unable to create logs dir: ${err}`
+            )
+          }
+        )
+      }
+      return okAsync("ok")
```
seaerchin, line 105: can we return `true` here? or the name of the folder itself?

src/services/review/RepoCheckerService.ts
```diff
@@ -77,33 +78,57 @@ export default class RepoCheckerService {
...
-
-    return ResultAsync.fromPromise(
-      fs.promises.access(lockFilePath, fs.constants.F_OK),
-      (err) => new SiteCheckerError(`${err}`)
-    ).map(() => true)
+    return this.createLogsFolder(repo)
+      .andThen(() =>
+        ResultAsync.fromPromise(
+          fs.promises.access(lockFilePath, fs.constants.F_OK),
+          (err) => new SiteCheckerError(`${err}`)
```
seaerchin, line 117: we don't have to fix this now but note that we can't differentiate between `SiteCheckerError` and (an imaginary) `FolderLockError`

src/utils/fs-utils.ts
```diff
@@ -0,0 +1,13 @@
...
+export async function doesDirectoryExist(path: string): Promise<boolean> {
+  let result = false
+  try {
+    const stat = await fs.stat(path)
+    result = stat.isDirectory()
+  } catch (e) {
+    result = false
+  }
+  return result
+}
```
seaerchin, line 13: since the caller here only really cares about whether the path, we could wrap this in a `ResultAsync`. the caller only has to call `isOk` and if it's an error, we could easily capture a log here. 

since we are moving to structured logging, this prevents silent failures in our system (`fs.stat` could fail for a wide variety of reasons) and allows us to capture contextual info. wdyt?

src/services/review/RepoCheckerService.ts
```diff
@@ -77,33 +78,57 @@ export default class RepoCheckerService {
...
+        )
+      )
+      .map(() => true)
```
seaerchin, line 120: double checking - we're checking if we can access the newly created lock file right? in the event that we can't, should we also perform some clean-up? (we could file a ticket for this, no worri3es)

src/services/review/RepoCheckerService.ts
```diff
@@ -375,91 +400,107 @@ export default class RepoCheckerService {
...
-            continue
-          }
-          const normalisedPermalink = this.normalisePermalink(filePermalink)
-
-          // we are only parsing out the front matter
-          const fileContent = file.split("---")?.slice(2).join("---")
-          if (!fileContent) {
-            continue
-          }
           findErrors.push(
```
seaerchin, line 403: ok with the `push` but it's difficult to tell what's being pushed exactly.

a quick way to solve this would just be to split the actual variable and the push 

```ts
const error = ResultAsync.fromPromise(...)
fileErrors.push(error)
```

src/services/review/RepoCheckerService.ts
```diff
@@ -493,52 +534,83 @@ export default class RepoCheckerService {
...
+  reporter(
+    repo: string,
+    errors: RepoError[]
+  ): ResultAsync<RepoError[], SiteCheckerError> {
+    return this.createLogsFolder(repo).andThen(() => {
+      const filePath = path.join(SITE_CHECKER_REPO_PATH, repo, REPO_LOG)
+      const stringifiedErrors = this.convertToCSV(errors)
+      return ResultAsync.fromPromise<fs.Stats, false>(
+        fs.promises.stat(filePath),
+        () => false
```
seaerchin, line 546: if it's the error branch, we discarded the error info to replace it w/ `false`, which is less descriptive than the error itself. perhaps we should let the caller decide this?

src/services/review/RepoCheckerService.ts
```diff
@@ -566,37 +638,38 @@ export default class RepoCheckerService {
...
-      return errAsync(new SiteCheckerError(`Folder does not exist for ${repo}`))
-    }
 
-    return this.isCurrentlyErrored(repo).andThen((isError) => {
-      if (isError) {
-        return errAsync(new SiteCheckerError(`Error checking repo ${repo}`))
-      }
+    return ResultAsync.fromPromise(
+      fs.promises.stat(filePath),
+      () => new SiteCheckerError(`Folder does not exist for ${repo}`)
```
seaerchin, line 644: maybe we can also log the error here?

seaerchin (commented):
can we make sure that users see a sensible error message if we do a `throw` here?

src/services/identity/ReposService.ts
kishore03109: hmm i am a bit confused by on what you refer to as useful.
this is called in the site create side form, and we return the error as is to the us, the admins of isomer. 

looking at the rest of of the calls, they seem to return something similar, inclined to keep the local consistent behaviour of the throws for this file

src/services/review/RepoCheckerService.ts
kishore03109: hey @seaerchin, this one i might want to push back. 
this function is responsible for persisting the report to disk, the caller should not be concerned on whether there already existed a pervious run of the report already existing in the disk. This is just a I/O call to see if the file exists. It just so happens that the api call is designed such that this will throw an error if such file does not exist, which just means that we need to create the file in the `orElse` branch.

src/services/review/RepoCheckerService.ts
kishore03109: nope, `this.createLogsFolder` is just a sanity check that the top level folder exists. 
the key thing that this function does is check if the `checker.lock` exists, which is done by the `fs.promise.access` call. 

the lock file creation is done in a separate function call. the clean up functionality already exists (if some error exists, create a error file + remove the `checker.lock` file)

src/services/review/RepoCheckerService.ts
kishore03109: thanks for this comment, after reading the documentation could just call write file directly for this, they recommend this approach instead

src/services/review/RepoCheckerService.ts
kishore03109: done

src/services/review/RepoCheckerService.ts
kishore03109: done

src/services/identity/ReposService.ts
seaerchin: Ok I thought this was user facing. If it doesn't cause the flow to just stop here without a useful error message then I'm ok w just throwing.

src/services/review/RepoCheckerService.ts
seaerchin: hmm what i'm hearing is that the point of `isCurrentlyLocked` is the check on `checker.lock`! seems abit weird to have the log creation folder here then? we can probably shift it out and this would be abit less confusing. (since this PR is scoped to fixing `fsSync`, we can do it later)

seaerchin (approved):
lgtm, checked using `\.[a-zA-Z]+sync` and only `statSync` in tests + neverthrow's `async` methods could be found.

seaerchin:
@kishore03109 as a follow-up, the next time we submit a form that triggers an infra flow on our backend, could we verify that CPU usage doesn't spike anymore?

kishore03109:
> @kishore03109 as a follow-up, the next time we submit a form that triggers an infra flow on our backend, could we verify that CPU usage doesn't spike anymore?

@seaerchin updated the pr desc for this

src/services/review/RepoCheckerService.ts
kishore03109: the dir structure is 
-  pa-corp (folder)
   - logs.csv (file)
   - checker.lock (file)

what we are doing here is ensuring that the top level folder exists before checking for the checker.lock ya

PR #1228, opened 2024-03-19
chore(growthbook): handle refreshes at module level

timotheeg:
## Problem

Using `autorefresh:true` in `growthbook.loadFeatures()` is an anti pattern.

`autorefresh:true` [translates to a subscribe on the instance](https://github.com/growthbook/growthbook/blob/v2.0.0/packages/sdk-js/src/GrowthBook.ts#L115-L122), which means the flags may be updated during the lifetime of a request.

In recent version of growthbook, `autorefresh:true` [has been deprecated](https://github.com/growthbook/growthbook/blob/v2.8.0/packages/sdk-js/src/GrowthBook.ts#L155-L158) in favour of a more explicit constructor setting `subscribeToChanges:true`, with a documented behaviour warning for middlewares [in documentation](https://docs.growthbook.io/lib/js#nodejs):
```
    // Important: make sure this is set to `false`, otherwise features may change in the middle of a request
    subscribeToChanges: false,
```

In addition, when the first request which loads the featureflag is in a middleware (typically the health endpoint at deployment), then the SSE call is created in context of the request, creating abnormally huge traces lasting days with thousands of spans. [Example here](https://ogp.datadoghq.com/apm/trace/8277262834568158153?colorBy=service&env=production&graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=3383986745756776406&spanViewType=metadata&timeHint=1709470545168.802&view=spans) or [here](https://ogp.datadoghq.com/apm/trace/65f0f2f9000000007add984d58603d36?colorBy=service&env=production&graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=7021684506432005181&spanViewType=metadata&timeHint=1710289657428.0059&view=spans)

image

## Solution

1. Upgrade growthbook to 0.36 to have have the new `subscribeToChanges` property available
2. Create a growthbook instance at startup time
3. load the feature immediately and subscribe to changes
4. Make the growthbook instances created in middleware NOT use `subscribeToChanges:true`

Bonus: Simplify the promise chain just a bit

**Breaking Changes**

## Tests

* Load the cms and verify functionality gated by feature flag still exists (TODO: find a specific feature flag to test on)

### DIFF
--- a/src/middleware/featureFlag.ts
+++ b/src/middleware/featureFlag.ts
@@ -4,6 +4,16 @@ import logger from "@root/logger/logger"
 import { RequestHandlerWithGrowthbook } from "@root/types"
 import { getNewGrowthbookInstance } from "@root/utils/growthbook-utils"
 
+// Keep one GrowthBook instance at module level
+// The instance will handle internal cache refreshes via a SSE connection
+const gb = getNewGrowthbookInstance(config.get("growthbook.clientKey"))
+
+gb.loadFeatures({ autoRefresh: true }).catch((e: unknown) => {
+  logger.error(
+    `Failed to load features from GrowthBook at startup: ${JSON.stringify(e)}`
+  )
+})
+
 // eslint-disable-next-line import/prefer-default-export
 export const featureFlagMiddleware: RequestHandlerWithGrowthbook<
   never,
@@ -21,12 +31,11 @@ export const featureFlagMiddleware: RequestHandlerWithGrowthbook<
 
   // Wait for features to load (will be cached in-memory for future requests)
   req.growthbook
-    .loadFeatures({ autoRefresh: true })
-    .then(() => next())
+    .loadFeatures()
     .catch((e: unknown) => {
       logger.error(
         `Failed to load features from GrowthBook: ${JSON.stringify(e)}`
       )
-      next()
     })
+    .then(() => next())
 }
--- a/src/middleware/featureFlag.ts
+++ b/src/middleware/featureFlag.ts
@@ -1,17 +1,21 @@
 import UserWithSiteSessionData from "@root/classes/UserWithSiteSessionData"
-import config from "@root/config/config"
+import { config } from "@root/config/config"
 import logger from "@root/logger/logger"
 import { RequestHandlerWithGrowthbook } from "@root/types"
 import { getNewGrowthbookInstance } from "@root/utils/growthbook-utils"
 
 // Keep one GrowthBook instance at module level
 // The instance will handle internal cache refreshes via a SSE connection
-const gb = getNewGrowthbookInstance(config.get("growthbook.clientKey"))
+const gb = getNewGrowthbookInstance({
+  clientKey: config.get("growthbook.clientKey"),
+  subscribeToChanges: true,
+})
 
-gb.loadFeatures({ autoRefresh: true }).catch((e: unknown) => {
-  logger.error(
-    `Failed to load features from GrowthBook at startup: ${JSON.stringify(e)}`
-  )
+gb.loadFeatures().catch((e: unknown) => {
+  logger.error({
+    error: e,
+    message: "Failed to load features from GrowthBook at startup",
+  })
 })
 
 // eslint-disable-next-line import/prefer-default-export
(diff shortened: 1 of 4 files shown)


src/middleware/featureFlag.ts
```diff
@@ -4,6 +4,16 @@ import logger from "@root/logger/logger"
 import { RequestHandlerWithGrowthbook } from "@root/types"
 import { getNewGrowthbookInstance } from "@root/utils/growthbook-utils"
 
+// Keep one GrowthBook instance at module level
+// The instance will handle internal cache refreshes via a SSE connection
```
seaerchin, line 8: not a problem with this PR but what does SSE mean?

src/middleware/featureFlag.ts
```diff
@@ -4,6 +4,16 @@ import logger from "@root/logger/logger"
...
+const gb = getNewGrowthbookInstance(config.get("growthbook.clientKey"))
+
+gb.loadFeatures({ autoRefresh: true }).catch((e: unknown) => {
+  logger.error(
```
seaerchin, line 12: should we just pass the `error` directly instead of doing `JSON.stringify`?

seaerchin (commented):
i think i'm lacking some context here - i get the need for removing `autoRefresh` but why would we create another gb instance with it? (and seemingly never use it).

if they both point to the same upstream, then will the `req.growthbook` itself get any streaming changes? (i'd assume it snapshots the feature flag state at request on backend?) 

not too sure what the module level `gb` does here - since we attach a new instance per request, do we still need server sided events given that we don't actually want the featuer flag to change halfway through a req?

src/middleware/featureFlag.ts
seaerchin: went to read the docs, this refers to server sided events. i assume the gb server issues events out to our client here in real time

src/middleware/featureFlag.ts
timotheeg: Good catch and yes we should!

If `e` is an `Error` here, `json.stringify()` will show us nothing at all. Better leave the logger to extract and format the info as needed.

I had copied the line from the middleware handling below to reduce the changeset to a minimum, but indeed we can make the logging much better.

I'll mark this PR as a dependent of the logger PR, so we can merge it later.

timotheeg:
@chin, so with growthbook, there's a shared internal cache for a given upstream. Such that even if multiple GB instances are created in multiple concurrent requests, they use the same set of feature flags. 

If the cache is stale, then a given GB instance calling loadFeatures() will make a HTTP call to fetch the flags, which will take some time on that request.

The idea is that we want the cache to be fresh for all (or most) of the incoming requests. So we need to have one instance in charge of refreshing the internal shared cache. That's what the module level instance is for. Its job is not to be accessed for flag values, its job is to be long lived (life of the node process), and update the shared cache.

In the middleware we still create GB instances, but these are short lived, just for the duration of the requests. They will reference an immutable set of flags constructed from the cache. So during the lifetime of a request, the flag values should be consistent, which is what we want if a given flag is accessed multiple times at different times over the life of the request.

The latest growthbook makes the subscription model more explicit, so I'm wondering if I shouldn't just add the upgrade as part of this PR too 🤔

seaerchin (approved):
lgtm! for full context:

1. growthbook internally has a cache upstream
2. the internal gb will handle subscription to upstream cache and update for any changes
3. this will **not** be seen by inflight gb (those alr initialised + on req) as the call to update cache gets passed the instance itself
4. new initialisation from this point onwards gets the updated cache.

impact is that we get a more updated cache which should minimise exp calls out to gb

timotheeg:
Just an update I dropped the ball on this. I want to upgrade growthbook to use the latest so the sync option is more obvious. I'll do that by end of the week.

timotheeg:
Upgraded GrowthBook to 0.36 and used the new more explicit instantiation parameter `subscribeToChanges:true|false`, which makes it clearer why the module level instance exists.

Also fixed the logging to log errors correctly.

@seaerchin

seaerchin (approved):

harishv7 (approved):

PR #1239, opened 2024-03-22
Safer pagination and refresh for site cache

timotheeg:
## Problem

The way isomercms retrieves and cache the list of repos is with the help of an environment variable `ISOMERPAGES_REPO_PAGE_COUNT`. That variable contains a number, which is used to create parallel requests to github.

In practice that works, but operationally this is not great:
* It introduces a mental load to update the variables if the number of supported repo grows beyond the current page count (easy to forget and create a potential incident!)
* The concurrency fetches optimize for time in all cases, while for the internal cache refreshes, optimizing for load distribution would be safer

Github APIs have a [very robust pagination system](https://docs.github.com/en/rest/using-the-rest-api/using-pagination-in-the-rest-api?apiVersion=2022-11-28#using-link-headers), so we can derive the number of pages from that to make the system dynamic, and remove the (small) operational risk of forgetting to update the env var.

Additionally, the cache only has one exposed API, which is to retrieve the lastUpdated time  by repoName. The list of repo is not stored in cache by name, but it is a list and so the lookup does a list search every time. 

## Solution

1. split the `getAllRepoData()` calls into 2 distinct use cases:
    a. exported API optimized for time: we fetch the first page, assess how many pages there in total, and fetch all remaining pages in parallel
    b. optimized for load distribution: we fetch the pages sequentially by following github pagination system's next link
2. populate the cache at bootstrap time with the time-optimized call
3. cache refresh uses the load-optimized call
4. Use a dictionary to store the repos indexes by name for `O(1)` lookup 

Notes:
1. The time-optimized fetch is still slower than the previous version, since the first page call is **always** made sequentially
2. We _could_ (to discuss) cache the number of pages, such that:
    a. only the first bootstrap call has the 2 step fetch
    b. every refresh call updates the number of pages
    c. when the number of pages is cached, the fast `getAllRepoData()` does all its calls in parallel again.
2. We add a custom span to instrument the time the cache refresh process takes.
   
**Breaking Changes**

Reference:
* [slack thread where this was discussed](https://opengovproducts.slack.com/archives/CK68JNFHR/p1710379981798289)

Results:
[Sample trace](https://ogp.datadoghq.com/apm/trace/66287efc000000001fb44bcddadb0cc5?colorBy=service&env=staging&graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=7336707215144269589&spanViewType=metadata&timeHint=1713929980189.008&view=spans) for the new refresh process that shows fetches the pages sequentially works:

image

## Post Deploy Work (after stability is observed)

- [ ] delete the SSM entries 'PROD_ISOMERPAGES_REPO_PAGE_COUNT' and `STAGING_ISOMERPAGES_REPO_PAGE_COUNT`. This should **ONLY** be done when we are sure we will not need to roll back possibly wait one full release.


src/services/identity/__tests__/SitesService.spec.ts
```diff
@@ -130,7 +130,10 @@ const mockDeploymentWithStagingLiteUrl = {
 
 describe("SitesService", () => {
   // Prevent inter-test pollution of mocks
-  afterEach(() => jest.clearAllMocks())
+  afterEach(() => {
+    jest.clearAllMocks()
+    mockAxios.get.mockReset()
```
timotheeg, line 135: Without this, the previous tests were not actually behaving as expected!! 😱

timotheeg:
Note: Typically another way to deal with pages and fetching data is by using the github sdk [octokit.js](https://www.npmjs.com/package/octokit), which has pagination handling built in. I did not introduce it in this PR because the goal was just to remove the env var `ISOMERPAGES_REPO_PAGE_COUNT`, and I kept the parallel fetch for pages 2-N, which octokit wouldn't do.

That should probably be revisited at some point.

src/services/identity/SitesCacheService.ts
```diff
@@ -14,81 +18,232 @@ import type { GitHubRepositoryData, RepositoryData } from "@root/types/repoInfo"
...
+  sort: "full_name",
+  page: 1,
+}
+
+type GitHubResponseHeaders = HeadersDefaults & {
+  link?: string
+}
+
+type Link = {
+  relid: "first" | "last" | "prev" | "next"
```
seaerchin, line 32: maybe we might wanna extract this out as a type called `LinkRelation`?

src/services/identity/SitesCacheService.ts
```diff
@@ -14,81 +18,232 @@ import type { GitHubRepositoryData, RepositoryData } from "@root/types/repoInfo"
...
+  url: string
+  pagenum: number
+}
+
+type LinkSet = {
+  first?: Link
+  last?: Link
+  prev?: Link
+  next?: Link
+}
```
seaerchin, line 42: we could simplify this with a type:

```ts
type LinkSest = { 
  [linkRelation: LinkRelation]?: Link
}
```

src/services/identity/SitesCacheService.ts
```diff
@@ -14,81 +18,232 @@ import type { GitHubRepositoryData, RepositoryData } from "@root/types/repoInfo"
...
+type FetchRepoPageResult = {
+  repos: RepositoryData[]
+  links?: LinkSet
+}
+
+type CacheStore = {
+  [key: string]: RepositoryData
+}
+
+function parseGitHubLinkHeader(linkheader: string) {
```
seaerchin, line 60: nit: add return type

src/services/identity/SitesCacheService.ts
```diff
@@ -14,81 +18,232 @@ import type { GitHubRepositoryData, RepositoryData } from "@root/types/repoInfo"
...
+  // example value: link: <https://api.github.com/organizations/40887764/repos?page=2>; rel="next", <https://api.github.com/organizations/40887764/repos?page=34>; rel="last"
+  const links: LinkSet = {}
+
+  const linkRe = /<(?<url>[^>]+)>; rel="(?<relid>[a-z]+)"(, )?/g
+  let regRes: LinkMatch | null = null
+
+  // eslint-disable-next-line no-cond-assign
+  while ((regRes = linkRe.exec(linkheader) as LinkMatch) !== null) {
```
seaerchin, line 68: style: i think this is a little bit confusing, what do you think about just calling `linkRe.exec(linkheader)` on the `while` then assigning inside the loop? 

```suggestion
  while (linkRe.exec(linkheader) !== null) {
    const regRes = linkRe.exec(linkheader) as LinkMatch
```

src/services/identity/SitesCacheService.ts
```diff
@@ -14,81 +18,232 @@ import type { GitHubRepositoryData, RepositoryData } from "@root/types/repoInfo"
...
+  private repoDataCache: CacheStore
 
   private refreshInterval: number
 
   constructor(refreshInterval: number) {
-    this.repoDataCache = []
-    this.refreshInterval = refreshInterval
-    this.renewCache()
+    this.repoDataCache = {} as CacheStore
+    this.refreshInterval = 90000
```
seaerchin, line 218: should this be `this.refreshInterval = refreshInterval`? we can give a default to get this behaviour

timotheeg: 😱 Oh yes! That was not supposed to be committed! 😱

src/services/identity/SitesCacheService.ts
timotheeg: No that would be wrong. Because `.exec()` traverses the input, consecutive calls to `.exec()` return different matches (i.e. not idempotent). Duplicating the calls would miss some of input matches.

It's a typical pattern of using regexes that is not supported by this annoying eslint rule 😑

src/services/identity/SitesCacheService.ts
timotheeg: Fixed in https://github.com/isomerpages/isomercms-backend/pull/1239/commits/c5aef04f88639ff759f8a3715611483547e2e70c
Good catch! 🙏

src/services/identity/SitesCacheService.ts
timotheeg: Hmm. the code structure in the function body makes it correctly extracted implicitly, but I'll add it anyway 👍

timotheeg: Added: https://github.com/isomerpages/isomercms-backend/pull/1239/commits/033783808a41601f3ecf6d438deb3aff82cb0974

src/services/identity/SitesCacheService.ts
timotheeg: Done in https://github.com/isomerpages/isomercms-backend/pull/1239/commits/e3ad470c4b86eb8bcf1e970819e958d21d0ce4e8

src/services/identity/SitesCacheService.ts
timotheeg: Hmm, That said I suppose using the "newer" API [`String.matchAll()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/matchAll) might work better here 🤔

src/services/identity/SitesCacheService.ts
timotheeg: Does this declaration work? eslint/typescript are not letting me do it 🤔

seaerchin: i got the syntax wrong oops

```ts
type LinkRelation = "first" | "last" | "prev" | "next"

type LinkSet = {
  [k in LinkRelation]: Link
}
```

src/services/identity/SitesCacheService.ts
timotheeg: I pushed a commit to use `String.MatchAll()` instead. Let me know if you find it more readable. 

Sadly I still needed to add a eslint disable rule because a tight loop over an iterator is considered dangerous, but in this case, we know the input set is limited (max 4 links), so it's fine (I added a comment for that)

src/services/identity/SitesCacheService.ts
timotheeg: Ah right, thanks! 😄 

Pushed commit to do that: https://github.com/isomerpages/isomercms-backend/pull/1239/commits/027bc2215b19ebeba5647f497b3aa462fe4547c7

timotheeg:
All comments addressed @seaerchin , Let me know if good to go. Thanks! 🙏

seaerchin (approved):

src/services/identity/SitesCacheService.ts
```diff
@@ -14,81 +16,225 @@ import type { GitHubRepositoryData, RepositoryData } from "@root/types/repoInfo"
...
+
+type CacheStore = {
+  [key: string]: RepositoryData
+}
+
+function parseGitHubLinkHeader(linkheader: string): LinkSet {
+  // example value: link: <https://api.github.com/organizations/40887764/repos?page=2>; rel="next", <https://api.github.com/organizations/40887764/repos?page=34>; rel="last"
+  const links: LinkSet = {}
+
+  const linkRe = /<(?<url>[^>]+)>; rel="(?<relid>[a-z]+)"(, )?/g
```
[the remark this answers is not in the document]
timotheeg: I'd rather leave the regexp loose tbh, in case github introduces more link relations in the future, they'd already be captured properly as part of the parsing routine. To be consistent then, we could upgrade the type to allow unknown keys too 🤔

timotheeg:
Result in prod:
* As expected: /v2/sites is slower now since the first page is always sequential ([sample trace here](https://ogp.datadoghq.com/apm/trace/4784715723176877813?colorBy=service&env=production&graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=119630317035733146&spanViewType=metadata&timeHint=1714096496691&view=spans))
image
* The refresh is a cheaper process now, since all the ages are sequential (sample internal trace)
image

PR #1243, opened 2024-03-27
fix: reduce log size to just last commit

timotheeg:
## Problem

To get the last commit, we issue a full `git log` command which can returns and parse thousands of lines and commits. 

Some sites have a HUGE amount of commits (e.g. `nlb-biblioasia` which has over 50k commits!)

```bash
git log | grep 'commit ' | wc -l
```

## Solution

Use the arg `--max-count=1` to restrict resultset to just one commit when fetching the last commit 😑 🤷 

## Benchmark

Assuming you have `nlb-biblioasia` locally (peer of your isomercms checkout), run this script to show the difference between the 2

```javascript
const SimpleGit = require("simple-git")
const git = SimpleGit()
const ITERATIONS = 20;

async function run() {
    // prime FS cache if needed
    await git.cwd('../nlb-biblioasia').log(['staging']);

    let startTs, latest;

    startTs = Date.now();

    for (let idx=ITERATIONS; idx--; ) {
        latest = (await git.cwd('../nlb-biblioasia').log(['staging'])).latest;
    }

    console.log('unfiltered:', Date.now() - startTs);
    console.log(latest)
    console.log('====================')

    startTs = Date.now();

    for (let idx=ITERATIONS; idx--; ) {
        latest = (await git.cwd('../nlb-biblioasia').log(['--max-count=1', 'staging'])).latest;
    }

    console.log('last only:', Date.now() - startTs);
    console.log(latest)
    console.log('====================')
}

run();
```

Result
```
unfiltered: 12699
{
  hash: 'ef9c558be46df2f0c03174eeb1209f53d14ff34c',
  date: '2024-03-25T01:17:25+00:00',
  message: '{"message":"Update file: Memory, History, Art: Yip Yew Chong’s “I Paint my Singapore”.md","userId":"341","fileName":"Memory, History, Art: Yip Yew Chong’s “I Paint my Singapore”.md"}',
  refs: 'HEAD -> staging, origin/staging, origin/HEAD',
  body: '',
  author_name: 'Isomer Admin',
  author_email: 'admin@isomer.gov.sg'
}
====================
last only: 164
{
  hash: 'ef9c558be46df2f0c03174eeb1209f53d14ff34c',
  date: '2024-03-25T01:17:25+00:00',
  message: '{"message":"Update file: Memory, History, Art: Yip Yew Chong’s “I Paint my Singapore”.md","userId":"341","fileName":"Memory, History, Art: Yip Yew Chong’s “I Paint my Singapore”.md"}',
  refs: 'HEAD -> staging, origin/staging, origin/HEAD',
  body: '',
  author_name: 'Isomer Admin',
  author_email: 'admin@isomer.gov.sg'
}
====================
```

On that site, fetching just the last commit is nearly 100x faster 😑 🎉

### DIFF
--- a/src/services/db/GitFileSystemService.ts
+++ b/src/services/db/GitFileSystemService.ts
@@ -421,6 +421,12 @@ export default class GitFileSystemService {
   ): ResultAsync<LogResult<DefaultLogFields>, GitFileSystemError> {
     const efsVolPath = this.getEfsVolPathFromBranch(branchName)
 
+    if (!Number.isInteger(maxCount) || maxCount < 1) {
+      return errAsync(
+        new GitFileSystemError(`Invalid maxCount value supplied: ${maxCount}`)
+      )
+    }
+
     return ResultAsync.fromPromise(
       this.git
         .cwd({ path: `${efsVolPath}/${repoName}`, root: false })
--- a/src/services/db/GitFileSystemService.ts
+++ b/src/services/db/GitFileSystemService.ts
@@ -417,18 +417,14 @@ export default class GitFileSystemService {
   getGitLog(
     repoName: string,
     branchName: string,
-    maxCount?: number
+    maxCount = 1
   ): ResultAsync<LogResult<DefaultLogFields>, GitFileSystemError> {
     const efsVolPath = this.getEfsVolPathFromBranch(branchName)
-
-    const logArgs = [branchName]
-
-    if (maxCount) logArgs.unshift(`--max-count=${maxCount}`)
 
     return ResultAsync.fromPromise(
       this.git
         .cwd({ path: `${efsVolPath}/${repoName}`, root: false })
-        .log(logArgs),
+        .log([`--max-count=${maxCount}`, branchName]),
 
       (error) => {
         logger.error(
@@ -1736,8 +1732,7 @@ export default class GitFileSystemService {
       .andThen((isBranchLocallyPresent) =>
         this.getGitLog(
           repoName,
-          isBranchLocallyPresent ? branchName : `origin/${branchName}`,
-          1
+          isBranchLocallyPresent ? branchName : `origin/${branchName}`
         )
       )
       .andThen((logSummary) => {
--- a/src/services/db/GitFileSystemService.ts
+++ b/src/services/db/GitFileSystemService.ts
@@ -416,13 +416,20 @@ export default class GitFileSystemService {
   // Get the Git log of a particular branch
   getGitLog(
     repoName: string,
-    branchName: string
+    branchName: string,
+    maxCount?: number
   ): ResultAsync<LogResult<DefaultLogFields>, GitFileSystemError> {
     const efsVolPath = this.getEfsVolPathFromBranch(branchName)
+
+    const logArgs = [branchName]
+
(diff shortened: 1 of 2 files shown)


src/services/db/GitFileSystemService.ts
```diff
@@ -416,13 +416,20 @@ export default class GitFileSystemService {
   // Get the Git log of a particular branch
   getGitLog(
     repoName: string,
-    branchName: string
+    branchName: string,
+    maxCount?: number
```
kishore03109, line 420: not a fan of exposing this flexibility when we no function needs this

timotheeg: Aye, I hesitated on it 🤔 

The method is called `getGitLog()`, so I didn't want to hardcode the `1` in the command since getGitLog could legitimately be used to retrieve more commits if we wanted to (even if it's not the case now).

If I hardcoded the `max-count:1`, the method should be renamed to `getLatestCommit` (or similar), but we already have a method named `getLatestCommitOfBranch()`. I suppose alternative is to remove the `getGitLog()` entirely and put the git command into `getLatestCommitOfBranch()` directly as an internal step 🤔 

Any preference/suggestion?

seaerchin: can we just give a default argument of `1`? i think we should have a sane default for this and only those who need extra power call it explicitly

seaerchin (commented):
tbh no big blockers, i'm also fine w/ this as is

src/services/db/GitFileSystemService.ts
timotheeg: > can we just give a default argument of 1? i think we should have a sane default for this and only those who need extra power call it explicitly

Done: made `maxCount` mandatory with default value of 1. Having a guaranteed value simplifies the cli args composition too: https://github.com/isomerpages/isomercms-backend/pull/1243/commits/7bae2e7033194ae0f354220b228e3b352c7fd2c6

Also added some sanity validation and tests as well: https://github.com/isomerpages/isomercms-backend/pull/1243/commits/16dc076ffd935afbbdb4a6b82765c87322c1a5b

In term of exposing functionality, I think that concern can be dismissed for now: there are no other callers to `getGitLog()` and the worst that can happen is the behaviour that was there before (returning the entire commit list). I'm keen in pushing the perf fix asap, and further refactors can be done later if you want to remove a method that you judge to be unnecessary.

src/services/db/GitFileSystemService.ts
```diff
@@ -416,13 +416,20 @@ export default class GitFileSystemService {
...
   ): ResultAsync<LogResult<DefaultLogFields>, GitFileSystemError> {
     const efsVolPath = this.getEfsVolPathFromBranch(branchName)
+
+    const logArgs = [branchName]
+
+    if (maxCount) logArgs.unshift(`--max-count=${maxCount}`)
```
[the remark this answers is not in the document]
timotheeg: done with making `maxCount` compulsory in https://github.com/isomerpages/isomercms-backend/pull/1243/commits/7bae2e7033194ae0f354220b228e3b352c7fd2c6

kishore03109 (approved):

PR #1244, opened 2024-03-27
fix(dockerfile): add dig to image

seaerchin:
## Problem
https://www.notion.so/opengov/2024-03-27-careers-gov-sg-misconfiguration-15469dbae95942588574db9d8e819177?pvs=4#c009bcd4e261452289d0d74463b37cb9

https://linear.app/ogp/issue/ISOM-881/add-dig-to-image-for-backend

## Solution
add `bind-tools` (`dig` + `nslookup` is part of this)

## Tests 
- [ ] `ecs exec` into staging
- [ ] run `dig www.google.com` 

## Images 
Screenshot 2024-03-27 at 10 36 32 PM

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,8 @@ WORKDIR /opt/isomercms-backend
 RUN apk update && \
   apk add --no-cache bash && \
   apk add git && \
-  apk add openssh-client
+  apk add openssh-client && \ 
+  apk add bind-tools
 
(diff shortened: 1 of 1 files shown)


seaerchin:
flakey timestamp test failed - not the first time, maybe we should relax the test

kishore03109:
lol i think we shouldnt use that library in the first place. can merge this for release tmr, but think we can just directly use node's dig

PR #1245, opened 2024-03-27
chore(ci): enhance mergify

seaerchin:
## Problem
see [here](https://github.com/isomerpages/isomercms-frontend/pull/1865)

## Solution
see [here](https://github.com/isomerpages/isomercms-frontend/pull/1865)

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -4,7 +4,7 @@ pull_request_rules:
       - author~=^dependabot\[bot\]$
       - check-success~=lint
       - check-success~=test
-      - title~=bump [^\s]+ from ([\d]+)\..+ to \1\.
+      - dependabot-update-type = version-update:semver-minor
     actions:
       review:
         type: APPROVE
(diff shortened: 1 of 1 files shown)


.github/mergify.yml
```diff
@@ -4,7 +4,7 @@ pull_request_rules:
       - author~=^dependabot\[bot\]$
       - check-success~=lint
       - check-success~=test
-      - title~=bump [^\s]+ from ([\d]+)\..+ to \1\.
+      - dependabot-update-type = version-update:semver-minor
```
kishore03109, line 7: this one does from minor all the way to major right? if not please dismiss my approval!

kishore03109 (approved):

.github/mergify.yml
seaerchin: no, only `semver-minor`

seaerchin: > Misunderstood this pr, what about patches ya?

i think lets just do 
```ts
- -dependabot-update-type = version-update:semver-major
```

which auto-merges everything less `semver-major`
Screenshot 2024-04-02 at 7 32 23 PM

timotheeg (approved):

PR #1246, opened 2024-03-27
fix(dig): dig not working

kishore03109:
## Problem

We do have some checks in the backend to check prior to any site launch that there are AAAA records/CAA records present. In this case, the dig commands to check the AAAA records [failed](https://ogp.datadoghq.com/logs?query=service%3Aisomer%20env%3Aproduction%20%22An%20error%20occurred%20while%20performing%20dig%20for%20domain%3A%22%20&cols=host%2Cservice&fromUser=true&index=%2A&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&from_ts=1711457521321&to_ts=1711543921321&live=true).

The reason for above is that we were using a library called `node-dig-dns`. This called the dig command [directly](https://github.com/StephanGeorg/node-dig-dns/blob/master/src/index.js#L78) at a system level. However, our docker container does not have the dig command out of the box. This resulted in the existence of AAAA records not being caught.

We are also codifying a check for CAA records and ensuring that if there exist at least one  caa record and it uses our redirection service, it should have letsencrypt as one of the caa record. 

To prevent accidental commits to live indirection repo during dev, also adding a check to only commit to the indirection repository iff it is in prod environment. 
 
## Solution

just use node's dns resolver directly. this way we dont have to depend on an external library's implementation of node and dont have to install unnecessary deps in the docker.

remove dep introduced in #1244 

**Breaking Changes**

  - Details ...

**Features**:
Screenshot 2024-03-27 at 11 21 44 PM

### Manual test (not to be copied over to deployment notes)

- [ ] add these lines of code at the end of server.js 
```
const formResponses = [
  {
    submissionId: "",
    requesterEmail: "kishore@open.gov.sg",
    repoName: "kishore-test-dev-emil",
    primaryDomain: "google.com",
    redirectionDomain: "www.google.com",
    agencyEmail: "kishore@open.gov.sg",
  },
]

formsgSiteLaunchRouter.handleSiteLaunchResults(formResponses, "test")
```  

- [ ] Assert that the email comes out to


kishore03109:
* **#1246** <a href="https://app.graphite.dev/github/pr/isomerpages/isomercms-backend/1246?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop`

<h2></h2>

src/routes/formsg/formsgSiteLaunch.ts
```diff
@@ -225,23 +227,91 @@ export class FormsgSiteLaunchRouter {
...
+        domain: launchResult.primaryDomainSource,
+        type: "AAAA",
+        value: record,
+      }))
+    } catch (e: any) {
+      if (e.code && e.code === "ENODATA") {
+        logger.info(
+          `Unable to get dig response for domain: ${launchResult.primaryDomainSource}. Skipping check for AAAA records`
+        )
+      }
```
seaerchin, line 252: could we change the `any`? separately, why are we not logging the `error` here? should it be `info` or should we set to `warn`?

src/routes/formsg/formsgSiteLaunch.ts
```diff
@@ -225,23 +227,91 @@ export class FormsgSiteLaunchRouter {
...
+    } catch (e: any) {
+      if (e.code && e.code === "ENODATA") {
+        logger.info(
+          `Unable to get dig response for domain: ${launchResult.primaryDomainSource}. Skipping check for CAA records`
         )
-        return null
-      })
+      }
+      throw e
+    }
```
seaerchin, line 313: ssee above comment

src/types/siteLaunch.ts
```diff
@@ -40,6 +40,19 @@ export interface SiteLaunchMessage {
...
+  SiteLaunchMessage,
+  | "status"
+  | "statusMetadata"
+  | "redirectionDomain"
+  | "appId"
+  | "requestorEmail"
+  | "agencyEmail"
+> & {
+  redirectionDomainSource?: string
+  redirectionDomainTarget?: string
```
seaerchin, line 53: is this just to facilitate the initial construction of `successResult` and setting the props later? it seems like we'll always have the properties (we throw on error and we have a default) so maybe we should set this as required to better reflect reality

src/routes/formsg/formsgSiteLaunch.ts
```diff
@@ -225,23 +227,91 @@ export class FormsgSiteLaunchRouter {
...
+    addAWSACMCertCAA: boolean
+    addLetsEncryptCAA: boolean
+  }> => {
+    try {
+      const caaDigResponses: CaaRecord[] = await dnsPromises.resolveCaa(
+        launchResult.primaryDomainSource
+      )
+
+      if (caaDigResponses.length > 0) {
+        const caaRecords = caaDigResponses
```
seaerchin, line 269: should we flip it so the default case is up top?

seaerchin (commented):
since we're removing the external dependency here, should we also update the `dockerfile` to remove the installation of `bind-tools`?

src/types/siteLaunch.ts
kishore03109: no ah no all site launches use redirection server leh

src/routes/formsg/formsgSiteLaunch.ts
kishore03109: [5e7d5e9](https://github.com/isomerpages/isomercms-backend/pull/1246/commits/5e7d5e9ca075851f0042b7bfc43852c8812bcf53)

src/routes/formsg/formsgSiteLaunch.ts
kishore03109: [5e7d5e9](https://github.com/isomerpages/isomercms-backend/pull/1246/commits/5e7d5e9ca075851f0042b7bfc43852c8812bcf53)

src/routes/formsg/formsgSiteLaunch.ts
kishore03109: added type guard [5e7d5e9](https://github.com/isomerpages/isomercms-backend/pull/1246/commits/5e7d5e9ca075851f0042b7bfc43852c8812bcf53)

src/types/nodeError.ts
```diff
@@ -0,0 +1,9 @@
+export function isErrnoException(obj: any): obj is NodeJS.ErrnoException {
+  return (
+    obj &&
+    (typeof obj.errno === "number" || obj.errno === undefined) &&
+    (typeof obj.code === "string" || obj.code === undefined) &&
+    (typeof obj.path === "string" || obj.path === undefined) &&
+    (typeof obj.syscall === "string" || obj.syscall === undefined)
+  )
+}
```
seaerchin, line 9:
```suggestion
export function isErrnoException(obj: unknown): obj is NodeJS.ErrnoException {
  const err = obj as NodeJS.ErrnoException
  return (
    err &&
    (typeof err.errno === "number" || err.errno === undefined) &&
    (typeof err.code === "string" || err.code === undefined) &&
    (typeof err.path === "string" || err.path === undefined) &&
    (typeof err.syscall === "string" || err.syscall === undefined)
  )
}
```

seaerchin (approved):
lgtm, just accept suggestion to remove `any`

PR #1255, opened 2024-03-28
ISOM-824 feat: don't fetch redundant author info from db

timotheeg:
## Problem

`computeShaMappings()` does one query `this.users.findByPk(userId)` for each author of all commits on the review. All the DB queries are issued in parallel. For large changeset with many commits, we expect the set of authors to be less than the number of commits, and so the process sometimes issues many redundant queries in parallel.

image
image

Goal: reduce the number of queries to the minimum required.

## Solution

* From the list of commits, identify the sets of unique authors
* Fetch author data in parallel as before on that minimized set
* Generate the mappings from the author data

**Breaking Changes**

TODO:
- [x] Log the gains the PR does (i.e. how many query we are making versus how many commits there were.

## Notes:
* ~~The same inefficient process of blindly retrieving user details (possibly with redundant calls) exist for `computeCommentData()`, the same fix should be done there as well.~~ DONE!
* We could investigate having a LRU cache for userId => userDetails that can be used to drop some queries entirely

### DIFF
--- a/src/services/review/ReviewRequestService.ts
+++ b/src/services/review/ReviewRequestService.ts
@@ -25,6 +25,7 @@ import PageParseError from "@root/errors/PageParseError"
 import PlaceholderParseError from "@root/errors/PlaceholderParseError"
 import RequestNotFoundError from "@root/errors/RequestNotFoundError"
 import logger from "@root/logger/logger"
+import { statsService } from "@root/services/infra/StatsService"
 import {
   BaseEditedItemDto,
   CommentItem,
@@ -457,13 +458,11 @@ export default class ReviewRequestService {
       )
     )
 
-    logger.info({
-      message: "computeShaMappings(): author query optimisation results",
-      meta: {
-        numCommits: commits.length,
-        numUniqueValidAuthors: validAuthorIds.size,
-      },
-    })
+    statsService.statsD.increment("review_requests.commits", commits.length)
+    statsService.statsD.increment(
+      "review_requests.commits.users",
+      validAuthorIds.size
+    )
 
     commits.forEach(({ commit, sha }, idx) => {
       const authorId = commitAuthorIds[idx]
@@ -490,13 +489,11 @@ export default class ReviewRequestService {
       await Promise.all(userIds.map((userId) => this.users.findByPk(userId)))
     )
 
-    logger.info({
-      message: "computeCommentData(): author query optimisation results",
-      meta: {
-        numComments: comments.length,
-        numUniqueValidUsers: userIds.length,
-      },
-    })
+    statsService.statsD.increment("review_requests.comments", comments.length)
+    statsService.statsD.increment(
+      "review_requests.comments.users",
+      userIds.length
+    )
 
     const mappings = comments.map(({ userId, message, createdAt }) => {
       const createdTime = new Date(createdAt)
--- a/src/services/review/ReviewRequestService.ts
+++ b/src/services/review/ReviewRequestService.ts
@@ -483,18 +483,32 @@ export default class ReviewRequestService {
     comments: GithubCommentData[],
     viewedTime: Date | null
   ) => {
-    const mappings = await Promise.all(
-      comments.map(async ({ userId, message, createdAt }) => {
-        const createdTime = new Date(createdAt)
-        const author = await this.users.findByPk(userId)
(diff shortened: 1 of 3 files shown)


timotheeg:
Oh boy... A bunch of tests had incorrect arrangements that were revealed by the fetch optimizations 😰😑 

The 2 constants struct `MOCK_GITHUB_COMMENT_DATA_ONE` and `MOCK_GITHUB_COMMENT_DATA_TWO` are actually referencing the same user id (`1`), but the tests were mocking different responses of different users with different email addresses (`MOCK_GITHUB_EMAIL_ADDRESS_ONE` and `MOCK_GITHUB_EMAIL_ADDRESS_TWO`).

When the old (inefficient) code was issuing one request per user, whether or not there were duplicates user ids, the mock responses could return different responses for each calls, regardless of whether the inbound args were actually different.

The tests passed in the sense they were using the response data for the final output, but they were internally inconsistent (i.e. user id 1 gets different user responses on successive mocked db calls)

I suggest we install [jest-when](https://www.npmjs.com/package/jest-when) to have the ability to mock responses based on the call args, so we could do:

```javascript
when(MockUsersRepository.findByPk)
    .calledWith('1')
    .mockReturnValue(MOCK_GITHUB_COMMIT_AUTHOR_ONE)
    
when(MockUsersRepository.findByPk)
    .calledWith('2')
    .mockReturnValue(MOCK_GITHUB_COMMIT_AUTHOR_TWO)
```

src/services/review/ReviewRequestService.ts
```diff
@@ -439,36 +440,74 @@ export default class ReviewRequestService {
...
-        const lastChangedTime = new Date(commit.author.date).getTime()
-        mappings[sha] = {
-          author: author?.email || commit.author.name,
-          unixTime: lastChangedTime,
-        }
-      })
+
+    // commitAuthorIds are linked to the commits by array index
+    const commitAuthorIds = commits.map(
+      ({ commit }) => fromGithubCommitMessage(commit.message).userId
```
kishore03109, line 446: could we filter out the possibly undefined `commitAuthorIds` so that we don't have to do a type cast below for `validAuthorIds`?

src/services/review/ReviewRequestService.ts
```diff
@@ -439,36 +440,74 @@ export default class ReviewRequestService {
...
     )
+
+    // Query DB for all valid author details
+    const validAuthorIds = new Set(
+      commitAuthorIds.filter(_.identity) as string[]
```
kishore03109, line 451: comment above, seems weird to type cast when we are not sure if this is a type `string[]`

kishore03109 (commented):
only blocking is querying the db for possibly undefined values

src/services/review/ReviewRequestService.ts
timotheeg: `commit.userId` is types as `string | undefined`, so the `.filter(_.identity)` guarantees  string. I wish TS would see it too 😢

src/services/review/ReviewRequestService.ts
timotheeg: Not at this step, unless we change the code more 🤔 

This is explained in the comment above:
```
// commitAuthorIds are linked to the commits by array index
```
This is done so when we iterate over the commits again further below, we have the userId of that commit without having to parse the commit message again. See line 468 below:
```
    const authorId = commitAuthorIds[idx]
```

An alternative might be to store some temporary structure for mappings by commit sha, but that also adds a bit of mess in the function.

timotheeg:
Thanks for the review, Kish!

> only blocking is querying the db for possibly undefined values

If you spot something that's really wrong in a PR, submit your review as `Request for Changes` to make sure the PR is blocked, and no one merges accidentally 😰 

That being said, in this PR, I think all user queries are now protected against querying with `undefined` (with the userIds being checked with `.filter(_.identity)`), while the old code was not 😰 .

src/services/review/ReviewRequestService.ts
timotheeg: The filter also drops empty string too btw, if that ever was an issue 🤷

kishore03109 (approved):
approving since there is a filter and the type is safe to `String[]`

PR #1256, opened 2024-03-31
fix(lock): fix early release of lock

kishore03109:
## Problem

Currently, we are releasing the lock before the promise gets resolved, which then leads to the lock becoming effectively useless as it does not guarantee that each mutation on the repo is isolated from one another. This would prob lead to more users complaining that there are concurrent editors, but there might be further optimisations that could be done to reduce the frequency of this. 

## Solution

dont release lock early

**Breaking Changes**

  - Details ...

## Before & After Screenshots

**BEFORE**:

Screenshot 2024-04-01 at 7 37 00 AM

**AFTER**:

Screenshot 2024-04-01 at 7 39 11 AM

## Tests

[ ] Go into any site and perform CRUD operations on a page level for a repo in efs.

### DIFF
--- a/src/middleware/routeHandler.ts
+++ b/src/middleware/routeHandler.ts
@@ -245,78 +245,80 @@ export const attachRollbackRouteHandlerWrapper: RouteWrapper<
     }
   }
 
-  Promise.resolve(routeHandler(req, res, next)).catch(async (err: Error) => {
-    try {
-      if (shouldUseGitFileSystem) {
-        await backOff(
-          () =>
-            convertNeverThrowToPromise(
-              gitFileSystemService.rollback(
-                siteName,
-                originalStagingCommitSha,
-                STAGING_BRANCH
-              )
-            ),
-          backoffOptions
-        )
-
-        if (shouldCheckStagingLite) {
-          // for quickie sites
+  Promise.resolve(routeHandler(req, res, next))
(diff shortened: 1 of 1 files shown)


kishore03109:
* **#1256** <a href="https://app.graphite.dev/github/pr/isomerpages/isomercms-backend/1256?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop`

<h2></h2>

alexanderleegs (approved):

timotheeg (approved):
LGTM for a very targeted fix 👍 

As discussed, a few things are "dirty" and could be refactored.
* We're mixing functional promise chains with `.then()`/`.catch()` and `async`/`await` style which make things confusing
* The express control flow is not being hijacked correctly: `next` is supplied to `routeHandler()`, which could call it, and so we calling `next()` again later (say, if we cannot release the lock for example) might lead to inconsistent behaviour. As a wrapper to a route handler, the code should take over management of the middleware sequence.

The above issues have potential for some run-time errors, but it looks like in practice, the system has been behaving ok 🤞😰

timotheeg:
Argh, I just spotted we need to do the same for `attachWriteRouteHandlerWrapper()`, it has the same early release issue. 
I'll open a PR for it now, and we should release urgently too.

PR #1259, opened 2024-04-02
fix: only release the lock after the handler is done

timotheeg:
## Problem

For write routes, the site lock is released too early, which makes the lock useless

This is a sibling PR to https://github.com/isomerpages/isomercms-backend/pull/1256, which introduced the same fix for the rollback routes.

## Solution

Only release the lock AFTER the action for the handler is completed. 

Notes:
1. For this to work (same a rollback routes), there is an assumption that ALL write route handlers are async and resolve when their job is done, but that is NOT enforced from typings. The handler is of type `RequestHandler` which is not specified to return a `Promise`, it returns `void`.
2. The handling of the express flow still looks wrong and should be discussed. When the handler is successful, and the lock fails to release, does it really make sense to call `next(err)`? The query was handled successfully, and failure to release the lock could be retried or just logged (and tracked!)

### DIFF
--- a/src/middleware/routeHandler.ts
+++ b/src/middleware/routeHandler.ts
@@ -108,16 +108,19 @@ export const attachWriteRouteHandlerWrapper: RouteWrapper<{
     return
   }
 
-  Promise.resolve(routeHandler(req, res, next)).catch(async (err: Error) => {
-    await unlock(siteName)
-    next(err)
-  })
-
-  try {
-    await unlock(siteName)
-  } catch (err) {
-    next(err)
-  }
+  Promise.resolve(routeHandler(req, res, next)).then(
+    async () => {
+      try {
+        await unlock(siteName)
+      } catch (err) {
+        next(err)
+      }
+    },
+    async (err) => {
+      await unlock(siteName)
+      next(err)
+    }
+  )
 }
 
 export const attachRollbackRouteHandlerWrapper: RouteWrapper<


src/middleware/routeHandler.ts
```diff
@@ -108,16 +108,19 @@ export const attachWriteRouteHandlerWrapper: RouteWrapper<{
...
-    await unlock(siteName)
-  } catch (err) {
-    next(err)
-  }
+  Promise.resolve(routeHandler(req, res, next)).then(
+    async () => {
+      try {
+        await unlock(siteName)
+      } catch (err) {
+        next(err)
```
timotheeg, line 116: Per PR description: does calling `next(err)` make sense here? The route handling was successful and it may have called `next()` itself.

src/middleware/routeHandler.ts
```diff
@@ -108,16 +108,19 @@ export const attachWriteRouteHandlerWrapper: RouteWrapper<{
...
+      }
+    },
+    async (err) => {
+      await unlock(siteName)
+      next(err)
```
timotheeg, line 121: Same vein as above: comparatively, why is this unlock not in a try..catch? This is the catch handler so the handler failed (and presumably did not call `next()`, so if the lock release also fails, should we not guarantee a call to `next(err)`?

timotheeg (commented):
Adding comments to illustrate the notes.

alexanderleegs (approved):
As discussed offline - will file a ticket to re-look into how we handle our route handler wrappers!

kishore03109 (approved):
>For this to work (same a rollback routes), there is an assumption that ALL write route handlers are async and resolve when their job is done, but that is NOT enforced from typings. The handler is of type RequestHandler which is not specified to return a Promise, it returns void.

hey i dont get this. if we wrap a promise around sync code, this is not a problem at all? we dont need to have the assumption that all the calls are async right?

> The handling of the express flow still looks wrong and should be discussed. When the handler is successful, and the lock fails to release, does it really make sense to call next(err)? The query was handled successfully, and failure to release the lock could be retried or just logged (and tracked!)

i get the concern, but rather than wrapping head around the complexity of this when using sub routers, can we just log something and push to prod line and monitor for two weeks? that way we track this without much eng effort + apply targeted fix.

src/middleware/routeHandler.ts
```diff
@@ -108,16 +108,19 @@ export const attachWriteRouteHandlerWrapper: RouteWrapper<{
     return
   }
 
-  Promise.resolve(routeHandler(req, res, next)).catch(async (err: Error) => {
-    await unlock(siteName)
-    next(err)
```
kishore03109, line 113: could we just log on next? that way we can monitor directly on prod if any of the route handlers call next before we do?

kishore03109 (approved):
am approving this with a nit of adding logging for a quick win

> For this to work (same a rollback routes), there is an assumption that ALL write route handlers are async and resolve when their job is done, but that is NOT enforced from typings. The handler is of type RequestHandler which is not specified to return a Promise, it returns void.

hey i dont get this. if we wrap a promise around sync code, this is not a problem at all? we dont need to have the assumption that all the calls are async right?

> The handling of the express flow still looks wrong and should be discussed. When the handler is successful, and the lock fails to release, does it really make sense to call next(err)? The query was handled successfully, and failure to release the lock could be retried or just logged (and tracked!)

i get the concern, but rather than wrapping head around the complexity of this when using sub routers, can we just log something and push to prod line and monitor for two weeks? that way we track this without much eng effort + apply targeted fix.

timotheeg:
I merged this first since it's a minimalistic fix that is obviously an improvement.

After release, we should both 1) monitor traces and, 2)  inspect the code to verify that the handlers have valid async behaviour.

On 2) It's OK when some work is done fire-and-forget on purpose (e.g. git push), but we should still verify that all routes exhibit their intended async behaviour. There are not that many routes that use the wrappers, so hopefully doing static code analysis on them shouldn't take too long 🤞

```bash 
git grep attachRollbackRouteHandlerWrapper | grep -v spec | wc -l
      54

git grep attachWriteRouteHandlerWrapper | grep -v spec | wc -l
      15 
```

PR #1260, opened 2024-04-02
feat: carry the name of the wrapped handler

timotheeg:
## Problem

APM traces do not show the name of the handlers because they are wrapped by anonymous functions

sample trace [here](https://ogp.datadoghq.com/apm/trace/4623130037215300159?colorBy=service&env=production&graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=2269893028902401593&spanViewType=metadata&timeHint=1712017202477&view=spans)

image

Note: The PR changes indentations, watch the changeset with the ignore whitespace setting on. [Direct link here](https://github.com/isomerpages/isomercms-backend/pull/1260/files?w=1).

## Solution

Carry the name of the wrapped function onto the wrapper.

Function names are typically read-only, so we must use [`Object. defineProperty()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty) to set the name.

## Tests

- [ ] Load a PR review in CMS UI
- [ ] Open DtaDog and find the trace, verify that the span is not anonymous

### DIFF
--- a/src/middleware/routeHandler.ts
+++ b/src/middleware/routeHandler.ts
@@ -76,49 +76,57 @@ type RouteWrapper<
   next: NextFunction
 ) => void
 
-// Used when there are no write API calls to the repo on GitHub
-export const attachReadRouteHandlerWrapper: RouteWrapper = (
-  routeHandler
-) => async (req, res, next) => {
-  Promise.resolve(routeHandler(req, res, next)).catch((err: Error) => {
-    next(err)
-  })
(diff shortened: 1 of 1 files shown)


timotheeg:
Woohoo! Awesome! Thanks for the typing fix @seaerchin! 🎉 🙏

alexanderleegs (approved):
lgtm!

PR #1267, opened 2024-04-03
Improve APM spans (no more <anonymous>)

timotheeg:
## Problem

We have a lot of spans in APM traces which are `<anonymous>`. While not a deal breaker to get value out of APM, it is annoying.

image

The anonymous functions are due to the way some methods are declared in classes, which is done to be able to specify types.

e.g. in [`ReviewsRouter.ts`](https://github.com/isomerpages/isomercms-backend/blob/develop/src/routes/v2/authenticated/review.ts#L92-L100), the assignment syntax creates anonymous methods
```typescript
  compareDiff: RequestHandler<
    { siteName: string },
    { items: EditedItemDto[] } | ResponseErrorBody,
    unknown,
    unknown,
  > = async (req, res) => {
    // ... 
  }
```
while in [`collectionPages.js`](https://github.com/isomerpages/isomercms-backend/blob/develop/src/routes/v2/authenticatedSites/collectionPages.js), this style of named method works:
```typescript
  async updateCollectionPage(req, res, next) {
    // ...
  }
```

Note: The first style of method declaration (assignment) does **NOT** create methods on the class' prototype. The methods are created on the instance themselves at construction (and so just FYI the call to `autoBind(this)` are actually useless for classes which only use that style of declaration)

TODO: We should ensure that all methods we use as express route handlers are named to have nice usable traces

## Solution

* Provide a utility `nameAnonymousMethods()` to ensure all methods of an instance and its class are named (invoked at construction)
* Update all router classes (ts files) to ensure all methods are named

Updating the Router classes was done via the script below, which was then cleaned up on commit by eslint/prettier:
```bash
grep 'autoBind(this)' $(git grep Router | grep class | grep Router | cut -d: -f1) | cut -d: -f1 | grep -v -E '.js$' | while read f; do
  echo 'import { nameAnonymousMethods } from "@root/utils/apm-utils"' > temp;
  sed -E -e 's/autoBind\(this\)/nameAnonymousMethods(this);autoBind(this)/' "${f}" >> temp;
  mv temp "${f}";
done
```

Verified on staging:
image

## Tests
- [x] Load the CMS and perform some actions, all happy paths should work
- [x] Find a span for a Review compare, and verify the span name is `compareDiff` (**NOT** `<anonymous>`)

### DIFF
--- a/src/routes/formsg/formsgSiteLaunch.ts
+++ b/src/routes/formsg/formsgSiteLaunch.ts
@@ -27,6 +27,7 @@ import {
 import TRUSTED_AMPLIFY_CAA_RECORDS from "@root/types/caaAmplify"
 import { isErrnoException } from "@root/types/nodeError"
 import { SiteLaunchResult } from "@root/types/siteLaunch"
+import { nameAnonymousMethods } from "@root/utils/apm-utils"
 import UsersService from "@services/identity/UsersService"
 import InfraService from "@services/infra/InfraService"
 
@@ -150,6 +151,7 @@ export class FormsgSiteLaunchRouter {
     this.usersService = usersService
     this.infraService = infraService
     // We need to bind all methods because we don't invoke them from the class directly
+    nameAnonymousMethods(this)
     autoBind(this)
   }
 
--- a/src/middleware/routeHandler.ts
+++ b/src/middleware/routeHandler.ts
@@ -88,7 +88,7 @@ const nameRouteHandlerWrapper = <
 ): RequestHandler<Params, ResBody, ReqBody, ReqQuery, Locals> => {
   if (original.name) {
     Object.defineProperty(wrapper, "name", {
-      value: original.name,
+      value: original.name.replace(/^bound /, ""),
       writable: false,
     })
   }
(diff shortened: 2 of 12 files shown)


src/routes/formsg/formsgSiteLaunch.ts
```diff
@@ -150,6 +151,7 @@ export class FormsgSiteLaunchRouter {
     this.usersService = usersService
     this.infraService = infraService
     // We need to bind all methods because we don't invoke them from the class directly
+    nameAnonymousMethods(this)
     autoBind(this)
```
timotheeg, line 155: Note on this for reviewer: `autoBind(this)`in this file (and all the router files thathave been converted to typescript) doesn't actually do what we think it does 😑😢 

When the methods are declared like this in the class:
```
 handleSiteLaunchFormRequest: RequestHandler<
    never,
    string,
    { data: { submissionId: string } },
    never,
    { submission: DecryptedContentAndAttachments }
  > = async (req, res) => {
   //. ...
}
```
then they are:
1) anonymous
2) **NOT added to the class' prototype**, so `autoBind()` will not get to them at all

alexanderleegs (approved):

PR #1269, opened 2024-04-04
ref(app): shift support flows into separate folder + ecs service

seaerchin:
**NOTE: production deploy stuff is in #1285**

## Problem
At present, some internal flows trigger a [huge spike](https://ogp.datadoghq.com/apm/trace/4475574368042772015?colorBy=service&env=production&graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=718787862051750205&spanViewType=infrastructure&timeHint=1711529331280.001&view=spans) in CPU usage, which will erroneously trigger an alarm. additionally, this CPU spike might cause a degradation on performance of the actual app. 

Closes ISOM-811

## Solution
1. use a separate ecs service to do it. we did **not** go with a sidecar container as was initially suggested. this is because ecs scaling is [done via metrics](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/service-autoscaling-targettracking.html) and the most fine-grained metric we have at present (that is relevant) is ecs cpu usage _by service_. as the sidecar is still part of the service, this would cause the support container to [trigger scaling](https://docs.aws.amazon.com/AmazonECS/latest/bestpracticesguide/application.html#task-definition) (and still alarms) on the service as a whole, which doesn't solve the problem. 
2. shift shared services out into `/common` -> this is to facilitate sharing of services between `support` and `src`. the new service listens on port 8082 - this is because on local, we want to expose both services ports to the host machine (hence cannot use same port). 
3.  added deployment step for **staging only** for `support` - added 2 more params to the `aws_deploy` workflow to support this. (as service + dockerfile is now different)
4. add `task-def` for `support` - we cannot reuse existing one because it's a new ecs service. 
 
**Breaking Changes**

  - not user facing, but the existing forms will **not** work and will require an update to `<stg | prod>-support-api/v2/infra/formsg`

## Tests
These tests have been done on `staging`, but the success has been checked **only via email** (ie, receive success email = good enough) 
1. submit the staging site launch form
2. should receive an email regarding the dns records to update
3. submit the staging site creation form
4. should receive a success email
5. submit staging audit logs form
6. should receive success email
7. submit staging site link checker form
8. should have logs (`"*link*"` in `isomer-infra` **or** `isomer` (not entirely sure why, suspect it's because our project tag in infra repo is `isomer-infra` which leads to the tag being `isomer-infra`)
Screenshot 2024-04-04 at 1 40 03 PM
9. submit staging site repair form 
10. should see success email


.aws/deploy/support-task-definition.staging.json
```diff
@@ -0,0 +1,311 @@
...
+        },
+        {
+          "name": "SITE_AUDIT_LOGS_FORM_KEY",
+          "valueFrom": "STAGING_SITE_AUDIT_LOGS_FORM_KEY"
+        }
+      ],
+      "logConfiguration": {
+        "logDriver": "awslogs",
+        "options": {
+          "awslogs-group": "/aws/elasticbeanstalk/cms-backend-staging-node18/var/log/web.stdout.log",
```
seaerchin, line 206: open to creating another log group, nbd

.aws/deploy/support-task-definition.staging.json
```diff
@@ -0,0 +1,311 @@
+{
+  "containerDefinitions": [
+    {
+      "name": "support",
+      "portMappings": [
+        {
+          "containerPort": 8082,
```
seaerchin, line 7: only change between this and `backend`

common/index.ts
seaerchin, line 1: no big changes, just copied out services from `server.js` -> here + changed from `require` to `import`

support/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+RUN adduser -u 900 webapp -D -h /home/webapp -s /bin/sh
+USER webapp
+
+COPY ../src ./src
+COPY ../common ./common
+COPY ../support ./support
+COPY ../package.json . 
+COPY ../package-lock.json .
+COPY ../scripts ./scripts
+COPY ../tsconfig.json .
```
seaerchin, line 18: change 1 - we selectively copy rather than just `COPY . .`; we should be able to omit `COPY ../src ./src` if we are rigorous about separation of concerns but for now this is ok. build times are not impacted too much cos both jobs run in parallel

support/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+
+# NOTE: We need to run the app as webapp, otherwise we will encounter
+# permissions issues on EFS, and all files will be erroneously owned by root.
+USER webapp
+
+RUN echo "[user]" > /home/webapp/.gitconfig
+RUN echo "  name = Isomer Admin" >> /home/webapp/.gitconfig
+RUN echo "  email = admin@isomer.gov.sg" >> /home/webapp/.gitconfig
+
+EXPOSE "8082"
```
seaerchin, line 50: need this because we using port 8082

.aws/deploy/support-task-definition.staging.json
kishore03109: think better to keep the same

support/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+
+RUN apk update && \
+    apk add --no-cache bash && \
+    apk add git && \
+    apk add openssh-client
+
+RUN adduser -u 900 webapp -D -h /home/webapp -s /bin/sh
+USER webapp
+
+COPY ../src ./src
```
kishore03109, line 12: if we are spiltting this up, should we split this up for the main docker file as well?

kishore03109 (approved):
lgtm, i felt that we should use the side car for more intensive operations only (eg site launch + create does not seem like intensive ops, but this is a nit)

.aws/deploy/support-task-definition.staging.json
```diff
@@ -0,0 +1,311 @@
...
+          "valueFrom": "STAGING_MOCK_AMPLIFY_DOMAIN_ASSOCIATION_CALLS"
+        },
+        { "name": "MUTEX_TABLE_NAME", "valueFrom": "STAGING_MUTEX_TABLE_NAME" },
+        {
+          "name": "NETLIFY_ACCESS_TOKEN",
+          "valueFrom": "STAGING_NETLIFY_ACCESS_TOKEN"
+        },
+        { "name": "NODE_ENV", "valueFrom": "STAGING_NODE_ENV" },
+        { "name": "OTP_EXPIRY", "valueFrom": "STAGING_OTP_EXPIRY" },
+        { "name": "OTP_SECRET", "valueFrom": "STAGING_OTP_SECRET" },
```
timotheeg, line 123: Should some of these variables removed from the support container? I assume they are not useful there? I don't have a sense of the full list of what's needed and not, but general principle is to only give the app what it needs.

.aws/deploy/support-task-definition.staging.json
```diff
@@ -0,0 +1,311 @@
...
+          "value": "staging"
+        },
+        {
+          "name": "DD_SERVICE",
+          "value": "isomer"
+        },
+        {
+          "name": "DD_TAGS",
+          "value": "service:isomer"
+        },
```
timotheeg, line 255: We should use a different service name for this (maybe `isomer-support`), so we can also see the ECS infra metrics and the endpoint performance under the service

.aws/deploy/support-task-definition.staging.json
timotheeg: +1 to creating another log group

Maybe we can drop `elasticbeanstalk` from the name too...

.aws/deploy/support-task-definition.staging.json
```diff
@@ -0,0 +1,311 @@
...
+          "value": "true"
+        },
+        {
+          "name": "DD_API_KEY",
+          "value": "<DD_API_KEY>"
+        }
+      ],
+      "dockerLabels": {
+        "com.datadoghq.tags.env": "staging",
+        "com.datadoghq.tags.service": "isomer",
```
timotheeg, line 275: When service is renamed, this will need to change as well.

timotheeg (commented):
Looking good!

Is the PR missing a file for prod?
```
.aws/deploy/support-task-definition.prod.json
```

.aws/deploy/support-task-definition.staging.json
seaerchin: i copied it because i didn't want to sift through and manually find out what env vars are/are not used.

seaerchin:
> Looking good!
> 
> Is the PR missing a file for prod?
> 
> ```
> .aws/deploy/support-task-definition.prod.json
> ```

yes it is! as mentioned, the CNAME setup + deploy workflow hasn't been done and will be in another PR

.aws/deploy/support-task-definition.staging.json
seaerchin: ya can, i think i'll change it to an ecs-prefixed log group **post tear down**  of eb

.aws/deploy/support-task-definition.staging.json
seaerchin: this deploys to `staging`! i think u mean prod? will do in a folloow up PR

.aws/deploy/support-task-definition.staging.json
timotheeg: Yes! Since we are creating a new service, we need to have the telemetry set up and linked well, so we all the drill down capability of DD work.

src/types/request.ts
```diff
@@ -35,13 +35,4 @@ export type RequestHandlerWithGrowthbook<
...
-  req: RequestWithGrowthBook & {
-    params: P
-    body: ReqBody
-    query: ReqQuery
-    locals: Locals
-  },
-  res: Response<ResBody, Locals>,
-  next: NextFunction
-) => void
+> = ExpressHandler<P, ResBody, ReqBody, ReqQuery, Locals>
```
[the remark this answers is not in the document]
seaerchin: the change was made to keep our types neater. we already have a global `express.d.ts`, which overrides the `Request` type exported originally! 

this means that the original typing here is superfluous and hence, removed

support/Dockerfile
seaerchin: to clarify, we want to omit copying of `support` into our main `Dockerfile`? 

we totally can honestly, will make the change. best is if we can omit copying `src` to the support dockerfile but it doesn't impact build times much because the `src` folder is small.

.aws/deploy/support-task-definition.staging.json
timotheeg: Bit painful yes, but we should do it anyway for good hygiene. If the service doesn't need to know of certain secrets or values, we shouldn't pass them in.

seaerchin: sounds good! you ok if i do this in a follow-up? my immediate priority now is to get functionality up and merged - will address this hygiene issue downstream

.aws/deploy/support-task-definition.staging.json
seaerchin: sgtm!

support/routes/v2/formsg/index.ts
```diff
@@ -0,0 +1,47 @@
...
+import { FormsgSiteCheckerRouter } from "./formsgSiteChecker"
+import { FormsgSiteCreateRouter } from "./formsgSiteCreation"
+import { FormsgSiteLaunchRouter } from "./formsgSiteLaunch"
+
+const formsgSiteCreateRouter = new FormsgSiteCreateRouter({
+  usersService,
+  infraService,
+  gitFileSystemService,
+})
+const formsgSiteLaunchRouter = new FormsgSiteLaunchRouter({
```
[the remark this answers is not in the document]
seaerchin: sgtm

seaerchin:
@timotheeg

for the issues listed below, i'll do in another PR
1. removing extra env vars -> functionality now is intact, want to get this out and fix hygiene issues later (will raise issue) 
2. creating separate log group -> this is because teh creation of a separate log group requires `pulumi up` + the PR itself to be approved on `infra` prior to the PR being raised

.aws/deploy/support-task-definition.staging.json
seaerchin: changed to `isomer-support` (keep things consistent throughout between files/service)

i've changed it on 3 areas: 
1.  `support` container itself (done via hard-coded env vars instead of via secrets injection through SSM - idt secrets injection adds much here and obfuscates the actual value itself, which is annoying)
2. updated env vars on `dd-agent` container too (in `support-task-def`) 
3. updated `dockerLabels` on `dd-agent` 

this should allow us to get the tagging separately, do check if i missed out anything!

.aws/deploy/support-task-definition.staging.json
seaerchin: updated, see above comment

.aws/deploy/support-task-definition.staging.json
```diff
@@ -0,0 +1,313 @@
...
+          "name": "DD_ENV",
+          "value": "staging"
+        },
+        {
+          "name": "DD_SERVICE",
+          "value": "isomer-support"
+        },
+        {
+          "name": "DD_TAGS",
+          "value": "team:isomer"
```
timotheeg, line 256: In case the infra linkage doesn't work (happened in armoury), we might need to set the service here too:
```suggestion
          "value": "team:isomer,service:isomer-support"
```

.github/workflows/deploy_staging.yml
```diff
@@ -21,11 +21,36 @@ jobs:
...
+  
+  deploy-support: 
+    name: Deploy support to staging
+    uses: ./.github/workflows/aws_deploy.yml
+    with:
+      aws-region: "ap-southeast-1"
+      cicd-role: "arn:aws:iam::095733531422:role/isomer-infra-github-oidc-role-16ea937"
+      ecr-repository: "isomer-infra-staging-ecr"
+      ecs-cluster-name: "isomer-stg-ecs"
+      ecs-web-service-name: "stg-support-ecs-service"
```
timotheeg, line 45: Naming convention doens't seem to match the deploy-app name? Did you mean
```suggestion
      ecs-web-service-name: "isomer-support-stg-ecs-service"
```

PR #1275, opened 2024-04-05
chore: index notifications table

alexanderleegs:
## Problem

Our notifications table is large, and we run frequent SELECT queries on it, but postgres does not automatically index foreign key columns, so WHERE clauses on these columns will require a full table scan, which results in inefficient db queries. This PR adds indexes to the foreign keys in the notifications table to speed up our queries.

Note that `site_id` is only indexed with `user_id`, since we never filter by `site_id` alone.

## Deploy Notes

- Run migration on staging and master before merge

### DIFF
--- a/src/database/migrations/20240405132731-add-notification-indexes.js
+++ b/src/database/migrations/20240405132731-add-notification-indexes.js
@@ -1,44 +1,24 @@
 /** @type {import('sequelize-cli').Migration} */
 module.exports = {
   async up(queryInterface, Sequelize) {
-    queryInterface.sequelize.transaction((t) =>
-      Promise.all([
-        queryInterface.addIndex(
-          "notifications", // name of Source model
-          ["site_id", "user_id"],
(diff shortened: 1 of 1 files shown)


src/database/migrations/20240405132731-add-notification-indexes.js
```diff
@@ -0,0 +1,44 @@
...
+          }
+        ),
+        queryInterface.addIndex(
+          "notifications", // name of Source model
+          ["user_id"],
+          {
+            transaction: t,
+          }
+        ),
+      ])
```
timotheeg, line 20: ⚠️ Warning: this might not work. Creating indexes on table locks, and so typically you do not want to do it in parallel, and not in a transaction either.

I recommend you drop the transaction entirely and await each lock creation.

Same comment for the the removal

timotheeg (approved):
LGTM

PR #1278, opened 2024-04-08
fix: name auth routers

timotheeg:
## Problem

The Auth middlewares are not named correctly:

Example trace:
image

[Trace link for reference](https://ogp.datadoghq.com/apm/trace/3862022928270485881?colorBy=service&env=production&graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=6340181816854815641&spanViewType=metadata&timeHint=1712392136957&view=spans)

## Solution

Use the method namer introduced in [this PR](https://github.com/isomerpages/isomercms-backend/pull/1267)

### DIFF
--- a/src/middleware/authorization.ts
+++ b/src/middleware/authorization.ts
@@ -5,6 +5,7 @@ import { ForbiddenError } from "@errors/ForbiddenError"
 import UserWithSiteSessionData from "@classes/UserWithSiteSessionData"
 
 import { RequestHandler } from "@root/types"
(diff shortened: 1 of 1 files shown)


seaerchin (approved):
lgtm; i checked this by just going into the `v2` routers and searching for arrow functions, which don't have a `this` binding (and hence, no `name`). 

i think `v2/authenticated/metrics.ts` was missing this but there's only 1 method, `collateUserFeedback` so maybe we just want to change it to a named function and merge it

PR #1280, opened 2024-04-08
fix(mergify): fix copy/paste mistake

seaerchin:
## Problem
merged previous mergify PR but managed to paste wrongly -> this resulted in **all non sem-ver minor** PRs up being merged.

this is 
1. [PR](https://github.com/isomerpages/isomercms-backend/commit/08ec28a66a954863135fa651ac3d338d1898ca7f) for `base-64` (we don't use this - maybe can remove) 
2. [PR](https://github.com/isomerpages/isomercms-backend/commit/a06afdddea52fa87124ec07ff228b0c4f6e8ec96) for `marked` -> we **do use this**! however, the change is to update `marked` to conform to changes in the markdown spec (to resolve ambiguity...)

## Solution
- change rule from `minor` -> `major`
- looked at marked's 12.0 update [here](https://github.com/markedjs/marked/releases/tag/v12.0.0); ss attached below for convenience; as most of the changes are helpful to us (we only call `marked.parse`), i'm **not** going to rollback the PR
Screenshot 2024-04-08 at 11 00 17 PM

**Breaking Changes**

  - Details ...

**Features**:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- Details ...

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Deploy Notes

- `env var` : env var details

**New scripts**:

- `script` : script details

**New dependencies**:

**New dev dependencies**:

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -5,7 +5,7 @@ pull_request_rules:
       - check-success~=lint
       - check-success~=build
       - check-success~=test
-      - -dependabot-update-type = version-update:semver-minor
+      - -dependabot-update-type = version-update:semver-major
     actions:
       review:
         type: APPROVE


.github/mergify.yml
```diff
@@ -5,7 +5,7 @@ pull_request_rules:
       - check-success~=lint
       - check-success~=build
       - check-success~=test
-      - -dependabot-update-type = version-update:semver-minor
+      - -dependabot-update-type = version-update:semver-major
```
alexanderleegs, line 8: Just checking, we use this so that we get both `patch` and `minor` updates right?

alexanderleegs (approved):
lgtm, minor clarification

.github/mergify.yml
seaerchin: Yeap, see original pr for details!

PR #1281, opened 2024-04-09
refactor(UserService): simplify login by using findOrCreate()

timotheeg:
## Problem

Incorrect comments on not being able to use `findOrCreate()` with a transaction.

(Also I think the old code for create might have been incorrect, because the first argument to `create()` must be an object, not a string (?) 🤔 

While `.findOrCreate()` does in fact accept a transaction (see [API documentation](https://sequelize.org/api/v6/class/src/model.js~model#static-method-findOrCreate)), I think the update to `lastLoggedIn` doesn't need to be in the transaction. Even if there is a conflict with multiple writes overwriting each other, the updates would be so close by that's the inaccuracy is not business critical.

Because of that, we can simplify the code by using the default transaction behaviour of `.findOrCreate()`, instead of providing our own transaction.

## Solution

1. Use `.findOrCreate()` in both `login` and `loginWithEmail`
2. Move the update to `lastLoggedIn` out of transaction
3. Update the tests

`.findOrCreate()` does provide its own transaction, [documentation](https://sequelize.org/api/v6/class/src/model.js~model#static-method-findOrCreate) states

> If no transaction is passed in the options object, a new transaction will be created internally, to prevent the race condition where a matching row is created by another connection after the find but before the insert call. 

Verified on staging ([sample trace here](https://ogp.datadoghq.com/apm/trace/6810154304698742342?colorBy=service&env=staging&graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=5390287072878530279&spanViewType=metadata&timeHint=1712797446427&view=spans)):
image

## Tests

- [x] Login successfully via email
- [ ] Login successfully via github

### DIFF
--- a/src/services/identity/UsersService.ts
+++ b/src/services/identity/UsersService.ts
@@ -136,30 +136,19 @@ class UsersService {
   }
 
   async login(githubId: string): Promise<User> {
-    const loginUpdate = { lastLoggedIn: new Date() }
-    const [user, created] = await this.repository.findOrCreate({
-      where: { githubId },
-      defaults: loginUpdate,
+    const [user] = await this.repository.upsert({
+      githubId,
+      lastLoggedIn: new Date(),
     })
-
-    if (!created) {
-      await user.update(loginUpdate)
-    }
 
     return user
   }
 
   async loginWithEmail(email: string): Promise<User> {
-    const parsedEmail = email.toLowerCase()
-    const loginUpdate = { lastLoggedIn: new Date() }
-    const [user, created] = await this.repository.findOrCreate({
-      where: { email: parsedEmail },
-      defaults: loginUpdate,
+    const [user] = await this.repository.upsert({
+      email: email.toLowerCase(),
+      lastLoggedIn: new Date(),
     })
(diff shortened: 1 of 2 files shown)


src/services/identity/UsersService.ts
```diff
@@ -137,28 +137,19 @@ class UsersService {
...
-      if (!user) {
-        user = await this.repository.create({
-          githubId,
-          transaction,
-        })
-      }
-
       user.lastLoggedIn = new Date()
+
       return user.save({ transaction })
```
timotheeg, line 146: Hmm, it's probably not necessary to run this save in the transaction 🤔 . The code could be simplified further by removing our custom transaction and use sequlize's internal transaction that is created when calling `findOrCreate()`. Per [documentation](https://sequelize.org/api/v6/class/src/model.js~model#static-method-findOrCreate):

> If no transaction is passed in the options object, a new transaction will be created internally, to prevent the race condition where a matching row is created by another connection after the find but before the insert call.

seaerchin (approved):

PR #1283, opened 2024-04-09
refactor(OTP): simplify code by using upsert() 

timotheeg:
## Problem

The code for `sendEmailOtp()` and `sendSmsOtp()` have a behaviour of find-then-create-or-update. However that not done in a transaction, and is not sync-safe.

Sequelize has `findOrCreate()` which does create transaction by default, so we can just use that instead. 

Additionally the method `createOtpEntry()` is not very useful since there are 2 code path which are conditional on the otp types, and these are already distinct in the methods `sendEmailOtp()` and `sendSmsOtp()` themselves.

Also there was one possible issue (not sure): The 2 OTP methods did not have the same behaviour to reset the OTP expiry (only the email OTP would reset the expiry). I'm not sure if it's on purpose or not, so it's worth having someone commenting on it. I'll start a thread in the PR to indicate the potential issue.

## Solution

* Use `upsert()` as needed
* remove unnecessary method `createOtpEntry()`

Verified on staging ([sample trace here](https://ogp.datadoghq.com/apm/trace/2318591157452590738?colorBy=service&env=staging&graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=8356505526243913396&spanViewType=metadata&timeHint=1712798060616))

image

## Tests

- [x] Login via email OTP

### DIFF
--- a/src/services/identity/UsersService.ts
+++ b/src/services/identity/UsersService.ts
@@ -189,20 +189,21 @@ class UsersService {
   async sendEmailOtp(email: string) {
     const parsedEmail = email.toLowerCase()
     const { otp, hashedOtp } = await this.otpService.generateLoginOtpWithHash()
+    const data = {
+      hashedOtp,
+      attempts: 0,
+      expiresAt: this.getOtpExpiry(),
+    }
 
     // Reset attempts to login
-    const otpEntry = await this.otpRepository.findOne({
+    // Reset attempts to login
+    const [otpEntry, created] = await this.otpRepository.findOrCreate({
       where: { email: parsedEmail },
+      defaults: data,
     })
-    if (!otpEntry) {
-      // create new entry
-      await this.createOtpEntry(parsedEmail, OtpType.Email, hashedOtp)
-    } else {
-      await otpEntry?.update({
-        hashedOtp,
-        attempts: 0,
-        expiresAt: this.getOtpExpiry(),
-      })
+
+    if (!created) {
+      await otpEntry?.update(data)
     }
 
     const subject = "One-Time Password (OTP) for IsomerCMS"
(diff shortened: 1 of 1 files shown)


src/services/identity/UsersService.ts
```diff
@@ -216,15 +217,22 @@ class UsersService {
...
+    const [otpEntry, created] = await this.otpRepository.findOrCreate({
+      where: {
+        mobileNumber,
+      },
+      defaults: data,
     })
-    if (!otpEntry) {
-      await this.createOtpEntry(mobileNumber, OtpType.Mobile, hashedOtp)
-    } else {
-      await otpEntry?.update({ hashedOtp, attempts: 0 })
```
timotheeg, line 227: Why was this update not updating `expiresAt`, like what `sendEmailOtp()` does?

src/services/identity/UsersService.ts
```diff
@@ -189,20 +189,21 @@ class UsersService {
...
     const { otp, hashedOtp } = await this.otpService.generateLoginOtpWithHash()
+    const data = {
+      hashedOtp,
+      attempts: 0,
+      expiresAt: this.getOtpExpiry(),
+    }
 
     // Reset attempts to login
-    const otpEntry = await this.otpRepository.findOne({
+    // Reset attempts to login
```
seaerchin, line 199: nit: this might be a duplicate comment

src/services/identity/UsersService.ts
```diff
@@ -189,20 +189,21 @@ class UsersService {
...
-      await this.createOtpEntry(parsedEmail, OtpType.Email, hashedOtp)
-    } else {
-      await otpEntry?.update({
-        hashedOtp,
-        attempts: 0,
-        expiresAt: this.getOtpExpiry(),
-      })
+
+    if (!created) {
+      await otpEntry?.update(data)
```
seaerchin, line 206: nit: actually because the data is already available prior and the important thing here is just to create, maybe `upsert` might be a better fit?

seaerchin (approved):
minor comment, behaviour seems consistent anyways

src/services/identity/UsersService.ts
timotheeg: You're right! `upsert` is the pattern being used in both the refactor PRT I opened, I'll do one more round of edits

src/services/identity/UsersService.ts
timotheeg: Thanks! Removed!

src/services/identity/UsersService.ts
timotheeg: Done https://github.com/isomerpages/isomercms-backend/pull/1283/commits/1a15704fa4832d49b3cc44973697bf2536f1d751

PR #1285, opened 2024-04-09
ci(sidecar): add deploy files for prod

seaerchin:
**NOTE: DO NOT MERGE, PENDING INFRA DEPLOY**

## Problem
This adds in the missing deployment files for our support ecs service **on production**. this is raised as a separate PR so we can go through these 2 files in excruciating detail. the base PR is #1269.

For reviewers, please help to crosscheck against what i did below for env vars + deployment workflow to make sure i haven't missed anything

## Solution
- create a new task definition for `support` for deployment to production. 
- update our existing `deploy_prod` workflow to have a step to deploy the support ecs service.

for verification, these are the steps that i took: 
**task def**
- copied over the existing `support-task-def` for `staging` to `production` 
- thereafter, i **deleted** the `secrets` section and replaced it with the one for `backend-task-def` (sans `DD_SERVICE/DD_TAGS`, which are done via `env` on the task def)
- replaced remaining `stg/staging` references with `prod/production`; referenced `backend-task-def` for the correct version (short or long) 
**deploy_prod**
- copied over the existing `deploy-support` workflow
- updated values (simple replacement cos this is deterministic via `pulumi`)

**Breaking Changes**

  - Details ...
- [x] No - this PR is backwards compatible with ALL of the following feature flags in this [doc](https://www.notion.so/opengov/Existing-feature-flags-518ad2cdc325420893a105e88c432be5) - should be fully back-compat, this should consume form webhook just fine

## Tests
See base PR!

### DIFF
--- a/.aws/deploy/support-task-definition.prod.json
+++ b/.aws/deploy/support-task-definition.prod.json
@@ -13,7 +13,7 @@
       "environment": [
         { "name": "ENV_TYPE", "value": "PROD" },
         { "name": "DD_SERVICE", "value": "isomer-support" },
-        { "name": "DD_TAGS", "value": "team:isomer" }
+        { "name": "DD_TAGS", "value": "team:isomer,service:isomer-support" }
       ],
       "mountPoints": [
         {
@@ -244,7 +244,7 @@
         },
         {
           "name": "DD_TAGS",
-          "value": "team:isomer"
+          "value": "team:isomer,service:isomer-support"
         },
         {
           "name": "DD_AGENT_MAJOR_VERSION",
--- a/.aws/deploy/support-task-definition.prod.json
+++ b/.aws/deploy/support-task-definition.prod.json
@@ -1,0 +1,304 @@
+{
+  "containerDefinitions": [
+    {
+      "name": "support",
+      "portMappings": [
+        {
+          "containerPort": 8082,
+          "hostPort": 8082,
+          "protocol": "tcp"
+        }
+      ],
+      "essential": true,
+      "environment": [
+        { "name": "ENV_TYPE", "value": "PROD" },
+        { "name": "DD_SERVICE", "value": "isomer-support" },
+        { "name": "DD_TAGS", "value": "team:isomer" }
+      ],
+      "mountPoints": [
+        {
+          "sourceVolume": "ggs-efs",
+          "containerPath": "/efs",
+          "readOnly": false
+        }
+      ],
+      "linuxParameters": {
+        "initProcessEnabled": true
+      },
+      "volumesFrom": [],
+      "secrets": [
+        {
+          "name": "AUTH_TOKEN_EXPIRY_DURATION_IN_MILLISECONDS",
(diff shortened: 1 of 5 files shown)


seaerchin:
Screenshot 2024-04-09 at 2 01 01 PM

timestamp flaky

.aws/deploy/support-task-definition.prod.json
```diff
@@ -0,0 +1,304 @@
...
+          "name": "DD_ENV",
+          "value": "production"
+        },
+        {
+          "name": "DD_SERVICE",
+          "value": "isomer-support"
+        },
+        {
+          "name": "DD_TAGS",
+          "value": "team:isomer"
```
timotheeg, line 247: /notblocker: In case we are not seeing the service reported correctly, the service tag can be added here too (we've seen the infra linkage didn't work without it for armoury):
```suggestion
          "value": "team:isomer,service:isomer-support"
```

.github/workflows/deploy_prod.yml
```diff
@@ -18,13 +18,38 @@ jobs:
...
+
+  deploy-support: 
+    name: Deploy support to staging
+    uses: ./.github/workflows/aws_deploy.yml
+    with:
+      aws-region: "ap-southeast-1"
+      cicd-role: "arn:aws:iam::095733531422:role/isomer-infra-github-oidc-role-16ea937"
+      ecr-repository: "isomer-infra-prod-ecr"
+      ecs-cluster-name: "isomer-prod-ecs"
+      ecs-web-service-name: "prod-support-ecs-service"
```
timotheeg, line 42: The naming convention is not consistent to the name for deploy-app above (`isomer-prod-ecs-service`). Did you mean `isomer-support-ecs-service` instead?

.aws/deploy/support-task-definition.prod.json
seaerchin: i actually removed cos i thought it might conflict iwth the `DD_SERVICE` tag @__@ good to know, will add in!

.github/workflows/deploy_prod.yml
seaerchin: no, this is very annoying and an oversight on my part! 

first, there's a strict upper limit on the number of characters we can use for the `name` here - this is because of pulumi (it respects AWS' [length conventions](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticloadbalancingv2-targetgroup.html#cfn-elasticloadbalancingv2-targetgroup-name) for load balancer target groups) 

second, because of the above, i opted to go with `isomer-<env>` for app last time (i didn't know there would be a separate ecs service...). hence now, when i ran into the same length issue, i opted to go with `<env>-<service>`

seaerchin: sadly enough, this leads to it being confusing...

.aws/deploy/support-task-definition.prod.json
seaerchin: updated

alexanderleegs (approved):

PR #1291, opened 2024-04-11
fix(5xx): should return 4xx when file not found

kishore03109:
## Problem

Described in linear: 
missing media files in pages leads to a 5xx when the resource name itself cannot be processed

as an example, consider a link (some_link)[/%2fa/b] -> this resolves to //a/b, which is not processable by our backend as it results in an error like so: Error when getting Git blob hash: Error: HEAD:images/technews//tvws-part2.JPG  and in turn causes a 5xx. 

this 5xx is caused because GitFileService (iirc) will try to determine the blob hash (see if can find the file on disk) and in doing so, it will get an error (as the file doesn't exist), leading to the 500.  

see here for an example where this occurs

in order to reproduce this: 
- [ ] go to any site with an image
- [ ] create or click on an existing markdown page
- [ ] in the page, add an image
- [ ] open network console
- [ ] update the link to be /images/%2f<existing>
- [ ] there should be a 5xx in the network console.

Additionally, we also see a `fatal: path '.git' does not exist in 'HEAD'` in the logs every time the workspace is refreshed. 
This is because we rely on `fileStats` to get a list of directories, but forget to remove special directories such as `.git`. 
## Solution

Reason for this is that git under the hood does a file path normalision for checking if the file exists in local disk, but does not when checking when it exists in head. 

consider these three examples:

Screenshot 2024-04-11 at 12 58 41 PM

We use fileStats to see if a file exists before passing it to `GetGitBlobHash`. issue with this is that while the file could exist in disk and still return a correct folder stats, this should still return a file not found error as the path given by the user is still incorrect. (am opting for this rather than normalising the path in the function `GetGitBlobHash` to keep the implementation of `GetGitBlobHash` lean and not think about potential path traversal attacks.)

for the special files, we simply do a filter to ignore files like `.git`. We know that the class of errors with the strings ` exists on disk` and `Error when getting Git blob hash` mostly comes from the .git folder  by looking at dd logs [here](https://ogp.datadoghq.com/logs?query=service%3Aisomer%20env%3Aproduction%20%40aws.awslogs.logStream%3A%2Aecs%2A%20source%3A%22elasticbeanstalk%2Fcms-backend-prod-node18%2Fvar%2Flog%2Fweb.stdout.log%22%20%22Error%20when%20getting%20Git%20blob%20hash%22%20%22exists%20on%20disk%22%20-%22.git%22%20&cols=host%2Cservice&fromUser=true&index=%2A&messageDisplay=inline&refresh_mode=sliding&saved-view-id=2492484&storage=hot&stream_sort=desc&viz=stream&from_ts=1711516320734&to_ts=1712812320734&live=true). 

ie all instances of `exists on disk` and `Error when getting Git blob hash` errors in [prod for the past two weeks](https://ogp.datadoghq.com/logs?query=service%3Aisomer%20env%3Aproduction%20%40aws.awslogs.logStream%3A%2Aecs%2A%20source%3A%22elasticbeanstalk%2Fcms-backend-prod-node18%2Fvar%2Flog%2Fweb.stdout.log%22%20%22Error%20when%20getting%20Git%20blob%20hash%22%20%22exists%20on%20disk%22%20&cols=host%2Cservice&fromUser=true&index=%2A&messageDisplay=inline&refresh_mode=sliding&saved-view-id=2492484&storage=hot&stream_sort=desc&viz=stream&from_ts=1711516320734&to_ts=1712812320734&live=true) came from trying to access the .git folders. 

**Breaking Changes**

  - Details ...

## Before & After Screenshots

**BEFORE**:
Screenshot 2024-04-11 at 12 51 22 PM

**AFTER**:
Screenshot 2024-04-11 at 12 52 17 PM

## Tests

in order to reproduce this: 
- [ ] go to any site with an image
- [ ] create or click on an existing markdown page
- [ ] in the page, add an image
- [ ] open network console
- [ ] update the link to be /images/%2f<existing>
- [ ] there should be a 4xx in the network console.
- [ ] go into datadog, there should not be any instance of `fatal: path '.git' exists on disk, but not in 'HEAD'`

### DIFF
--- a/src/services/db/GitFileSystemService.ts
+++ b/src/services/db/GitFileSystemService.ts
@@ -319,7 +319,7 @@ export default class GitFileSystemService {
     repoName: string,
     filePath: string,
     isStaging: boolean
-  ): ResultAsync<string, GitFileSystemError> {
+  ): ResultAsync<string, GitFileSystemError | NotFoundError> {
     const efsVolPath = isStaging
       ? EFS_VOL_PATH_STAGING
       : EFS_VOL_PATH_STAGING_LITE
@@ -333,6 +333,14 @@ export default class GitFileSystemService {
         )
 
         if (error instanceof GitError) {
+          // NOTE: While some path can be potentially normalised by Git (eg. images//path.png),
+          // it is not guaranteed to exist in HEAD. We simply return a not found error in this case.
+          if (
+            error.message.includes("fatal: path") &&
+            error.message.includes("exists on disk, but not in 'HEAD'")
+          ) {
+            return new NotFoundError("File/Directory does not exist")
+          }
           return new GitFileSystemError("Unable to determine Git blob hash")
         }
 
@@ -1121,7 +1129,11 @@ export default class GitFileSystemService {
         )
       )
       .andThen((directoryContents) => {
-        const resultAsyncs = directoryContents.map((directoryItem) => {
+        const IGNORED_FILES = [".git"]
+        const filteredDirectoryContents = directoryContents.filter(
+          (directoryItem) => !IGNORED_FILES.includes(directoryItem.name)
+        )
+        const resultAsyncs = filteredDirectoryContents.map((directoryItem) => {
           const isDirectory = directoryItem.isDirectory()
           const { name } = directoryItem
           const path = directoryPath === "" ? name : `${directoryPath}/${name}`
--- a/src/services/db/__tests__/GitFileSystemService.spec.ts
+++ b/src/services/db/__tests__/GitFileSystemService.spec.ts
@@ -56,6 +56,11 @@ const dirTree = {
   },
 }
 
+const dirTreeWithIgnoredFiles = {
+  ...dirTree,
+  ".git": "fake git directory",
+}
+
 describe("GitFileSystemService", () => {
   beforeEach(() => {
     mockFs({
@@ -171,6 +176,18 @@ describe("GitFileSystemService", () => {
       )
 
       expect(result._unsafeUnwrapErr()).toBeInstanceOf(NotFoundError)
+    })
+
+    it("should ignore .git folder", async () => {
+      mockFs({
+        [EFS_VOL_PATH_STAGING]: dirTreeWithIgnoredFiles,
+      })
+      const result = await GitFileSystemService.listDirectoryContents(
+        "fake-repo",
+        "fake-empty-dir",
+        DEFAULT_BRANCH
+      )
+      expect(result._unsafeUnwrap()).toHaveLength(0)
     })
   })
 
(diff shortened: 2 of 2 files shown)


kishore03109:
* **#1291** <a href="https://app.graphite.dev/github/pr/isomerpages/isomercms-backend/1291?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop`

<h2></h2>

src/services/db/GitFileSystemService.ts
```diff
@@ -333,6 +333,12 @@ export default class GitFileSystemService {
         )
 
         if (error instanceof GitError) {
+          if (
```
harishv7, line 336: would you know why the request even contains things like //filename?

are these actually broken links that cms tries to resolve?

kishore03109: > would you know why the request even contains things like //filename?
we have a markdown editor, so there are possibility for typo + the logs suggest this has occurred at least once. 
we dont debounce too, to this could occur while the user is typing the correct string. 

> are these actually broken links that cms tries to resolve?
yes

harishv7: can we add comments for this check here so that we don't lose context of when this error statements occur?

src/services/db/GitFileSystemService.ts
```diff
@@ -333,6 +333,12 @@ export default class GitFileSystemService {
...
+            error.message.includes("fatal: path") &&
+            error.message.includes("exists on disk, but not in 'HEAD'")
```
seaerchin, line 338: not a blocker for now, because we can raise a PR to improve this + bug is fixed here but i wonder if `error.message` is a robust enough property for this?

src/services/db/GitFileSystemService.ts
```diff
@@ -1121,7 +1127,11 @@ export default class GitFileSystemService {
         )
       )
       .andThen((directoryContents) => {
-        const resultAsyncs = directoryContents.map((directoryItem) => {
+        const IGNORED_FILES = [".git"]
```
seaerchin, line 1130: i think the pattern we might wanna aim for here is to respect the filesystem's convention! in here, this means that hidden files/folders (prefixed by a `.) should be ignored rather than a specific instance.

wwdyt?

seaerchin (commented):
should we also have tests out for this PR since it's fixing a bug and we want to avoid a regression?

src/services/db/GitFileSystemService.ts
kishore03109: hmm i rather keep this as .git since this actually is not the file system convention, but rather the .git system's convention. since they dont have a direct mapping, opting for the stricter check here.

src/services/db/GitFileSystemService.ts
kishore03109: seems like best effort, since this is directly reliant on the error message that git returns us
do have an alternative in mind?

PR #1295, opened 2024-04-12
Fix + Refactor: improve verify otp flows

timotheeg:
## Problem

The verifyOTP flow has multiple defficiencies 😰 

* db transaction should not be started if the provided otp is invalid
* if the initial find otp entry fails, transactions will not be committed ⚠️ 😱 
* Transaction creation error is reported as a BadRequestError (4XX), it should be a DataBaseError (5XX)
* Error when finding the OTP entry is reported as a BadRequestError (4XX), it should be a DataBaseError (5XX)
* validating the otp db entry (e.g. expiry) should be done first, to allow early close of the transaction if anything is not right!
* The neverThrow promise chain carries the otpDbEntry at every step, but that's not required considering the object is available from the function context
* there's a lot of duplicate code
* Error logging with `JSON.stringify()`on the the error object would not output anything in logs because Error instances are not serializable

TODO: refactor to 
* reduce transaction usage to bare minimum
* guarantee the transaction is always closed
* make code more readable
* Use the new better logger capabilities

## Solution

Refactor the code to:
* Extract and share common code between mobile and email otp verification
* Correctly split internal server errors (5XX) from request errors (4XX)
* Handle the closing the transaction correctly
* if otp is invalid: don't start a transaction at all!
* if anything about the otp db entry is invalid, abort early to close the transaction
* Simplify code to not carry the otpEntry for every step (that removes unnecessary combine calls)
* Create new private function to extract common code

### Results:

Transaction still works as expected.

[Sample trace of succesful login](https://ogp.datadoghq.com/apm/trace/5566567161567526847?colorBy=service&env=staging&graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=8950173092715292615&spanViewType=metadata&timeHint=1712909148736&view=spans)
image

[Sample trace of an invalid OTP login](https://ogp.datadoghq.com/apm/trace/5917740591885955589?colorBy=service&env=staging&graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=1306121409681557315&spanViewType=metadata&timeHint=1712909171522)
image

[Sample trace of max attempt reached](https://ogp.datadoghq.com/apm/trace/7476393482834566997?colorBy=service&env=staging&graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=6375220807564881848&spanViewType=metadata&timeHint=1712909266633)
image

### Potential extra TODOs 

To be discussed if relevant / useful
- [ ] Enforce `otp: string` in the signature of `verifyEmailOtp()` and `verifyMobileOtp()` instead of `otp: string | undefined` (implies do the validation in the route handler instead!
- [ ] clean up route handlers for the verify OTP endpoints (in `/v2/authenticated/user.ts`, because at the moment, there's a mixture of exception throwing and neverthrow chains (e.g. [here](https://github.com/isomerpages/isomercms-backend/blob/develop/src/routes/v2/authenticated/users.ts#L139-L141))

Things that were added for discussion but discarded:
- ~~remove `otpEntry` from DB when maxAttempts is reached (?)~~ if we remove the entry, submitting an invalid OTP entry again will now show "OTP not found" instead of "max number of attempts reached".

## Tests

- [x] Log in with email OTP
- [x] Log in with OTP but supply invalid OTP, verify an error message shows as "OTP is not valid"
- [x] Submit the same invalid OTP multiple times till the message becomes "Max number of attempts reached", subsequent attempt will keep showing that same message


src/services/identity/UsersService.ts
```diff
@@ -208,201 +212,169 @@ class UsersService {
...
-          ),
-        ])
+        )
+      })
+      .andThen(() =>
+        ResultAsync.fromPromise(
+          this.otpService.verifyOtp(otp, otpEntry!.hashedOtp),
+          (error) =>
+            this.otpGetAndLogError(
+              BadRequestError,
```
timotheeg, line 333: ⚠️ This should not be a `BadRequestError`. Failure to perform the verification is not the same as failing the verification (which is done by checking the boolean output at the next step).

Failure to perform the verification should be an unexpected internal server error, maybe we could just use `BaseIsomerError` here? Or make an new Error type 🤔

src/services/identity/UsersService.ts
```diff
@@ -208,201 +212,169 @@ class UsersService {
...
-              )
-
-              return new DatabaseError("Error destroying OTP entry in database")
-            }
-          ).andThen(() => errAsync(new BadRequestError("Hashed OTP not found")))
+        if (!otpEntry.hashedOtp) {
+          return this.otpDestroyEntry(otpEntry, transaction!).andThen(() =>
+            errAsync(new BadRequestError("Hashed OTP not found"))
+          )
         }
```
timotheeg, line 313: This should probably be done in the DB query itself where we we only want record where `hashedOtp` is a truthy value.

timotheeg: Possible implementation for the query:
```
            where: {
              ...findConditions,
              hashedOtp: {
                [Op.regexp]: '\\S+' // at least one non-white character
              }
            },
```

harishv7:
@timotheeg thanks Tim! as discussed offline, I think we don't even require a transaction in this flow. Removing that will help simplify the code further

timotheeg:
Some extra note:

The traces show that the transaction has a big idle wait because it needs to wait for `bcrypt.compare()` to run, which is very slow!

Adding a short summary of discussion with @harishv7 

Is the transaction even needed at all here? We're both on the same page that, No, the transaction are just overkill here
* The deletions are atomic
* The increments are atomic

I'll open a separate ticket to schedule a cleanup.

timotheeg:
@harish: I dropped the transaction wrapper! More code deletion for the PR 💪 
https://github.com/isomerpages/isomercms-backend/pull/1295/commits/1585f8898ca45abae913874a1397fe6220e9fb3c

src/services/identity/UsersService.ts
timotheeg: done

timotheeg:
Thanks for bringing that up @dcshzj !

I went to check checked **GTA-24-012 WP3**, and I agree the current state of the PR sadly re-introduces the issue 😢 .

That said, I think introducing the transaction was an expensive addition, and we should find a better solution 🤔 . The issue we have here is that `bcrypt.compare()` which we ran  while the transaction is ongoing is expensive, so the transaction and lock take quite a bit of time (north of 80ms!).

I reckon we should have implemented the second recommendation of the report, which is a much cheaper alternative:

> Another method would be to use a WHERE clause during updates to ensure that the data is
not altered during the process. For example, UPDATE otps SET attempts = x WHERE
attempts = x - 1 AND id = y.

Let me add a commit for that.

One alternative is that I revert the removal of the transaction in this PR, just to get the legit cleanups released ASAP (unnecessary transaction started, transaction potentially left dangling, incorrect error types reported), and we can schedule a refactor of the transaction itself for later.

timotheeg:
@dcshzj @harishv7 Implemented the robust update in this commit:
https://github.com/isomerpages/isomercms-backend/pull/1295/commits/34fd4456f102f781617dd5cc91ce7a910e4a21c8

harishv7:
> Hey @timotheeg @harishv7, just noticed this PR. The transaction was originally added due to **GTA-24-012 WP3** in our latest VAPT, so if we were to remove it we would have to ensure that we don't reintroduce this back again.

Thanks for highlighting this @dcshzj!

harishv7 (approved):

PR #1300, opened 2024-04-13
fix: fix comments

timotheeg:
## Problem

Incorrect comment says `prefix`, when it should say `suffix` 🤦 

## Solution

Fix comment

### DIFF
--- a/src/services/identity/UsersService.ts
+++ b/src/services/identity/UsersService.ts
@@ -166,13 +166,13 @@ class UsersService {
     const whitelistDomains = whitelistEntries.map((entry) => entry.email)
     const hasMatchDomain =
       whitelistDomains.filter((domain) => {
-        // if domain is really just a domain (does not include a @ OR starts with a @), we can do a prefix match
+        // if domain is really just a domain (does not include a @ OR starts with a @), we can do a suffix match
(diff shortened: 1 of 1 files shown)


src/services/identity/UsersService.ts
```diff
@@ -166,13 +166,13 @@ class UsersService {
     const whitelistDomains = whitelistEntries.map((entry) => entry.email)
     const hasMatchDomain =
       whitelistDomains.filter((domain) => {
-        // if domain is really just a domain (does not include a @ OR starts with a @), we can do a prefix match
+        // if domain is really just a domain (does not include a @ OR starts with a @), we can do a suffix match
         if (/^@|^[^@]+$/.test(domain)) {
```
timotheeg, line 170: Not related to this PR, but we could flip the condition to use a simpler regexp (didn't think of it during incident yesterday 😑🤦)

```typescript
// if domain a full email address (it has content before the @), we can ONLY do exact match on it
if (/^.+@/.test(domain)) {
    return normalizedEmail === domain
}

// otherwise we can do a suffix match
return normalizedEmail.endsWith(domain)
```

timotheeg:
Closing PR. SUperseeded by https://github.com/isomerpages/isomercms-backend/pull/1301

PR #1301, opened 2024-04-13
Feat: Optimize canSendEmailOtp

timotheeg:
## Problem

`UserService.canSendEmailOtp()` uses a whitelist to determine if emails are allowed to receive OTPs.

The whitelist contains
1) domains that the input can be matched against by suffix match
2) full email that the input can be matched again with exact match

At the moment the code basically download the full content of the whitelist table, and does a filter and check in code locally. 

It would be much better to leverage the query capability of the DB, to ask if a whitelist entry exists and get a small yes/no type result.
 

## Solution

Replace app-level code filtering by a single DB query.

The DB query will return: either:
* the first whitelist entry that matches
* nothing if there is no match

The PR will log the whitelist match, so we can verify if anything is wrong.

#### Old query (for reference)
[Sample trace](https://ogp.datadoghq.com/apm/trace/6688008335714158923?colorBy=service&env=staging&graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=3260853590528579031&spanViewType=metadata&timeHint=1713259938105)
image

[Sample trace for success](https://ogp.datadoghq.com/apm/trace/9058953070850640436?colorBy=service&env=staging&graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=4403824909672225076&spanViewType=metadata&timeHint=1712978730516&view=spans)
image
image
image

[Sample trace for failure](https://ogp.datadoghq.com/apm/trace/460028362167627232?colorBy=service&env=staging&graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=779548380314975616&spanViewType=metadata&timeHint=1712979224210.0051&view=spans)
image
image

### Performance

This particularly is not very heavy, returning most of the time in less than 2 milliseconds (link to [performance tracking graph here](https://ogp.datadoghq.com/dashboard/8tj-6k5-p2v/isomer-cm-s?fromUser=false&refresh_mode=paused&view=spans&from_ts=1713088170314&to_ts=1713260970314&live=false&tile_focus=3522012591074081)). The purpose of the PR is not to get a great performance gain but instead to 1) correct a bad pattern of fetching a large resultset to do filtering in code when that could be done in the query itself; 2) show that we can use raw queries to implement efficient queries.

But so concretely:

When no whitelist entry matches, the new query does a full table scan and returns nothing.

The previous query also did a full table scan to find active whitelist rows, but also incurred the transport cost to return the data, whether or not there was a whitelist match.

The new query is slightly heavier on the DB because it makes it do more work, but at the size of the whitelist table, it's negligible. The network transit time of the resulset should still be a net benefit in total time spent.

Overall, we expect the raw query time to be equivalent, and this can hopefully be visualized after release in [this tracker in the isomer dashboard](https://ogp.datadoghq.com/dashboard/8tj-6k5-p2v/isomer-cm-s?fromUser=false&refresh_mode=paused&view=spans&from_ts=1713088787609&to_ts=1713261587609&live=false&tile_focus=7015368340178734).

### Important note:

With the whitelisting filtering logic done in the query rather than code, the unit tests are not representative of the behaviour any more. Great care should be spent on reviewing/testing the query, and ideally we should have synthetic tests that validate the behaviour as e2e tests (some happy path tests and some expected failure mode).

## Tests

Note the traces for the query will be at [this link](https://ogp.datadoghq.com/apm/traces?query=%40_top_level%3A1%20service%3Aisomer-postgres%20env%3Astaging%20%40db.statement%3A%22SELECT%20email%20FROM%20whitelist%20WHERE%20%28%20expiry%20is%20%3F%20OR%20expiry%20%3E%3D%20NOW%20%28%20%29%20%29%20AND%20CASE%20WHEN%20email%20~%20%3F%20THEN%20email%20%3D%20%3F%20ELSE%20%3F%20LIKE%20CONCAT%20%28%20%3F%20email%20%29%20END%20LIMIT%20%3F%22&cols=core_service%2Ccore_resource_name%2Clog_duration%2Clog_http.method%2Clog_http.status_code&fromUser=false&graphType=flamegraph&historicalData=true&messageDisplay=inline&query_translation_version=v0&shouldShowLegend=true&sort=time&spanType=service-entry&traceQuery=&view=spans&start=1711772720605&end=1712982320605&paused=false).

- [ ] Try to get an OTP for your ogp email address -> should succeed. 
    * Locate trace for `isomer-postgres` with the new whitelist search query, check the log for the match (e.g. `Email valid for OTP by whitelist`)
- [ ] Try to get an OTP for `badguy.jeneszh@gmail.com` -> should appear to succeed but not send an email
    * Locate trace for `isomer-postgres` with the new whitelist search query, check the log for the error (e.g. `Error occurred when attempting to login user <email>`)

### DIFF
--- a/src/services/identity/UsersService.ts
+++ b/src/services/identity/UsersService.ts
@@ -157,13 +157,23 @@ class UsersService {
     const normalizedEmail = email.toLowerCase()
 
     // raw query because ORMs suck!
+    // why: we want to leverage the DB to see if a single record exists that whitelists the input
+    // we do not want to download the whole table locally to filter in local code
+    // query logic:
+    // - if whitelist entry is a full email (something before the @), then do exact match
+    // - if whitelist entry is a domain (no @, or starting with @), then do suffix match
+    // Limit 1 is added to allow the query to exit early on first match
     const records = (await this.sequelize.query(
       `
-        SELECT email AS found
+        SELECT email
         FROM whitelist
         WHERE
           (expiry is NULL OR expiry >= NOW())
-          AND :email LIKE CONCAT('%', email)
+          AND
+          CASE WHEN email ~ '^.+@'
+            THEN email = :email
+            ELSE :email LIKE CONCAT('%', email)
+          END
         LIMIT 1
       `,
       {
--- a/src/services/identity/UsersService.ts
+++ b/src/services/identity/UsersService.ts
@@ -156,12 +156,17 @@ class UsersService {
   async canSendEmailOtp(email: string) {
     const normalizedEmail = email.toLowerCase()
 
-    // raw query because ORMs suck!
-    // why: we want to leverage the DB to see if a single record exists that whitelists the input
+    // Raw query for readability because it uses 2 "unusual" query patterns
+    // - a CASE WHEN in the WHERE clause
+    // - a where condition of the form 'input like cell' (with a concat() call thrown in), as opposed to the more common 'cell like input'
+    //
+    // Why? We want to leverage the DB to see if a single record exists that whitelists the input
     // we do not want to download the whole table locally to filter in local code
+    //
     // query logic:
     // - if whitelist entry is a full email (something before the @), then do exact match
     // - if whitelist entry is a domain (no @, or starting with @), then do suffix match
+    //
     // Limit 1 is added to allow the query to exit early on first match
     const records = (await this.sequelize.query(
       `
--- a/src/services/identity/UsersService.ts
+++ b/src/services/identity/UsersService.ts
@@ -168,7 +168,7 @@ class UsersService {
         SELECT email
         FROM whitelist
         WHERE
-          (expiry is NULL OR expiry >= NOW())
+          (expiry is NULL OR expiry > NOW())
           AND
           CASE WHEN email ~ '^.+@'
             THEN email = :email
--- a/src/services/identity/UsersService.ts
+++ b/src/services/identity/UsersService.ts
@@ -1,5 +1,5 @@
 import { ResultAsync, errAsync, okAsync } from "neverthrow"
-import { Op, ModelStatic, Transaction } from "sequelize"
+import { Op, ModelStatic, Transaction, QueryTypes } from "sequelize"
 import { Sequelize } from "sequelize-typescript"
 import { RequireAtLeastOne } from "type-fest"
 
@@ -155,26 +155,24 @@ class UsersService {
 
   async canSendEmailOtp(email: string) {
     const normalizedEmail = email.toLowerCase()
-    const whitelistEntries = await this.whitelist.findAll({
-      attributes: ["email"],
-      where: {
-        expiry: {
-          [Op.or]: [{ [Op.is]: null }, { [Op.gt]: new Date() }],
-        },
-      },
-    })
-    const whitelistDomains = whitelistEntries.map((entry) => entry.email)
-    const hasMatchDomain =
-      whitelistDomains.filter((domain) => {
-        // if domain is really just a domain (does not include a @ OR starts with a @), we can do a prefix match
-        if (/^@|^[^@]+$/.test(domain)) {
-          return normalizedEmail.endsWith(domain)
-        }
 
-        return normalizedEmail === domain
-        // otherwise we can ONLY do an exact match
-      }).length > 0
-    return hasMatchDomain
+    // raw query because ORMs suck!
+    const records = (await this.sequelize.query(
+      `
+        SELECT 1 AS found
+        FROM whitelist
+        WHERE
+          (expiry is NULL OR expiry >= NOW())
+          AND :email LIKE CONCAT('%', email)
+        LIMIT 1
+      `,
+      {
+        replacements: { email: normalizedEmail },
+        type: QueryTypes.SELECT,
+      }
+    )) as { found: 1 }[]
+
+    return records.length >= 1
   }
 
   async sendEmailOtp(email: string) {
--- a/src/services/identity/UsersService.ts
+++ b/src/services/identity/UsersService.ts
@@ -159,7 +159,7 @@ class UsersService {
     // raw query because ORMs suck!
     const records = (await this.sequelize.query(
       `
-        SELECT 1 AS found
+        SELECT email AS found
         FROM whitelist
         WHERE
           (expiry is NULL OR expiry >= NOW())
@@ -170,9 +170,21 @@ class UsersService {
         replacements: { email: normalizedEmail },
         type: QueryTypes.SELECT,
       }
-    )) as { found: 1 }[]
+    )) as { email: string }[]
 
-    return records.length >= 1
+    if (records.length >= 1) {
+      logger.info({
+        message: "Email valid for OTP by whitelist",
+        meta: {
+          email,
+          whitelistEntry: records[0].email,
+        },
+      })
+
+      return true
+    }
+
+    return false
   }
 
   async sendEmailOtp(email: string) {
--- a/src/services/identity/__tests__/UsersService.spec.ts
+++ b/src/services/identity/__tests__/UsersService.spec.ts
@@ -32,6 +32,7 @@ const MockRepository = {
 
 const MockSequelize = {
   transaction: jest.fn((closure) => closure("transaction")),
+  query: jest.fn(),
 }
 
 const MockWhitelist = {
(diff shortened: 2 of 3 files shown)


src/services/identity/UsersService.ts
```diff
@@ -155,26 +155,46 @@ class UsersService {
...
+    // - if whitelist entry is a domain (no @, or starting with @), then do suffix match
+    // Limit 1 is added to allow the query to exit early on first match
+    const records = (await this.sequelize.query(
+      `
+        SELECT email
+        FROM whitelist
+        WHERE
+          (expiry is NULL OR expiry >= NOW())
+          AND
+          CASE WHEN email ~ '^.+@'
```
timotheeg, line 173: Note: compared to the previous code, this is a simpler regexp. It detects the "full email", while the previous more complicated regexp was detecting the domains. Obviously the match strategies have been swapped accordingly.

src/services/identity/__tests__/UsersService.spec.ts
```diff
@@ -120,71 +121,31 @@ describe("User Service", () => {
...
-        email: "foo@accenture.com",
-      },
-    ]
-    MockWhitelist.findAll.mockResolvedValueOnce(mockWhitelistEntries)
+    it("should return false when the db query returns no record", async () => {
+      // Arrange
+      const expected = false
+      // NOTE: This ends with gov.sg not .gov.sg (lacks a dot after the @)
+      const emailWithoutDot = "user@gov.sg"
+      MockSequelize.query.mockResolvedValueOnce([])
```
alexanderleegs, line 142: I think we should still be adding the mock for `[{ email: ".gov.sg" }]` - our test suite clears the db state between tests, so this test is checking for `user@gov.sg` against an empty whitelist

src/services/identity/UsersService.ts
```diff
@@ -1,5 +1,5 @@
 import { ResultAsync, errAsync, okAsync } from "neverthrow"
```
harishv7, line 1: Could we share the before and after trace in the pr description to see the difference and also maybe if the latency improvements

src/services/identity/UsersService.ts
```diff
@@ -155,26 +155,46 @@ class UsersService {
 
   async canSendEmailOtp(email: string) {
     const normalizedEmail = email.toLowerCase()
-    const whitelistEntries = await this.whitelist.findAll({
-      attributes: ["email"],
-      where: {
-        expiry: {
-          [Op.or]: [{ [Op.is]: null }, { [Op.gt]: new Date() }],
+
+    // raw query because ORMs suck!
```
harishv7, line 159: Just curious we used raw query here because the ORM way was more complicated or?

src/services/identity/__tests__/UsersService.spec.ts
timotheeg: Oh wait what? the tests can populate the DB? I would love indeed  to run the query for real rather than arbitrarily mocking a db response!

alexanderleegs: Whoops sorry i meant the jest mocks are reset! So we should be mocking the return of query here as `[{ email: ".gov.sg" }] `

That said we do have population of DB for some of our tests in `integration` - we don't do this for the tests in each service which we consider unit tests

The `/integration/Users.spec.ts` test suite does cover some whitelist tests though (under the `/email/otp` suite), so if we want to test this against the db we could add this test there as well!

src/services/identity/UsersService.ts
timotheeg: > Could we share the before and after trace in the pr description

The after traces are already there, I'll add a before trace as well.


> if the latency improvements

Aye, so on this a few things:
* The current query currently runs under 2ms most of the time, so the optimization isn't all that interesting from a latency perspective to be honest, and we are not expecting a big performance boost or anything
* A dev pattern of downloading the full table content is bad generally, so the PR serves as illustration of how more complex query that are not necessarily supported by the ORM. The ORM is a tool, it should not be a blocker to run inefficient queries.
* I already added in the IsomerCMS dashboard a block to monitor the performance of the old query versus the new query for comparison [here](https://ogp.datadoghq.com/dashboard/8tj-6k5-p2v?fromUser=false&refresh_mode=paused&view=spans&from_ts=1713230653828&to_ts=1713245053828&live=false&tile_focus=7015368340178734)

src/services/identity/UsersService.ts
timotheeg: Well, so this query has 2 query patterns which are "unusual" for typical ORM use-cases:
1. a `CASE WHEN` in the `WHERE` clause
2. a where condition of the form `input like cell` (with a `concat()` call thrown in), where most of the times, queries are the other way around `cell like input`

I wasn't sure if sequelize could even do 2), but even if it could, I reckon the sequelize syntax would make the query very hard to read the query (likely mixing object and literal syntax in one statement ?)

In addition to that, this query does not return entities as instance of models or fetches dependencies, we just want 1 record of one email, if any theinput matches any, so the ORM brings no additional value here (minus maybe one level of indirection for the table name and field name).

There's also a bit of personal preference here too I suppose. Personally I think 
```
(expiry is NULL OR expiry > NOW())
```
is much easier to read than:
```
[Op.or]: [{ [Op.is]: null }, { [Op.gt]: new Date() }],
```
and that's just a small part of the statement in this case.

So I was keen to show that it's OK to have efficient readable raw queries in app code. ORM is just a tool, and shold not get in the way of good query patterns.

But so with all that said, If you can write the query in sequelize syntax, and it reads well, I don't mind pushing that in at all. It's your code base after all! 😄

src/services/identity/__tests__/UsersService.spec.ts
timotheeg: Ah I see, I'll take a look at the integration tests (I could probably use some help for that 😅)

But so specifically in this case you are commenting on here, the mock does what we expect. For reference, the new query no longer return the full content of the table, it returns either:
- no match if the input matches none of the whitelist entries
OR
- the first whitelist entry the input matches

These 2 distinct response types represent a boolean-ish yes/no state of whether the input is whitelisted.

And in this case, the input `user@gov.sg` is not expecting to match `.gov.sg` as a whitelist entry, so that cannot be in the mocked response.

alexanderleegs: Ah you're right, I got the intent of the test wrong sorry! But yeah i think we can then introduce an integration test with this scenario (db contains `.gov.sg`, user tried to login via `blah@gov.sg` which should not match)

src/services/identity/UsersService.ts
harishv7: No worries Tim, agreed on the points you mentioned. I had some thoughts about mixing raw queries and ORMs, but understandability of code is more important than just have "standardised" code

src/services/identity/UsersService.ts
timotheeg: > Could we share the before and after trace in the pr description

Done! 😄  I added a sample trace for the before in the PR description and I updated the performance section too. 

So again, in the grand scheme of things, this is a very cheap query so this PR is not bringing visible impact to users. It's more to illustrate how to query the DB with good patterns :) 

Also Just for reference, here is a screenshot on the variability of max run time of the current query in prod. We'll be comparing the performance of the new query against that ([link to widget here](https://ogp.datadoghq.com/dashboard/8tj-6k5-p2v/isomer-cm-s?fromUser=false&refresh_mode=paused&view=spans&from_ts=1713089089262&to_ts=1713261889262&live=false&tile_focus=3522012591074081))

image

src/services/identity/__tests__/UsersService.spec.ts
timotheeg: I've updated the comments to make it clearer why a raw query is used: https://github.com/isomerpages/isomercms-backend/pull/1301/commits/db15356463310f164287a44132d39c9eca2bd888

src/services/identity/UsersService.ts
timotheeg: I've updated the comment to better explain why the raw query was used: https://github.com/isomerpages/isomercms-backend/pull/1301/commits/db15356463310f164287a44132d39c9eca2bd888

harishv7 (approved):

harishv7:
@timotheeg ckuld we add tests in the PR description so that we can verify this flow during release by testing on staging

timotheeg:
Code is released, I'll be adding some info here to monitor results. I take screenshots because these spans will soon disappear from DD:

For now, p99 and max of the old request for the past ~2 weeks:

## old query
max ([link](https://ogp.datadoghq.com/apm/traces?query=%40_top_level%3A1%20service%3Aisomer-postgres%20env%3Aproduction%20%40db.statement%3A%22SELECT%20email%20FROM%20whitelist%20WHERE%20%28%20Whitelist%20.%20expiry%20IS%20%3F%20OR%20Whitelist%20.%20expiry%20%3E%20%3F%20%29%22&agg_m=%40duration&agg_m_source=base&agg_q=&agg_t=max&cols=core_service%2Ccore_resource_name%2Clog_duration%2Clog_http.method%2Clog_http.status_code&fromUser=true&graphType=flamegraph&historicalData=true&messageDisplay=inline&query_translation_version=v0&shouldShowLegend=true&sort=time&spanType=service-entry&step=1200000&top_n=10&traceQuery=&view=spans&viz=timeseries&start=1712199228000&end=1713333945124&paused=true))
image

p99 ([link](https://ogp.datadoghq.com/apm/traces?query=%40_top_level%3A1%20service%3Aisomer-postgres%20env%3Aproduction%20%40db.statement%3A%22SELECT%20email%20FROM%20whitelist%20WHERE%20%28%20Whitelist%20.%20expiry%20IS%20%3F%20OR%20Whitelist%20.%20expiry%20%3E%20%3F%20%29%22&agg_m=%40duration&agg_m_source=base&agg_q=&agg_t=pc99&cols=core_service%2Ccore_resource_name%2Clog_duration%2Clog_http.method%2Clog_http.status_code&fromUser=true&graphType=flamegraph&historicalData=true&messageDisplay=inline&query_translation_version=v0&shouldShowLegend=true&sort=time&spanType=service-entry&step=1200000&top_n=10&traceQuery=&view=spans&viz=timeseries&start=1712199228000&end=1713333945124&paused=true))
image

## new query, we need to wait for a week or to aggregate more data
max ([link](https://ogp.datadoghq.com/apm/traces?query=%40_top_level%3A1%20service%3Aisomer-postgres%20env%3Aproduction%20%40db.statement%3A%22SELECT%20email%20FROM%20whitelist%20WHERE%20%28%20expiry%20is%20%3F%20OR%20expiry%20%3E%20NOW%20%28%20%29%20%29%20AND%20CASE%20WHEN%20email%20~%20%3F%20THEN%20email%20%3D%20%3F%20ELSE%20%3F%20LIKE%20CONCAT%20%28%20%3F%20email%20%29%20END%20LIMIT%20%3F%22&agg_m=%40duration&agg_m_source=base&agg_t=max&cols=core_service%2Ccore_resource_name%2Clog_duration%2Clog_http.method%2Clog_http.status_code&fromUser=false&graphType=flamegraph&historicalData=true&messageDisplay=inline&query_translation_version=v0&shouldShowLegend=true&sort=time&spanType=service-entry&step=1200000&traceQuery=&view=spans&viz=timeseries&start=1712198990948&end=1713408590948&paused=false))
image

p99 ([link](https://ogp.datadoghq.com/apm/traces?query=%40_top_level%3A1%20service%3Aisomer-postgres%20env%3Aproduction%20%40db.statement%3A%22SELECT%20email%20FROM%20whitelist%20WHERE%20%28%20expiry%20is%20%3F%20OR%20expiry%20%3E%20NOW%20%28%20%29%20%29%20AND%20CASE%20WHEN%20email%20~%20%3F%20THEN%20email%20%3D%20%3F%20ELSE%20%3F%20LIKE%20CONCAT%20%28%20%3F%20email%20%29%20END%20LIMIT%20%3F%22&agg_m=%40duration&agg_m_source=base&agg_t=pc99&cols=core_service%2Ccore_resource_name%2Clog_duration%2Clog_http.method%2Clog_http.status_code&fromUser=false&graphType=flamegraph&historicalData=true&messageDisplay=inline&query_translation_version=v0&shouldShowLegend=true&sort=time&spanType=service-entry&step=1200000&traceQuery=&view=spans&viz=timeseries&start=1712198990948&end=1713408590948&paused=false))
image


TODO: take screenshot for the graphs for the new query in 2 weeks.

Notes:
As mentioned in the commen threads, this optimization is not a big gain from the user's perspective (most of the times both old and new query runs under 2ms!), but fetching a whole table to do filtering in code is not a good pattern, so the PR illustrate how the DB filtering capabilities can be better utilized.

PR #1311, opened 2024-04-16
ISOM-942: add user info to traces

timotheeg:
## Problem

We do not have the user information in traces. Since DataDog offers [an API](https://github.com/DataDog/dd-trace-js/blob/master/docs/API.md#user-identification) to track the user, we should use the capability.

## Solution

Use the [dd-trace setUser() API ](https://github.com/DataDog/dd-trace-js/blob/master/docs/API.md#user-identification) to populate user id in traces.

This is done in a middleware that will run right after the session middleware, so we can report whatever user is in session.

### Question

Should we track the user email and github id in traces as well? (basically the entire content of the session's UserInfo object ?) Working with emails make its faster for us to identify people, but is that a privacy concern? (potentially not since we also have a lot of emails in logs?)

### Results

Verified working on staging:
[Sample trace here](https://ogp.datadoghq.com/apm/trace/6897147640340843683?colorBy=service&env=staging&graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=7613602567414186727&spanViewType=metadata&timeHint=1713260101003.0051)
image

Traces and spans can be filtered by show the user id and session id:
[Sample search here](https://ogp.datadoghq.com/apm/traces?query=%40_top_level%3A1%20env%3Astaging%20service%3Aisomer%20%40usr.id%3A119&cols=service%2Cresource_name%2C%40duration%2C%40http.method%2C%40http.status_code%2C%40usr.id%2C%40usr.session_id&fromUser=false&graphType=flamegraph&historicalData=false&messageDisplay=inline&query_translation_version=v0&shouldShowLegend=true&sort=time&spanType=service-entry&traceQuery=&view=spans&start=1713259340314&end=1713260240314&paused=false)
image

### DIFF
--- a/src/middleware/traceUser.ts
+++ b/src/middleware/traceUser.ts
@@ -14,7 +14,8 @@ function isUserSession(session: Session): session is UserSession {
     "userInfo" in session &&
     typeof session.userInfo === "object" &&
     session.userInfo &&
-    "isomerUserId" in session.userInfo
+    "isomerUserId" in session.userInfo &&
+    "email" in session.userInfo
   )
 }
 
--- a/src/middleware/traceUser.ts
+++ b/src/middleware/traceUser.ts
@@ -1,0 +1,21 @@
+import tracer from "@utils/tracer"
+
+import { RequestHandlerWithGrowthbook } from "@root/types"
+
+// eslint-disable-next-line import/prefer-default-export
+export const traceUserMiddleware: RequestHandlerWithGrowthbook<
+  never,
+  unknown,
+  unknown,
+  never
+> = async (req, res, next) => {
+  if (req.session?.userInfo?.isomerUserId) {
+    tracer.setUser({
+      id: req.session.userInfo.isomerUserId,
+      session_id: req.session.id,
+
+      // Could consider adding the entire userInfo object if we are comfortable having that data in DD traces
+    })
+  }
+  next()
+}
--- a/src/middleware/traceUser.ts
+++ b/src/middleware/traceUser.ts
@@ -1,15 +1,30 @@
+import { NextFunction, Request, Response } from "express"
+import { Session } from "express-session"
+
 import tracer from "@utils/tracer"
 
-import { RequestHandlerWithGrowthbook } from "@root/types"
+import { SessionData } from "@root/types/express/session"
+
+type UserSession = Session & SessionData
+
+function isUserSession(session: Session): session is UserSession {
+  return !!(
+    session &&
+    "id" in session &&
+    "userInfo" in session &&
+    typeof session.userInfo === "object" &&
+    session.userInfo &&
+    "isomerUserId" in session.userInfo
+  )
+}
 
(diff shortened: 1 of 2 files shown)


src/middleware/traceUser.ts
```diff
@@ -0,0 +1,36 @@
...
+function isUserSession(session: Session): session is UserSession {
+  return !!(
+    session &&
+    "id" in session &&
+    "userInfo" in session &&
+    typeof session.userInfo === "object" &&
+    session.userInfo &&
+    "isomerUserId" in session.userInfo
+  )
+}
```
timotheeg, line 19: Typescript is a pain in my ass 😢 

It feels way overkill to do this, but I couldn't get the types to work otherwise 😢

src/middleware/traceUser.ts
```diff
@@ -0,0 +1,36 @@
+import { NextFunction, Request, Response } from "express"
+import { Session } from "express-session"
+
+import tracer from "@utils/tracer"
+
+import { SessionData } from "@root/types/express/session"
+
+type UserSession = Session & SessionData
```
kishore03109, line 8: seems quite similar to UserWithSiteSessionData or RequestWithSession, could we just use that type instead?

src/middleware/traceUser.ts
kishore03109: hmm wait this is not a strict validation tho right? should we just add the remaining fields? Feels like this is more risky than just doing a type assertion of just `req: Request<UserSession>` since this in first impression feels like 'correct and internal type checking'

src/middleware/traceUser.ts
timotheeg: Yeah, I was hoping to reuse some existing types too, but I wasn't satisfied with the existing types 😞
* `UserWithSiteSessionData` requires a sitename, which may not be there, and is not needed by this middleware (although I suppose if it is present, we could shove it in the trace too 🤔 )
* `RequestWithSession` extends `RequestWithGrowthBook` which also felt a bit weird wrong (not sure?) . growthbook is not needed in this middleware (at the moment anyway).

timotheeg:
Thanks for the review so far @kishore03109 ! 🙏 

Any input on the question in the PR description?

> Should we track the user email and github id in traces as well? (basically the entire content of the session's UserInfo object ?) Working with emails make its faster for us to identify people, but is that a privacy concern? (potentially not since we also have a lot of emails in logs?)

And extended to one of the other comment, if we have the `sitename`, should we shove it into the trace as well in this PR? Probably in the root span directly, rather than the user object.

kishore03109:
Should we track the user email and github id in traces as well? (basically the entire content of the session's UserInfo object ?) Working with emails make its faster for us to identify people, but is that a privacy concern? (potentially not since we also have a lot of emails in logs?)

> hmm we dont seem to regularly log emails, prefer to go with the safer approach here. We iterate from here ah

PR #1332, opened 2024-04-25
ISOM-1006: Optimize verifyASiteMember()

timotheeg:
## Problem

`verifySiteMember()` is a potential cause for the DB load we see. It is very costly for what it does, and we run it all the time as part of the auth flow.

verifySiteMember() is a potential cause for the DB load we see

Some flaws exhibited
* Multiple consecutive query on the isomeradmin tables for the same users
* Fetching more data than needed when we actually care for a `exists()` boolean output
* `getRole()` issues an extremely costly query, involving multiple full table inner joins and json processing. 

[Sample trace](https://ogp.datadoghq.com/apm/trace/839518262592971788?colorBy=service&env=production&graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=5077096174128205426&spanViewType=metadata&timeHint=1714009172442&view=spans)
image

The 3 queries are the most costly operations IsomerCMS runs, both in term of number of queries and in term of time spent:
image

## Solution

This PR will be fully backward compatible but aims to reduce the cost of `verifySiteMember()`.

Expected outcomes:
* No more duplicate queries for the same user
* faster middleware processing for `verifySiteMember()`
* Reduced DB CPU load

## Reference:
* [Linear ticket](https://linear.app/ogp/issue/ISOM-1006/optimize-verifysitemember)


timotheeg:
I tried 2 possible query optimization (see [this commit ](https://github.com/isomerpages/isomercms-backend/pull/1332/commits/0fd73a36de871f324bec07c253ca4db4c30578a1)for details)

Results are as below in [this sample trace](https://ogp.datadoghq.com/apm/trace/5500350286498939905?colorBy=service&env=staging&graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=307981445858476553&spanViewType=metadata&timeHint=1714023027531.003)
image

original: 15.8ms
optimization option 1: 3.49ms
optimization option 2: 2.07ms

Optimization 1 retains the check for soft deletion of `users`, `repos`, and `sites`, while optimization 2 makes an efficient query on just `repos` and `site_members`, and assumes both the site and the user are not soft deleted (which may or may not cause trouble).

Optimization 1 is still 4x more performant than the original, so we could  start by releasing that and we can keep optimization 2 as a potential improvement for the future.

@harishv7 any comment on the continuous check for soft deletions in every query?

timotheeg:
Query plan reports from postgres on the 3 queries:

original:
image

optimization 1:
image

optimization 2:
image

timotheeg:
Also for reference, below is an example of the actual query sequelize sent to the DB (on staging)
(It's easier to read than the normalized version in the DD traces)

```sql
SELECT 
    "Site".*,
    "site_members"."id" AS "site_members.id",
    "site_members"."email" AS "site_members.email",
    "site_members"."github_id" AS "site_members.githubId",
    "site_members"."contact_number" AS "site_members.contactNumber",
    "site_members"."last_logged_in" AS "site_members.lastLoggedIn",
    "site_members"."created_at" AS "site_members.createdAt",
    "site_members"."updated_at" AS "site_members.updatedAt",
    "site_members"."deleted_at" AS "site_members.deletedAt",
    "site_members->SiteMember"."id" AS "site_members.SiteMember.id",
    "site_members->SiteMember"."user_id" AS "site_members.SiteMember.userId",
    "site_members->SiteMember"."site_id" AS "site_members.SiteMember.siteId",
    "site_members->SiteMember"."role" AS "site_members.SiteMember.role",
    "site_members->SiteMember"."created_at" AS "site_members.SiteMember.createdAt",
    "site_members->SiteMember"."updated_at" AS "site_members.SiteMember.updatedAt"
FROM
(
    SELECT
        "Site"."id",
        "Site"."name",
        "Site"."site_status" AS "siteStatus",
        "Site"."job_status" AS "jobStatus",
        "Site"."is_private" AS "isPrivate",
        "Site"."created_at" AS "createdAt",
        "Site"."updated_at" AS "updatedAt",
        "Site"."deleted_at" AS "deletedAt",
        "Site"."creator_id" AS "creatorId",
        "repo"."id" AS "repo.id",
        "repo"."name" AS "repo.name",
        "repo"."url" AS "repo.url",
        "repo"."created_at" AS "repo.createdAt",
        "repo"."updated_at" AS "repo.updatedAt",
        "repo"."deleted_at" AS "repo.deletedAt",
        "repo"."site_id" AS "repo.siteId"
    
    FROM
        "sites" AS "Site" INNER JOIN "repos" AS "repo"
            ON "Site"."id" = "repo"."site_id" AND ("repo"."deleted_at" IS NULL AND "repo"."name" = 'kishore-test')
            
    WHERE
        ("Site"."deleted_at" IS NULL) 
        AND ( 
            SELECT
                "SiteMember"."id"

            FROM "site_members" AS "SiteMember" INNER JOIN "users" AS "User"
                ON "SiteMember"."user_id" = "User"."id" AND ("User"."deleted_at" IS NULL AND "User"."id" = '119')

            WHERE ("Site"."id" = "SiteMember"."site_id")
            LIMIT 1
        ) IS NOT NULL
    LIMIT 1

)
AS "Site" INNER JOIN
(
    "site_members" AS "site_members->SiteMember" INNER JOIN "users" AS "site_members"
        ON "site_members"."id" = "site_members->SiteMember"."user_id"
)
ON "Site"."id" = "site_members->SiteMember"."site_id" AND ("site_members"."deleted_at" IS NULL AND "site_members"."id" = '119');
```

src/services/identity/CollaboratorsService.ts
```diff
@@ -225,30 +231,51 @@ class CollaboratorsService {
...
+    //     r.site_id = sm.site_id
+    //     AND r.name = :siteName
+    //     AND sm.user_id = :userId
+    //     AND r.deleted_at IS NULL
+
+    const records = (await this.sequelize.query(
+      `
+        SELECT sm.role
+        FROM
+            site_members sm,
```
kishore03109, line 250: nit: not a fan of these renames rather keep them as is for readability

src/services/identity/CollaboratorsService.ts
```diff
@@ -225,30 +231,51 @@ class CollaboratorsService {
...
+            AND u.deleted_at IS NULL
+      `,
+      {
+        replacements: { siteName, userId },
+        type: QueryTypes.SELECT,
+      }
+    )) as { role: CollaboratorRoles }[]
+
+    if (records.length > 0) {
+      return records[0].role
```
kishore03109, line 271: for clarification: when does this occur ah? by right this should never be the case right?

timotheeg: Hmm? No, it's possible there's no record for a particular user in a particular site, right? 

Exactly in the same sense the previous code was doing this:

```typescript
    return site?.site_members?.[0]?.SiteMember?.role ?? isomerAdminRole
```

II suppose specifically: if an isomer admin visits a site they are not an official collaborator of, there'd be no record in the DB, but `getRole()` will still return `CollaboratorRoles.IsomerAdmin`.

Also if the repo/site/user are soft-deleted, the query will also return an empty set (just like in the previous code)

timotheeg: Basically, this is to block access for folks trying to crawl resource they are not supposed to have access to. I mean, it's what `verifySiteMember()` is there to do, right? 😅

kishore03109: hmm for clarity records.length should by right = 1 or 0 right? we are just making a statement > 0 defensively correct?

timotheeg: Yes, correct!

timotheeg: Ah, so you were commenting on the `> 0` ? I thought you were asking about the conditional check itself 😅 

I can make this more explicit with a `LIMIT 1` and doing an explicit check for length 1

src/services/identity/CollaboratorsService.ts
timotheeg: Sure, I can drop the aliases 🤷 . For short queries, I think readability is still OK, but it's your codebase! 😄

kishore03109 (approved):
nit: can either limit 1 or can just add a comment stating that the if statement is done defensively, practically we only expect to see one or 0 elements in the array

src/services/identity/CollaboratorsService.ts
timotheeg: Just to drop it here in fact. Personally I find that when the table name is longer than the field names, the query reads worse 😰

timotheeg: Done! Dropped the short aliases: https://github.com/isomerpages/isomercms-backend/pull/1332/commits/43621ae20891d872630d06aa08138699b9e28ac1

timotheeg:
Hey @kishore03109 btw, I'm a litttle confused with something: the original query had a check for `site_members.deleted_at`, but I do not see that field in prod DB.. How can this be? 🤔 

Oh I also note from looking at the query plans that we are missing an index for `repos.name`. I opened a ticket to add it.

timotheeg:
I updated the IsomerCMS dashboard graph to show the DB CPU  load this week versus last week, so we can see at glance whether we're doing better or worse.

[Direct link to widget here](https://ogp.datadoghq.com/dashboard/8tj-6k5-p2v/isomer-cm-s?fromUser=true&refresh_mode=paused&view=spans&from_ts=1713925271268&to_ts=1714098071268&live=false&tile_focus=5042119162208894)

image

PR #1334, opened 2024-04-25
feat: optimise local diff

jacobkwan:
## Problem
Closes https://linear.app/ogp/issue/ISOM-833/benchmark-impact-of-local-git-diff-on-cpu-usage
Local diff was implemented a while back but kept behind a feature flag.
Feature flag has not been enabled in production due to concerns about performance especially for larger diffs.

### Understanding bottlenecks via Datadog
**Everything below is in the context of motac2 with ~444 files changed in one RR**
[Trace for unoptimised implementation](https://ogp.datadoghq.com/apm/traces?query=%40_top_level%3A1%20env%3Astaging%20service%3A%2Aisomer%2A%20resource_name%3A%2Acompare%2A&agg_m=count&agg_m_source=base&agg_t=count&cols=core_service%2Ccore_resource_name%2Clog_duration%2Clog_http.method%2Clog_http.status_code&fromUser=false&graphType=flamegraph&historicalData=false&messageDisplay=inline&query_translation_version=v0&shouldShowLegend=true&sort=time&spanID=3675313136803532925&spanType=service-entry&spanViewType=errors&timeHint=1714031343944&trace=90028220631642002043675313136803532925&traceID=9002822063164200204&traceQuery=&view=spans&start=1714051891906&end=1714052791906&paused=false)
Significant bottlenecks (in order of priority):
1. `getFilesChanged` 27s / 40s or ~66%
2. calling `getLatestCommitOfPath` for every single file changed
    - called in parallel but still each call is about 10s / 40s or 25%
3. querying `Users` for every commit to get `email` (@timotheeg had already fixed this for the GitHub diff flow), actually doesn't add much to latency

### Optimisations
1. Refactor `getFilesChanged` to call `git diff-tree -r --name-only master..staging`
    - this change alone cuts the 27s to ~70ms
2. Refactor the entire flow. Like @timotheeg  mentioned, writing "clean" and "elegant" code in the original implementation - mapping over the filenames and performing the operations on that individual level - is not always the best when it comes to efficiency. Additionally, there is extra network cost from interacting with EFS many times.
    - instead `getCommitsBetweenMasterAndStaging` once
    - loop over these commits, constructing a `filenameToLogMap`. At the same time, get unique `userIds` from commit messages to dedupe queries
    - query DB for the unique users and construct a `userIdToUserMap`
    - finally use the two maps to construct the full response for each filename - minimal changes here just that it no longer needs to query DB and get latest commit from EFS for every single filename, as the two data structures are sufficient.

### Final flame graph after optimisations:
Bottleneck is now the many parallel reads on EFS. Out of scope for this PR and its also current behaviour for GitHub flow.
TLDR: reads are needed to get the permalink from frontmatter, which is then used to construct the `stagingUrl`

https://github.com/isomerpages/isomercms-backend/assets/77665182/44d530aa-da0c-4a94-88a1-b24978d13cdf

**Breaking Changes**

  - Details ...

## Tests
Copied over tests from https://github.com/isomerpages/isomercms-backend/pull/1172
- [ ] Toggle feature flag `is_local_diff_enabled` for account being tested https://app.growthbook.io/features/is_local_diff_enabled
- [ ] Create a page
- [ ] Edit a page
- [ ] Delete a page
- [ ] Move a page
- [ ] Try to create an RR and check that the diff shown is correct for all files (this tests that /compare works)
- [ ] After creating the RR, navigate to the RR page and check that the diff shown is correct (this tests that /listFullReviewRequest works)
- [ ] Approve and publish the RR
- [ ] Check that diff is empty now (tests that master has been fastforwarded)
- [ ] Finally, last sanity check that the diff still works after making another edit

### DIFF
--- a/src/services/db/GitFileSystemService.ts
+++ b/src/services/db/GitFileSystemService.ts
@@ -385,7 +385,7 @@ export default class GitFileSystemService {
   }
 
   /**
-   * Wrapper over `git diff --name-only` that also creates `master` branch if it does not exist.
+   * Wrapper over `git diff-tree -r --name-only master..staging` that also creates `master` branch if it does not exist.
    */
   getFilesChanged(repoName: string): ResultAsync<string[], GitFileSystemError> {
     return this.createLocalTrackingBranchIfNotExists(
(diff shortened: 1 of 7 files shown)


src/services/db/GitFileSystemService.ts
```diff
@@ -396,7 +396,7 @@ export default class GitFileSystemService {
       return ResultAsync.fromPromise(
         this.git
           .cwd({ path: `${efsVolPath}/${repoName}`, root: false })
-          .diff(["master..staging", "--name-only"]),
+          .raw(["diff-tree", "-r", "--name-only", "master..staging"]),
```
kishore03109, line 399: hmm what is the `-r` for ah? is this added defensively?

kishore03109, line 399: hmm what is the -r for ah? is this added defensively?

src/services/db/__tests__/GitFileSystemService.spec.ts
```diff
@@ -777,49 +777,6 @@ describe("GitFileSystemService", () => {
     })
   })
 
-  describe("getLatestLocalCommitOfPath", () => {
```
kishore03109, line 780: hmm should add a `getCommitsBetweenMasterAndStaging` test case hor

src/services/review/ReviewRequestService.ts
```diff
@@ -150,15 +151,65 @@ export default class ReviewRequestService {
...
-            userWithSiteSessionData,
-            stagingLink
-          )
-        )
-        return ResultAsync.combine(editMetadata)
+        if (filenames.length === 0) {
+          return okAsync([])
+        }
+        const filenamesSet = new Set(filenames)
+        return this.apiService
```
kishore03109, line 158: hmm is this needed? we have a set and a map, but we could use just a map insead here right?

kishore03109 (approved):

PR #1337, opened 2024-04-26
feat: disable issues on new repositories

qin-guan:
## Problem

Issues on public repositories have been recently used to distribute malware which look like they're from official sources.

Discussion on Slack (internal) : https://opengovproducts.slack.com/archives/C055FUACLJU/p1714113816195799

## Solution

For the teams consideration to disable issues for repos on new Isomer sites, though it is still possible to upload files using Pull Requests 😔


kishore03109 (approved):
am ok with this, dont recall this feature being used in our operations anyways
thank you for raising this pr @qin-guan! :)

PR #1351, opened 2024-04-29
Fix/whitespace and new line chars

alexanderleegs:
This PR fixes issues we have with trailing whitespace and unicode line breaks. See [this issue](https://linear.app/ogp/issue/ISOM-929/tighten-validations-for-subfolder-names-not-to-have-special-characters), [this issue](https://linear.app/ogp/issue/ISOM-928/tighten-validations-for-urls-in-components) and [this issue](https://linear.app/ogp/issue/ISOM-1013/cms-should-validate-trailing-spaces) for more details.

Concretely, this PR creates a new util method to trim whitespace and replace unicode new line characters with spaces, which is used to parse user provided input for all page/folder/settings create/update endpoints. Note that this means existing folders with a trailing space continue to have their new pages affected until the folder is renamed - this is expected and should be resolved manually as they come up.

## Tests

- [ ] Copy a unicode character for new line (e.g. from [here](https://unicode-explorer.com/c/2028))
- [ ] Create a new folder with the unicode character in it + trailing and preceding spaces
- [ ] Verify that the created folder does not have the unicode character or trailing spaces
- [ ] Repeat for pages, media, and special pages e.g. homepage


src/utils/yaml-utils.ts
```diff
@@ -2,6 +2,26 @@ import yaml from "yaml"
 
 import { sanitizer } from "@services/utilServices/Sanitizer"
 
+const LINE_BREAK_REGEX = "[\r\n\x0B\x0C\u0085\u2028\u2029]"
```
kishore03109, line 5: rather than trying to reimplement white space character check, i suggest only caring about new line characters and let `String.prototype.trim()` handle spaces. 

Currently, in the attempt of implementing both, the check for spaces is not as [complete](https://github.com/v8/v8/blob/7f55dc42841c93dbf095995ffd2f6a7ca41bf51e/src/builtins/string-trim.tq#L151) as the one implemented for the function call trim. 

Do we care about any space here? ie ` hi there ` should be `hithere`?

kishore03109 (commented):
implementation of `recursiveTrimAndReplaceLineBreaks` is incomplete, suggest to use something like
```
const newLineRegex = /\r?\n|\r/
"hi \n there \x0C \x0D ".split(newLineRegex).map((str) => str.trim()).join("") // returns 'hithere'
```

src/utils/yaml-utils.ts
alexanderleegs: Hmm the new line regex is really just to catch the new line characters, and not to replace other whitespace - I could replace it with `[\r\n\u2028\u2029]` instead. The in between spaces should be preserved

src/utils/yaml-utils.ts
```diff
@@ -2,6 +2,26 @@ import yaml from "yaml"
...
+
+export const recursiveTrimAndReplaceLineBreaks = (obj: any): any => {
+  if (typeof obj === "string") {
+    return obj.replaceAll(GLOBAL_LINE_BREAK_SEARCH, " ").trim()
+  }
+  if (Array.isArray(obj)) {
+    return obj.map((item) => recursiveTrimAndReplaceLineBreaks(item))
+  }
+  if (typeof obj === "object" && obj !== null) {
+    const newObj: any = {}
```
kishore03109, line 16: possible to use some sort of recursive types and avoid the use of `any`?

kishore03109: eh i re-requested a review by accident, just ping me on slack once ready

alexanderleegs: Fixed in https://github.com/isomerpages/isomercms-backend/pull/1351/commits/1e212c836cfdea573a5c264df2e9f977a7332f2d!

seaerchin (approved):

PR #1355, opened 2024-04-30
fix(sitelaunch): do not throw for no dig results

kishore03109:
## Problem

When there was no records that were found, dns:node throws an error. While we did handle it, there should have been an additional return statement to exit early rather than propagating into an error state. 

## Solution

If the error is of type `ENODATA`, return early with some default values instead.

**Breaking Changes**

  - Details ...

## Tests
Locally enter these values into `support/routes/v2/formsg/index.ts`

```
formsgSiteLaunchRouter
  .digAAAADomainRecords({
    primaryDomainSource: "singaopreoae.sg",
  } as SiteLaunchResult)
  .then(console.log)
  .catch(console.error)

formsgSiteLaunchRouter
  .digCAADomainRecords({
    primaryDomainSource: "singaopreoae.sg",
  } as SiteLaunchResult)
  .then(console.log)
  .catch(console.error)

```

assert that the console log outputs 

## Deploy Notes

This fix is meant to go into prod line asap

### DIFF
--- a/support/routes/v2/formsg/formsgSiteLaunch.ts
+++ b/support/routes/v2/formsg/formsgSiteLaunch.ts
@@ -252,6 +252,7 @@ export class FormsgSiteLaunchRouter {
         logger.info(
           `Domain ${launchResult.primaryDomainSource} does not have any AAAA records.`
         )
+        return [] // no AAAA records found
       }
       logger.error(
         `Error when trying to get AAAA records for domain ${launchResult.primaryDomainSource}: ${e}`
@@ -315,6 +316,12 @@ export class FormsgSiteLaunchRouter {
         logger.info(
           `Domain ${launchResult.primaryDomainSource} does not have any CAA records.`
         )
+
+        // if no CAA records, no need to add Amazon CAA and letsencrypt.org CAA
+        return {
+          addAWSACMCertCAA: false,
+          addLetsEncryptCAA: false,
+        }
       }
       logger.error(
         `Error when trying to get CAA records for domain ${launchResult.primaryDomainSource}: ${e}`


kishore03109:
* **#1355** <a href="https://app.graphite.dev/github/pr/isomerpages/isomercms-backend/1355?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop`

<h2></h2>

support/routes/v2/formsg/formsgSiteLaunch.ts
```diff
@@ -315,6 +316,12 @@ export class FormsgSiteLaunchRouter {
         logger.info(
           `Domain ${launchResult.primaryDomainSource} does not have any CAA records.`
         )
+
+        // if no CAA records, no need to add Amazon CAA and letsencrypt.org CAA
```
kishore03109, line 320: without this return, this ended up throwing an error instead

support/routes/v2/formsg/formsgSiteLaunch.ts
```diff
@@ -252,6 +252,7 @@ export class FormsgSiteLaunchRouter {
         logger.info(
           `Domain ${launchResult.primaryDomainSource} does not have any AAAA records.`
         )
+        return [] // no AAAA records found
```
kishore03109, line 255: without this return, this ended up throwing an error instead

harishv7 (approved):

support/routes/v2/formsg/formsgSiteLaunch.ts
```diff
@@ -252,6 +252,7 @@ export class FormsgSiteLaunchRouter {
...
       }
       logger.error(
         `Error when trying to get AAAA records for domain ${launchResult.primaryDomainSource}: ${e}`
```
timotheeg, line 258: /nit since you're editing this file, you could take the opportunity to clean up the neighbouring log statements. The existing statements does not print the error (since errors are not serializable).

```suggestion
      logger.error({
        message: 'Error when trying to get AAAA records for domain',
        error: e,
        meta: {
            domain: launchResult.primaryDomainSource
        }
      })
```

Same can be done for the other error() and info() logs seen as context code in the PR. 

But yeah, so main message with enahnced logging:
- report errors as-is, let the logger figure out how to export it
- log parameters as fields rather than interpolated content in the message

kishore03109: since merged will create a separate pr ya

timotheeg: Yes Sir 👍 . Thank you 🙏 .

PR #1358, opened 2024-04-30
chore(sitelaunch): better error message

kishore03109:
## Problem
The existing statements does not print the error (since errors are not serializable) as mentioned [here](https://github.com/isomerpages/isomercms-backend/pull/1355#discussion_r1584027034)

In addition, we have not been using the child loggers as effectively as possible to provide the context needed to debug efficiently when errors arise. A obvious example is our constant noise of `Github API call made`, but we have no context surrounding that particular call being made.

To fix this issue, this pr introduces a small introduction of using child loggers, which has already been declared in our codebase by @seaerchin. A child logger is used to create a new logger instance that inherits all of its parent's metadata (e.g. logger name) and attach new metadata for all logs written by that child logger. This way we avoid having to keep repeating our context in the multiple calls that are needed. 

## Help needed, but not a blocker to this particular PR.
Moving forward, my stance is that we would want to inject the logger directly onto the child functions in different modules. 

eg in this trace in the happy path, we can see that the trace for all the calls are not instantiated with the meta that I would like it to have. 

Screenshot 2024-05-02 at 11 12 45 AM

The reason for this is because the other logs are done in seperate modules, and if we were to pass the child logger over, this means that for every call, we would need to pass in the logger with bindings as needed. But, I see how this could lead to verbose code. 

To me, it seems like there is a balance to be strived for. I personally think using child loggers just the module is not very helpful. What I found more useful is the ability to view the context around the call useful. Ie, `with what set of inputs did this module behave in the manner that we see here`. I think this can help with debugging, rather than just the module name. But I am not too sure where to draw the line between passing the context form the parent service vs depending on the child service to log the relevant meta as needed during debugging. Happy to hear your thoughts on this. Note that regardless of the direction that we chose, this pr should still go in as it is only scoped to one module + it is an incremental improvement to introducing the usage of child loggers in our codebase.  
 
## Solution

use pino to serialize the errors instead + use child loggers with relevant context so that subsequent loggers do not need to keep repeating the context over and over again.

### DIFF
--- a/support/routes/v2/formsg/formsgSiteLaunch.ts
+++ b/support/routes/v2/formsg/formsgSiteLaunch.ts
@@ -8,7 +8,7 @@ import { err, ok } from "neverthrow"
 
 import { config } from "@config/config"
 
-import logger from "@logger/logger"
+import parentLogger from "@logger/logger"
 
 import InitializationError from "@errors/InitializationError"
 
@@ -85,9 +85,15 @@ export class FormsgSiteLaunchRouter {
       },
     ]
 
-    logger.info(
-      `Launch site form submission [${submissionId}] (repoName '${repoName}', domain '${primaryDomain}') requested by <${requesterEmail}>`
-    )
+    this.siteLaunchLogger.info({
+      message: "Launch site form submission",
+      meta: {
+        submissionId,
+        repoName,
+        primaryDomain,
+        requesterEmail,
+      },
+    })
 
     // 2. Check arguments
     if (!requesterEmail) {
@@ -146,6 +152,10 @@ export class FormsgSiteLaunchRouter {
   private readonly usersService: FormsgSiteLaunchRouterProps["usersService"]
 
   private readonly infraService: FormsgSiteLaunchRouterProps["infraService"]
+
+  private readonly siteLaunchLogger = parentLogger.child({
+    module: "formsgSiteLaunch",
+  })
 
   constructor({ usersService, infraService }: FormsgSiteLaunchRouterProps) {
     this.usersService = usersService
@@ -249,14 +259,21 @@ export class FormsgSiteLaunchRouter {
       }))
     } catch (e) {
       if (isErrnoException(e) && e.code === "ENODATA") {
-        logger.info(
-          `Domain ${launchResult.primaryDomainSource} does not have any AAAA records.`
-        )
+        this.siteLaunchLogger.info({
+          message: `Domain does not have any AAAA records.`,
+          meta: {
+            domain: launchResult.primaryDomainSource,
+          },
+        })
         return [] // no AAAA records found
       }
-      logger.error(
-        `Error when trying to get AAAA records for domain ${launchResult.primaryDomainSource}: ${e}`
-      )
+      this.siteLaunchLogger.error({
+        message: "Error when trying to get AAAA records for domain",
+        error: e,
+        meta: {
+          domain: launchResult.primaryDomainSource,
+        },
+      })
       throw e
     }
   }
@@ -313,9 +330,12 @@ export class FormsgSiteLaunchRouter {
       return result
     } catch (e) {
       if (isErrnoException(e) && e.code === "ENODATA") {
-        logger.info(
-          `Domain ${launchResult.primaryDomainSource} does not have any CAA records.`
-        )
+        this.siteLaunchLogger.info({
+          message: `Domain does not have any CAA records.`,
+          meta: {
+            domain: launchResult.primaryDomainSource,
+          },
+        })
 
         // if no CAA records, no need to add Amazon CAA and letsencrypt.org CAA
         return {
@@ -323,9 +343,14 @@ export class FormsgSiteLaunchRouter {
           addLetsEncryptCAA: false,
         }
       }
-      logger.error(
-        `Error when trying to get CAA records for domain ${launchResult.primaryDomainSource}: ${e}`
-      )
+
+      this.siteLaunchLogger.error({
+        message: "Error when trying to get CAA records for domain",
+        error: e,
+        meta: {
+          domain: launchResult.primaryDomainSource,
+        },
+      })
       throw e
     }
   }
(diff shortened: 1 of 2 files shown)


kishore03109:
* **#1358** <a href="https://app.graphite.dev/github/pr/isomerpages/isomercms-backend/1358?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop`

<h2></h2>

timotheeg:
Thanks @kishore03109 🙏 

My recommendation when updating logging in a given file (as part of "clean up your neighbourhood" activities) is:
* update ALL the log call sites in the file (not just `error()` calls, `info()`/`warn()` calls as well)
* remove message interpolation in favour of data fields at all call sites
* create a [child logger](https://github.com/isomerpages/isomercms-backend/blob/develop/src/logger/logger.ts#L50-L55) for the file / module to specify the module

kishore03109:
>Thanks @kishore03109 🙏
My recommendation when updating logging in a given file (as part of "clean up your neighbourhood" activities) is:
> - update ALL the log call sites in the file (not just error() calls, info()/warn() calls as well)
> - remove message interpolation in favour of data fields at all call sites
> - create a [child logger](https://github.com/isomerpages/isomercms-backend/blob/develop/src/logger/logger.ts#L50-L55) for the file / module to specify the module

@timotheeg, apologies, have done changes to address points one and two. 

i am unclear of the third point, wondering if this pattern is used anywhere in the current codebase that i can reference the pattern from!

timotheeg:
> i am unclear of the third point, wondering if this pattern is used anywhere in the current codebase that i can reference the pattern from!

Ah yes, the intention to use child loggers was there ever since the new logger was merged, but I don’t think any module was updated to make use of it yet (@seaerchin intended to do it) Apologies for dumping that on you as a “little surprise” :sweat_smile:

So it’s a very minor stuff, but it’s something every dev should be aware of, so we can multiply the effects of doing incremental cleanup whenever we touch code.

The idea is that log entries have broad contextual info (e.g. which module is doing the logging), and very specific info for the call sites (code flows, error condition, etc.)

Rather than pollute the call sites with the broad context everywhere, child loggers are a way to encapsulate the context in a new logger instance. Like so:

```typescript
import logger as baseLogger from "@root/logger/logger"

const logger = baseLogger.child({ module: 'SiteService' })

// later in the file
logger.info({message: 'foo', /* etc... */})
```

This means that in the file, your log entries will automatically have some additional information added to them, but your call sites will be as lean as possible.

It’s the same principle as DD adding the service name and environment transparently to log entries, but in this case, the base loggers and child loggers are explicitly controlled by us.

You are able to dump put more info into the child logger when you instantiate it btw (if you so wish), and/or you can have child’s child loggers (if that is useful for whatever reason).

Note that having the “module” as broad context doesn’t necessarily have a lot of value... APM is our killer tool to know the true context of a log call, BUT APM traces expire and we might need to revisit log entries in the future (e.g. police summon). So, having the log entries self-contained with context is useful. The context is also helpful for cases where we have duplicated log message in different places of the code base (e.g. “Record updated”), which is terrible because we can’t know where a given log line is from, and so the extra context help disambiguating.

support/routes/v2/formsg/formsgSiteLaunch.ts
```diff
@@ -204,6 +214,12 @@ export class FormsgSiteLaunchRouter {
       ) || []
 
     res.sendStatus(200) // we have received the form and obtained relevant field
+    this.siteLaunchLogger.setBindings({
+      submissionId,
+      primaryDomain: Array.from(
+        new Set(formResponses.map((x) => x.primaryDomain))
+      ),
+    })
```
timotheeg, line 222: ⚠️ ⚠️ Warning: this is not a correct pattern to apply here 😰 

`FormsgSiteLaunchRouter` is a router class, and so its method `FormsgSiteLaunchRouter.handleSiteLaunchFormRequest()` may be invoked multiple times to handle multiple requests (possibly in parrallel?). We should **not** be storing request-specific data on the logger of the router instance.

The assumption that the logger will be used specifically by the call to `handleSiteLaunchResults()` that follows below does not hold. The method is async and has multiple async tasks therein, and so it is possible that the logger gets modified by a fresher request, and corrupts the logging that is done in the older request 😱 

As tempting as it was to simplify the call sites by having all required information will be logged form the context bindings, we need to ensure correctness first! In this case, I think the call site themselves probably still need to log the information themselves.

OR, you can create a new specific child logger not attached to the instance and pass it into `handleSiteLaunchResults()` below, but that feels a little dirty, since it pollutes the signature of the method 🤔

timotheeg (changes requested):
Hey Kishore, I added a comment. I might be misunderstanding, but I see a potential red flags. Please check :)

support/routes/v2/formsg/formsgSiteLaunch.ts
kishore03109: you are absolutely right here, thank you for the catch. 
i have [ccaf5b0](https://github.com/isomerpages/isomercms-backend/pull/1358/commits/ccaf5b0b0c053e0c0986a119f1c32fe7ac0228d8), seems like those only work (usefully) if they are tied within the same function or if the logger is passed down to the various function calls, which does pollute the signature of the method, which does seem too much at this point of time.

timotheeg (approved):
LGTM

support/routes/v2/formsg/formsgSiteLaunch.ts
timotheeg: Thanks Kishore, the change looks good now! 👍 

Not something to spend time on now, but there may be ways to hook into nodejs's context system to store data there. That could allow to store and access pre-configured assets (a child logger in this case) without polluting method signatures.

Looking at the source code for [dd-trace-js](https://github.com/DataDog/dd-trace-js) might be a good source of info on this, since it's basically how tracers "magically" handle contextual scopes (e.g. with `const span = tracer.scope().active()`)

Perhaps a good Learning-Friday activity 😅

PR #1365, opened 2024-05-09
ISOM-1037 - Create automation for prod ops to whitelist

harishv7:
## Problem

On-call queries have repetitive issues such as whitelisting of emails etc. which can be solved with automation.

Closes ISOM-1037

## Solution

Introduce a new bot service which serves a private org Slackbot with shortcut commands. This will empower Ops to perform these repetitive queries themselves. Going forward, any automations can be added to this service.

To whitelist emails, we can now enter `/whitelist-emails email1@xyz.com,2024-06-22 email2@abc.com,2025-01-31` and so on. Note that the exp date is in the format of YYYY-MM-DD.

**Breaking Changes**

  - Details ...

**Features**:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- Details ...

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Tests

- After deployment to staging, update the URL of the app to be the staging BE API on Slack's config page. E.g. `<staging-api>/v2/isobot/whitelist-emails`
- On Slack, find `IsoBot` app
- Click on the Messages tab and type /whitelist-email
Screenshot 2024-05-19 at 10 14 54 PM
- Enter a sample whitelist query as shown in above image as per instructions
- Check staging DB to ensure the entries are in correctly

## Deploy Notes

Note that for this to work, we need to edit the endpoint for the /whitelist-emails command on Slack's config pages. This should be updated to the CMS BE endpoint instead.

- `SLACK_SIGNING_SECRET`: required to verify the incoming message is from our Slackbot. Note that this has yet to be added on 1PW or SSM. Requires to be added before deployment to staging & prod.


support/routes/v2/index.ts
```diff
@@ -1,9 +1,12 @@
 import express from "express"
 
 import { formSgRouter } from "./formsg"
+import { isobotRouter } from "./isobot"
```
seaerchin, line 4: love the name

.platform/hooks/predeploy/06_fetch_ssm_parameters.sh
```diff
@@ -84,6 +84,7 @@ ENV_VARS=(
     "GGS_REPAIR_FORM_KEY"
     "SITE_CHECKER_FORM_KEY"
     "SITE_AUDIT_LOGS_FORM_KEY"
+    "SLACK_SIGNING_SECRET"
```
seaerchin, line 87: sanity check here - not entirely sure if this is correct, since this is for eb which has been deprecated

src/services/identity/WhitelistService.ts
```diff
@@ -0,0 +1,30 @@
...
+}
+
+class WhitelistService {
+  private readonly repository: WhitelistServiceProps["repository"]
+
+  constructor({ repository }: WhitelistServiceProps) {
+    this.repository = repository
+  }
+
+  addWhitelist = async (emails: WhitelistEntry[]) => {
```
seaerchin, line 21: just wanted to see if we should also check the structure of the email? ie, that it's a valid email or that it's a string beginning iwth `@...`

support/routes/v2/isobot/index.ts
```diff
@@ -0,0 +1,44 @@
...
+)
+
+const handleWhitelistEmails: RequestHandler<
+  {},
+  { message: string },
+  SlackPayload,
+  {},
+  {}
+> = async (req, res) => {
+  console.log(req.headers)
```
seaerchin, line 21: nit: swap to using `logger`?

support/routes/v2/isobot/index.ts
```diff
@@ -0,0 +1,44 @@
...
+  const slackTimestamp = req.headers["x-slack-request-timestamp"] as string
+  const slackSig = req.headers["x-slack-signature"] as string
+
+  if (!slackTimestamp || !slackSig)
+    return res.send({ message: "Missing header/signature" })
+
+  const isVerifiedMessage = botService.verifySignature(
```
seaerchin, line 28: nit: i think we should also limit this method to 
1. isomer prod ops
2. isomer engineers 

so that the potential for accidents are minimised. additionally, we might benefit from writing any such changes into the `isomer-alarm` channel so we get clarity on who does what

support/routes/v2/isobot/ops/botService.ts
```diff
@@ -0,0 +1,97 @@
...
+        reqBody,
+      },
+    })
+
+    if (timingSafeEqual(Buffer.from(signature), Buffer.from(computedSig))) {
+      logger.info({ message: "Signature verified" })
+      return true
+    }
+    logger.error({ message: "Signature verification failed" })
+    return false
```
seaerchin, line 66: something we could do here is return a `VerifiedSignature`, where `VerifiedSignature = Tagged<"verified">` (see [here](https://github.com/sindresorhus/type-fest/blob/main/source/opaque.d.ts)). 

this is an example of parse, don't validate (blog post [here](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/)), which encodes this idea of "validity" into our typesystem and prevents duplicate work

support/routes/v2/isobot/ops/botService.ts
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  public async whitelistEmails(payload: SlackPayload) {
+    // Sample user input:
+    // email1,expDate email2,expDate
+    // email1@xyz.com,2024-06-22 email2@abc.com,2025-01-31
```
seaerchin, line 72: this seems abit odd - why not go with csv + newline delineation? can't imagine this will be too readable on prod ops side for white-listing too

support/routes/v2/isobot/ops/botService.ts
```diff
@@ -0,0 +1,97 @@
...
+    const rawEmails = payload.text.split(" ")
+    const emails = rawEmails.map((email) => {
+      const [emailStr, expStr] = email.split(",")
+      const expDate = new Date(expStr)
+      if (expDate.toString() === "Invalid Date") {
+        logger.error({
+          message: "Invalid date format when attempting to whitelist emails",
+          meta: { expStr },
+        })
+        throw new Error(`Invalid date format: ${expStr}`)
```
seaerchin, line 82: wondering if we are throw safe here?

support/routes/v2/isobot/ops/botService.ts
```diff
@@ -0,0 +1,97 @@
...
+    // email1,expDate email2,expDate
+    // email1@xyz.com,2024-06-22 email2@abc.com,2025-01-31
+    const rawEmails = payload.text.split(" ")
+    const emails = rawEmails.map((email) => {
+      const [emailStr, expStr] = email.split(",")
+      const expDate = new Date(expStr)
+      if (expDate.toString() === "Invalid Date") {
+        logger.error({
+          message: "Invalid date format when attempting to whitelist emails",
+          meta: { expStr },
```
seaerchin, line 80: since we're the consumer of the logs, might i suggest also logging email? 

additionally, the keys here should be **unambiguous** and i suggest renaming it to `expiry`

support/routes/v2/isobot/ops/botService.ts
```diff
@@ -0,0 +1,97 @@
...
+        })
+        throw new Error(`Invalid date format: ${expStr}`)
+      }
+      // Update timing of the expiry to be 16:00:00 +00
```
seaerchin, line 84: i think this could be inferred from teh code - what we might wanna add here is why we're setting it to 16 (since we're in GMT +8, this means that it's expiring at 0000)

seaerchin: there's also abit of ambiguity here - does expiry at 2024-01-01 means 2024-01-01 0000 (as is given here) or 2024-01-01 2359? 

suspect the definition here is correct but just a clarity thing

support/routes/v2/isobot/ops/botService.ts
```diff
@@ -0,0 +1,97 @@
...
+}
+
+class BotService {
+  whitelistService: WhitelistService
+
+  constructor(whitelistService: WhitelistService) {
+    this.whitelistService = whitelistService
+  }
+
+  public verifySignature(
```
seaerchin, line 29: @spaceraccoon  - could i just trouble you to check this? not entirely sure on this and is definitely not my area of expertise HAHA

support/routes/v2/isobot/index.ts
```diff
@@ -0,0 +1,44 @@
...
+  try {
+    await botService.whitelistEmails(req.body)
+    return res.send({ message: "Emails whitelisted successfully" })
+  } catch (e) {
+    logger.error({ error: e })
+    return res.send({ message: "Failed to whitelist emails" })
+  }
+}
+
+isobotRouter.post("/whitelist-emails", handleWhitelistEmails)
```
seaerchin, line 44: wondering here if we should extract `verifySignature` since it seems like something we'll be doing repeatedly as part of verifying slack requests

support/routes/v2/isobot/ops/botService.ts
```diff
@@ -0,0 +1,97 @@
...
+    // Verify timestamp is within 5 minutes
+    const currentTime = new Date().getTime() / 1000
+    const slackTimestamp = parseInt(timestamp)
+
+    if (Math.abs(currentTime - slackTimestamp) > 300) {
+      return false
+    }
+
+    // Create HMAC with the signing secret and the timestamp
+    const signingSecret = config.get("slackbot.secret")
```
kishore03109, line 43: hmm why not use [bolt](https://github.com/slackapi/bolt-js/blob/main/src/receivers/verify-request.ts#L46)? rather than re-implement this logic?

kishore03109 (commented):
hmm is there a reason why we never use bolt ah?

support/routes/v2/isobot/ops/botService.ts
seaerchin: resolving, gonna swap in #1391

support/routes/v2/isobot/index.ts
seaerchin: resolving, done in #1391

support/routes/v2/isobot/ops/botService.ts
seaerchin: resolving, done in #1391

support/routes/v2/isobot/index.ts
seaerchin: resolving, done in #1391

.platform/hooks/predeploy/06_fetch_ssm_parameters.sh
seaerchin: removed

support/routes/v2/isobot/ops/botService.ts
seaerchin: not doing - we'll deprecate this method cos bolt does it for us

support/routes/v2/isobot/ops/botService.ts
seaerchin: keeping for now and changing later

support/routes/v2/isobot/ops/botService.ts
seaerchin: updated

seaerchin (approved):

PR #1367, opened 2024-05-10
feat: move DNS checker slackbot from lambda

dcshzj:
## Problem

The DNS checker Slackbot currently lives inside a lambda.

## Solution

**Breaking Changes**

**Features**:

- Move the lambda function into the CMS support container.
    - There's quite a bit of permutations on the possible states, so the code is a little jank. Open to suggestions on how we can improve it.

## Tests

- Use the slash command `/stagingasjklasd` inside our Slack (this is connected to our staging instance)
- Verify the output for a few domains

## Deploy Notes

*None*

### DIFF
--- a/src/constants/constants.ts
+++ b/src/constants/constants.ts
@@ -79,7 +79,8 @@ export const REDIRECTION_SERVER_IPS = [
 export const ALLOWED_DNS_ERROR_CODES = ["ENOTFOUND", "ENODATA"]
 
 export const DNS_INDIRECTION_DOMAIN = "hostedon.isomer.gov.sg"
-export const DNS_CNAME_SUFFIXES = ["cloudfront.net", "kxcdn.com"]
+export const DNS_KEYCDN_SUFFIX = "kxcdn.com"
+export const DNS_CNAME_SUFFIXES = ["cloudfront.net", DNS_KEYCDN_SUFFIX]
 export const DNS_INDIRECTION_REPO = "isomer-indirection"
 export const ISOMER_ADMIN_EMAIL = "admin@isomer.gov.sg"
 export const ISOMER_SUPPORT_EMAIL = "support@isomer.gov.sg"
--- a/src/constants/constants.ts
+++ b/src/constants/constants.ts
@@ -79,6 +79,7 @@ export const REDIRECTION_SERVER_IPS = [
 export const ALLOWED_DNS_ERROR_CODES = ["ENOTFOUND", "ENODATA"]
 
 export const DNS_INDIRECTION_DOMAIN = "hostedon.isomer.gov.sg"
+export const DNS_CNAME_SUFFIXES = ["cloudfront.net", "kxcdn.com"]
 export const DNS_INDIRECTION_REPO = "isomer-indirection"
 export const ISOMER_ADMIN_EMAIL = "admin@isomer.gov.sg"
 export const ISOMER_SUPPORT_EMAIL = "support@isomer.gov.sg"
--- a/support/routes/v2/isobot/index.ts
+++ b/support/routes/v2/isobot/index.ts
@@ -3,6 +3,7 @@ import express, { RequestHandler } from "express"
 import { Whitelist } from "@database/models"
 import logger from "@root/logger/logger"
 import WhitelistService from "@root/services/identity/WhitelistService"
+import type { DnsCheckerResponse } from "@root/types/dnsChecker"
 
 import BotService, { SlackPayload } from "./ops/botService"
 
@@ -40,4 +41,39 @@ const handleWhitelistEmails: RequestHandler<
   }
 }
 
+const handleDnsChecker: RequestHandler<
+  never,
+  DnsCheckerResponse | { message: string },
+  SlackPayload,
+  unknown,
+  never
+> = async (req, res) => {
+  const slackTimestamp = req.headers["x-slack-request-timestamp"] as string
+  const slackSig = req.headers["x-slack-signature"] as string
+
+  if (!slackTimestamp || !slackSig)
+    return res.status(400).send({ message: "Missing header/signature" })
+
+  const isVerifiedMessage = botService.verifySignature(
+    slackSig,
+    slackTimestamp,
+    req.body
+  )
+  if (!isVerifiedMessage)
+    return res.status(400).send({ message: "Invalid signature" })
+
+  const validatedDomain = botService.getValidatedDomain(req.body)
+  if (!validatedDomain)
+    return res.status(200).send({
+      message: `Sorry, \`${req.body.text}\` is not a valid domain name. Please try again with a valid one instead.`,
+    })
+
+  // We need to return 200 OK to Slack before we process the request
+  // The response will be done via the response_url provided in the payload
+  return botService
+    .dnsChecker(req.body)
+    .map((response) => res.status(200).send(response))
+}
+
+isobotRouter.post("/dns-checker", handleDnsChecker)
 isobotRouter.post("/whitelist-emails", handleWhitelistEmails)
--- a/support/routes/v2/isobot/index.ts
+++ b/support/routes/v2/isobot/index.ts
@@ -54,8 +54,6 @@ const handleDnsChecker: RequestHandler<
       message: `Sorry, \`${req.body.text}\` is not a valid domain name. Please try again with a valid one instead.`,
     })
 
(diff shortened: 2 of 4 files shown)


support/routes/v2/isobot/index.ts
```diff
@@ -41,5 +42,26 @@ const handleWhitelistEmails: RequestHandler<
   }
 }
 
+const handleDnsChecker: RequestHandler<
```
kishore03109, line 45: trouble you to remove the console.log in line 22 please

support/routes/v2/isobot/index.ts
```diff
@@ -41,5 +42,26 @@ const handleWhitelistEmails: RequestHandler<
...
+  unknown,
+  never
+> = async (req, res) => {
+  const validatedDomain = botService.getValidatedDomain(req.body)
+  if (!validatedDomain)
+    return res.status(200).send({
+      message: `Sorry, \`${req.body.text}\` is not a valid domain name. Please try again with a valid one instead.`,
+    })
+
+  // We need to return 200 OK to Slack before we process the request
```
kishore03109, line 58: is this comment misplaced? seems like we only return a 200 after we actually to the dns calls and get a proper response. 

also why are we optimising for api response times here ah? shouldn't this just be a ping to the domain name, should be at most 2-3 network calls no?

support/routes/v2/isobot/ops/botService.ts
```diff
@@ -19,13 +27,84 @@ export interface SlackPayload {
   text: string
 }
 
+// Slack only gives us 3 seconds to respond, but usually DNS resolution is fast,
```
kishore03109, line 30: uhm. why put an artificial limit for us here ah? is it for graceful error recovery?

support/routes/v2/isobot/ops/botService.ts
```diff
@@ -19,13 +27,84 @@ export interface SlackPayload {
...
   constructor(whitelistService: WhitelistService) {
     this.whitelistService = whitelistService
   }
 
+  private checkCname(domain: string) {
+    return ResultAsync.fromPromise(
+      Promise.race([
+        dns.resolveCname(domain),
+        new Promise<null>((_, reject) =>
+          setTimeout(() => reject(null), DNS_TIMEOUT)
```
kishore03109, line 46: rather than null can just put something like `timeout exceeded`

kishore03109: arh unless this was a typing thing

support/routes/v2/isobot/ops/botService.ts
```diff
@@ -91,6 +170,276 @@ class BotService {
 
     await this.whitelistService.addWhitelist(emails)
   }
+
+  getValidatedDomain(payload: SlackPayload) {
+    const { text: domain } = payload
+    const DOMAIN_NAME_REGEX = /^[a-zA-Z0-9][a-zA-Z0-9-]{0,61}[a-zA-Z0-9](?:\.[a-zA-Z]{2,})+$/
```
kishore03109, line 176: damn where did 61 come from

support/routes/v2/isobot/ops/botService.ts
```diff
@@ -91,6 +170,276 @@ class BotService {
...
+        // check if our indirection domain is still correct
+        if (!cnameRecord) {
+          const indirectionDomain = domain.startsWith("www.")
+            ? `${domain
+                .slice(4)
+                .replaceAll(".", "-")}.${DNS_INDIRECTION_DOMAIN}`
+            : `${domain.replaceAll(".", "-")}.${DNS_INDIRECTION_DOMAIN}`
+
+          return ResultAsync.combine([
+            okAsync(null),
```
kishore03109, line 389: i believe this is misplaced

support/routes/v2/isobot/ops/botService.ts
```diff
@@ -91,6 +170,276 @@ class BotService {
...
+        // check if the CNAME record is valid
+        return ResultAsync.combine([
+          okAsync(cnameDomain),
+          okAsync(cnameRecord),
+          okAsync(cnameRecord),
+          this.checkA(cnameRecord),
+        ])
+      })
+      .andThen(
+        ([cnameDomain, cnameRecord, indirectionDomain, indirectionRecords]) => {
```
kishore03109, line 406: nit: prefer obj over passing in an array so we dont mix up the elements accidently

support/routes/v2/isobot/ops/botService.ts
```diff
@@ -91,6 +170,276 @@ class BotService {
...
+    // Domain is directly pointing to our indirection layer
+    const isDomainOnIndirection =
+      cnameRecord && cnameRecord.endsWith(`.${DNS_INDIRECTION_DOMAIN}`)
+
+    // The intermediate layer (indirection or CNAME) is resolving to valid IPs
+    // If on KeyCDN, then there should only be 1 IP, otherwise there should be 4
+    const isIntermediateValid =
+      intermediateRecords &&
+      ((cnameRecord &&
+        cnameRecord.endsWith(".kxcdn.com") &&
```
kishore03109, line 209: i think we declared the suffixes in a different `constants.ts` file

support/routes/v2/isobot/index.ts
dcshzj: Oh yeah the comment was for a previous iteration where the response was done later, but opted to return immediately for simplicity.

Removed in 15d5512dc78e9bce861968a01f86caf51172741b!

support/routes/v2/isobot/index.ts
dcshzj: Removed in 15d5512dc78e9bce861968a01f86caf51172741b!

support/routes/v2/isobot/ops/botService.ts
dcshzj: Clarified offline. This is because we have to handle the case where the nameserver itself is taking a very long time to respond (or potentially does not respond at all). The Slack command will timeout after 3 seconds and any additional time spent processing would have been wasted work, so opted to terminate early and get the user to retry the command.

support/routes/v2/isobot/ops/botService.ts
dcshzj: Ah yeah the `logger.info` would have captured this.

support/routes/v2/isobot/ops/botService.ts
dcshzj: Domain names have a character limit of 63 characters per part, so the 61 comes from the 63 minus the first and last character.

support/routes/v2/isobot/ops/botService.ts
dcshzj: Hmm no, this one is explicit to be null because the cnameDomain does not actually exist, so we know that it is not valid.

support/routes/v2/isobot/ops/botService.ts
dcshzj: No strong opinions here but was thinking of structuring it this way so that each step of the checking process is separate. Alternatively doing the object method means they would be in a single `ResultAsync.combine()`. Possible but would prefer keeping it the current way for better understanding of the process flow, let me know your thoughts.

dcshzj: Adjusted in d33c2d0a!

src/constants/constants.ts
```diff
@@ -77,6 +77,7 @@ export const REDIRECTION_SERVER_IPS = [
   "18.139.47.66",
 ]
 export const DNS_INDIRECTION_DOMAIN = "hostedon.isomer.gov.sg"
+export const DNS_CNAME_SUFFIXES = ["cloudfront.net", "kxcdn.com"]
```
kishore03109, line 80: sir I meant that its declared in the constants file here haha

kishore03109 (approved):

support/routes/v2/isobot/ops/botService.ts
kishore03109: https://github.com/isomerpages/isomercms-backend/pull/1367#discussion_r1599217406

src/constants/constants.ts
dcshzj: Ah okay, updated in 3ea7d34d8fbb597b2b1826cff233f5b856aa8a43!

PR #1376, opened 2024-05-15
feat(monitoring): add dns reporter

kishore03109:
## Problem

This is a first pr that is up to add some level of sane reporting. 
While scheduling is part of this feature, it is not within the scope of this pr. This pr only adds (currently dead code) logic to grab the domains that we own in isomer, and do a dns dig. This is meant to be verbose, and in the future alarms can be added based on the results of this. 

This is not meant to replace monitoring, it is just meant to fine tune some blind spots that uptime robot currently has + some sane checker during incident response to show history of dns records for a site that we manage.

I am opting to log it directly in our backend to keep things simple. will add alarms + the scheduler in subsequent prs. 

## Solution

grab ALL domains from keycdn + amplify + redirection records + log dns records on them. 

**Breaking Changes**

  - Details ...

## Tests

in server.ts add: 
`monitoringService.driver()`

should see this in the logs:

## Deploy Notes

- `KEYCDN_API_KEY` : to get all the zones that we own in keycdn
- `S3_BUCKET_NAME`: bucket name 
    - [ ] HAVE NOT added env var to 1PW + SSM script

 (`fetch_ssm_parameters.sh`)

**New scripts**:

- `script` : script details

**New dependencies**:

**New dev dependencies**:

### DIFF
--- a/.aws/deploy/backend-task-definition.staging.json
+++ b/.aws/deploy/backend-task-definition.staging.json
@@ -112,6 +112,7 @@
           "valueFrom": "STAGING_INCOMING_QUEUE_URL"
         },
         { "name": "JWT_SECRET", "valueFrom": "STAGING_JWT_SECRET" },
+        { "name": "KEYCDN_API_KEY", "valueFrom": "STAGING_KEYCDN_API_KEY" },
         {
           "name": "MAX_NUM_OTP_ATTEMPTS",
           "valueFrom": "STAGING_MAX_NUM_OTP_ATTEMPTS"
--- a/.aws/deploy/backend-task-definition.staging.json
+++ b/.aws/deploy/backend-task-definition.staging.json
@@ -112,7 +112,6 @@
           "valueFrom": "STAGING_INCOMING_QUEUE_URL"
         },
         { "name": "JWT_SECRET", "valueFrom": "STAGING_JWT_SECRET" },
-        { "name": "KEYCDN_API_KEY", "valueFrom": "STAGING_KEYCDN_API_KEY" },
         {
           "name": "MAX_NUM_OTP_ATTEMPTS",
           "valueFrom": "STAGING_MAX_NUM_OTP_ATTEMPTS"
--- a/src/monitoring/index.ts
+++ b/src/monitoring/index.ts
@@ -1,0 +1,295 @@
+import dns from "dns/promises"
+
+import { Octokit } from "@octokit/rest"
+import autoBind from "auto-bind"
+import { errAsync, okAsync, ResultAsync } from "neverthrow"
+import Papa from "papaparse"
+
+import parentLogger from "@logger/logger"
+
+import config from "@root/config/config"
+import MonitoringError from "@root/errors/MonitoringError"
+import LaunchesService from "@root/services/identity/LaunchesService"
+
+interface MonitoringServiceInterface {
+  launchesService: LaunchesService
+}
+
+const IsomerHostedDomainType = {
+  REDIRECTION: "redirection",
+  INDIRECTION: "indirection",
+  KEYCDN: "keycdn",
+  AMPLIFY: "amplify",
+} as const
+
+interface IsomerHostedDomain {
+  domain: string
+  type: typeof IsomerHostedDomainType[keyof typeof IsomerHostedDomainType]
+}
+
+type keyCdnZoneAlias = {
+  name: string
+}
+
+interface KeyCdnResponse {
+  data: {
+    zonealiases: keyCdnZoneAlias[]
+  }
+}
+
+interface RedirectionDomain {
+  source: string
+  target: string
+}
+
+interface ReportCard {
+  domain: string
+  type: typeof IsomerHostedDomainType[keyof typeof IsomerHostedDomainType]
+  aRecord: string[]
+  quadArecord: string[]
+  cNameRecord: string[]
+  caaRecord: string[]
+}
+
+function isKeyCdnZoneAlias(object: unknown): object is keyCdnZoneAlias {
+  return "name" in (object as keyCdnZoneAlias)
+}
+
+function isKeyCdnResponse(object: unknown): object is KeyCdnResponse {
+  return "data" in (object as KeyCdnResponse)
+}
+
+export default class MonitoringService {
+  private readonly launchesService: MonitoringServiceInterface["launchesService"]
+
+  private readonly monitoringServiceLogger = parentLogger.child({
+    module: "monitoringService",
+  })
+
+  constructor({ launchesService }: MonitoringServiceInterface) {
+    autoBind(this)
+    this.launchesService = launchesService
+  }
+
+  getKeyCdnDomains() {
+    const keyCdnApiKey = config.get("keyCdn.apiKey")
+
+    return ResultAsync.fromPromise(
+      fetch(`https://api.keycdn.com/zonealiases.json`, {
+        headers: {
+          Authorization: `Basic ${btoa(`${keyCdnApiKey}:`)}`,
+        },
+      }),
+      (error) => new MonitoringError(`Failed to fetch zones: ${error}`)
+    )
+      .andThen((response) => {
+        if (!response.ok) {
+          return errAsync(
+            new MonitoringError(
+              `Failed to retrieve zones: ${response.statusText}`
+            )
+          )
+        }
+        return okAsync(response)
+      })
+      .andThen((response) =>
+        ResultAsync.fromPromise(
+          response.json(),
+          (error) => new MonitoringError(`Failed to parse response: ${error}`)
+        )
+      )
+      .andThen((data: unknown) => {
+        if (!isKeyCdnResponse(data)) {
+          return errAsync(new MonitoringError("Failed to parse response"))
+        }
+
+        const domains = data.data.zonealiases
+          .filter(isKeyCdnZoneAlias)
+          .map((zone: keyCdnZoneAlias) => zone.name)
+          .map(
+            (domain) =>
+              ({
+                domain,
+                type: IsomerHostedDomainType.KEYCDN,
+              } as IsomerHostedDomain)
+          )
+
+        return okAsync(domains)
+      })
+  }
+
+  getAmplifyDeployments() {
+    return this.launchesService.getAllDomains().map((domains) =>
+      domains.map(
+        (domain) =>
+          ({
+            domain,
+            type: IsomerHostedDomainType.AMPLIFY,
+          } as IsomerHostedDomain)
+      )
+    )
+  }
+
+  /**
+   * While most of our redirections are in our DB, we do have ad-hoc redirections.
+   * @returns List of redirection domains that are listed in the isomer-redirection repository
+   */
+  getRedirectionDomains() {
+    const SYSTEM_GITHUB_TOKEN = config.get("github.systemToken")
+    const OctokitRetry = Octokit.plugin()
+    const octokitWithRetry = new OctokitRetry({
+      auth: SYSTEM_GITHUB_TOKEN,
+      request: { retries: 5 },
+    })
+
+    return ResultAsync.fromPromise(
+      octokitWithRetry.request(
+        "GET /repos/opengovsg/isomer-redirection/contents/src/certbot-websites.csv"
+      ),
+      (error) =>
+        new MonitoringError(`Failed to fetch redirection domains: ${error}`)
+    )
+      .andThen((response) => {
+        const content = Buffer.from(response.data.content, "base64").toString(
+          "utf-8"
+        )
+        return ResultAsync.fromPromise(
+          new Promise<RedirectionDomain[]>((resolve, reject) => {
+            Papa.parse(content, {
+              header: true,
+              complete(results) {
+                // validate the csv
+                if (!results.data) {
+                  reject(new MonitoringError("Failed to parse csv"))
+                }
+                resolve(results.data as RedirectionDomain[])
+              },
+              error(error: unknown) {
+                reject(error)
+              },
+            })
+          }),
+          (error) => new MonitoringError(`Failed to parse csv: ${error}`)
+        )
+      })
+      .map((redirectionDomains) =>
+        redirectionDomains
+          .map((domain) => domain.source)
+          .map(
+            (domain) =>
+              ({
+                domain,
+                type: IsomerHostedDomainType.REDIRECTION,
+              } as IsomerHostedDomain)
+          )
+      )
+  }
+
+  /**
+   * This is in charge of fetching all the domains that are are under Isomer, inclusive
+   * of any subdomains and redirects.
+   */
+  getAllDomains() {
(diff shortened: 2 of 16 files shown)


kishore03109:
* **#1376** <a href="https://app.graphite.dev/github/pr/isomerpages/isomercms-backend/1376?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop`

<h2></h2>

src/monitoring/index.ts
```diff
@@ -0,0 +1,295 @@
...
+        )
+        .map((value) =>
+          reportCard.push({
+            ...value,
+          })
+        )
+        .andThen(() => okAsync(reportCard))
+    })
+
+    return ResultAsync.combineWithAllErrors(domainResolvers)
```
seaerchin, line 277: actually inspecting the signature for `domainResolvers`, the error type is `never`. 

this is because we've discarded the error earlier (`.orElse(() => okAsync([]))`). this seems to suggest that either
1. this won't error out or that
2. all errors are equivalent to an empty response?

wondering if this assumption is true and if it's not, whether there's value in actually logging the error rather than just giving up and returning empty arrays

src/server.ts
```diff
@@ -362,6 +363,10 @@ const authV2Router = new AuthRouter({
   sgidAuthRouter,
 })
 
+const monitoringService = new MonitoringService({
+  launchesService,
+})
+
```
seaerchin, line 369: could i check why's this initialised here and not the `support` container?

src/monitoring/index.ts
```diff
@@ -0,0 +1,295 @@
...
+        )
+      )
+    })
+  }
+
+  // todo: once /siteup logic is merged into dev, we can add that as to alert isomer team
+  generateReportCard(domains: IsomerHostedDomain[]) {
+    const reportCard: ReportCard[] = []
+
+    const domainResolvers = domains.map(({ domain, type }) => {
```
seaerchin, line 231: actually this function appears to be doing domain resolution and reporting the resolved records. 

should we be splitting the method up so it's clearer?

src/monitoring/index.ts
```diff
@@ -0,0 +1,295 @@
...
+        return okAsync(response)
+      })
+      .andThen((response) =>
+        ResultAsync.fromPromise(
+          response.json(),
+          (error) => new MonitoringError(`Failed to parse response: ${error}`)
+        )
+      )
+      .andThen((data: unknown) => {
+        if (!isKeyCdnResponse(data)) {
```
kishore03109, line 102: we need the data here since we want to type cast directly from an `unknown`

src/monitoring/index.ts
```diff
@@ -0,0 +1,295 @@
...
+            cNameRecord: resolvedDns[2],
+            caaRecord: resolvedDns[3],
+          })
+        )
+        .map((value) =>
+          reportCard.push({
+            ...value,
+          })
+        )
+        .andThen(() => okAsync(reportCard))
```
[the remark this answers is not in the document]
kishore03109: hmm concretely what are you suggesting ah? if we just return we will return the array directly, we will get an empty array since we resolve before the promise resolves?

src/monitoring/index.ts
kishore03109: we go with the second assumption. We are later intending to map this over /siteup for correctness, this is just a from of reporting for some level of sanity check that we have x number of domains under our control and they resolve to these dns values

src/server.ts
kishore03109: hmm that would require creating launches service on the support container. no strong opinions on this, since this is a pretty light operation anyways

seaerchin: to probe - when would this latency/cost matter? the support container is not user facing so latency doesn't really matter as opposed to creating it on the `app` container. (not that it's a big cost. i think the separation between user-facing and internal is more important here and what we want to maintain)

src/monitoring/index.ts
seaerchin: not sure i'm understanding you correctly here! if the intent is as a form of reporting, i think preserving the errors and formatting them -> displaying has value. discarding the errors seems to be counter to the intent of reporting?

src/server.ts
kishore03109: not that it's a big cost. i think the separation between user-facing and internal is more important here and what we want to maintain)

so i am assuming we dont care about singleton pattern in our codebase ya? 
since in this case, launchesService will be created twice?

src/monitoring/index.ts
kishore03109: hmm we are doing discarding any errors here. we are simply reporting that we could not resolve to anything by returning an empty string. this is like a dns dig on the terminal and getting that it resolves to nothing mah

src/monitoring/index.ts
```diff
@@ -0,0 +1,295 @@
...
+    )
+  }
+
+  /**
+   * While most of our redirections are in our DB, we do have ad-hoc redirections.
+   * @returns List of redirection domains that are listed in the isomer-redirection repository
+   */
+  getRedirectionDomains() {
+    const SYSTEM_GITHUB_TOKEN = config.get("github.systemToken")
+    const OctokitRetry = Octokit.plugin()
```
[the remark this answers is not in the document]
seaerchin: going off docs [here](https://github.com/octokit/plugin-retry.js/#readme) 

```ts
const MyOctokit = Octokit.plugin(**retry**);
const octokit = new MyOctokit({ auth: "secret123" });

// retries request up to 3 times in case of a 500 response
octokit.request("/").catch((error) => {
  if (error.request.request.retryCount) {
    console.log(
      `request failed after ${error.request.request.retryCount} retries`,
    );
  }

  console.error(error);
});```

File: src/monitoring/index.ts [hunk 7]
```diff
@@ -0,0 +1,295 @@
...
+    return ResultAsync.combine([
+      this.getAmplifyDeployments().mapErr(
+        (error) => new MonitoringError(error.message)
+      ),
+      this.getRedirectionDomains(),
+      this.getKeyCdnDomains(),
+    ]).andThen(([amplifyDeployments, redirectionDomains, keyCdnDomains]) => {
+      this.monitoringServiceLogger.info("Fetched all domains")
+      return okAsync(
+        [...amplifyDeployments, ...redirectionDomains, ...keyCdnDomains].sort(
```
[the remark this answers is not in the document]
seaerchin: i think this might be a miscommunication - since our comparison is always based on the stripped length (ie, `www.isomer.gov.sg` is taken as equivalent to `isomer.gov.sg`), why not just use `sortBy(array, val => val.startsWith("www.") ? val.slice(0, 4) : val`.

in this case, our comparator compares via the stripped value so both `www.isomer.gov.sg` will be treated identically to `isomer.gov.sg`. lmk if this makes sense

src/monitoring/index.ts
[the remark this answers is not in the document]
seaerchin: Hmm, if we just return without the `map`, it'll be in a `ResultAsync` which we can ask downstream to consume isn't it? not sure if i'm misunderstanding you here

src/monitoring/index.ts
seaerchin: could i just check how this feature is meant to be used? in my head, the flow is something like this: 
- user queries `/siteup <domain>` 
- we hit the backend here 
- we query for all the records
- if any error 
  - discard error and return []
- if no errors
  - return the resolved records?

i think it's also worth nothing here that the code structure adopted here doesn't quite fit in with `neverthrow` - we're meant to recover and fix errors (you can observe this through the example for `combineWithAllErrors` [here](https://github.com/supermacro/neverthrow?tab=readme-ov-file#resultasynccombinewithallerrors-static-class-method)). this method actually discards successful results and returns **only** the errors.

i don't think there's any actionable wrt updating this to be idiomatic `neverthrow` but it's something we should consider

src/monitoring/index.ts
seaerchin: yeap, i think in my head, the flow is something like: 
```ts
generateReportcard = (domains) => { 
  generateARecords().andThen(generateQuadA).andThen(generateCaa).andThen(generateCname).orElse(someError)
}
```

but this would potentially be expensive to do. ok w/ just shifting the domains out: 

```ts
return generateDomains().andThen(() => {
        this.monitoringServiceLogger.info({
          message: "Report card generated",
          meta: {
            reportCard,
            date: new Date(),
          },
        })
        return okAsync(reportCard)
      })
      .orElse(() => okAsync([]))
```

src/server.ts
seaerchin: i think this might be due to a mistake in rebase! **all** shared services should be exported from `"common/"`.

src/monitoring/index.ts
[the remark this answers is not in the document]
seaerchin: just wanted to nudge on this! if we care about retries, we might want to add the plugin in!

seaerchin (commented):
forgot to submit as review (sorry) - i think i'm not entirely sure how the errors will play out so i'm holding off on approval. if i could understand how you'd intend this code to function, i think it might be better

src/monitoring/index.ts
```diff
@@ -116,8 +112,11 @@ export default class MonitoringService {
    */
   getRedirectionDomains() {
     const SYSTEM_GITHUB_TOKEN = config.get("github.systemToken")
-    const OctokitRetry = Octokit.plugin()
-    const octokitWithRetry = new OctokitRetry({
+    // seems to be a bug in typing, this is a direct
+    // copy paste from the octokit documentation
+    // https://octokit.github.io/rest.js/v20#automatic-retries
+    const OctokitRetry = Octokit.plugin(retry as any)
```
seaerchin, line 118: not sure if this is required! saw your comment but going to the linked site doesn't seem to suggest this.

image

this might be due to our `tsconfig` not changing, which [seems](https://github.com/octokit/plugin-retry.js/) to be required by the `retry` library 
image

i think we might want to update our `tsconfig` and check again

seaerchin (changes requested):
potential breakage with the plugin - should either update `tsconfig` or drop

src/monitoring/index.ts
kishore03109: > user queries /siteup <domain>
we hit the backend here
we query for all the records
if any error
discard error and return []
if no errors
return the resolved records?


nope, siteup functionality is to remain as is. once we get a list of domains, the idea is to map through the list of domains via the site up so that every morning, we know if anything is not as what we expect it to be.

all this does is gets a list of domains -> then dig on them. note that the figuring out if the values returned form the dig are errors is not in this pr as of yet. again, the reason why we are ignoring this is because node:dig throws an error if it resolves to nothing, which does not mean that it is an error. it is valid, and expected for for eg isomer.gov.sg to not resolve anything for CAA results.

src/monitoring/index.ts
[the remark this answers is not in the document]
kishore03109: wait ah `value` is of type `ReportCard`, but what we want is a list. 

The distinction to make here is that `ReportCard` is for one domain, and we want to map over every domains to get this.

src/monitoring/index.ts
kishore03109: sadly i get this even after changing 
Screenshot 2024-06-06 at 3 54 25 PM
Screenshot 2024-06-06 at 3 55 10 PM

src/services/identity/LaunchesService.ts
```diff
@@ -353,6 +353,12 @@ export class LaunchesService {
           new SiteLaunchError(`Failed to update site status for ${siteName}`)
       )
     })
+
+  getAllDomains = () =>
+    ResultAsync.fromPromise(
+      this.launchesRepository.findAll(),
```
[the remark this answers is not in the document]
kishore03109: yes, which is why we also fetch domains from keycdn

src/monitoring/index.ts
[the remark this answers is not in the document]
kishore03109: this has been modified

src/monitoring/index.ts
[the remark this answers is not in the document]
kishore03109: oo this is neater code, thanks

src/server.ts
kishore03109: ok, moved already

.aws/deploy/backend-task-definition.staging.json
```diff
@@ -112,6 +112,7 @@
           "valueFrom": "STAGING_INCOMING_QUEUE_URL"
         },
         { "name": "JWT_SECRET", "valueFrom": "STAGING_JWT_SECRET" },
+        { "name": "KEYCDN_API_KEY", "valueFrom": "STAGING_KEYCDN_API_KEY" },
```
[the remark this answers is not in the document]
kishore03109: moved

src/monitoring/index.ts
```diff
@@ -0,0 +1,242 @@
...
+  getKeyCdnDomains() {
+    const keyCdnApiKey = config.get("keyCdn.apiKey")
+
+    return ResultAsync.fromPromise(
+      axios.get(`https://api.keycdn.com/zonealiases.json`, {
+        headers: {
+          Authorization: `Basic ${btoa(`${keyCdnApiKey}:`)}`,
+        },
+      }),
+      (error) => new MonitoringError(`Failed to fetch zones: ${error}`)
```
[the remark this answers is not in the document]
kishore03109: modified

src/monitoring/index.ts
```diff
@@ -0,0 +1,242 @@
...
+  KEYCDN: "keycdn",
+  AMPLIFY: "amplify",
+} as const
+
+interface IsomerHostedDomain {
+  domain: string
+  type: typeof IsomerHostedDomainType[keyof typeof IsomerHostedDomainType]
+}
+
+type keyCdnZoneAlias = {
```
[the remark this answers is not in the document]
kishore03109: changed

src/monitoring/index.ts
```diff
@@ -0,0 +1,242 @@
...
+        headers: {
+          Authorization: `Basic ${btoa(`${keyCdnApiKey}:`)}`,
+        },
+      }),
+      (error) => new MonitoringError(`Failed to fetch zones: ${error}`)
+    )
+      .map((response) => response.data.data.zonealiases)
+      .andThen((data) => {
+        if (!isKeyCdnResponse(data)) {
+          return errAsync(new MonitoringError("Failed to parse response"))
```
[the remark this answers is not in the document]
kishore03109: changed

src/monitoring/index.ts
kishore03109: hey @seaerchin, this has since been update by increasing the version number, thanks for the catch

seaerchin (approved):

kishore03109:
### Merge activity

* **Jun 27, 2:08 AM EDT**: @kishore03109 started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/isomerpages/isomercms-backend/1376).

PR #1383, opened 2024-05-17
feat(monitoring): add scheduler functionality

kishore03109:
## Problem

This is the second part of the monitoring feature that we want to build. This PR only cares about adding a scheduler + the related infra needed for this to function. this will make the monitor run once every 5 mins, for oncalls to pick any related alarms from this.

Adding the alarms is done in the downstream PR . 

## Solution
Using bullmq to conveniently create a queue, a worker and a repeatable job over multiple instances. We do some level of exponential backoff retries since it is a nice to have and easy to implement.  The original `/site-up` code has since been refactored to return an `err` or a `ok`, depending on whether the configuration is ideal. 
Unfortunately, this caused quite a number of edge cases to pop up. Due to the nature of this, a more loose check of whether the isomer logo is present is being used to determine if a site is up. 
Even with this loose check, we have a `workplacelearning.gov.sg` who have modified their site to not have the Isomer logo. Have used gb to code white list this weird site. Potentially, if tomorrow we have an alarm of a site going down, but this is expected to prolong, we can go to growthbook and change the config for this to be whitelisted. 

**Breaking Changes**

  - Details ...

## Tests
Screenshot 2024-05-21 at 11 20 59 AM
on deployment, assert that you see these logs. it is ok for there to be multiple instances of this log (it directly corresponds to the number of instances that we have) since bullmq is smart enough to only create one queue, and one repeatable job over multiple instances. 

## Deploy Notes

corresponding infra pr should be deployed to production and only then should the redis host value be populated into the 1pw for production. 

Additionally, post approval of this pr, add two alarms, one for 
`Error running monitoring service` and another for `Monitoring service has failed`. These are errors when the job has failed to be initalised, and when there is a new error. 

- `REDIS_HOST` : Redis host 

**New dependencies**:

- `bullmq` : scheduler of choice


kishore03109:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/isomerpages/isomercms-backend/1383?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1383** <a href="https://app.graphite.dev/github/pr/isomerpages/isomercms-backend/1383?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1376** <a href="https://app.graphite.dev/github/pr/isomerpages/isomercms-backend/1376?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `develop`

<h2></h2>

seaerchin:
test failure from a missing `keyCdn.apiKey`! could i trouble you to fix this @kishore03109

kishore03109:
> test failure from a missing keyCdn.apiKey! could i trouble you to fix this @kishore03109

added from downstream pr

src/monitoring/index.ts
```diff
@@ -65,9 +67,74 @@ export default class MonitoringService {
...
     autoBind(this)
+    const jobName = "dnsMonitoring"
     this.launchesService = launchesService
+    this.worker = new Worker(
+      this.queue.name,
+      async (job: Job) => {
+        this.monitoringServiceLogger.info(`Monitoring Worker ${job.id}`)
+        if (job.name === jobName) {
+          // The retry's work on a thrown error, so we need to convert the neverthrow to a promise
+          const res = await convertNeverThrowToPromise(this.driver())
```
seaerchin, line 102: just off the top of my head, this seems odd. we're defining driver in `MonitoringService` but actually `Worker` uses it. This seems to blur the lines between the service and the worker. 

i think this PR might be better if we split the concerns here! logically speaking, i'd expect the worker to be the one to actually _do_ the work, and for the monitoring service to just handle the worker scheduling/linking up the worker to the queue. 

i'm not sure if this delineation makes sense (it might not) but i think having a method defined on the service that's **only called by the worker** is definitely a code smell

seaerchin (changes requested):

.aws/deploy/backend-task-definition.prod.json
```diff
@@ -133,6 +133,7 @@
           "name": "REDIRECT_URI",
           "valueFrom": "PROD_REDIRECT_URI"
         },
+        { "name": "REDIS_HOST", "valueFrom": "PROD_REDIS_HOST" },
```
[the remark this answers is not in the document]
kishore03109: this has since been modified. only support needs this.

src/monitoring/index.ts
```diff
@@ -65,9 +67,74 @@ export default class MonitoringService {
...
+      backoff: {
+        type: "exponential",
+        delay: 60000, // this operation is not critical, so we can wait a minute
+      },
+    },
+  })
+
+  private readonly worker: Worker<unknown, string, string> | undefined
+
   constructor({ launchesService }: MonitoringServiceInterface) {
```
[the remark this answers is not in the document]
kishore03109: good catch, need error handler, added it

src/monitoring/index.ts
```diff
@@ -65,9 +67,74 @@ export default class MonitoringService {
...
+      },
+      {
+        connection: {
+          ...this.REDIS_CONNECTION,
+        },
+        lockDuration: 60000, // 1 minute, since this is a relatively expensive operation
+      }
+    )
+
+    const dailyCron = "0 0 9 * *"
```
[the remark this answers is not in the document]
kishore03109: wah. good catch, ecs timezone is utc. BUT, actually following our incident actually, our actual monitor (uptime robot) sucks. am going to use 5 min interval instead, lmk your thoughts on that.

src/monitoring/index.ts
```diff
@@ -65,9 +67,74 @@ export default class MonitoringService {
...
+      removeOnFail: true,
+      attempts: 3,
+      backoff: {
+        type: "exponential",
+        delay: 60000, // this operation is not critical, so we can wait a minute
+      },
+    },
+  })
+
+  private readonly worker: Worker<unknown, string, string> | undefined
```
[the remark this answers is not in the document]
kishore03109: do you know how to? the way that i see this, this is a limitation of the library, as is allows jobs of any shape to be scheduled. in this case, we want to keep the type to unknown and actually we are not using the data at all (since now the invariant is that any job is a monitoring job, so start the scheduled driver)

.aws/deploy/backend-task-definition.prod.json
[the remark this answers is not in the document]
seaerchin: sorry not quite understanding! why would they share the same env code?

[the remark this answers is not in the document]
kishore03109: they share the same `config.ts` code? either we would separate them or make the env var as optional right?

src/monitoring/index.ts
```diff
@@ -65,9 +67,74 @@ export default class MonitoringService {
...
+          const res = await convertNeverThrowToPromise(this.driver())
+          return res
+        }
+        throw new MonitoringError("Invalid job name")
+      },
+      {
+        connection: {
+          ...this.REDIS_CONNECTION,
+        },
+        lockDuration: 60000, // 1 minute, since this is a relatively expensive operation
```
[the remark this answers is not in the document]
kishore03109: this is renamed!

src/monitoring/index.ts
kishore03109: hey, this has been refactored

.aws/deploy/backend-task-definition.prod.json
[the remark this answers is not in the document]
seaerchin: ah ok, sounds good

src/monitoring/index.ts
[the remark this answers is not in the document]
seaerchin: i think we should not. our status page is based off uptime robot + that's what we use to determine our metrics, so i think we should stick with it.

additionally, there's drift between what's on uptime robot and redirection/indirection (uptime robot has some that's not present on the other) so i think we should just go with uptime robot unless there are breaking issues w/ uptime robot

src/monitoring/index.ts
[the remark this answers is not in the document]
seaerchin: the library exposes typings on `Worker` and `Job`. 

```ts
  private readonly worker: Worker<number, number, string>
  this.worker = new Worker<number, number>
  job: Job<number, number>
```

these allowed me to type it correctly. since we're just using the queue as a cronjob, the value add from `unknown -> void` might not be that much. fine w/ either

.aws/deploy/backend-task-definition.prod.json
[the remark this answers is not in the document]
seaerchin: i realised we give a default in `config.ts` so i think there's still value in removing. leave the choice to you but fwiw, removing here is clearer

src/monitoring/MonitoringService.ts
```diff
@@ -0,0 +1,109 @@
...
+      },
+      {
+        connection: {
+          ...this.REDIS_CONNECTION,
+        },
+        lockDuration: ONE_MINUTE, // since this is a relatively expensive operation
+      }
+    )
+
+    this.worker.on("failed", (job: Job | undefined, error: Error) => {
```
seaerchin, line 99: we also need a handler for the `error` event:

`this.worker.on("error", () => {...})`

this only handles failed jobs but because the worker runs in process, an error on the worker itself will bring down our whole service.

src/monitoring/MonitoringService.ts
```diff
@@ -0,0 +1,109 @@
...
+      logger.error({
```
seaerchin, line 100: [FYI, no action required] the failure to consume a job might be concerning. this is as good as a silent failure here because we'll never take note.

wondering if we should just ping a channel? or alternatively, set alarms on `logger.error`

src/utils/dns-utils.ts
```diff
@@ -0,0 +1,327 @@
...
+  })
+    .andThen((cname) => {
+      if (!cname || cname.length === 0) {
+        return okAsync(null)
+      }
+
+      //! todo: remove, this is some colima weird logic
+      if (cname.length === 1 && cname[0] === domain) {
+        return okAsync(null)
+      }
```
seaerchin, line 34: this seems like a todo? should this be here

src/middleware/featureFlag.ts
```diff
@@ -6,7 +6,7 @@ import { getNewGrowthbookInstance } from "@root/utils/growthbook-utils"
 
 // Keep one GrowthBook instance at module level
 // The instance will handle internal cache refreshes via a SSE connection
-const gb = getNewGrowthbookInstance({
+export const gb = getNewGrowthbookInstance({
```
seaerchin, line 9: are we sure this is what we want? this subscribes to changes via SSE and **during the job's lifecycle**, we can update the changes.

eg:
t = 0, config has `whiteListedRepos: [xyz], isEnabled: true`  
t = 1, worker gets job
t = 2, worker runs
t = 3, config updated so that `whitelistedRepos: [], isEnabled: true` 
t = 4, we don't run for xyz.

ok either way but i suggest unexporting this and create a **local, scoped** growthbook instance in `worker` and `dns-utils`

src/utils/dns-utils.ts
seaerchin, line 1: i'm honestly not quite sure how to review this so i'm skipping for now

support/index.ts
```diff
@@ -24,7 +20,17 @@ const app = express()
 // poller site launch updates
 infraService.pollMessages()
 
-monitoringService.driver()
+// only needed for support container
+export const monitoringWorker = new MonitoringWorker({
+  launchesService,
+})
+// dnsMonitor("isomer.gov.sg").mapErr(console.log).map(console.log)
+// todo: remove after testing
```
seaerchin, line 28: do we need this here?

seaerchin (approved):
i think only blocker from me is the handling of worker errors!

this seems like a single producer, potentially multiple consumer case but it should be ok since we don't care about ordering.

.aws/deploy/backend-task-definition.prod.json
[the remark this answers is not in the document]
kishore03109: ok, i make var optional, then remove in task def

src/utils/dns-utils.ts
kishore03109: ps, sorry this has been removed

src/monitoring/MonitoringService.ts
kishore03109: yes, part of the deploy step is to add alarm for this

kishore03109:
@seaerchin 
> this seems like a single producer, potentially multiple consumer case but it should be ok since we don't care about ordering.

hmm wait, no this is why we are using bullmq, right? it implements some sort of locking mechanism to ensure that only one consumer (instance) is doing the job at one time?

support/index.ts
kishore03109: sorry, removed

kishore03109: sorry this has been removed

src/monitoring/MonitoringService.ts
kishore03109: good catch, missed this. has been added

src/middleware/featureFlag.ts
kishore03109: oki, changed so that now they both have a local scoped gb instance

src/monitoring/MonitoringService.ts
```diff
@@ -96,9 +102,15 @@ export default class MonitoringService {
       }
     )
 
+    this.worker.on("error", (error: Error) => {
+      logger.error({
+        message: "Monitoring service has errored",
+        error,
+      })
+    })
```
seaerchin, line 110: no issues on using `logger.error` here, just a reminder that this should ideally ping the alarms channel or call on-call as this implies the service has already gone down

.aws/deploy/backend-task-definition.prod.json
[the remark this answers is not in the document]
seaerchin: i don't think we need to make it optional! i thought required w/ defaults will cause it to fall back to the default if it isn't specified? 

if this isn't the case, both are about the same imo

src/monitoring/MonitoringService.ts
seaerchin: sgtm, not resolving for visibility but not a blcoker

src/middleware/featureFlag.ts
seaerchin: could i be annoying and suggest **not exporting** this? intentional choice so we have to go through the helper method

src/monitoring/MonitoringWorker.ts
```diff
@@ -177,6 +179,10 @@ export default class MonitoringWorker {
   }
 
   driver() {
+    const gb = getNewGrowthbookInstance({
+      clientKey: config.get("growthbook.clientKey"),
+      subscribeToChanges: true,
```
seaerchin, line 184: just a note here, because we set `subscribeToChanges: true`, the same issue where the config changes in-processing still remains. 

this isn't really an issue for me because it only concerns whitelisted sites + whether this is enabled. leave it to you to make the final judgement. how it behaves now is that it initialises 2 new instances and gets the config at that time (whilst listening to changes); if we want to say that the config doesn't change at all in-processing, we could instead initialise growthbook once and get the config at the start and pass it downwards.

tl;dr: behaviour still doesn't change, which means that config can change in-processing. not a biggie for me but leave it to you to decide if you wanna make this change

seaerchin:
> @seaerchin
> 
> > this seems like a single producer, potentially multiple consumer case but it should be ok since we don't care about ordering.
> 
> hmm wait, no this is why we are using bullmq, right? it implements some sort of locking mechanism to ensure that only one consumer (instance) is doing the job at one time?

how bullmq works is (according to their docs, at least) that a worker must continually inform the queue that it's still working on the job. if this doesn't happen (thread is always busy on worker), then the job will be stalled and it's possible to have another worker working on it. 

but for clarification, this isn't what i'm concerned about! my comment was wrt _out of order consumption_ of jobs. i think in this case, we don't really care since the job is just to monitor all the site and ordering doesn't really matter

kishore03109 (commented):
@seaerchin 
As discussed, thowing is deprior for now.
after upgrading the types, all work, verified that the breaking changes were node upgades, and since we are already using node18, this should no be an issue.

kishore03109:
### Merge activity

* **Jun 27, 2:08 AM EDT**: @kishore03109 started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/isomerpages/isomercms-backend/1383).

PR #1391, opened 2024-05-27
ref(bot): use bolt for app

seaerchin:
## Problem
At present, we are using express' router for interfacing with slack, which is less ergonomic than using bolt. this is problematic because we want to offload some tasks from form -> slack bot, which might be more cumbersome using bolt.

## Solution
Shift from using express to using `bolt` + `ExpressReceiver`. This simplifies code-flow abit, but note that there si some validation logic which isn't captured here but will be downstream

## Tests 
**NOTE:** This is untested at present

## Deploy Notes
- go to SSM 
- search for `SLACK` 
- check that there are four secrets (2 tokens * 2 envs)

### DIFF
--- a/.aws/deploy/support-task-definition.staging.json
+++ b/.aws/deploy/support-task-definition.staging.json
@@ -174,6 +174,10 @@
           "valueFrom": "STAGING_SITE_PASSWORD_SECRET_KEY"
         },
         {
+          "name": "SLACK_SIGNING_SECRET",
+          "valueFrom": "STAGING_SLACK_SIGNING_SECRET"
+        },
+        {
           "name": "STAGING_SSH_PUBLIC_KEY",
           "valueFrom": "STAGING_SSH_PUBLIC_KEY"
         },
--- a/.aws/deploy/support-task-definition.staging.json
+++ b/.aws/deploy/support-task-definition.staging.json
@@ -178,6 +178,10 @@
           "valueFrom": "STAGING_SLACK_SIGNING_SECRET"
         },
         {
+          "name": "SLACK_TOKEN",
+          "valueFrom": "PROD_SLACK_TOKEN"
+        },
+        {
           "name": "STAGING_SSH_PUBLIC_KEY",
           "valueFrom": "STAGING_SSH_PUBLIC_KEY"
         },
--- a/support/routes/v2/isobot/index.ts
+++ b/support/routes/v2/isobot/index.ts
@@ -19,9 +19,7 @@ const botReceiver = new ExpressReceiver({ signingSecret, endpoints: "/" })
 export const isobotRouter = botReceiver.router
 
 const bot = new App({
-  signingSecret,
   token,
-  endpoints: "/",
   receiver: botReceiver,
 })
 
@@ -73,6 +71,5 @@ const handleDnsChecker: RequestHandler<
 }
 
 isobotRouter.post("/dns-checker", handleDnsChecker)
-isobotRouter.post("/whitelist-emails", handleWhitelistEmails)
 // FIXME: update this to proper signature
 // bot.command("/whitelist-emails", handleWhitelistEmails)
--- a/support/routes/v2/isobot/index.ts
+++ b/support/routes/v2/isobot/index.ts
@@ -1,13 +1,11 @@
 import { App, ExpressReceiver } from "@slack/bolt"
-import { RequestHandler } from "express"
 
 import { Whitelist } from "@database/models"
 import config from "@root/config/config"
 import logger from "@root/logger/logger"
 import WhitelistService from "@root/services/identity/WhitelistService"
-import type { DnsCheckerResponse } from "@root/types/dnsChecker"
 
-import BotService, { SlackPayload } from "./ops/botService"
+import BotService from "./ops/botService"
 
 const botService = new BotService(
   new WhitelistService({ repository: Whitelist })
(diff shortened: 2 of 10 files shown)


support/routes/v2/isobot/index.ts
```diff
@@ -1,15 +1,28 @@
...
 const botService = new BotService(
   new WhitelistService({ repository: Whitelist })
 )
+const signingSecret = config.get("slackbot.secret")
+const token = config.get("slackbot.token")
+
+const botReceiver = new ExpressReceiver({ signingSecret, endpoints: "/" })
+export const isobotRouter = botReceiver.router
+
+const bot = new App({
```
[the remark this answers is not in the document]
seaerchin: i'm not very sure here. in this case, we want bolt to act as a router, not as a server. this means that we want bolt to capture a subset of routes and respond to those with a registered handler rather than capturing all incoming network data over the specified port (what `.start` does i think)

again, i haven't tested/deployed this so i can't be sure but i think this is my understanding

seaerchin:
pending actions on my end: spin up local `ngrok` instance, register this endpoint and check to see if it works

support/routes/v2/isobot/index.ts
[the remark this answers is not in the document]
seaerchin: tested with a `echo` command (not committed) against a ngrok forwarding to my local instance. this worked but with a huge caveat - we omit using middleware for the bot. this is because bolt runs their own middleware (see below ss) which causes forwarding from express -> bolt to fail 

i'm assuming this is ok because bolt _should_ enforce the security measures on their end via their inbuilt middleware. an alternative is we could manually check the middlewares until we find the one that causes the failure (not worth imo). an exhaustive list of the middleware: 

```ts

  if (isSecure) {
    // Our server only receives requests from the alb reverse proxy, so we need to use the client IP provided in X-Forwarded-For
    // This is trusted because our security groups block all other access to the server
    app.set("trust proxy", true)
  }
  app.use(helmet())

  // use growthbook across routes
  app.use(featureFlagMiddleware)

  app.use(
    cors({
      origin: FRONTEND_URL,
      credentials: true,
    })
  )
  app.use(express.json({ limit: "7mb" }))
  app.use(express.urlencoded({ extended: false }))
  app.use(cookieParser())
  app.use(express.static(path.join(__dirname, "public")))
  app.use(nocache())
}
```

image

image

support/routes/v2/isobot/index.ts
```diff
@@ -1,63 +1,47 @@
...
-  { message: string },
-  SlackPayload,
-  {},
-  {}
-> = async (req, res) => {
-  const slackTimestamp = req.headers["x-slack-request-timestamp"] as string
-  const slackSig = req.headers["x-slack-signature"] as string
-
-  if (!slackTimestamp || !slackSig)
-    return res.send({ message: "Missing header/signature" })
```
seaerchin, line 26: we don't need this as `bolt` provides this by default - see [here](https://api.slack.com/authentication/verifying-requests-from-slack#SDK-support)

support/routes/v2/isobot/ops/botService.ts
```diff
@@ -16,22 +13,6 @@ import logger from "@root/logger/logger"
...
-  token: string
-  team_id: string
-  team_domain: string
-  channel_id: string
-  channel_name: string
-  user_id: string
-  user_name: string
-  command: string
-  text: string
-}
```
seaerchin, line 29: used inbuilt type `SlashCommand`

support/routes/v2/isobot/ops/botService.ts
```diff
@@ -40,21 +21,13 @@ class BotService {
...
-        ),
-      ]),
-      () => {
-        logger.info({
-          message: "Error resolving CNAME",
-          meta: { domain, method: "checkCname" },
-        })
-        return new Error()
-      }
-    )
```
seaerchin, line 57: we don't need to `Promise.race` because we `await ack()` in the caller - that's the ack within 3s that slack is waiting for.

see [here](https://slack.dev/bolt-js/concepts#commands) under listening and responding to commands

support/routes/v2/isobot/ops/botService.ts
```diff
@@ -66,21 +39,13 @@ class BotService {
...
-      () => {
-        logger.info({
-          message: "Error resolving A record",
-          meta: { domain, method: "checkA" },
-        })
-        return new Error()
-      }
-    )
+    return ResultAsync.fromPromise(dns.resolve4(domain), () => {
+      logger.info({
```
harishv7, line 43: message says error but the log level is info?

support/routes/v2/isobot/ops/botService.ts
```diff
@@ -106,51 +71,11 @@ class BotService {
...
-    }
-    logger.error({ message: "Signature verification failed" })
-    return false
-  }
-
-  public async whitelistEmails(payload: SlackPayload) {
+  public async whitelistEmails(text: string) {
     // Sample user input:
     // email1,expDate email2,expDate
     // email1@xyz.com,2024-06-22 email2@abc.com,2025-01-31
```
harishv7, line 77: does this format make sense, feel free to suggest alternatives

support/routes/v2/isobot/index.ts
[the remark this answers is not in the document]
harishv7: to close up the loop on this, we are using express as the server and bolt as the router? bolt has it's own internal (unexposed) middlewares which does things like hmac verification etc.?

.aws/deploy/support-task-definition.staging.json
```diff
@@ -173,6 +173,14 @@
           "name": "SITE_PASSWORD_SECRET_KEY",
           "valueFrom": "STAGING_SITE_PASSWORD_SECRET_KEY"
         },
+        {
+          "name": "SLACK_SIGNING_SECRET",
```
harishv7, line 177: need to add to SSM also right. Rmb when deploying!

support/routes/v2/isobot/index.ts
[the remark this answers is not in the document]
seaerchin: yes, check the source [here](https://github.com/slackapi/bolt-js/blob/ddeb601b0d140b47e93e08305c23ad2cef4508b9/src/receivers/ExpressReceiver.ts#L430)

support/routes/v2/isobot/ops/botService.ts
seaerchin: sure, i'll mkae the change

support/routes/v2/isobot/ops/botService.ts
seaerchin: we can run with this first and let prod ops decide. tbh i think this won't quite work for >2-3 emails but idk how rare that would be

harishv7 (approved):

PR #1392, opened 2024-05-28
feat(isobot): add locking/cloning functionalities

seaerchin:
## Problem
We want to shift locking/unlocking functionality into a slack bot because using the form is unwieldy. This is because we have to log into form sg -> otp -> find form -> otp again. 

## Solution
- just shift related APIs/endpoint into a slackbot endpoint. copied stuff as-is from the existing form repair service so there shouldn't be too much differences
- not quite sure if it's safe to set `error.cause` in `unlock` for `mutex-utils`. as is, the current return from `unlock` is pretty useless

### DIFF
--- a/support/routes/v2/isobot/index.ts
+++ b/support/routes/v2/isobot/index.ts
@@ -4,7 +4,6 @@ import {
   Middleware,
   SlackCommandMiddlewareArgs,
 } from "@slack/bolt"
-import { RequestHandler } from "express"
 import { okAsync } from "neverthrow"
 
 import { repairService } from "@common/index"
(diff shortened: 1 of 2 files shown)


support/routes/v2/isobot/index.ts
```diff
@@ -16,11 +23,37 @@ const token = config.get("slackbot.token")
 const botReceiver = new ExpressReceiver({ signingSecret, endpoints: "/" })
 export const isobotRouter = botReceiver.router
 
+// TODO: add slack ids of isomer user
+const ISOMER_USERS_ID = ["U01HTSFC0RY"]
```
harishv7, line 27: should we be hardcoding this? Or fetch it dynamically? Else every member movement in team will require a change

An alternative I can think of is to fetch members of the `isomer-team` and then check if the user is within this team. I believe the team will be more permanent than a user id

support/routes/v2/isobot/index.ts
```diff
@@ -45,3 +78,112 @@ bot.command("/siteup", async ({ payload, respond, ack }) => {
...
 })
+
+bot.command(
+  "/clone",
+  validateIsomerUser,
+  async ({ command, ack, respond, payload, client }) => {
+    await ack()
+
+    const HELP_TEXT =
+      "Usage: `/clone <github_repo_name>`. Take note that this locks the repo for 15 minutes by default. To bypass this behaviour, add `-n` at the end of the command"
```
harishv7, line 89: us clone the right term? we usually refer to it as a repair?

support/routes/v2/isobot/index.ts
```diff
@@ -45,3 +78,112 @@ bot.command("/siteup", async ({ payload, respond, ack }) => {
...
+)
+
+bot.command(
+  "/lock",
+  validateIsomerUser,
+  async ({ command, ack, respond, payload, client }) => {
+    await ack()
+
+    const HELP_TEXT =
+      "Usage: `/lock <github_repo_name> -d <duration_in_minutes>`. Take note that this locks the repo for 15 minutes by default if `-d` is not specified"
```
harishv7, line 128: should d also have a max time? what if someone puts in an accidental 10000000 mins or something?

support/routes/v2/isobot/index.ts
seaerchin: the repair does a `lock -> clone -> unlock` under the hood!

support/routes/v2/isobot/index.ts
seaerchin: ok this sounds good to me, i'll make this change!

support/routes/v2/isobot/index.ts
seaerchin: makes sense, will make this change

support/routes/v2/isobot/index.ts
harishv7: correct, hence the 3-step process is called `repair`

PR #1402, opened 2024-06-03
chore: update deps

seaerchin:
## Problem
We have an outdated `dompurify` dependency which exposes us to template injection

## Solution
update `isomorphic-dompurify` to 2.x. This drops MSIE support and changes the way they detect server/web env but i don't think it's applicable to us

## Tests 
- [ ] go into any repo
- [ ] update some content
- [ ] click save
- [ ] the save should be successful

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,7 +73,7 @@
     "hot-shots": "^10.0.0",
     "http-errors": "~1.8.0",
     "is-svg": "^4.4.0",
-    "isomorphic-dompurify": "^0.27.0",
+    "isomorphic-dompurify": "^2.0.0",
     "isomorphic-git": "^1.25.10",
     "joi": "^17.13.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -73,7 +73,7 @@
     "hot-shots": "^10.0.0",
     "http-errors": "~1.8.0",
     "is-svg": "^4.4.0",
-    "isomorphic-dompurify": "^0.27.0",
+    "isomorphic-dompurify": "^2.0.0",
```
harishv7, line 76: seems to be 2 major versions ahead, any breaking changes?

harishv7 (approved):
approved with a comment on any breaking changes

package.json
seaerchin: Idt it affects us tbh

PR #1449, opened 2024-09-24
chore: add datadog test action

seaerchin:
## Problem
we want to have visibility on our tests on datadog.

## Solution
1. use dd test visibility action to forward to dd
2. add the action right before the test step as per their docs to avoid interference with the tracers

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -86,8 +86,19 @@ jobs:
           path: ~/.npm
           key: ${{ runner.OS }}-node-${{ hashFiles('**/package-lock.json') }}
       - run: npm ci
-      - run: npm run test:docker
-      - run: . .env.test && npx jest --runInBand
+      - run: npm run test:docker      
+      - name: Configure Datadog Test Visibility
+        env:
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -86,7 +86,15 @@ jobs:
...
+      - run: npm run test:docker      
+      - name: Configure Datadog Test Visibility
+        env:
+          DD_API_KEY: ${{ secrets.DD_API_KEY }}
+        uses: datadog/test-visibility-github-action@v1
+        with:
+          languages: js
+          service: isomer
+          api_key: ${{ secrets.DD_API_KEY }}
       - run: . .env.test && npx jest --runInBand
```
karrui, line 98: need to add the `NODE_OPTIONS` env var to the test step so dd agent can inject itself

```
  env:
    NODE_OPTIONS: -r ${{ env.DD_TRACE_PACKAGE }}
```

seaerchin: added

karrui (approved):

PR #1454, opened 2025-01-28
fix: use admin token for privatisation

alexanderleegs:
Uses our admin token to privatise repos, to fix github's changes to disallow write tokens to change repo status.

### DIFF
--- a/src/services/db/GitHubService.ts
+++ b/src/services/db/GitHubService.ts
@@ -17,6 +17,8 @@ import logger from "@root/logger/logger"
 import { GitCommitResult } from "@root/types/gitfilesystem"
 import { GitHubRepoInfo, RawGitTreeEntry, RepoState } from "@root/types/github"
 
+import { E2E_TEST_GH_TOKEN } from "../middlewareServices/AuthenticationMiddlewareService"
+
 import * as ReviewApi from "./review"
 
 export default class GitHubService {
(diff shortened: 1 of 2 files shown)


src/services/db/GitHubService.ts
```diff
@@ -17,6 +17,8 @@ import logger from "@root/logger/logger"
 import { GitCommitResult } from "@root/types/gitfilesystem"
 import { GitHubRepoInfo, RawGitTreeEntry, RepoState } from "@root/types/github"
 
+import { E2E_TEST_GH_TOKEN } from "../middlewareServices/AuthenticationMiddlewareService"
```
seaerchin, line 20: do we actually have this token at run time or is it only during e2e on local

alexanderleegs: Run time - we currently also use it for pr approval

seaerchin (approved):

dcshzj (approved):

src/services/db/GitHubService.ts
harishv7: Not a blocker but The name of the token is very misleading

src/services/db/GitHubService.ts
```diff
@@ -861,9 +863,9 @@ export default class GitHubService {
     const endpointTemplate = urlTemplate.parse(`{siteName}`)
     const endpoint = endpointTemplate.expand({ siteName })
 
-    // Privatising a repo is restricted to repo admins - an admin token will be inserted in via our axios interceptor
+    // Privatising a repo is restricted to repo admins - our main admin token is used
     const headers = {
-      Authorization: "",
```
harishv7, line 866: How come this was previously empty? Comment seems to say this was injected by the axios interceptor 

If we are modifying to inject token here is the interceptor still needed

alexanderleegs: interceptor always exists, and inserts in the token in cases where it doesn't exist (i.e. all email login actions). normally this is just to catch the github flow, where the user's personal github token wold be used (but they would lack perms)

In this case we're using the admin token for both flows, so no change

PR #1460, opened 2025-05-21
fix: comment logic

seaerchin:
## Problem
currently, the way our comments work is getting the associated `reviewMeta` via the `pullRequestNumber` then usign that to get the `reviewId`. however, we conflate the `pullRequestNumber` (actual pr number in github) with the `reviewId` given by the user (logical id for our review request, unique across every review, even if pr number is the same)

## Solution
1. ~use `reviewId` provided by the request to both create and get comments~ -> the `reviewId` provided by the request actually isn't the `reviewId` in our database. this is the **PR number**. because of this, we have to use the `(prNumber, siteName)` tuple to extract the `reviewMeta`!!! i've added a test for this under `GET comments`
2. still preserve getting `reviewMeta` as a check to ensure that the review request exists

### DIFF
--- a/src/services/review/ReviewRequestService.ts
+++ b/src/services/review/ReviewRequestService.ts
@@ -1077,47 +1077,49 @@ export default class ReviewRequestService {
 
   createComment = async (
     sessionData: UserWithSiteSessionData,
-    pullRequestNumber: number,
+    reviewId: number,
     message: string
   ): Promise<ReviewComment> => {
     const { siteName, isomerUserId } = sessionData
 
-    logger.info(
-      `Creating comment for PR ${pullRequestNumber}, site: ${siteName}`
-    )
-    // get id of review request
+    logger.info(`Creating comment for PR ${reviewId}, site: ${siteName}`)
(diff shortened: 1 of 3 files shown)


src/services/review/ReviewRequestService.ts
```diff
@@ -1077,55 +1077,57 @@ export default class ReviewRequestService {
 
   createComment = async (
```
harishv7, line 1078: Have we updated the caller functions to pass in the right arguments?

src/services/review/ReviewRequestService.ts
```diff
@@ -1077,55 +1077,57 @@ export default class ReviewRequestService {
...
         )
         throw new DatabaseError("Error creating comment in DB")
       }
     }
-    logger.info(`No review request found for PR ${pullRequestNumber}`)
+    logger.info(`No review request found for PR ${reviewId}`)
     throw new RequestNotFoundError("Review Request not found")
   }
 
   getComments = async (
```
harishv7, line 1113: need to update the test spec files

